PR #4, opened 2017-01-30
Line hints on parser exception

jarradh:
Introduces `ParserException` class that is parent to `InvalidTypeException`, `TypeMismatchException`, `VariableDeclarationException`, `VariableDeclarationException` & `ConstancyViolationException`.   

ParserException attempts to highlight the violating code segment in the parsed viper source. 

Exception string examples:
```
line 2: Invalid type: 'Num(n=5)'
x: 5
---^

line 2: Argument must have type
def foo(x): pass
--------^

line 5: Undeclared variable: x
    send("0x1234567890123456789012345678901234567890", x)
-------------------------------------------------------^
```

It is used two ways, one via passing an ast node directly to the exception at raise, it is also implemented as a more general `try: ... except ParserException:` which was introduced in sections where either LLL has some parse error but no directly accessible ast node or there was a general `item` or `stmt` that can be used in a catch-all.

Also I'm reintroducing the cli tool `bin/viper` to `setup.py` which will added to PATH on install. 

Also a minor fix to `.gitignore` which broke on last merge.

Future plans for next PR will be to create examples:
`multisig`, `erc20 token`, `vault`, `auth` & `ENS resolver`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,14 +1,13 @@
-<<<<<<< HEAD
-viper.egg-info/
-__pycache__/
-build/
-dist/
-.eggs/
-=======
 # Byte-compiled / optimized / DLL files
 __pycache__/
 *.py[cod]
 *$py.class
+
+# egg-related
+viper.egg-info/
+build/
+dist/
+.eggs/
 
 # pyenv
 .python-version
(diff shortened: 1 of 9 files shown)


jarradh:
This should be ready for review.

jarradh:
Updated to remove global

jarradh:
Rebased, conflicts fixed.

PR #266, opened 2017-04-22
update README.md

NIC619:
update on: map, struct, public


yograterol:
Any reason to don't merge this?

fubuloubu:
Yeah, agree with @yograterol, I was wondering why the readme had incorrect (old?) syntax

PR #275, opened 2017-07-18
Add Solidity-compatible ERC20 Example

pdaian:
**NOT READY FOR REVIEW!**


AlexXiong97:
Hi Phil, 

I fixed some failed tests in your test.py 
https://github.com/AlexXiong97/Viper-ERC/blob/master/note_test_py.txt

Specifically:
NOTE:

88 : self.assertEqual(self.c.allowance(self.t.a1, self.t.a2, sender=self.t.k2), 1)
88*: self.assertEqual(self.c.allowance(self.t.a2, self.t.a1, sender=self.t.k2), 1)

89 : By line 88, k3 approved k2 1 token, k2 approved k1 1 token, but, only k1 has 1 token
      therefore, it should return false.
89*: self.assertFalse(self.c.transferFrom(self.t.a2, self.t.a3, 1, sender=self.t.k2))

92 : Should return false, because k2 once approved k1 1 token, so without resetting it
      to 0, k2 are not allowed to approve again.
92*: self.assertFalse(self.c.approve(self.t.a1, 1, sender=self.t.k2))

93 :self.assertEqual(self.c.allowance(self.t.a1, self.t.a2, sender=self.t.k2), 1)
93*:self.assertEqual(self.c.allowance(self.t.a2, self.t.a1, sender=self.t.k2), 1)

After changing all code above, and then run python3.6 test.py, it passes all the test

pdaian:
Thanks so much @AlexXiong97!  I've updated the tests with a full coverage test suite that was missing before, and am pretty much ready to mark this for review (will check a few more things first).  You should pull in the new version, as I found a mistake while finishing the new test suite (namely that deposit, when it returned first, would keep Ether and not throw, thus leading to potentially stuck Ether).

Glad you and @glambeth found the example useful!

fubuloubu:
I just quickly reviewed this again and realized you have 2 overflow functions and a payable function to basically do what the language is already supposed to do inherently (adding the `@payable` decorator where relevant e.g. `self.payable(true, msg.value)` and not where it is false). I'm also not familiar with what `num256_sub`, but just based on intuition it shouldn't be necessary for the same reason as overflows should be checked, so all in all you should be able to significantly simplify this example while still being functionally compatible with the ERC20 token (you can confirm this if all your failures are still raising in your tests)

Also, `num256` is the same as just `num` I believe

pdaian:
Closed and will re-file with updated/finished/fixed code soon.  Please don't use this code, it has a few issues.

pdaian:
@fubuloubu num256 is definitely not the same as num, nor is it overflow protected afaics.

Also, the payable decorator did not exist when I wrote this code ;).

DavidKnott:
@pdaian Are you still working on this?

pdaian:
@AlexXiong97 yup!  just pushed #366 let me know what you think!

pdaian:
(sorry I meant @DavidKnott ... long day :))

Issue #292, opened 2017-07-28
Need a .travis.yml file to properly run travis tests


vbuterin:
^

javimaravillas:
If it works for you, I will pick up this task during the next weekend.

PR #293, opened 2017-07-28
Add .travis.yml

beaugunderson:
You can verify that the [build was successful on my fork](https://travis-ci.org/beaugunderson/viper/builds/258661293).

Closes #292 .

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -20,7 +20,7 @@ setup(
     license=license,
     packages=find_packages(exclude=('tests', 'docs')),
     install_requires=[
-        'ethereum == 1.3.7',
+        'ethereum==2.0.4',
         'serpent',
         'pytest-cov',
         'pytest-runner', # Must be after pytest-cov or it will not work
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,13 @@
+language: python
+
+python:
+  - "3.6"
+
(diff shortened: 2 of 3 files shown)


beaugunderson:
Closes #293.

yograterol:
@beaugunderson great! Coud you add the travis integration with the README.md?

setup.py
```diff
@@ -20,7 +20,7 @@
     license=license,
     packages=find_packages(exclude=('tests', 'docs')),
     install_requires=[
-        'ethereum == 1.3.7',
+        'ethereum==2.0.4',
         'serpent',
```
yograterol: It would be good to add the version for every dependency, what do you think?

beaugunderson: Yes, I think that's a best practice; though `serpent` here is incorrect (it should be `ethereum-serpent`) and testing requirements should not be in `install_requires`, they should actually be in `tests_require`.

yograterol: We could to implement [Semantic Version](http://semver.org/) and freeze the dependencies version with every Viper version.

beaugunderson:
@yograterol OK, I added the Travis badge ✨

setup.py
beaugunderson: 👍 for using semver

PR #299, opened 2017-07-29
Simple company stock example

fubuloubu:
I wrote a very simple example


vbuterin:
Looks good, though I recommend one modification: change the type of `avail_shares` and `total_shares` from `num` to `currency_value`. Then change `price` from `wei_value` to `num (wei / currency)`. This gets the type system to do some of the work of ensuring the variables point to the right things and get combined in the right ways.

fubuloubu:
So, "currency_value" is a type that represents the quantity or "token/coin" represented by the contract? In my case, company shares.

Also, the syntax for representing that more complicated type is totally not obvious, why is it `avail_shares: currency_value` when it's a simple type, but `price: num (wei / currency)` when it's more complex (e.g. what happens to **`_value`**)? Why do I need to specify `num` as a basetype when composing a type, can't that be inferred?

vbuterin:
`currency_value` is an alias for `num(currency)`. You can also use `num(currency)`.

PR #301, opened 2017-07-31
Stock example

fubuloubu:
@vbuterin, made your suggested changes.


DavidKnott:
@fubuloubu Hey, it might be good if you rebased this (19 commits with 1 file changed).

fubuloubu:
Lulz, I made a mistake when I fooling around with branches on my laptop and it like double-counted my prior commits. There's really only 1 commit (b00d6fa) that matters, and it was to address comments @vbuterin made about types to the prior example that was already excepted. I can rebase it for convenience, but no updates have been made to that one file so it doesn't really do much to rebase it.

DavidKnott:
I see, makes sense

fubuloubu:
I'm working on removing those excess commits. I learned the hard way that GitHub doesn't track rebasing well...

fubuloubu:
Wow, that was a PITA. @DavidKnott, can you review my update. I made some more changes here, due to the realization that you don't have to initialize the value for an uninitialized key in a map.

fubuloubu:
Added tests

examples/stock/company.v.py
```diff
@@ -25,29 +27,23 @@ def buy_stock():
...
     # Take the shares off the market and give to stockholder
-    self.avail_shares -= buy_order
-    if self.holdings[msg.sender]:
-        # Account exists, so add to that account
-        self.holdings[msg.sender] += buy_order
-    else:
-        # Not currently a stockholder, so add them
-        self.holdings[msg.sender] = buy_order
+    self.holdings[self.company] -= buy_order
+    self.holdings[msg.sender] += buy_order
```
DavidKnott: Maybe it's too complex for this example but should partial buys be an option? e.g. I wanna buy 10 shares and there only 5 shares left.  As it is right now it'll just throw, it might be nice if I could try to buy 10 and get 5 with a half refund.

examples/stock/company.v.py
```diff
@@ -25,29 +27,23 @@ def buy_stock():
...
+    self.holdings[self.company] += sell_order
     send(msg.sender, sell_order * self.price)
 
-    # Put the stock back on the market
-    self.avail_shares += sell_order
-
 @constant
 def valuation() -> wei_value:
-    return (self.total_shares - self.avail_shares) * self.price
+    return (self.total_shares - self.holdings[self.company]) * self.price
```
DavidKnott: I could be wrong but doesn't valuation mean the total shares that exist time the price and not just the shares that the company isn't holding? (as in just `self.total_shares * self.price`)

examples/stock/test_company.py
```diff
@@ -0,0 +1,54 @@
...
+    # You can't sell more than you own
+    test_shares = tester.total_shares / 10
+    tester.c.buy_stock(sender=t.k1, value=int(test_shares * tester.share_price))
+    assert_tx_failed(tester, lambda: tester.c.sell_stock(tester.total_shares, sender=t.k1))
+
+def test_valuation(tester):
+    # Valuation is number of shares held times price
+    assert tester.c.valuation() == 0
+    tester.c.buy_stock(sender=t.k1, value=int(tester.total_shares * tester.share_price))
+    assert tester.c.valuation() == (tester.total_shares * tester.share_price)
```
DavidKnott: Not, that important but all your tests are done using `t.k1` .  I'd use another sender to just to make sure that the behavior's the same.

examples/stock/company.v.py
```diff
@@ -1,21 +1,23 @@
 # Own shares of a company!
 company: address
-avail_shares: num
-total_shares: num
-price: wei_value
+total_shares: currency_value
+price: num (wei / currency)
```
DavidKnott: I checked and public works the same as it does in solidity with variables (automatically creating getters).  I'd make `total_shares` and `price` public so the user can call `get_total_shares()` and `get_price()`.

fubuloubu: Ah thanks, I was wondering what that did

fubuloubu: Are those getters `@constant` e.g. cost zero gas? because the data is available to be read?

examples/stock/company.v.py
fubuloubu: True, and also this property is basically `self.balance`. Maybe I'll change the name to `worth()`

examples/stock/company.v.py
fubuloubu: The estimates are definitely non-zero for `@constant` functions, I guess I was mistaken to think they'd cost zero gas. In my head that made sense though, the blockchain data is publicly available and should in theory not cost anything to compute something that doesn't modify state. Is this a bug with the gas estimation?

examples/stock/test_company.py
fubuloubu: Addressing this in my next revision

examples/stock/company.v.py
DavidKnott: I checked it out and both `@constant` and `public` functions cost gas (every evm call does).  The issue is that the gas should only be taken from the sender when the state is changed.  Simply calling the blockchain costs gas but is free because it can be done locally (accessing 1 node instead of all of them)

fubuloubu: Every EVM call, or every Viper call?

Interesting, so some of the overestimating viper has is for checks that don't require state modification and therefore do not cost gas, is that correct?

PR #302, opened 2017-07-31
Create voting example

DavidKnott:
This is a voting example written in Viper (based on Solidity docs example).  I've added it to the docs locally but would love to get some eyes on the code to make sure it's right before making a PR for the docs.
@pdaian I used the same style for testing that you did, it definitely sped things up.  Thanks a lot!
@fubuloubu Could you take a look?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,10 +1,22 @@
-.PHONY: docs
+.PHONY: docsclean-pyc clean-build docs
 
 init:
 	python setup.py install
 
 test:
 	python setup.py test
+
+clean: clean-build clean-pyc
+
+clean-build:
+	rm -fr build/
+	rm -fr dist/
+	rm -fr *.egg-info
+
+clean-pyc:
+	find . -name '*.pyc' -exec rm -f {} +
+	find . -name '*.pyo' -exec rm -f {} +
+	find . -name '*~' -exec rm -f {} +
 
 docs:
 	rm -f docs/viper.rst
--- a/examples/voting/ballot.v.py
+++ b/examples/voting/ballot.v.py
@@ -1,0 +1,100 @@
+# Voting with delegation.
+
+# Information about voters
+voters: public({
+    # weight is accumulated by delegation
+    weight: num,
+    # if true, that person already voted
+    voted: bool,
+    # person delegated to
+    delegate: address,
+    # index of the voted proposal
+    vote: num
+}[address])
+
+# This is a type for a list of proposals.
+proposals: public({
+    # short name (up to 32 bytes)
+    name: bytes32,
+    # number of accumulated votes
+    vote_count: num
+}[num])
+
+voter_count: public(num)
+chairperson: public(address)
+
+# Setup global variables
+def __init__(_proposalNames: bytes32[5]):
+    self.chairperson = msg.sender
+    self.voter_count = 0
+    for i in range(5):
+        self.proposals[i] = {
+            name: _proposalNames[i],
+            vote_count: 0
+        }
+
+# Give `voter` the right to vote on this ballot.
+# May only be called by `chairperson`.
+def give_right_to_vote(voter: address):
+    # Throws if sender is not chairpers
+    assert msg.sender == self.chairperson
+    # Throws if voter has already voted
+    assert not self.voters[voter].voted
+    # Throws if voters voting weight isn't 0 
+    assert self.voters[voter].weight == 0
+    self.voters[voter].weight = 1
+    self.voter_count += 1
+
+# Delegate your vote to the voter `to`.
+def delegate(_to: address):
+    to = _to
+    # Throws if sender has already voted
+    assert not self.voters[msg.sender].voted 
+    # Throws if sender tries to delegate their vote to themselves
+    assert not msg.sender == to
+    # loop can delegate votes up to the current voter count
+    for i in range(self.voter_count, self.voter_count+1):
+        if self.voters[to].delegate:
+        # Because there are not while loops, use recursion to forward the delegation
+        # self.delegate(self.voters[to].delegate)
+            assert self.voters[to].delegate != msg.sender 
+            to = self.voters[to].delegate
+    self.voters[msg.sender].voted = True
+    self.voters[msg.sender].delegate = to
+    if self.voters[to].voted:
+        # If the delegate already voted,
+        # directly add to the number of votes
+        self.proposals[self.voters[to].vote].vote_count += self.voters[msg.sender].weight
+    else:
+        # If the delegate did not vote yet,
+        # add to her weight.
+        self.voters[to].weight += self.voters[msg.sender].weight
+
+# Give your vote (including votes delegated to you)
+# to proposal `proposals[proposal].name`.
+def vote(proposal: num):
+    assert not  self.voters[msg.sender].voted
+    self.voters[msg.sender].voted = True
+    self.voters[msg.sender].vote = proposal
+    # If `proposal` is out of the range of the array,
+    # this will throw automatically and revert all
+    # changes.
+    self.proposals[proposal].vote_count += self.voters[msg.sender].weight
+
+# Computes the winning proposal taking all
+# previous votes into account.
+@constant
+def winning_proposal() -> num:
+    winning_vote_count = 0
+    for i in range(5):
+        if self.proposals[i].vote_count > winning_vote_count:
+            winning_vote_count = self.proposals[i].vote_count
+            winning_proposal = i
+    return winning_proposal
+
+# Calls winning_proposal() function to get the index
+# of the winner contained in the proposals array and then
+# returns the name of the winner
+@constant
+def winner_name() -> bytes32:
+    return self.proposals[self.winning_proposal()].name
--- a/tests/examples/voting/test_ballot.py
+++ b/tests/examples/voting/test_ballot.py
@@ -1,0 +1,138 @@
+from ethereum.tools import tester
+import ethereum.utils as utils
+import ethereum.abi as abi
+
+def assert_tx_failed(ballot_tester, function_to_test, exception = tester.TransactionFailed):
+    """ Ensure that transaction fails, reverting state (to prevent gas exhaustion) """
+    initial_state = ballot_tester.s.snapshot()
+    ballot_tester.assertRaises(exception, function_to_test)
+    ballot_tester.s.revert(initial_state)
+
+class TestVoting(unittest.TestCase):
+    def setUp(self):
+        # Initialize tester, contract and expose relevant objects
+        self.t = tester
+        self.s = self.t.Chain()
+        from viper import compiler
+        self.t.languages['viper'] = compiler.Compiler()
+        contract_code = open('examples/voting/ballot.v.py').read()
+        self.c = self.s.contract(contract_code, language='viper', args=[["Clinton", "Trump"]])
+
+    def test_initial_state(self):
+        # Check chairperson is msg.sender
+        self.assertEqual(utils.remove_0x_head(self.c.get_chairperson()), self.t.a0.hex())
+        # Check propsal names are correct
+        self.assertEqual(self.c.get_proposals__name(0), b'Clinton\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00')
+        self.assertEqual(self.c.get_proposals__name(1), b'Trump\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00')
+        # Check proposal vote_count is 0
+        self.assertEqual(self.c.get_proposals__vote_count(0), 0)
+        self.assertEqual(self.c.get_proposals__vote_count(1), 0)
+        # Check voter_count is 0
+        self.assertEqual(self.c.get_voter_count(), 0)
+        # Check voter starts empty
+        self.assertEqual(self.c.get_voters__delegate(), '0x0000000000000000000000000000000000000000')
+        self.assertEqual(self.c.get_voters__vote(), 0)
+        self.assertEqual(self.c.get_voters__voted(), False)
+        self.assertEqual(self.c.get_voters__weight(), 0)
+
+    def test_give_the_right_to_vote(self):
+        self.c.give_right_to_vote(self.t.a1)
+        # Check voter given right has weight of 1
+        self.assertEqual(self.c.get_voters__weight(self.t.a1), 1)
+        # Check no other voter attributes have changed
+        self.assertEqual(self.c.get_voters__delegate(self.t.a1), '0x0000000000000000000000000000000000000000')
+        self.assertEqual(self.c.get_voters__vote(self.t.a1), 0)
+        self.assertEqual(self.c.get_voters__voted(self.t.a1), False)
+        # Chairperson can give themselves the right to vote
+        self.c.give_right_to_vote(self.t.a0)
+        # Check chairperson has weight of 1
+        self.assertEqual(self.c.get_voters__weight(self.t.a0), 1)
+        # Check voter_acount is 2
+        self.assertEqual(self.c.get_voter_count(), 2)
+        # Check several giving rights to vote
+        self.c.give_right_to_vote(self.t.a2)
+        self.c.give_right_to_vote(self.t.a3)
+        self.c.give_right_to_vote(self.t.a4)
+        self.c.give_right_to_vote(self.t.a5)
+        # Check voter_acount is now 6
+        self.assertEqual(self.c.get_voter_count(), 6)
+        # Check chairperson cannot give the right to vote twice to the same voter
+        assert_tx_failed(self, lambda: self.c.give_right_to_vote(self.t.a5))
+        # Check voters weight didn't change
+        self.assertEqual(self.c.get_voters__weight(self.t.a5), 1)
+
+    def test_delegate(self):
+        self.c.give_right_to_vote(self.t.a0)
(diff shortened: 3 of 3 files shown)


fubuloubu:
I was just making my own voting example lol.

Personally, multiple AND'd assertion statements I like to break out into multiple assertion statements for clarity. Functionally identical though.

Also, for testing I was thinking about integrating the testing of these examples into the testing folder? This way these more complex examples are verified in CI to pass and we might get some coverage boost.
Good idea / bad idea?

`tests/` uses pytest instead of unittest right now though

DavidKnott:
@fubuloubu I like the idea of putting all the tests into the tests dir, I'll move my tests later today.

fubuloubu:
To integrate with pytest, I believe you just need to remove the class, indent the contents to the left, remove all the selfs, and unencapsulate the setup function so it's executed before any tests

DavidKnott:
@DustinWehr 
I just ran this test case and it passed
```
    def test_complex_delegation(self):
        self.c.give_right_to_vote(self.t.a0)
        self.c.give_right_to_vote(self.t.a1)
        self.c.give_right_to_vote(self.t.a2)
        self.c.give_right_to_vote(self.t.a3)
        self.c.delegate(self.t.a2, sender=self.t.k1)
        self.c.delegate(self.t.a3, sender=self.t.k2)
        self.c.vote(0, sender=self.t.k3)
        self.assertEqual(self.c.get_proposals__vote_count(0), 3)
        self.c.delegate(self.t.a1, sender=self.t.k0)
        self.assertEqual(self.c.get_proposals__vote_count(0), 4)
```
Did I correctly implement the situation you described?  If not I'd love to figure out how, as I'm not 100% sure what's wrong with the loop (it's currently carrying each delegation up the chain, adding weight or to a vote count at the end).

DustinWehr:
Considering I'm having trouble understanding your test case, probably there's nothing wrong and I was just being an overconfident newb. What is the relationship between the `k` and `a` instance variables? I'm going to assume `k1` and `a1` both correspond to voter 1, etc. If that's basically right, then your test case seems to correctly implements the situation I described:
```
voter 1 delegates to voter 2
voter 2 delegates to voter 3
voter 3 votes for proposal 0
Now there are 3 votes for proposal 0
voter 0 delegates to voter 1
Now there are 4 votes for proposal 0
```
However, I still don't see why the code works in that situation! 

Doesn't the body of the `for` loop execute only once? And the variable `i` isn't mentioned in the body. So why not just replace the `for` loop with its body?

DustinWehr:
Just had another look. Try changing your test case so that voter 3 votes for proposal 1, rather than 0. I think the only reason the test pasts now is that `.vote` is initialized to 0.

DustinWehr:
I wrote another version in pseudocode. It's a bit more complicated. Using pattern matching on your code to convert it to viper.

DustinWehr:
I have the new version ready, and I believe I've updated your tests, but `python setup.py test` still crashes for me. I will try reinstalling.

DavidKnott:
The test still passes when voting for proposal 1, t.a1 and t.k1 correspond with accounts (address and private key).  The for loop runs based on however many voters there are (usually more then once), running up the chain of delegation with this line.
```
to = self.voters[to].delegate
```
These lines are what add to a proposals votes if the final delegate has already voted:
```
 if self.voters[to].voted:
# If the delegate already voted,
# directly add to the number of votes
       self.proposals[self.voters[to].vote].vote_count += self.voters[msg.sender].weight
```
Also, if you keep having troubles with getting things working (`python setup.py test` etc) I'd be happy to pair.

DustinWehr:
How can this possibly run more than once?
> for i in range(self.voter_count, self.voter_count+1):

The will be like `range(10,11)` at run time, no?

DustinWehr:
If `range` is inclusive on both ends, you'd have to add a 5th voter to make it fail.

Let vi[W,D] represent "voter i's weight is W and delegate is D". And ai is voter i's address. I'll give each action, followed by the (relevant part of the) state that I believe the contract is in after action.

Initial state
v0[1,0] v1[1,0] v2[1,0] v3[1,0] v4[1,0]
v1 delegates to v2
v0[1,0] v1[0,a2] v2[2,0] v3[1,0] v4[1,0]
v2 delegates to v3
v0[1,0] v1[0,a2] v2[0,a3] v3[3,0] v4[1,0]
v3 delegates to v4
v0[1,0] v1[0,a2] v2[0,a3] v3[0,a4] v4[4,0]
v0 delegates to v1
If `range` is exclusive of the second argument, then the state would be
v0[0,a2] v1[0,a2] v2[1,a3] v3[0,a4] v4[4,0] 
If `range` is inclusive of the second argument, then the state would be
v0[0,a3] v1[0,a2] v2[0,a3] v3[1,a4] v4[4,0]

DustinWehr:
In general, imagine a long sequence of delegations
v1 -> v2
v2 -> v3
...
v999 -> v1000
At this point, the contract has the correct weight of 1000 for v1000. It has explicitly recorded that v1 delegates to v2, ..., v999 delegates to v1000. It does **NOT** have explicitly recorded that v0 indirectly delegates to v1000.
Now v1000 votes for proposal 1 No surprises there.
Now v0 delegates to v1. 

_Whatever your `delegate` function is doing_, it must have bounded gas cost (if I understand that design criteria of Viper correctly). I claim there is no way that `delegate` can compute that v0's vote needs to go to the proposal that v1000 votes for, unless it uses on the order of 1000 operations, in which case its gas cost is linear in the number of voters.

Can we get a second opinion? @fubuloubu?

Here is a gist with, I believe, a fix to the basic algorithmic problem (though it probably still has other bugs). https://gist.github.com/DustinWehr/b4f3e0d3b1d29a55f4c3042e0485bd69

fubuloubu:
@DavidKnott you should split the Makefile updates to a separate PR

fubuloubu:
I'm not sure I fully understand what @DavidKnott was attempting in `delegate()` in my brief review but I agree it looks sorta fishy. Why not only allow delegating to someone who hasn't voted yet, and allocating all your votes (including delegated ones from others) to the new delegate should avoid any problems requiring recursive calls or loops. You can track the delegation tree for posterity to show who delegated to who. That suggestion is sort of a linear, one-way operation and analysis of the tree can happen off-chain.

fubuloubu:
As an aside, I've been sort of learning that it is better to have push-based ("poke-based"?) operations requiring participants in a contract to do typical array/loop operations, and that the limited utility of the looping contructs in viper is actually a really good thing making you rethink in a more distributed sense of how do perform some operation at scale.

DustinWehr:
@fubuloubu Hmm I think I see what you mean by tracking the delegation tree off chain, but that is putting a significant part of the code into javascript or whatever. Is that normal? Check out the approach I took in the gist. It might be what you mean by "push-based ("poke-based"?) operations requiring participants in a contract to do typical array/loop operations"? It adds a function `shorten_delegation_chain` that anyone can call to make the contract make progress.

DustinWehr:
I'm writing a version now that can break cycles.

fubuloubu:
It's more like "what is the basic operation I am trying to do here" and tracking who delegated to who doesn't actually help with voting, so do that off chain (if you're even interested in that information). Like, make sure the information to compute more complicated results is available (keeping `voter[].delegate`) but it doesn't need to re-track the changes in delegation over time to accurately track the overall voting strategy. Don't make a simple thing complicated lol.

This is how I would rewrite that function:
```python
# Delegate your vote to the voter `to`.
def delegate(_to: address):
    # Throws if sender has already voted
    assert not self.voters[msg.sender].voted 
    # Throws if sender tries to delegate their vote to themselves
    assert not msg.sender == to
    # Throws if delegate has already voted
    assert not self.voters[to].voted 

    self.voters[to].weight += self.voters[msg.sender].weight
    self.voters[msg.sender].voted = True
    self.voters[msg.sender].delegate = to # Track for off-chain analysis
```

DustinWehr:
But then why even bother with `.weight`?
The other issue, which the solidity contract doesn't have, is that with that implementation there is no way to prevent or recover from delegation cycles.

fubuloubu:
`.weight` is necessary to ensure that your vote(s) get delegated correctly overall. That is like the most necessary part lol.

There should be no delegation cycles here because when you delegate that counts as you voting, so you can only do it once (and only to someone who hasn't voted). It is impossible to create a cycle in this scenario.

DustinWehr:
You don't need `.weight` since you have `.vote` and `.delegate`.

Yes I see why there can't be delegation cycles. Again, though, not being able to delegate to someone who's already voted is a limitation that the solidity contract doesn't have.

DustinWehr:
Ok, I think I see that with your restriction of only being able to delegate to someone who hasn't voted, then you can still compute the winner on chain if you use weight.

DustinWehr:
Version with cycle breaking https://gist.github.com/DustinWehr/d178b3a0b50dcb217a26ad1f26be82e1

fubuloubu:
Gotcha. You acn use `.weight` like that for analytical purposes as you suggested, although probably not necessary as you suggest. This is what I meant by it's easier to do "poke-based" operations.

A modification to the above that could allow those who have voted to be come delegates is that if you set `.weight` to zero when you delegate and vote, then reset voted if you are delegated to after you've voted/delegated. Still avoids on-chain delegation cycles (you could theoretically cycle off-chain, but I don't care if you're wasting your own compute cycles or creating more transactions for no reason lol)

DavidKnott:
As this is an example I'd like to keep it as simple as possible and agree that the linear gas costs are not scalable.  @DustinWehr I looked into the viper for loop source code and think you're correct.  So far I think the simplest solution is the one @fubuloubu made a few minutes ago (basically taking out the delegate chain functionality).  What do you guys think?

DustinWehr:
@fubuloubu that's an improvement. But what about delegating to someone who's delegated to someone who's voted?
@DavidKnott I could provide a prose proof of correctness.

DustinWehr:
How about both? One that's simple, but imposes limitations that the Solidity version doesn't, and the other that's more complicated?

fubuloubu:
@DustinWehr in my suggestion, if you've delegated your votes then it counts as you voting. Someone could delegate to you and then you can vote again, and you would have to create another transaction to delegate your votes to that same person if you wanted to. This could create a cycle of you delegating to the same person for each vote that is delegated to you, but that is happening off chain so it's not the contract's fault someone keeps delegating to you after you've already delegated. Definitely a flaw that could be exploited, but a simple method to allow re-delegations.

Alternatively in that scenario, you could just see who the person you're delegating to had delegated their votes to (assuming that information is visible) and just delegate to them. I sort of don't understand why that costs gas to compute because you're doing the work of recursion to find out who's at the end of that chain, then creating a transaction at that point. I want to believe it is only necessary to charge gas for that last transaction because that is all that matters for submitting the transaction (the final person in the chain). That is my opinion though, I don't fully understand it to be able to say for sure.

fubuloubu:
Further thoughts, `constant` functions cost no gas, so if you could create such a function that looped through and found the correct person to delegate to, that would be cool. Just gave me the thought that perhaps you can use additional looping constructs but only if the function is `constant`. Food for thought.

DustinWehr:
That sounds like a very good idea.
> Further thoughts, constant functions cost no gas, so if you could create such a function that looped through and found the correct person to delegate to, that would be cool. Just gave me the thought that perhaps you can use additional looping constructs but only if the function is constant. Food for thought.

DustinWehr:
I like this approach too, when combined with stripping the contract to its bare minimum by eliminating `.weight` also. It's not the same as the Solidity contract though, and requires an argument that it's just as useful (or better).
> Alternatively in that scenario, you could just see who the person you're delegating to had delegated their votes to (assuming that information is visible) and just delegate to them. I sort of don't understand why that costs gas to compute because you're doing the work of recursion to find out who's at the end of that chain, then creating a transaction at that point. I want to believe it is only necessary to charge gas for that last transaction because that is all that matters for submitting the transaction (the final person in the chain). That is my opinion though, I don't fully understand it to be able to say for sure.

DustinWehr:
However, we would need to warn programmers about race conditions if they use that approach.

DustinWehr:
I think I like the idea of allowing less-restricted loops in constant functions the best. We can even still prevent infinite looping. Hell we could even ensure polynomial run time of constant functions:

`for i in range(F(‹size of some collection))` where F is restricted to be a polynomial.

DavidKnott:
@DustinWehr Less restricted for loops in constant functions seems like a great viper feature (should improving functionality) while not decreasing security (out of gas errors in particularly).  What do you mean by race conditions in this context?

DustinWehr:
Sorry by "that approach" I wasn't talking about the less restricted for loops in constant functions idea. I was talking about the approach of more-or-less moving everything that's not strictly necessary off chain. It wouldn't be a problem in the voting example, since once someone votes they can't change their vote.

DavidKnott:
That makes sense, thanks for clarifying

DustinWehr:
@fubuloubu it's your idea. Make an issue for further discussion?

DustinWehr:
@DavidKnott in the meantime I think we could make ballot2.v.py (the one without cycle breaking) simple enough to be a good example.

fubuloubu:
The only race condition I could perceive is delegating at the same time someone is voting. In that situation, either the delegation is rejected or the voting is rejected. Under rejection, the network will coalesce around one scenario or the other, so I don't think there is any particular race situation that loses data, only gas. Is that correct thinking @DavidKnott?

DustinWehr:
I wasn't suggesting a race condition for the voting alg, just for the idea of gathering information off chain (not to be confused with the idea of using constant functions... hopefully I'm using "off chain" correctly).
> It wouldn't be a problem in the voting example, since once someone votes they can't change their vote.

fubuloubu:
@DustinWehr I was thinking about it some more, and I believe gas would get consumed the loop if you called a `@constant` function with a loop inside a transaction call. It provides a good method for providing code in a contract for free only if you are not using it elsewhere in the contract, e.g. you run `get_delegate(to: address)` on some starting address to retrieve a final address for your delegation (based on the tree), and then separately delegate to that person on-chain (so that calculation never occurs on-chain). Definitely more complicated than I previously assumed.

fubuloubu:
@DustinWehr, yes totally but as I was stating there is no data loss scenario so you should just make sure your data is up-to-date and consistent, and perhaps externally discuss with that address owner your intentions to make sure your delegation is accepted.

DustinWehr:
@fubuloubu are you going to flesh out your approach to fixing the voting example?

DustinWehr:
About gas being used if you called a `@constant` function inside your transaction call: I was imagining the ledger would just contain an operation saying which constant function was called, and with what inputs, and the output. Then it would be a simple matter for miners to verify that other miners used the correct function output. But I'm still quite new to this. What am I missing?

fubuloubu:
Well, imagine that you used the result of that `@constant` function to modify the underlying data structure, now it doesn't come up with the same result anymore. IIRC you're basically asking everyone to do the same full calculation to mark it consistent, therefore gas is used for that reason. Viper compiles to EVM bytecode etc. so it wouldn't be off-chain unless there was no usage of that function inside the main transaction pieces of the contract.

fubuloubu:
@DustinWehr I was more doing thought experiments. Is there a particular idea here you'd like me to write some code for? I think the one rewrite I did above is what @DavidKnott will do to simplify his example. I don't think we need more than one simple voting example to demonstrate viper's capabilities.

DustinWehr:
I thought @constant functions have to be pure? D

fubuloubu:
They are, in that with the same input it produces the same output. The problem is that contracts contain mutable data by definition, so that same function could produce different results if it makes use of the global data associated with the contract to compute its result.

DustinWehr:
I don't see the problem. The state of the contract's mutable data is verified before the function call.

fubuloubu:
Other's need to use that same function to do the verification. I'm not positive about it, this is just my understanding.

DustinWehr:
@DavidKnott can you set us straight? ;-)

fubuloubu:
@DavidKnott to clarify, do `@constant` functions ever cost gas if used inside another function?

DavidKnott:
@fubuloubu You're correct `@constant`  functions called from other functions do cost gas because of the reasons you stated above (the `@constant` function cannot be run locally because it's output leads to a change in state.  As far as I can tell the way to use `@constant` here would be to use it as a helper function, getting the final delegate and then enter it into the original delegate function (this seems suboptimal due to it's off-chain reliance).  As for as I understand `@constant` functions still cost gas, it's only when they're run locally that they're free.

fubuloubu:
Alright, thanks for confirming my understanding. Although I think you misspoke a little because isn't the point of `@constant` that it does not change state?

DavidKnott:
I meant `@constant` itself doesn't change state but it leads to a state change if it's called from a non-constant function (causing a cost in gas).

DustinWehr:
Just to check my understanding: this is an (in principle) improvable
deficiency in the EVM right?

fubuloubu:
It wouldn't change state when called from another function either, but the results would be used for that purpose. My understanding is that it costs gas to run the `@constant` function when confirming the state for the calling function which is actually making state changes.

Minor technicality, I just want to make sure I'm not misunderstanding something.

DustinWehr:
For sure you need gas to confirm the state for the calling function. I thought we were inquiring about whether the operations that happen _inside_ the `@constant` function cost gas. My read of @DavidKnott's answer is yes.

DavidKnott:
Just reconfirming, it's a yes.

fubuloubu:
Responding to this: "... getting the final delegate and then enter it into the original delegate function (this seems suboptimal due to it's off-chain reliance)"

@DavidKnott the client (or test case) could make immediate use of this value e.g. `delegate(get_delegate(to))` in order to get the final delegate from the tree to provide to the transaction. I don't think this is much extra work for the client. It would be cool to specify like `@local` or something where the value, when called in the function, is computed locally and provided to a function via call-by-value for the actual state change, if that makes sense.

PR #305, opened 2017-08-02
Fixes #303 Change error message in Map type in functions

### LINKED ISSUE #303: Array of bytes has infinite type size
I am pretty sure this is an error, although I am a bit unclear behind the purpose of declaring the size of a type to be infinite. Is this a part of the gas estimate?

Anyways, if you define an array of bytes using another type, it throws an error e.g.:
```python
def __init__(test: bytes32[bool]):
    pass
```

even though that data size should only be 64 bytes max (32 bytes x 2 possible values for bool)

Would appreciate some understanding behind the purpose of this check and if there is supposed to be a deeper check occuring, like some sort of maximum (or reasonable) number of bytes we should be allowed to use? (distinct from their economic cost to store them)

yograterol:
I like the idea of https://github.com/ethereum/viper/issues/303 to change the error copy.

### DIFF
--- a/viper/types.py
+++ b/viper/types.py
@@ -310,7 +310,7 @@ def get_size_of_type(typ):
     elif isinstance(typ, ListType):
         return get_size_of_type(typ.subtype) * typ.count
     elif isinstance(typ, MappingType):
-        raise Exception("Type size infinite!")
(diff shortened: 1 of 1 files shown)


fubuloubu:
Perfect!

PR #307, opened 2017-08-02
Coverage testing

fubuloubu:
Moved some of the utility code and imports to their own setup module to break apart that stuff from the tests, and also so I can use the same setup in other tests. Working on other test modules...

### DIFF
--- a/tests/test_parser.py
+++ b/tests/test_parser.py
@@ -1563,8 +1563,8 @@ def foo(x: bytes <= 32) -> num:
         success = False
     print('Passed bytes_to_num tests')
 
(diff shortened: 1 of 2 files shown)


tests/test_parser.py
```diff
@@ -1606,6 +1563,7 @@ def foo(x: bytes <= 32) -> num:
         success = False
     print('Passed bytes_to_num tests')
 
+import rlp
```
yograterol: This import could be a local import within `test_rlp_decoder_code` function.

fubuloubu:
Made that update

Issue #306, opened 2017-08-02
Get complete listing of gas estimates for contract


fubuloubu:
Some way to get a gas estimate for each function in a contract would be sweet, allowing you to profile the gas-intensive functions you might have but are unaware of and also see it in client programs/editors.

@vbuterin Can this be included in the ABI output? Does that make sense? I am not familiar with the requirements or uses of that format enough

vbuterin:
Yeah, I think it would be sensible to add the gas estimate into the ABI, alongside the other fields.

fubuloubu:
Easier than I thought, #308

Feel free to change the string used to store it in the ABI if you so choose.

PR #308, opened 2017-08-03
Modification of abi format to include gas cost estimate for functions…

fubuloubu:
… (skips constructor)

### DIFF
--- a/viper/compiler.py
+++ b/viper/compiler.py
@@ -12,11 +12,6 @@ def compile(code, *args, **kwargs):
     return compile_lll.assembly_to_evm(compile_lll.compile_to_assembly(lll))
 
 
-def mk_full_signature(code, *args, **kwargs):
-    o = parser.mk_full_signature(parser.parse(code))
-    return o
-
-
 def gas_estimate(origcode, *args, **kwargs):
     o = {}
     code = optimizer.optimize(parser.parse_to_lll(origcode))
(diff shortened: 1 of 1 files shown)


viper/compiler.py
```diff
@@ -32,6 +27,19 @@ def gas_estimate(origcode, *args, **kwargs):
     return o
 
 
+def mk_full_signature(code, *args, **kwargs):
+    abi = parser.mk_full_signature(parser.parse(code))
+    # Add gas estimates for each function to ABI
+    gas_estimates = gas_estimate(code)
+    for idx, func in enumerate(abi):
+        func_name = func['name'].split('(')[0]
+        print(func_name)
```
fubuloubu: Duh, fixed. Thanks!

fubuloubu:
Had to reorder the commits because I tried re-basing this branch and made a mistake.

fubuloubu:
@vbuterin see any problems with this?

fubuloubu:
@DavidKnott What do you think of this update? Is there a potential problem with adding a field to the ABI output?

DavidKnott:
I don't think so, I'll look into it

DavidKnott:
@fubuloubu I think as of now gas estimation isn't accurate enough to be included in the abi.  Right now gas estimation fails on quite a few the tests in `test_parser.py`.  I think gas estimation functionality needs to be improved before this gets merged.

fubuloubu:
I think it would be easier to debug if that information was actually available in the ABI as it at least shows a bigger picture, and also if the gas estimation feature were working correctly then this update still works the same way and provides those more accurate updates in the same way. I don't believe this update creates any new bugs as it is fairly simple and unobtrusive, and #306 shows it is a useful feature to have in the long term.

This is alpha software after all, no guarantees on correct output just yet.

DavidKnott:
@fubuloubu Looked at this again and I think you're right, merging it in.

Issue #297, opened 2017-07-29
What should the syntax for (i) ABI-formatted logs, and (ii) calls to external contracts be?


vbuterin:
For ABI-formatted logs, we'd need to have syntax for both event type declaration and making the actual logs. For event type declaration, one option is for the parser to expect a list of declarations between storage variable declarations and function declarations that look like this:

```
MyLog: __log__(arg_1: num, arg2: indexed(bytes32), arg3: timestamp, arg4: bytes <= 100)
```
For the actual logging, I personally highly disagree with Solidity's syntax of making event logs and function calls syntactically the same; it goes against Viper's goals of making it hard to write misleading code.

One option is to copy Serpent syntax:

```
log(type=MyLog, 5, sha3("cow"), block.timestamp, "moose")
```

Another is:

```
log.MyLog(5, sha3("cow"), block.timestamp, "moose")
```

For external contract calls, we already have `raw_call`, but it would be good to have something ABI-formatted. We have the ABI formatting call in place already for self-calls, but to extend that to external calls we need a syntax for declaring external contract types.

One option is for the parser to expect another list of declarations, like:

```
class Foo():
    def bar(arg1: num, arg2: bytes32): pass
    @constant
    def baz(arg1: bytes <= 100, arg2: num256): pass
```

Another is to just directly pass in JSON ABI:

```
Foo = extern([{"constant": false, "type": "function", "name": "bar(int128,bytes32)", "outputs": [], "inputs": [{"type": "int128", "name": "arg1"}, {"type": "bytes32", "name": "arg2"}]}, {"constant": true, "type": "function", "name": "baz(bytes,uint256)", "outputs": [], "inputs": [{"type": "bytes", "name": "arg1"}, {"type": "uint256", "name": "arg2"}]}])
```

Then, one could declare anything from then on as having type Foo (we'd require contract types to start with a capitalized letter).

Any other ideas?

yograterol:
We could create a reservated variable, by example "Logs" into it variable we can set a dict with every Event.

```python
Logs: {
    Log1: __log__(arg_1: num),
    Log2: __log__(arg_1: num)
}
```

And we call:

```
Logs.Log1(1)
```

yograterol:
About of external contracts, I vote to ABI JSON directly.

jarradh:
Reading the pythonic ABI class definitions is just, well, it's a pleasure. Having that said - JSON ABI is probably preferred because of interoperability with other developer tools. Still thinking about logs

[...]

PR #309, opened 2017-08-04
Create open auction example

DavidKnott:
This is an open auction example written in Viper (based on Solidity docs example).  Once again, I've added it to the docs locally but would love to get some eyes on the code to make sure it's right before making a PR for the docs.
@fubuloubu Would you mind taking a look at this as well?  (I used pytest from the start this time!) Also I'd be more than happy to review any of your Viper code, just tag me (a second set of eyes always helps)

### DIFF
--- a/examples/auctions/simple_open_auction.v.py
+++ b/examples/auctions/simple_open_auction.v.py
@@ -1,0 +1,67 @@
+# Open Auction
+
+# Auction params
+# Beneficiary recieves money from the highest bidder
+beneficiary: public(address)
+auction_start: public(timestamp)
+auction_end: public(timestamp)
+
+# Current state of auction
+highest_bidder: public(address)
+highest_bid: public(wei_value)
+
+# Set to true at the end, disallows any change
+ended: public(bool)
+
+# Create a simple auction with `_bidding_time`
+# seconds bidding time on behalf of the
+# beneficiary address `_beneficiary`.
+def __init__(_beneficiary: address, _bidding_time: timedelta):
+    self.beneficiary = _beneficiary
+    self.auction_start = block.timestamp
+    self.auction_end = self.auction_start + _bidding_time
+
+# Bid on the auction with the value sent
+# together with this transaction.
+# The value will only be refunded if the
+# auction is not won.
+@payable
+def bid():
+    # Check if bidding period is over.
+    assert block.timestamp < self.auction_end
+    # Check if bid is high enough
+    assert msg.value > self.highest_bid
+    if not self.highest_bid == 0:
+        # Sends money back to the previous highest bidder
+        send(self.highest_bidder,self.highest_bid)
+    self.highest_bidder = msg.sender
+    self.highest_bid = msg.value
+
+
+# End the auction and send the highest bid
+# to the beneficiary.
+def auction_end():
+    # It is a good guideline to structure functions that interact
+    # with other contracts (i.e. they call functions or send Ether)
+    # into three phases:
+    # 1. checking conditions
+    # 2. performing actions (potentially changing conditions)
+    # 3. interacting with other contracts
+    # If these phases are mixed up, the other contract could call
+    # back into the current contract and modify the state or cause
+    # effects (Ether payout) to be performed multiple times.
+    # If functions called internally include interaction with external
+    # contracts, they also have to be considered interaction with
+    # external contracts.
+
+    # 1. Conditions
+    # Check if auction endtime has been reached
+    assert block.timestamp >= self.auction_end
+    # Check if this function has already been called
+    assert not self.ended
+
+    # 2. Effects
+    self.ended = True
+
+    # 3. Interaction
+    send(self.beneficiary, self.highest_bid)
--- a/tests/examples/auctions/test_simple_open_auction.py
+++ b/tests/examples/auctions/test_simple_open_auction.py
@@ -1,0 +1,86 @@
+import pytest
+from ethereum.tools import tester
+import ethereum.utils as utils
+
+FIVE_DAYS = 432000
+
+@pytest.fixture
+def auction_tester():
+    t = tester
+    tester.s = t.Chain()
+    from viper import compiler
+    t.languages['viper'] = compiler.Compiler()
+    contract_code = open('examples/auctions/simple_open_auction.v.py').read()
+    tester.c = tester.s.contract(contract_code, language='viper', args=[tester.accounts[0], FIVE_DAYS])
+    return tester
+
+@pytest.fixture
+def assert_tx_failed():
+    def assert_tx_failed(tester, function_to_test, exception = tester.TransactionFailed):
+        initial_state = tester.s.snapshot()
+        with pytest.raises(exception):
+            function_to_test()
+        tester.s.revert(initial_state)
+    return assert_tx_failed
+
+
+def test_initial_state(auction_tester):
+    # Check beneficiary is correct
+    assert utils.remove_0x_head(auction_tester.c.get_beneficiary()) == tester.accounts[0].hex()
+    # Check bidding time is 5 days
+    assert auction_tester.c.get_auction_end() == auction_tester.s.head_state.timestamp + 432000
+    # Check start time is current block timestamp
+    assert auction_tester.c.get_auction_start() == auction_tester.s.head_state.timestamp
+    # Check auction has not ended
+    assert auction_tester.c.get_ended() == False
+    # Check highest bidder is empty
+    assert auction_tester.c.get_highest_bidder() == '0x0000000000000000000000000000000000000000'
+    # Check highest bid is 0
+    assert auction_tester.c.get_highest_bid() == 0
+
+
+def test_bid(auction_tester, assert_tx_failed):
+    # Bidder cannot bid 0
+    assert_tx_failed(auction_tester, lambda: auction_tester.c.bid(value=0, sender=auction_tester.k1))
+    # Bidder can bid
+    auction_tester.c.bid(value=1, sender=tester.k1)
+    # Check that higest bidder and highest bid have changed accordingly
+    assert utils.remove_0x_head(auction_tester.c.get_highest_bidder()) == auction_tester.accounts[1].hex()
+    assert auction_tester.c.get_highest_bid() == 1
+    # Bidder bid cannot equal current highest bid
+    assert_tx_failed(auction_tester, lambda: auction_tester.c.bid(value=1, sender=auction_tester.k1))
+    # Higher bid can replace current highest bid
+    auction_tester.c.bid(value=2, sender=tester.k2)
+    # Check that higest bidder and highest bid have changed accordingly
+    assert utils.remove_0x_head(auction_tester.c.get_highest_bidder()) == auction_tester.accounts[2].hex()
+    assert auction_tester.c.get_highest_bid() == 2
+    # Multiple bidders can bid
+    auction_tester.c.bid(value=3, sender=tester.k3)
+    auction_tester.c.bid(value=4, sender=tester.k4)
+    auction_tester.c.bid(value=5, sender=tester.k5)
+    # Check that higest bidder and highest bid have changed accordingly
+    assert utils.remove_0x_head(auction_tester.c.get_highest_bidder()) == auction_tester.accounts[5].hex()
+    assert auction_tester.c.get_highest_bid() == 5
+    auction_tester.c.bid(value=1 * 10**10, sender=tester.k1)
+    balance_before_out_bid = auction_tester.s.head_state.get_balance(auction_tester.accounts[1])
+    auction_tester.c.bid(value=2 * 10**10, sender=tester.k2)
+    balance_after_out_bid = auction_tester.s.head_state.get_balance(auction_tester.accounts[1])
+    # Account has more money after its bid is out bid
+    assert balance_after_out_bid > balance_before_out_bid
+
+
+def test_auction_end(auction_tester, assert_tx_failed):
+    # Fails if auction end time has not been reached
+    assert_tx_failed(auction_tester, lambda: auction_tester.c.auction_end())
+    auction_tester.c.bid(value=1 * 10**10, sender=tester.k2)
+    # Move block timestamp foreward to reach auction end time
+    auction_tester.s.head_state.timestamp += FIVE_DAYS
+    balance_before_end = auction_tester.s.head_state.get_balance(auction_tester.accounts[0])
+    auction_tester.c.auction_end(sender=tester.k2)
+    balance_after_end = auction_tester.s.head_state.get_balance(auction_tester.accounts[0])
+    # Beneficiary receives the highest bid
+    assert balance_after_end == balance_before_end + 1 * 10 ** 10
+    # Bidder cannot bid after auction end time has been reached
+    assert_tx_failed(auction_tester, lambda: auction_tester.c.bid(value=10, sender=auction_tester.k1))
+    # Auction cannot be ended twice
+    assert_tx_failed(auction_tester, lambda: auction_tester.c.auction_end())


examples/auctions/simple_open_auction.v.py
```diff
@@ -0,0 +1,76 @@
...
+    self.beneficiary = _beneficiary
+    self.bidding_time = _bidding_time
+    self.auction_start = block.timestamp
+
+# Bid on the auction with the value sent
+# together with this transaction.
+# The value will only be refunded if the
+# auction is not won.
+@payable
+def bid():
```
fubuloubu: If someone had bid previously, and they were overbid but decided to bid again, they now carry an extra balance in `pending_returns` in addition to what they are now bidding. Can you ADD the value sent in the bid (e.g. new bid = old bid + msg.value, asserting that new bid is over the highest bidding amount and removing the old bid from `pending_returns`) so the users wouldn't have to tie up an exponentially increasing amount of wei depending on the amount of bids there are, only to have that excess wei returned when the auction is over? Do you follow what I am saying?

examples/auctions/simple_open_auction.v.py
```diff
@@ -0,0 +1,76 @@
...
+    # Check if auction endtime has been reached
+    assert block.timestamp >= self.auction_start + self.bidding_time
+    # Check if this function has already been called
+    assert not self.ended
+
+    # 2. Effects
+    self.ended = True
+
+    # 3. Interaction
+    send(self.beneficiary, self.highest_bid)
```
fubuloubu: This is in effect sending the bidder back what their bid was, instead of transferring ownership of the bid item to the highest bidder. You would have to have some mechanism to assign ownership to an item that is being bid on, e.g. a contract owner or something? Unclear what this might mean in practice.

fubuloubu: You should automatically refund everyone's money who didn't win the auction at this point, and perhaps consider removing the `withdraw` function above (somewhat in conjunction with my comment about being able to re-use funds that were outbid)

examples/auctions/simple_open_auction.v.py
```diff
@@ -0,0 +1,76 @@
...
+    # Check if bid is high enough
+    assert msg.value > self.highest_bid
+    if not self.highest_bid == 0:
+        # Allows the previous highest bidder to take their money back
+        self.pending_returns[self.highest_bidder] += self.highest_bid
+    self.highest_bidder = msg.sender
+    self.highest_bid = msg.value
+
+# Withdraw a bid that was overbid
+def withdraw():
```
fubuloubu: In conjunction with comment about auction completion and refunding money, perhaps this function can have an argument of the person to refund with a default of `msg.sender` or something similar so that the same function to refund money automatically can be used by a person during the auction to get their money back.

examples/auctions/simple_open_auction.v.py
```diff
@@ -0,0 +1,76 @@
...
+
+# Set to true at the end, disallows any change
+ended: public(bool)
+
+# Create a simple auction with `_bidding_time`
+# seconds bidding time on behalf of the
+# beneficiary address `_beneficiary`.
+def __init__(_beneficiary: address, _bidding_time: num(sec)):
+    self.beneficiary = _beneficiary
+    self.bidding_time = _bidding_time
```
DavidKnott: I like that idea,anything that saves gas is a plus

examples/auctions/simple_open_auction.v.py
```diff
@@ -0,0 +1,76 @@
...
+    self.auction_start = block.timestamp
+
```
DavidKnott: `self.ended` does default to `False`  but I agree that setting it to `False` in `__init__` is clearer and hence should be done.

examples/auctions/simple_open_auction.v.py
```diff
@@ -0,0 +1,76 @@
...
+# auction is not won.
+@payable
+def bid():
+    # Check if bidding period is over.
+    assert block.timestamp < (self.auction_start)+ self.bidding_time
+    # Check if bid is high enough
+    assert msg.value > self.highest_bid
+    if not self.highest_bid == 0:
+        # Allows the previous highest bidder to take their money back
+        self.pending_returns[self.highest_bidder] += self.highest_bid
```
DavidKnott: I have it tested and it doesn't error because the default value is 0, which allows for the old_bid to be added to it without an error

examples/auctions/simple_open_auction.v.py
DavidKnott: Now that I think about it, I think it makes sense to get rid of withdraw and pending_returns and just refund the bid right away.  What do you think?

examples/auctions/simple_open_auction.v.py
DavidKnott: I'm a little confused.  `send(self.beneficiary, self.highest_bid)`  sends the winning bid to the `beneficiary`, the `beneficiary` (the person who holds the auction) is meant to transfer the item/good won to the winner of chain.

examples/auctions/simple_open_auction.v.py
fubuloubu: That's clever, that way you don't have to worry about refunding at the end. Although, this may end up costing more fees overall because more transactions occur per person? I'm picturing a bidding war where two people are going back and forth on an item, and the constant refund -> re-bid transaction fees might be more than the fees to add more value to your bid each time. Maybe there is a compromise where you maintain a 5 person queue, and you get refunded if you fall off of that queue, but if you're in the queue, the value adds? Lots of potential issues with that thought maybe. A very rich set of trades to explore there I think.

examples/auctions/simple_open_auction.v.py
fubuloubu: Hmm, well as long as it's defined as being zero for an uninitialized key in a map in Viper instead of undefined behavior (I'm thinking C where undefined can occasionally mean whatever value that variable address held before) then it should be okay or even preferred NOT to check for uninitialized values as it reduces gas cost overall. I.E. it's a guarantee of the language

examples/auctions/simple_open_auction.v.py
fubuloubu: Oh yes, you're right! My apologies

fubuloubu: Are `self.highest_bid` and `self.balance` equal? Under what conditions? I have to think about that.

examples/auctions/simple_open_auction.v.py
fubuloubu: Well, that costs extra gas so if the language explicitly maintains unintialized values as 0 or False, then perhaps it's not such a terrible idea? I have the opinion that reading uninitialized values is bad idea typically because that behavior could be undefined, but if the values are clearly set by Viper or EVM then it is not dangerous (only potentially obscure). I am not familiar enough to say either way, @vbuterin?

examples/auctions/simple_open_auction.v.py
fubuloubu: @vbuterin Need guidance about this, similar to `self.ended` above. More broadly, since there is a higher gas cost associated with explicitly checking for an uninitialized key in a map, is there a benefit from being able to set the default value in that scenario in case the user desires non-zero initializations or references before initialization e.g.

```python
accounts: currency_value[address] = 100 # New users get 100 coins to start
    ...

def sell(amt: currency_value):
    if self.accounts[msg.sender] >= amt:
        self.accounts[msg.sender] -= amt
    ...

def buy(amt: currency_value):
    self.accounts[msg.sender] += amt
```

examples/auctions/simple_open_auction.v.py
```diff
@@ -0,0 +1,76 @@
...
+    # external contracts.
+
+    # 1. Conditions
+    # Check if auction endtime has been reached
+    assert block.timestamp >= self.auction_start + self.bidding_time
+    # Check if this function has already been called
+    assert not self.ended
+
+    # 2. Effects
+    self.ended = True
```
fubuloubu: You don't use this in your other functions to determine if the auction is over, only the time. This could mean that someone could add value to this contract after the auction is over, and because of `assert not self.ended` above, that value would get orphaned. I would suggest you check for both in your functions above (`bid()`) e.g. `assert not self.ended` and `assert block.timestamp < self.endtime`

examples/auctions/simple_open_auction.v.py
JackieXu: I'm confused, where is the mentioned `withdraw` function? Removed in an amend?

examples/auctions/simple_open_auction.v.py
```diff
@@ -0,0 +1,67 @@
...
+    if not self.highest_bid == 0:
+        # Sends money back to the previous highest bidder
+        send(self.highest_bidder,self.highest_bid)
+    self.highest_bidder = msg.sender
+    self.highest_bid = msg.value
+
+
+# End the auction and send the highest bid
+# to the beneficiary.
+def auction_end():
```
JackieXu: Does this method not allow for a call depth attack?

If anyone can directly call this method, and artificially sets the call stack so that this call is the last possible function call, the `self.ended` boolean will be set to `True` and the `send` method will fail right? Afterwards the funds appear to be stuck.

examples/auctions/simple_open_auction.v.py
DavidKnott: @JackieXu  I'm sorry, I probably should've made a comment explaining what I did.   I removed `withdraw` and rebased, now when the highest bid is displaced it sends the second highest bid back right away (eliminating the need for a withdraw method and keeping it simple).

examples/auctions/simple_open_auction.v.py
DavidKnott: Could you explain this a bit more?  My understanding is that the function will fail if it's called when `self.ended == True` but on the call when `self.ended` is set to true the `send` should still be able to go through.

JackieXu: This could be a misunderstanding from my side, but I figured that this occurs if something somehow does 1023 calls, and then calls `auction_end()`. This would cause the call depth to be at 1024.

If this is the first time `auction_end` is called (and `timestamp >= auction_end`), then it will successfully pass both asserts, and go on to setting `self.ended = True` (which does nothing to the call stack).

However, getting to `send(self.beneficiary, self.highest_bid)` will cause it to try and up the call depth by 1 again, which is not possible. This causes `send` to return `False`, effectively throwing the funds in limbo.

DavidKnott: Great explanation!  `send` is currently like transfer in Solidity (it `throws` as opposed to returning `false` on failure).  Knowing that is there still a problem? (it seems to me that throwing may prevent the call depth attack)

JackieXu: I wasn't aware that it threw an error instead of simply returning `False`! That's great, in that case the issue no longer happens. Ignore me. 😄

fubuloubu: I didn't know about the call depth issue. So, 1024 is the stack depth. If you made 1023 calls that weren't `auction_end()` and then called `auction_end()` (which makes a call to `send()`) that inner call would increment the stack 1 beyond the limit and therefore fail the stack limit. The questions are then:
1. Does `send()` throw an error (I was just thinking about whether it ALWAYS throws an error when it fails for any reason, or if this call depth thing supercedes `send()` and therefore doesn't throw)
2. Does throwing an error reverse the entire call stack (including `self.ended = True`), or just the last attempted transaction. (I believe it would stop `auction_end()` from being called, or should all 1023 prior calls also not happen?)

Am I understanding this correctly? Some very import nuance here (at least to me), and I am thinking about how to ensure a basic user can't make this mistake. I know one of the tenants of Viper is that explicit recursion is not allowed (you can only call methods defined prior to the method you are defining).

DavidKnott: @fubuloubu @JackieXu  To answer your questions.
1) Send will always throw on failure (the way it's coded in Viper)
2) Throwing will rollback all state changes made by a given transaction (in this case `self.ended` will go back to `False`)

fubuloubu: 1. Does the stack exceedence get checked before or during `send()`. I think either way would throw the error, and as you noted, rollback the transaction (thereby preventing the attack from being successful).
2. All state changes from the failed transaction (e.g. `auction_end()`), or all transactions made by the sender (the 1024, exceeding the stack). I think both prevent the stack call attack from being successful, but for clarity's sake...

DavidKnott: 1) Pyethereum makes the check [here](https://github.com/ethereum/pyethereum/blob/develop/ethereum/vm.py#L642), I believes this means that it'll get checked while the `send` opcodes are being processed.
2)  I believe all state changes made by the `tx.origin` are considered one transaction and are therefore rolled back (can rollback through multiple contracts and functions calls)

fubuloubu: Very interesting.

So, in summary, due to the implementation choice of `send()` throwing on error, and the implementation choice of pyethereum to completely rollback on a throw, this attack vector is not useful? Does that logically make sense?

DavidKnott: At least in this situation the aforementioned attack doesn't work.  It makes sense logically to me.

fubuloubu: @JackieXu the attack vector is basically:
1. Create N-1 transactions (where N is the stack limit)
2. Make call to target function where function changes the desired global values, and then makes subsequent call to another function exceeding stack length N
3. Leverage modified global values in situation where target function did not successfully complete, even though global values were modified as if it did complete
4. ...
5. Profit

Is that a correct summary of what can occur in Solidity?

JackieXu: @fubuloubu Yes, that's correct regarding Solidity.

I'm still interested in checking out whether this actually doesn't happen, but I suppose a simple test case could be implemented to check this:

 - Create a separate contract (`call_depth_attack.v.py`)
 - In this contract, up the call stack to 1023 and then do an external call to `auction_end`
 - Check whether any funds have been sent.
 - Check whether the `ended` flag has been set to `True`.

The reason for my interest is that it feels strange to have different implementations - as I understand the current case to be -  of what the underlying 'EVM' does, depending on whether it is running `pyethereum`, `go-ethereum` etc.

fubuloubu: Yes, that would be very helpful to prove this is the case. Maybe this could be a more general test case present in `tests/test_parser.py`. IIRC pyethereum and go-ethereum are definitely different, I believe pyethereum is some sort of future implementation that implements some fixes go-ethereum and the other implementations do not.

DavidKnott: The different implementations all have to pass the same [test suite](https://github.com/ethereum/tests/tree/738cf17893af31ed5f4a52fb70165b4b40316beb) (this ensures high-level uniformity between implementations).  Though I'd definitely be interested in having a test within `tests/test_parser.py` to ensure the aforementioned call depth attack doesn't work.  Would either of you like to take that on?

I don't believe the call depth attack is an issue with evm's but rather with the opcodes being used.  Sending `ether` using the `call` opcode doesn't inherently `throw` on failure.  The reason it does in Viper is because there's an`assert` after the call (essentially making Viper's `send` similiar to Solidities `transfer`).

fubuloubu: My mistake, I've been thinking pyetheapp was the python implementation and pyethereum was some sort of testbed. pyethapp is just the client duh :1st_place_medal:

I can write that test if @JackieXu hasn't already.

JackieXu: @fubuloubu I haven't started on that test yet. I took a peek just now, and it seems that attack vector is not possible from a Viper perspective, unless you - for some reason - decide to write it into your own contract (most likely by accident, but that's quite hard to do).

As far as I understand, the contracts written using the current Viper version don't allow you to make an external contract call, so there is no way for a different Viper contract to call the `auction_end` method in this case.

It's actually a bit tricky to try and test this, because there is no recursion allowed in Viper. To test this you would need to define 1023 dummy functions that call the one above them, and then make a call to `auction_end`. This seems fine from a testing perspective (because you're checking all the options that might change your contract state into something undesirable), but I'd like to hear your thoughts on it.

fubuloubu: Haha, that's exactly the test I was writing! (1023 dummy functions)

My thoughts is that this attack can be provably unlikely given the lengths you would have to go to allow it in your contract, as you stated. Additionally, the way it operates under that failure scenario doesn't seem to allow any undesirable results (setting values when part of the transaction fails) because our `send()` function throws on failure as described. An important thing to look out for are core Viper functions that do not behave in that way (throw on failure), or if something about how the language operates would allow a situation where a failure occurs but does not throw. I think that should be made a design goal generally.

An interesting thing you said is external contract calls, we're working on adding that feature currently and that has a whole different set of concerns about recursion I noted in #297.

fubuloubu: I tried for a few hours today, and I couldn't figure out exactly how to get this to work. @JackieXu if you come up with something for this feel free to commit the test

JackieXu: I tried a bit just now, but ran into gas limits. I think I ran into gas limits, at least; the error messages aren't all that descriptive.

Will take a more in-depth look later.

JackieXu: I tried for a while and was unable to get a contract of that size to work, mine simply wouldn't initialize. I tried setting the starting gas and gas limit to 2**128 - 1, without much added success.

fubuloubu: Haha, well sounds like we need to show that it's not possible via proof by contradiction. Sounds like a good time, maybe we can collborate next week.

DavidKnott: @fubuloubu I feel comfortable merging this in as I don't think the discussed exploit scenario can occur in this example.

fubuloubu: Yup, that's fair. @JackieXu and I can't write a failing test case right now to show this behavior in action.

PR #312, opened 2017-08-07
Add automatic versioning

fubuloubu:
After this pull request is accepted, create a new tag with `bumpversion`

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -7,7 +7,7 @@ tag = True
 search = version = "{current_version}"
 
 [aliases]
-test=pytest
+test = pytest
 
 [tool:pytest]
 addopts = --cov-report term --cov-report html --cov=viper
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,3 +1,11 @@
+[bumpversion]
+current_version = 0.0.1
+commit = True
+tag = True
+
+[bumpversion:file:setup.py]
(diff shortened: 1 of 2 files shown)


setup.cfg
```diff
@@ -1,5 +1,13 @@
+[bumpversion]
+current_version = 0.0.1
```
yograterol: Should Viper start from 0.1 or 0.0.1?

fubuloubu: 0.1.0, would be when features are solidified and ready for the first beta IMO.

I would start at 0.0.1 for now.

yograterol: Viper is ready to use right now, not in production environment but if in test environment. So, VIper has solid features :) At least compile a whole contract.

But it's ok start from 0.0.1 :+1:

fubuloubu: Sure agreed, but there are unimplemented features that I would argue need to be there for a complete first draft to alpha/beta with. Perhaps that is more conservative than necessary, but that's my opinion.

I'm not in any position to be dictating what the first version is here anyways since that still needs to be set up initially after this PR is merged, and that's not up to me haha. I just used 0.0.1 because that is what it was.

yograterol: Hahaha, agree .. Well, maybe @vbuterin could tell you what is the right version for now.

But the PR is OK. : +1:

PR #313, opened 2017-08-07
Wrapper contract_with_gas_estimation

yograterol:
PR related to https://github.com/ethereum/viper/issues/295

Added two functions:

1 - `gas_estimation_decorator`: wrap the ABI function and call `check_gas` after the contract function is called.
2 - `get_contract_with_gas_estimation` create the contract from the source code adding by default `language="viper"`, get  the functions from the contract and wrap the contract function to use `gas_estimation_decorator`. 
3 - `get_contract` create the contract from the source code adding by default `language="viper"`.

TODO

- [x] Change each tests that use `check_gas` function.

Note: We need to review `check_gas` in the future to make the function more accurate.

### DIFF
--- a/tests/setup_transaction_tests.py
+++ b/tests/setup_transaction_tests.py
@@ -58,14 +58,26 @@ def gas_estimation_decorator(fn, source_code, func):
     return decorator
 
 
-def contract_with_gas_estimation(source_code, target_func_check_gas=None):
-    contract = chain.contract(source_code, language="viper")
+def set_decorator_to_contract_function(contract, source_code, func):
+    func_definition = getattr(contract, func)
+    func_with_decorator = gas_estimation_decorator(
+        func_definition, source_code, func
+    )
+    setattr(contract, func, func_with_decorator)
 
-    if hasattr(contract, target_func_check_gas):
-        func_definition = getattr(contract, target_func_check_gas)
-        func_with_decorator = gas_estimation_decorator(
-            func_definition, source_code, target_func_check_gas
+
+def get_contract_with_gas_estimation(
+        source_code,
+        *args, **kwargs):
+    contract = chain.contract(source_code, language="viper", *args, **kwargs)
+
+    for func_name in contract.translator.function_data:
+        set_decorator_to_contract_function(
+            contract, source_code, func_name
         )
-        setattr(contract, target_func_check_gas, func_with_decorator)
 
     return contract
+
+
+def get_contract(source_code, *args, **kwargs):
+    return chain.contract(source_code, language="viper", *args, **kwargs)
--- a/tests/setup_transaction_tests.py
+++ b/tests/setup_transaction_tests.py
@@ -58,14 +58,14 @@ def gas_estimation_decorator(fn, source_code, func):
     return decorator
 
 
-def contract_with_gas_estimation(source_code, func=None):
+def contract_with_gas_estimation(source_code, func_target_check_gas=None):
     contract = chain.contract(source_code, language="viper")
 
-    if hasattr(contract, func):
-        func_definition = getattr(contract, func)
+    if hasattr(contract, func_target_check_gas):
+        func_definition = getattr(contract, func_target_check_gas)
         func_with_decorator = gas_estimation_decorator(
-            func_definition, source_code, func
+            func_definition, source_code, func_target_check_gas
         )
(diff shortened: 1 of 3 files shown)


fubuloubu:
Interesting, but there is at least one example where that isn't the desired behavior i.e. testing a contract with multiple methods and getting the gas estimate for all calls made, instead of a specific function.

I was sort of thinking about this when I created that function a while ago, it would be cool that with every call made to the contract, that the gas estimation feature was checked e.g. `c.foo()` calls `check_gas()` with function `foo` after that call was made. This contrasts with specifying which function you want to check on contract creation, as you may want to run `c.foo()` and `c.bar()` and check them both of each call instead of only one function. This might look something like:
```python
c = compile_viper_contract( code )
check_call(c, 'foo', result, *args, **kwargs)
```
and `check_call()` would basically do (figure out how to actually use args and kwargs correctly here):
```python
assert c.foo(args) == result
```
This would give nicely instrumented test functions for us to use. I don't think your proposed changes helps in that regard right now.

yograterol:
@fubuloubu Thank you for your comment.

I could do that `target_func_check_gas` a list of functions to check gas.

So, you define your contract like:

```python
c = contract_with_gas_estimation(source_code, ['foo']) #  For a single function

OR

c = contract_with_gas_estimation(source_code, ['foo', 'bar']) # For multiples functions
```

When you call `foo` or `bar` automatically `check_gas` is called.

fubuloubu:
If you are able to get it to check the gas on a function call automatically, why would you have to supply a list of functions to check? Is that not possible?

I would aim to make it as simple as possible for the developer. My ideal would be that any call automatically calls the gas estimation check after with the given function.

yograterol:
@fubuloubu I can get every function from `ABIContract.translator.function_data` but do we want that behavior?

fubuloubu:
My concern is having to type `foo` twice in two locations for a given test, and remembering to keep things updated and using unnatural syntax for testing. The closer the syntax for testing is to regular use, the easier testing is to figure out and maintain.

yograterol:
@fubuloubu your concern is valid.

What do you think of the next proposal?

```python

def set_decorator_to_contract_function(contract, source_code, func):
    func_definition = getattr(contract, func)
    func_with_decorator = gas_estimation_decorator(
        func_definition, source_code, func
    )
    setattr(contract, func, func_with_decorator)


def contract_with_gas_estimation(
        source_code,
        target_func_check_gas=None,
        *args, **kwargs):
    contract = chain.contract(source_code, language="viper", *args, **kwargs)

    # Check single function
    if target_func_check_gas and hasattr(contract, target_func_check_gas):
        set_decorator_to_contract_function(
            contract, source_code, target_func_check_gas
        )
    elif not target_func_check_gas:
        # Check all contract functions
        for func_name in contract.translator.function_data:
            set_decorator_to_contract_function(
                contract, source_code, func_name
            )

    return contract

# ------ In test file ---------

# For single function
def test_x():
    ...
    c = contract_with_gas_estimation(null_code, 'foo')
    c.foo() # Run internally check_gas

# For every function
def test_x_y():
    ...
    c = contract_with_gas_estimation(null_code)
    c.x()
    c.y()
    # x and y function call `check_gas`
```

fubuloubu:
I like that. Definitely much closer to what I was thinking.

I think I still don't see the point of being able to specify which functions get the decoration to check for gas usage, e.g. `contract_with_gas_estimation(null_code, 'foo')`. The decorator is only called to check the gas if the code calls that function (I believe), and we want to check gas usage for every function. So I don't see the savings behind explicitly saying which ones get checked and which ones don't.

Does that make sense?

yograterol:
@fubuloubu yes, make sense. I did a little change to the source code above, can you check it again?

fubuloubu:
Yes, perfect. `get_contract()` doesn't need to run `check_gas()` because you would be assuming that you're not checking the gas for that specific contract. That would be useful if you're running into problems with a certain contract leading to a test failure. That's also something that py.test has a means to circumventing too, so maybe the utility of the `get_contract` function without checking the gas is limited, but it works for now.

yograterol:
With `get_contract` you will be able to run `check_gas` function if you want, is not mandatory.

I'm doing the changes.

yograterol:
@fubuloubu Done :+1:

Issue #310, opened 2017-08-06
Failed `send` doesn't error or return false


DavidKnott:
Right now when I run:
```
send(msg.sender, self.balance + 1)
```
It neither errors nor returns false.  I would like to implement functionality similar to Solidity where a failed `send` returns false and implement `transfer`  which will throw on failure.  Thoughts?

fubuloubu:
I'd say that's a major issue. Does the balance remain what it was, or is everything lost?

DavidKnott:
The balance remains which is good, I'm making a PR so that send returns a `bool`.

fubuloubu:
I think someone was discussing `send()` and how it should throw an exception instead of return `True / False`. I'm not sure of the subtlties, but I would prefer any transaction that is supposed to send me money to fail if the transaction doesn't work.

DavidKnott:
I agree, that people should have that functionality but I like the idea of doing it like Solidity does by having `send`  which returns a `bool` and `transfer` which throws so that people can choose.

PR #314, opened 2017-08-08
`send` throws on failure

DavidKnott:
This PR is in response to issue #310.  This PR changes `send` so that it throws on failure and tests the added functionality.  I tried to be as minimalistic as possible.

### DIFF
--- a/tests/test_parser.py
+++ b/tests/test_parser.py
@@ -1,3 +1,4 @@
+import pytest
 from .setup_transaction_tests import chain as s, tester as t, ethereum_utils as u, check_gas
 
 def test_null_code():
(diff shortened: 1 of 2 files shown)


yograterol:
@DavidKnott this PR is not ready to merge, true? I think it because @fubuloubu ask to talk more about this change https://github.com/ethereum/viper/issues/310

DavidKnott:
@yograterol I think this PR is better then the current way `send` works in which nothing happens.  Though I agree that it should definitely by discussed further (sending Ether is a huge part of the language).

fubuloubu:
I think @DavidKnott has a point with `transfer()` throwing and `send()` returning a bool. Although, one of the benefits of viper to me is sort of being Pythonic "simple is better than complex" and two different ways to send Ether is not simpler.

If instead the rules are that `send()` (or whatever function sends money) throws, and all asset transfer fails back to prior state (which I believe EVM does inherantly?), I believe that is simpler than asking people to do it themselves.

How many use cases are there for allowing Ether transfer to fail but still transfer assets?

DavidKnott:
@fubuloubu If the send fails the user my still want to transfer assets, but I see the main issue being that the user may want to change state even if  `send` has failed and if `send` were to throw on failure then the user may have to create more transactions (one for changing state and one for a `send` that might throw reverting everything).  In terms of security and simplicity I agree with you 100% in terms of just throwing every time.   The more I think about it I think you're right because I can't think of a use case off the top of my head where someone would want `send` to return `False` as opposed to throwing.

How would you feel about me changing `send` to throw?

fubuloubu:
Does `send` not throw right now? I remember testing and getting `TransactionFailed` from pyethereum, but I'm not sure about the whole path there.

DavidKnott:
Right now it doesn't throw.  If you wanna confirm, run this (it should always fail):
```
send(msg.sender, self.balance + 1)
```

PR #317, opened 2017-08-15
Added auction example.

Dexaran:


fubuloubu:
@DavidKnott also had an auction example, similarly using the solidity docs. I suggest you guys combine efforts here,

Also, can you make the extension `.v.py` so we can see source code highlighting?

examples/auction/auction.vy
```diff
@@ -0,0 +1,94 @@
...
+    assert msg.value > self.highestBid
+    null_address: address
+    if self.highestBidder != null_address:
+        # Sending back the money by simply using
+        # highestBidder.send(highestBid) is a security risk
+        # because it could execute an untrusted contract.
+        # It is always safer to let the recipients
+        # withdraw their money themselves.
+
+        self.pendingReturns[self.highestBidder] += self.highestBid;
```
fubuloubu: Your check above `self.highestBidder != null_address` is after the face, and means that the previous bid is still in the contract. I suggest changing your check to an assertion that the msg.sender is not the null address (which, based on my above comment, may not be necessary)

examples/auction/auction.vy
```diff
@@ -0,0 +1,94 @@
...
+def withdraw() -> bool:
+    amount = self.pendingReturns[msg.sender]
+    if amount > 0:
+        # It is important to set this to zero because the recipient
+        # can call this function again as part of the receiving call
+        # before `send` returns to avoid re entry attack
+
+        self.pendingReturns[msg.sender] = 0
+
+    send(msg.sender, amount)
```
fubuloubu: Why allow someone to withdraw if they have no amount to send? Can this be changed to `assert amount > 0`?

examples/auction/auction.vy
```diff
@@ -0,0 +1,94 @@
...
+    return True
+
```
fubuloubu: If `self.balance == 0`, can this contract be destroyed?

Dexaran:
I see it is duplicated https://github.com/ethereum/viper/pull/309.

fubuloubu:
Feel free to review #309 if you have any comments.

PR #318, opened 2017-08-15
Added example of contract-wallet with passphrase access restriction.

Dexaran:


fubuloubu:
Interesting example, although I'm not sure the `call()` function is something that should be included.

Either way, can you make the extension `.v.py` so we can see source code highlighting?

examples/wallet_with_pincode/wallet_with_pincode.v.py
```diff
@@ -0,0 +1,77 @@
...
+# Owners wallet.
+owner: public(address)
+
+# Hash of a secret number that allows user
+# to operate with funds from this contract.
+secret_hash: public(bytes32)
+
+# Assign owner to contract creator.
+def __init__():
+    self.owner = msg.sender
```
fubuloubu: `secret_hash` should be set here, as the first time you go to use `set_secret()` below it might not work properly.

examples/wallet_with_pincode/wallet_with_pincode.v.py
```diff
@@ -0,0 +1,77 @@
...
+@payable
+def deposit():
+    return
+
+# A function that allows anyone
+# to check which hash matches the entered secret.
+# `@constant` decorator allows call this function
+# without broadcasting a transaction to the blockchain.
+@constant
+def secret_from_num(_secret: num) -> bytes32:
```
fubuloubu: Name is a little confusing.

Can this be restructured into `check_secret()` that performs the same function `set_secret()`, `withdraw()`, and `call()` check for in their execution, that way 1. you're not repeating yourself and 2. This operation is clearer to it's intent?

examples/wallet_with_pincode/wallet_with_pincode.v.py
```diff
@@ -0,0 +1,77 @@
...
+    return sha3(as_bytes32(_secret))
+
+# This function allows `owner` to withdraw funds
+# if he will successfully prove ownership with a `secret`
+def withdraw(_secret: num):
```
fubuloubu: Can `withdraw()` be `transfer()`? With `transfer()` (adding `_to: address` as an arg) we can send money to someone else without resorting to `call()` below, and we can replicate your functionality by just calling `transfer()` with `msg.sender` as `_to`. I think `withdraw()` is too limiting to include.

examples/wallet_with_pincode/wallet_with_pincode.v.py
```diff
@@ -0,0 +1,77 @@
...
+
+# Assign owner to contract creator.
+def __init__():
+    self.owner = msg.sender
+
+# This function allows to reset secret
+# each secret should be used only once
+# since everyone can track your transaction data
+# in blockchain.
+def set_secret(_secret: num, _new_secret: bytes32):
```
fubuloubu: Make `_new_secret` into `_new_secret_hash` so it is clearer what it's purpose is

examples/wallet_with_pincode/wallet_with_pincode.v.py
```diff
@@ -0,0 +1,77 @@
...
+    # Only owner can execute this.
+    assert msg.sender == self.owner
+
+    # Delivers funds from this contract to the `owner`
+    # in case of successful unlocking of this contract.
+    send(self.owner, self.balance)
+
+    # NOTE: if the execution of this function was successful
+    # then the execution data will remain in the blockchain
+    # `secret` needs to be updated.
```
fubuloubu: Wouldn't it be a better idea to add `_new_secret_hash` as an argument and call `set_secret()` after success here to enforce the pin update? Same with `call()`

examples/wallet_with_pincode/wallet_with_pincode.v.py
```diff
@@ -0,0 +1,77 @@
...
+
+# Perform a call from this contract
+# with given parameters.
+def call(_to: address, _value: wei_value, _data: bytes <= 4096, _secret: num) -> bytes <= 4096:
+    assert sha3(as_bytes32(_secret)) == self.secret_hash
+    assert msg.sender == self.owner
+
+    # Executes a raw call from this wallet
+    # with the following parameters
+    return raw_call(_to, _data, outsize=4096, gas=3000000, value=_value)
```
fubuloubu: More a comment in general: it would be nice if we could set default values for parameters (such as `gas=` here) so that items like that can be either inherited or set directly. Thoughts? @DavidKnott @vbuterin

jacqueswww:
Closing for the time being, feel free to re-open once it's been updated to latest syntax :smiley_cat:

Issue #315, opened 2017-08-09
Calling constant functions that use msg.sender in non-constant functions fails


fubuloubu:
I was testing some updates I was making with one of my example contracts and I think I may have stumbled upon a bug, or at least something I don't understand why it should fail (and it's not very clear why)

To summarize, I have a contract that contains a constant method to return the balance of some map (address -> record) for the address associated with the message sender. I decided to use this function in my other functions so I didn't duplicate this access in multiple spots, basically assuming that if I called this `@constant` function, the `msg.sender` attribute would be passed to that `@constant` function. However, when I do this I get the `TransactionFailed` exception from pyethereum.

Here's an example:
```python
from ethereum.tools import tester
from viper import compiler

chain = tester.Chain()
tester.languages['viper'] = compiler.Compiler()
chain.head_state.gas_limit = 10**9

code="""
bal: num[address]

def __init__():
    self.bal[msg.sender] = 1

@constant
def get_my_bal() -> num:
    return self.bal[msg.sender]

def foo():
    assert self.bal[msg.sender] > 0

def bar():
    assert self.get_my_bal() > 0
"""

c = chain.contract(code, language='viper')
c.get_my_bal() # Works fine
c.foo() # Works fine as well
c.bar() # Fails with ethereum.tools.tester.TransactionFailed()
```

Please let me know if I am misunderstanding how the `@constant` is supposed to be used (I am assuming you should be allowed to call `@constant` functions from other functions) (I am also assuming that implicit values such as `msg.sender` are passed to `@constant` functions as well).

Not totally certain what the problem is...

fubuloubu:
I believe that `msg.sender` and other "internal"quantities are not passed to an internal call to a function in a Viper program. Still trying to figure out why, would appreciate any help!

DavidKnott:
@fubuloubu Great catch!  I'll help with this, off the top of my head I'm guessing it's an issue with how function contexts are being made.

fubuloubu:
I narrowed it down to a test case that we can use to debug. I'll create a pull request for that, and then maybe you can help me debug the issue further.

fubuloubu:
Created #320

Further collaboration on that please.

PR #320, opened 2017-08-16
Failing test case for debugging constant inherited arguments

fubuloubu:
See #315

Looking for help resolving this problem as the compiler should be passing through all the inherited arguments (e.g. `msg.sender`) in the context call to a function from within another function.

### DIFF
--- a/tests/test_parser.py
+++ b/tests/test_parser.py
@@ -156,6 +156,33 @@ def foo() -> num:
     print('Passed basic state accessor test')
 
 
+def test_use_constant_function():
+    code="""
+bal: num[address]
+
+def __init__():
+    self.bal[msg.sender] = 1
+
+@constant
+def get_my_bal() -> num:
+    return self.bal[msg.sender]
+
+def foo():
+    assert self.bal[msg.sender] > 0
+    self.bal[msg.sender] += 1
+
+def bar():
+    assert self.get_my_bal() > 0
+    self.bal[msg.sender] += 1
+    """
+    
+    c = s.contract(code, language='viper')
+    assert c.get_my_bal() == 1
+    c.foo() # equivalent code to bar()
+    c.bar() # This transaction should estimate less gas than foo(), but take the same amt
+    assert c.get_my_bal() == 3
+    check_gas(code, num_txs=4)
+
 def test_arbitration_code():
     arbitration_code = """
 buyer: address
--- a/tests/test_parser.py
+++ b/tests/test_parser.py
@@ -156,32 +156,26 @@ def foo() -> num:
     print('Passed basic state accessor test')
 
 
-def test_use_constant_function():
+def test_reuse_function():
     code="""
 bal: num[address]
 
 def __init__():
     self.bal[msg.sender] = 1
 
-@constant
-def get_my_bal() -> num:
+def my_bal() -> num:
     return self.bal[msg.sender]
 
-def foo():
-    assert self.bal[msg.sender] > 0
-    self.bal[msg.sender] += 1
-
 def bar():
-    assert self.get_my_bal() > 0
+    assert self.my_bal() > 0
     self.bal[msg.sender] += 1
     """
     
     c = s.contract(code, language='viper')
-    assert c.get_my_bal() == 1
-    c.foo() # equivalent code to bar()
-    c.bar() # This transaction should estimate less gas than foo(), but take the same amt
-    assert c.get_my_bal() == 3
-    check_gas(code, num_txs=4)
+    assert c.my_bal() == 1
+    c.bar()
+    assert c.my_bal() == 2
+    check_gas(code, num_txs=3)
 
 def test_arbitration_code():
     arbitration_code = """
--- a/tests/test_parser.py
+++ b/tests/test_parser.py
@@ -163,6 +163,7 @@ bal: num[address]
 def __init__():
     self.bal[msg.sender] = 1
 
+@constant
 def my_bal() -> num:
     return self.bal[msg.sender]
 
@@ -171,11 +172,10 @@ def bar():
     self.bal[msg.sender] += 1
     """
     
-    c = s.contract(code, language='viper')
+    c = get_contract_with_gas_estimation(code)
     assert c.my_bal() == 1
     c.bar()
     assert c.my_bal() == 2
-    check_gas(code, num_txs=3)
 
 def test_arbitration_code():
     arbitration_code = """


DavidKnott:
@fubuloubu After doing some research it's not only an issue `@constant` but with normal functions as well.  The issue is that the `caller` changes to the contract itself which is why the following code passes:
```
bal: num[address]

def __init__():
    self.bal[msg.sender] = 1
    self.bal[self] = 1

@constant
def get_my_bal() -> num:
    return self.bal[msg.sender]

def foo():
    assert self.bal[msg.sender] > 0

def bar():
    assert self.get_my_bal() > 0
```
I should be to the bottom of this shortly, looking into the opcode calls now.

fubuloubu:
Yeah, I just added `@constant` because it made sense in the scenario that you would call smaller constant functions to access things commonly. Do you think I should simplify the scenario?

DavidKnott:
It makes sense, I was just noting it.

DavidKnott:
@fubuloubu I figured out the opcodes to get it working properly, mind if I make a PR or I can push to your repo if you'd prefer.

fubuloubu:
Push to my repo probably best

DavidKnott:
Hey, I got msg.sender persist by using `delegatecall` but it broke internal functions because both use the `CALLER` opcode (in my implementation if msg.sender is still an outside address then internal functions will throw).  Any ideas?

fubuloubu:
@DavidKnott not sure I entirely understand what you are saying, but from what I understand the context shifts from the client (`msg.sender`) to the contract (`self`) when the call is made internally of function b from function a. Is that correct?

Curious as to why this context shift is necessary and what might be causing the shift if it is unintentional.

fubuloubu:
Do you have a specific line you can point to where the context shift occurs?

DavidKnott:
Yes, exactly.  Basically when I change the context to keep client(msg.sender) `internal` doesn't work (because internal means it can only be called from within).  I think I'm just implementing it wrong.

fubuloubu:
Man that file is super large and needs to be cleaned up.

Hmm. I could see `internal` being implemented like `inline` in C where the compiled function referencing `internal` actually gets a copy of that code inserted atomically instead of a function created requiring the `self` caller context. Inlining that code probably wouldn't be too hard since we mostly use global variables to track state, and perhaps any locals can be either checked for collisions or name-mangled via function name to avoid namespace issues.

Summary:
1. `internal` doesn't generate any code in the final IR/Bytecode (ABI can still generate though?)
2. when a call is made to an `internal` function, the compiler overwrites that call with the code inlined
3. This has the effect of naturally precluding any calls to that code externally, either via clients or contract inheritance, because the method doesn't actually exist

Definitely a significant change proposal there worth discussion. @vbuterin, what do you think?

fubuloubu:
@DavidKnott can you post what your currently solution is? (diff fmt?)

DavidKnott:
Sure, though I think the change I made is wrong as I'm not sure how to get `internal` working with it.  It's replacing line 772 with this:
```
['assert', ['delegatecall', ['gas'], ['address'],
```
`DELEGATECALL` is like `CALL`  but persists the current values for sender and value.

fubuloubu:
Rebased this to latest updates.

fubuloubu:
Analyzing the IR it appears that whenever `msg.sender` is used, `caller` is put in it's place in the produced code. As you noted, `caller` for an internal function call is set to the contract's address by definition. So, in order to get this to pass you would have to introduce a slightly different construct for `msg.sender`: instead of replacing with `caller`, it would need to retrieve the original calling party from the context (`msg.sender` as an additional field) in order to reference correctly in internal functions.

I believe this implicit context change is somewhat dangerous if the contract coder does not understand this nuance, e.g. using a function directly produces different results than if used internally inside another function. As an example:
```python
balance: currency_value[address]
pending: {from: address, to: address, amt: currency_value}[num]

def pay(_to: address, _amt: currency_value):
    assert self.balance[msg.sender] >= _amt
    self.balance[msg.sender] -= _amt
    self.balance[_to] += _amt

def accept(tx_id: num):
    assert msg.sender == self.pending[tx_id].from
    self.pay(self.pending[tx_id].to, self.pending[tx_id].amt)

...
```

If you called `pay()` directly, you would be paying from your own account to the relevant parties account; however, if you called `accept()` on some requested transaction (stored in `self.pending` from some unimplemented billing function), it would actually pay from the contract's own account instead.

An obvious solution here is to specify both parties in the `pay()` transaction, but in the interest of code savings this can be a useful alternative. I could probably come up with a more complex example given time, but I hope my point is clear that `msg.sender` is set to in different calling contexts in different scenarios, and that it is a potential source of user bugs.

@vbuterin @DavidKnott thoughts?

fubuloubu:
Figured it out. `msg.sender` is set to caller of that function, `tx.origin` tracks the initiator of the original transaction (outer level caller). This is in line 494-495 of parser.py [here](https://github.com/ethereum/viper/blob/b080a33185aed58a2b07812b6c612e5290ffd0e4/viper/parser.py#L494)

I'm assuming the original intent was that `tx.origin` is tracking which address initiated the whole transaction, which includes all function calls made. so `msg.sender` is then misleading to me, because it sounds like it's talking about a user, but in reality it is tracking the caller of that function instead. I'm debating whether `msg.caller` would more clearly state the purpose behind that, and/or whether a better name for `tx.origin` might also help. Also had a random thought about whether `msg.value` would be passed through correctly to the inner function? I'll have to write a test case for that

@vbuterin @DavidKnott thoughts?

P.S. When I make the update from `msg.sender` to `tx.origin` the issue gets resolved, but gas estimation fails for the function call with the inner call.

DavidKnott:
@fubuloubu I think Viper needs both, I prefer using msg.sender over tx.origin because it's easier to control.  Tx.origin messes things up when you have contracts interacting with one another see ([this](https://github.com/ethereum/solidity/issues/683) and [this](https://ethereum.stackexchange.com/questions/1891/whats-the-difference-between-msg-sender-and-tx-origin)).  Basically msg.sender shouldn't change between function calls but should change between contract calls (I believes msg.value should behave similarly).

fubuloubu:
Kind of revisited this for a bit today. @DavidKnott are you trying to say that there is a potential problem here with how `msg.sender` is working? If so, is that issue with viper, solidity, or EVM?

DavidKnott:
@fubuloubu I'll look into this again tomorrow and get back to you tomorrow, right now I suspect it's a fixable issue with viper.

fubuloubu:
Looking at [these](https://github.com/ethereum/viper/blob/b080a33185aed58a2b07812b6c612e5290ffd0e4/viper/parser.py#L494) lines in parser.py again, It seems like it's making a call to something that is in EVM. Namely, I think that references to [these](https://github.com/ethereum/viper/blob/b080a33185aed58a2b07812b6c612e5290ffd0e4/viper/opcodes.py#L28) opcodes. If solidity didn't have this same problem, then I would agree something in Viper has to be wrong.

That, or I am totally misunderstanding what the problem is here.

Let me know what you find.

fubuloubu:
I copied the failing test case over to Solidity [here](http://remix.ethereum.org/#version=soljson-v0.4.16+commit.d7661dd9.js) and it seems to work fine. You must be right there is probably an issue with something on the Viper end.

Although, doesn't viper use a different implementation of the EVM?

DavidKnott:
@fubuloubu I looked into how Solidity maintains `msg.sender` while supporting `internal` and though I'm not very familiar with C++ it seems to be similar to inlining as you mentioned above.  So far the best possible solution I've find is to not include unused internal functions in the evm code and then inserting them where they're used within functions. 

Generally speaking, I think the `call` opcode is meant to call other contracts functions as opposed to it there own.

DavidKnott:
@fubuloubu I talked to Vitalik and as of now the proper way to use `msg.sender` after the first function call is to pass it through as an argument.  The other option would be to load functions directly through memory (Solidity does this) as opposed to calling them (`mload` vs `call` opcodes) as loading them wouldn't change the `msg.sender`.  But, it was agreed that loading functions from memory would increase code complexity more then it's worth.

fubuloubu:
I will close this PR then because there's nothing unique in this additional test case.

PR #329, opened 2017-09-01
Create Viper Docs Skeleton

DavidKnott:
Adds `sphinx_rtd_theme` to the documentation, adds in contribution guidlines and sets the format for creating future docs files.
@fubuloubu I'd really appreciate it if you took a look at this.


fubuloubu:
I don't know much about sphinx, but the text seems alright with me!

DavidKnott:
Thanks for looking it over!

PR #334, opened 2017-09-07
Fix test path

JackieXu:
### - What I did
I updated the test discovery option in `setup.cfg` so that pytest also looks into the subdirectories of `tests/`.

Additionally a small update to resolve an `ImportError` that was raised in one of the tests that wasn't being tested.

### - How I did it
Changed `setup.cfg`

### - How to verify it
If you run `python setup.py test` without these changes, and again with these changes, you will see more tests being run in the latter case.

N/A

### - Cute Animal Picture
cute cat

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -11,4 +11,6 @@ test = pytest
 
 [tool:pytest]
 addopts = --cov-report term --cov-report html --cov=viper
-python_files = tests/test_*.py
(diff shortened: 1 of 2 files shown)


JackieXu:
Seems like the build got stuck after trying to download and decompress Python 3.6. 😞

PR #338, opened 2017-09-08
Stock example

fubuloubu:
### - What I did
Updates to my stock example to provide a means to transfer stock between stockholders
Also updated tests

~~NOTE: This fails due to `msg.sender` bug (see #320)~~

Example update

### - Cute Animal Picture
baby tapir

### DIFF
--- a/examples/stock/company.v.py
+++ b/examples/stock/company.v.py
@@ -72,7 +72,7 @@ def transfer_stock(receiver: address, transfer_order: currency_value):
     self.holdings[receiver] += transfer_order
 
 # Allows the company to pay someone for services rendered
-def give(vendor: address, amount: wei_value):
+def pay_bill(vendor: address, amount: wei_value):
     # Only the company can pay people
     assert msg.sender == self.company
     # And only if there's enough to pay them with
--- a/examples/stock/company.v.py
+++ b/examples/stock/company.v.py
@@ -40,8 +40,8 @@ def buy_stock():
 
 # So someone can find out how much they have
 @constant
-def get_holding() -> currency_value:
-    return self.holdings[msg.sender]
+def get_holding(_stockholder: address) -> currency_value:
+    return self.holdings[_stockholder]
 
 # The amount the company has on hand in cash
 @constant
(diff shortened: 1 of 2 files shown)


fubuloubu:
@DavidKnott this updated example now works and passes testing

examples/stock/company.v.py
```diff
@@ -27,23 +32,61 @@ def buy_stock():
...
 
+# So someone can find out how much they have
+@constant
+def get_holding(_stockholder: address) -> currency_value:
+    return self.holdings[_stockholder]
+
+# The amount the company has on hand in cash
+@constant
+def cash() -> wei_value:
+    return self.balance
```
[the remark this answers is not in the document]
DavidKnott: `.balance` is a built in function that gets the balance of an address, you can also write `msg.sender.balance` or `_insert_any_address.balance`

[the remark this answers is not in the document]
JackieXu: That clears things up, thanks!

examples/stock/company.v.py
```diff
@@ -27,23 +32,61 @@ def buy_stock():
...
+
+    # Pay the bill!
+    send(vendor, amount)
+
+# The amount a company has raised in the stock offering
 @constant
-def valuation() -> wei_value:
+def worth() -> wei_value:
     return (self.total_shares - self.holdings[self.company]) * self.price
+
```
DavidKnott: @fubuloubu I'd call `worth` something else or change how it's calculated as when I think of a companies worth I think of all the stock including the shares that the company owns.  What do you think?

fubuloubu: I was thinking about the valuation as the amount owed to every stockholder if all of a sudden they all wanted to sell to the company. I think this whole example is a bit different then how stocks traditionally work in that you don't usually buy and sell directly to the company, this idea was more representative of what could be possible if you changed the paradigm a bit.

That being said, if you have a better name, I'd consider it. It's not really a problem to change it.

DavidKnott: It seems like the company is beholden to the shareholders for `worth()`, what do you think about calling it `debt()`?

fubuloubu: I had a debt function below that represent the value held in cash versus the amount held by stock. I think I could have different names if it would make it clearer

DavidKnott: That'd be great! Thanks

fubuloubu:
@DavidKnott I just swapped the two. I think it's clearer now actually. Thanks!

DavidKnott:
@fubuloubu Assume! Thanks a lot.

PR #340, opened 2017-09-09
Format the ABI output to be JSON.

jacqueswww:
### - What I did

- For easy copy-paste purposes, I found that I constantly had to convert the python ABI output from the command line, this just does a simple json.dumps to make it easier.
- Minor pep8 fixes in the same file.

### - How I did it

Inserted a json.dumps ;)

### - How to verify it

Run `./viper_bin.py examples/crowdfund.vy -f json` and check the output.

Use JSON ABI output instead of python output (-f abi).

### DIFF
--- a/bin/viper
+++ b/bin/viper
@@ -1,13 +1,14 @@
 #!/usr/bin/env python3.6
+import argparse
+import json
 
 import viper
 from viper import compiler, optimizer
 from viper.parser import parse_to_lll
-import argparse
 
 parser = argparse.ArgumentParser(description='Viper {0} programming language for Ethereum'.format(viper.__version__))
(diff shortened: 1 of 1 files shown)


fubuloubu:
I beileve we had decided on Python output for `abi`, but maybe you can add a forth formatting option for `json` instead? I have no problems with adding another format

jacqueswww:
Ah I see, sure will do 😉

jacqueswww:
@fubuloubu updated to add the `json`  option.

fubuloubu:
Perfect, I like it!

DavidKnott:
@jacqueswww Thanks a lot for doing this, LGTM

PR #341, opened 2017-09-09
Move parser type tests into a types folder

DavidKnott:
### - What I did
Tests are an essential part of every project and can help users understand a project's functionality.  I'm planning on breaking all the tests out of `test_parser.py` to make Viper's test suite easier to navigate for both users and auditors.
I broke all type related tests in `test_parser.py` into a `types` folder and divided the tests by type (num, bytes, etc).
@JackieXu Started breaking out the tests, like you mentioned.
### - How I did it
I did this by moving type tests into their appropriate folders.
### - How to verify it
Look at the tests that've been removed from `test_parsery.py` and check that they've been moved to the appropriate folder/file.
None
### - Cute Animal Picture
image


fubuloubu:
I like it. Ultimately, I was thinking by feature more than by type, but types are features so... lol

I think `test_invalids.py` is a poor name. That file is ACTUALLY testing the parser, I wanted to call it something like `test_front_end.py` or maybe even break that out into separate files by type inside a `front_end/` folder. It's kind of a large file as is.

DavidKnott:
@fubuloubu I agree with you, I already broke out all of `test_parsery.py` and plan on doing the same with `test_invalids.py`.

JackieXu:
Thumbs up! Looks good to me, would indeed be great if you could do the same with `test_invalids.py` as well.

PR #345, opened 2017-09-10
Updates to function name and signature in ABI

### LINKED ISSUE #344: Remove inputs from abi name field.
### What's your issue about?
Viper's abi `name` field should not have a functions inputs attached to it for two reasons:
1) Duplicate information (the inputs are already in that abi elsewhere)
2) Abi uniformity, now the `name` field will match Solidity's

### How can it be fixed?

Change abi `name` to not have inputs

#### Cute Animal Picture

image

fubuloubu:
### - What I did
Update to function signature to have name be function name, added signature field to abi, fixed abi signature field to reflect return type as well. Fixes #344

ABI Change

### - Cute Animal Picture

baby walrus

### DIFF
--- a/viper/function_signature.py
+++ b/viper/function_signature.py
@@ -83,13 +83,16 @@ class FunctionSignature():
             assert canonicalize_type(output_type)
         # Get the canonical function signature
         sig = name + '(' + ','.join([canonicalize_type(parse_type(arg.annotation, None)) for arg in code.args.args]) + ')'
+        if output_type is not None:
+            sig += " -> " + canonicalize_type(output_type)
         # Take the first 4 bytes of the hash of the sig to get the method ID
         method_id = fourbytes_to_int(sha3(bytes(sig, 'utf-8'))[:4])
         return cls(name, args, output_type, const, payable, internal, sig, method_id)
 
     def to_abi_dict(self):
         return {
-            "name": self.sig,
+            "name": self.name,
+            "sig": self.sig,
             "outputs": [{"type": canonicalize_type(self.output_type), "name": "out"}] if self.output_type else [],
             "inputs": [{"type": canonicalize_type(arg.typ), "name": arg.name} for arg in self.args],
             "constant": self.const,
--- a/viper/function_signature.py
+++ b/viper/function_signature.py
@@ -90,7 +90,6 @@ class FunctionSignature():
     def to_abi_dict(self):
         return {
             "name": self.name,
-            "sig": self.sig,
             "outputs": [{"type": canonicalize_type(self.output_type), "name": "out"}] if self.output_type else [],
             "inputs": [{"type": canonicalize_type(arg.typ), "name": arg.name} for arg in self.args],
             "constant": self.const,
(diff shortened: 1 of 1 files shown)


fubuloubu:
Welp, I should've checked that. Apparently it was being used for gas estimation somehow in the tests...

fubuloubu:
@DavidKnott it passes now, but `sig.name` is used in a lot of suspect places in `parser.py`, I think as a proxy for determining uniqueness, so the meaning might have changed for it's usage.

fubuloubu:
I highlighted a few potential problem areas...

```diff
diff --git a/viper/parser.py b/viper/parser.py
index 623e035..dd986b6 100644
--- a/viper/parser.py
+++ b/viper/parser.py
@@ -238,6 +238,7 @@ def parse_tree_to_lll(code, origcode):
             add_gas += 30
             sig = FunctionSignature.from_definition(_def)
             sig.gas = sub[-1].total_gas
+            #Problem?
             sigs[sig.name] = sig
         o.append(['return', 0, ['lll', sub, 0]])
     return LLLnode.from_list(o, typ=None)
@@ -285,6 +286,7 @@ def parse_func(code, _globals, sigs, origcode, _vars={}):
         if arg.name in _globals:
             raise VariableDeclarationException("Variable name duplicated between function arguments and globals: "+arg.name)
     # Create a context
+    #Problem?
     context = Context(vars=_vars, globals=_globals, sigs=sigs,
                       return_type=sig.output_type, is_constant=sig.const, is_payable=sig.payable, origcode=origcode)
     # Copy calldata to memory for fixed-size arguments
@@ -293,6 +295,7 @@ def parse_func(code, _globals, sigs, origcode, _vars={}):
     if not len(sig.args):
         copier = 'pass'
     elif sig.name == '__init__':
+        #Problem?
         copier = ['codecopy', RESERVED_MEMORY, '~codelen', copy_size]
     else:
         copier = ['calldatacopy', RESERVED_MEMORY, 4, copy_size]
@@ -315,6 +318,7 @@ def parse_func(code, _globals, sigs, origcode, _vars={}):
     if sig.name == '__init__':
         o = LLLnode.from_list(['seq'] + clampers + [parse_body(code.body, context)], pos=getpos(code))
     else:
+        #Problem?
         method_id_node = LLLnode.from_list(sig.method_id, pos=getpos(code), annotation='%s' % sig.name)
         o = LLLnode.from_list(['if',
                                   ['eq', ['mload', 0], method_id_node],
@@ -322,6 +326,7 @@ def parse_func(code, _globals, sigs, origcode, _vars={}):
                                ], typ=None, pos=getpos(code))
     o.context = context
     o.total_gas = o.gas + calc_mem_gas(o.context.next_mem)
+    #Problem?
     o.func_name = sig.name
     return o
```

DavidKnott:
I don't think this is an issue because changing the `to_abi_dict` won't change the `FunctionSignature` itself (which is what's used in `parser.py`).  Why did you add the `sig` key?

fubuloubu:
@DavidKnott just so the prior usage was available. It does not have to be there however. Adding a field to the ABI isn't itself dangerous, but I guess being inconsistent can be a problem.

fubuloubu:
That's a fair point and probably why testing passes okay after I removed my change in the `from_definition()` function

DavidKnott:
@fubuloubu I don't think `sig` should be kept because all it's doing is duplicating information.  Also, I'd like to get Viper's abi as close to Solidity's as possible so that integrating with development frameworks (primarily Populous) goes as smoothly as possible.

fubuloubu:
Agreed, it's gone

DavidKnott:
@fubuloubu Thanks a lot! LGTM.

PR #353, opened 2017-09-17
Move readme to docs

DavidKnott:
### - What I did
Added `structure of contract` and `types` pages to the docs.

### - How I did it
I did it by looking at the Solidity docs for formatting and then added parts of the old readme file to the docs.

### - How to verify it
Clone down this branch and enter `make docs`

### - Cute Animal Picture
image

### DIFF
--- a/docs/structure-of-a-contract.rst
+++ b/docs/structure-of-a-contract.rst
@@ -1,0 +1,67 @@
+.. index:: contract, state variable, function, event, struct, enum, function;modifier
+
+.. _contract_structure:
+
+***********************
+Structure of a Contract
+***********************
+
+Contracts in Viper are contained within files, with each file being one smart-contract.  Files in Viper are similar to classes in object-oriented languages.
+Each file can contain declarations of :ref:`structure-state-variables`, :ref:`structure-functions`, and :ref:`structure-structs-types`.
+
+.. _structure-state-variables:
+
+State Variables
+===============
+
+State variables are values which are permanently stored in contract storage.
+
+::
+
+  storedData: num
+
+See the :ref:`types` section for valid state variable types and
+:ref:`visibility-and-getters` for possible choices for
+visibility.
+
+.. _structure-functions:
+
+Functions
+=========
+
+Functions are the executable units of code within a contract.
+
+::
+
+  @payable
+  function bid(): // Function
+    // ...
+  }
+
+:ref:`function-calls` can happen internally or externally
+and have different levels of visibility (:ref:`visibility-and-getters`)
+towards other contracts.
+
+
+
+.. _structure-structs-types:
+
+Structs Types
+=============
+
+Structs are custom defined types that can group several variables (see
+:ref:`structs` in types section).
+
+::
+
+    # Information about voters
+    voters: public({
+        # weight is accumulated by delegation
+        weight: num,
+        # if true, that person already voted
+        voted: bool,
+        # person delegated to
+        delegate: address,
+        # index of the voted proposal
+        vote: num
+    })
--- a/docs/structure-of-a-contract.rst
+++ b/docs/structure-of-a-contract.rst
@@ -1,4 +1,4 @@
-.. index:: contract, state variable, function, event, struct, enum, function;modifier
+.. index:: contract, state variable, function;
 
 .. _contract_structure:
 
@@ -42,26 +42,3 @@ Functions are the executable units of code within a contract.
 and have different levels of visibility (:ref:`visibility-and-getters`)
 towards other contracts.
 
-
-
-.. _structure-structs-types:
-
-Structs Types
-=============
-
-Structs are custom defined types that can group several variables (see
-:ref:`structs` in types section).
-
-::
-
-    # Information about voters
-    voters: public({
-        # weight is accumulated by delegation
-        weight: num,
-        # if true, that person already voted
-        voted: bool,
-        # person delegated to
-        delegate: address,
-        # index of the voted proposal
-        vote: num
-    })
--- a/docs/types.rst
+++ b/docs/types.rst
@@ -34,8 +34,7 @@ Operators:
 *  ``and`` (logical conjunction, "&&")
 *  ``or`` (logical disjunction, "||")
 *  ``==`` (equality)
-*  ``not ... == ... `` (inequality)
-*  ``!=`` (inequality)
+*  ``not ... == ... ``, ``!=`` (inequality)
 
 The operators ``or`` and ``and`` apply the common short-circuiting rules. This means that in the expression ``f(x) or g(y)``, if ``f(x)`` evaluates to ``true``, ``g(y)`` will not be evaluated even if it may have side-effects.
 
@@ -45,7 +44,7 @@ The operators ``or`` and ``and`` apply the common short-circuiting rules. This m
 Integers
 --------
 
-``num``:  a signed integer strictly between -2\*\*128 and 2\*\*128.
+``num``:  equivalent to``int128``, a signed integer strictly between -2\*\*127 and 2\*\*127-1.
 
 Operators:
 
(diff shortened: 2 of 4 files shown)


docs/structure-of-a-contract.rst
```diff
@@ -0,0 +1,67 @@
...
+
+:ref:`function-calls` can happen internally or externally
+and have different levels of visibility (:ref:`visibility-and-getters`)
+towards other contracts.
+
+
+
+.. _structure-structs-types:
+
+Structs Types
```
fubuloubu: I don't consider `structs` to be different than any other `global`, is there any reason to mention them separately in `docs/structure-of-a-contract.rst`? At the very least, if they were adjacent or in the same `global` sub-section, that would work for me

docs/types.rst
```diff
@@ -0,0 +1,145 @@
...
+
+``bool``: The possible values are constants ``true`` and ``false``.
+
+Operators:
+
+*  ``not`` (logical negation)
+*  ``and`` (logical conjunction, "&&")
+*  ``or`` (logical disjunction, "||")
+*  ``==`` (equality)
+*  ``not ... == ... `` (inequality)
```
fubuloubu: is `==` a typo here?

I would write `not`, `!=` (inequality)

docs/types.rst
```diff
@@ -0,0 +1,145 @@
...
+
+The operators ``or`` and ``and`` apply the common short-circuiting rules. This means that in the expression ``f(x) or g(y)``, if ``f(x)`` evaluates to ``true``, ``g(y)`` will not be evaluated even if it may have side-effects.
+
+.. index:: ! uint, ! int, ! integer
+
+
+Integers
+--------
+
+``num``:  a signed integer strictly between -2\*\*128 and 2\*\*128.
```
fubuloubu: `num` is `int128` behind the scenes, which makes the range [`-2**127`, `2**127-1`] technically (one bit for sign, and 0 counts as a value so highest is less one value)

docs/types.rst
```diff
@@ -0,0 +1,145 @@
...
+
+Operators:
+
+* Comparisons: ``<=``, ``<``, ``==``, ``!=``, ``>=``, ``>`` (evaluate to ``bool``)
+* Arithmetic operators: ``+``, ``-``, unary ``-``, unary ``+``, ``*``, ``/``, ``%`` (remainder)
+
+
+Decimals
+--------
+``decimal``:  a decimal fixed point value with the integer component being a signed integer strictly between -2\*\*128 and 2\*\*128 and the fractional component being ten decimal places
```
fubuloubu: simplify and say `decimal` has an integer component represented as a `num`, and 10 decimal places after that. (so you don't have to re-write the size of the integer portion here again)

docs/types.rst
```diff
@@ -0,0 +1,145 @@
...
+
+
+Time
+-----
+``timestamp``:  a timestamp value
+
+``timedelta``:  a number of seconds (note: two timedeltas can be added together, as can a timedelta and a timestamp, but not two timestamps)
```
fubuloubu: What underlying data types are `timedelta` and `timestamp`? (i.e. max `2**127` seconds, or whatever)

docs/types.rst
```diff
@@ -0,0 +1,145 @@
...
+
+
+Value
+------
+``wei_value``:  an amount of wei
+
+``currency_value``:  an amount of currency
```
fubuloubu: Explain "`currency_value` should be used to represent token values, or other assets where ether is traded for value" or something similar.

docs/types.rst
```diff
@@ -0,0 +1,145 @@
...
+Time
+-----
+``timestamp``:  a timestamp value
+
+``timedelta``:  a number of seconds (note: two timedeltas can be added together, as can a timedelta and a timestamp, but not two timestamps)
+
+
+Value
+------
+``wei_value``:  an amount of wei
```
fubuloubu: point to the Ether denomination chart [like this](https://github.com/ethereum/web3.js/blob/0.15.0/lib/utils/utils.js#L40) (but don't use that post, not official)

docs/types.rst
```diff
@@ -0,0 +1,145 @@
...
+``currency_value``:  an amount of currency
+
+
+
+.. _address:
+
+Address
+-------
+
+``address``: Holds a 20 byte value (size of an Ethereum address).
```
fubuloubu: Rewrite "Holds an Ethereum address (20 byte value)."

docs/types.rst
```diff
@@ -0,0 +1,145 @@
...
+
+.. _members-of-addresses:
+
+Members of Addresses
+^^^^^^^^^^^^^^^^^^^^
+
+* ``balance`` and ``send``
+
+It is possible to query the balance of an address using the property ``balance``
+and to send Ether (in units of wei) to an address using the ``send`` function:
```
fubuloubu: I didn't know this was possible! (`x.send(amt)`) Interesting.

docs/types.rst
```diff
@@ -0,0 +1,145 @@
...
+
+``type[length]``: finite list
+
+``bytes <= maxlen``: a byte array with the given maximum length
+
+
+Structs
+-------
+
+``{arg1:type, arg2:type...}``:  struct (can be accessed via struct.argname)
```
fubuloubu: Needs more. Move above section (`docs/structure-of-a-contract.rst`) here?

docs/types.rst
```diff
@@ -0,0 +1,145 @@
...
+    def foo(x: adress):
+        if (x.balance < 10 and self.balance >= 10):
+            x.send(10)
+
+
+.. index:: byte array, bytes32
+
+
+Fixed-size byte arrays
+----------------------
```
fubuloubu: Needs more info

docs/types.rst
```diff
@@ -0,0 +1,145 @@
...
+
+
+.. index:: !mapping
+
+Mappings
+========
+
+Mapping types are declared as ``_ValueType[_KeyType]``.
+Here ``_KeyType`` can be almost any type except for mappings, a contract, or a struct.
+``_ValueType`` can actually be any type, including mappings.
```
fubuloubu: I thought nested mappings was disallowed e.g. `num[address][address]` doesn't work?

docs/types.rst
DavidKnott: It's not, I'm so used to Solidity that I mix it up sometimes, though I do think it's clearer because it makes it a lot harder to mix up the `addr` and the `amt` then the current `send(addr, amt)`.

docs/types.rst
DavidKnott: As of now the following works:
```
hm: address[num][num]

def foo(x: num) -> address:
    self.hm[1][1] = self
    return self.hm[1][1]
```
Is that what you're talking about?

docs/types.rst
fubuloubu: Well, maybe it's worth considering changing that syntax if it matches solidity. Although, I think it should either be one or the other.

docs/types.rst
fubuloubu: That's a double-ly nested array, which I wasn't sure about it being explicitly allowed, but didn't hear that it was disallowed. I just tested:
```python
test: num[address][address]

def foo(x: address, y: address) -> num:
    return self.test[x][y]
```

No problems, which means I must be mistaken, but I could've sworn I remembered that @vbuterin said it should not be allowed because it would create an infinite keyspace or something. Anyways, I tested 10 levels deep and it still compiles:
```python
test: num[address][address][address][address][address][address][address][address][address][address]

def foo(x0: address, x1: address, x2: address, x3: address, x4: address, x5: address, x6: address, x7: address, x8: address, x9: address) -> num:
    return self.test[x0][x1][x2][x3][x4][x5][x6][x7][x8][x9]
```

DavidKnott:
@fubuloubu I made the changes you requested.  Could you take another look?

fubuloubu:
I like it, looks good to me!

DavidKnott:
@fubuloubu Thanks for the help, merging this in.

PR #354, opened 2017-09-18
Move parser integration tests into an integration folder

DavidKnott:
### - What I did
I broke the integration tests in `test_parser.py` into separate folders/files.
### - How I did it
I went through all the integration tests in `test_parser.py` and figured out where they belong.

### - How to verify it
Pull down my branch and run all the tests / go through my PR to make sure where I moved the integration tests makes sense.

None
### - Cute Animal Picture
image


JackieXu:
Looks good to me, are you also looking at `test_invalids.py`?

DavidKnott:
@JackieXu Yes I am, just trying to make small enough PR's that they're reviewable.

fubuloubu:
I like the changes.

I believe `test_invaids.py` should be in a different section(s) e.g. `front_end/`, `exceptions/`, etc. To more clearer delineate the subsets being tested. I also think there is a part of the current `test_parser.py` that belongs with the same sections of tests as the invalids because they test the front end more then they test functionality in `parser.py` (e.g. duplicate)

DavidKnott:
@fubuloubu I agree with you, merging this in.

Issue #352, opened 2017-09-14
Improve quality of `wallet` example


fubuloubu:
### What's your issue about?

I was working on a black box tester for the examples and realized the wallet example is pretty difficult to understand (in my opinion). It may also suffer from some older syntax rules that might not work anymore as it was written 2 months ago. I think there is a lot of potential to show off some of the more obscure methods available in viper through this contract, so I think if someone just went and cleaned it up (and added a LOT of comments about why certain hard numbers are chosen), it would be very helpful.

#### Cute Animal Picture

![chipmonks holding hands](https://i.pinimg.com/736x/d2/14/0e/d2140ea52dd96cbb329c503c986632c4--chip-and-dale-holding-hands.jpg)

fubuloubu:
Closing due to #356

PR #356, opened 2017-09-21
Add comments / format wallet example

DavidKnott:
### - What I did
I looked through the wallet example and wallet tests and tried to make them clearer and easier to read in response to #352.  @fubuloubu Happy to add more comments if you see anything I missed.

### - How I did it
1) I changed `addrs` in the wallets example to `owners`
2) Moved the wallet test file into the tests folder
3) Added comments to wallet and wallet tests
4) Added `assert_tx_failed()` to `setup_transaction_tests.py`

### - How to verify it
1) Look over my PR
2) Pull down my branch and run make test (now the wallets tests are running)
None
### - Cute Animal Picture
image


examples/wallet/wallet.v.py
```diff
@@ -1,25 +1,46 @@
...
-def __init__(_addrs: address[5], _threshold: num):
+def __init__(_owners: address[5], _threshold: num):
     for i in range(5):
-        if _addrs[i]:
-            self.addrs[i] = _addrs[i]
+        if _owners[i]:
+            self.owners[i] = _owners[i]
     self.threshold = _threshold
 
 def approve(_seq: num, to: address, value: wei_value, data: bytes <= 4096, sigdata: num256[3][5]) -> bytes <= 4096:
```
fubuloubu: This function approves transactions (potentially sending `value` wei out). How does this wallet gain value to send?

examples/wallet/wallet.v.py
```diff
@@ -1,25 +1,46 @@
...
+            assert ecrecover(h2, sigdata[i][0], sigdata[i][1], sigdata[i][2]) == self.owners[i]
+            # For every valid signature increase the number of approvals by 1
             approvals += 1
+    # Throw if the number of approvals is less then the number of approvals required (the threshold)
     assert approvals >= self.threshold
+    # The transaction has been approved
+    # Increase the number of approved transactions by 1
     self.seq += 1
+    # Use raw_call to send the transaction
     return raw_call(to, data, outsize=4096, gas=3000000, value=value)
```
fubuloubu: `gas` seems randomly chosen. What if I don't send enough to afford that gas cost in conjunction with that value? I like the idea of `approve()` being `@payable`, and `gas` is set to `msg.value - value`. Wouldn't this mean the sender should get their refund? I'm not sure.

DavidKnott: I agree with you, I'll make `approve()` `@payable` .  I also like the idea of making it so that the `msg.sender` gets any excess `msg.value` leftover after their transaction calls.

DavidKnott:
@fubuloubu I made some of the changes you suggested, but chose to keep the gas priced fixed for simplicity sake.

examples/wallet/wallet.v.py
fubuloubu: @DavidKnott is there a method to obtain/return the leftover gas from a `raw_call`?

fubuloubu:
Awesome, looks great. As long as it's understood that not exploring the gas cost setting is a potential source of error. Are there functional/technical roadblocks to implementing that?

I think as it stands now, if there is less value supplied to make the call then there is gas spent, `raw_call` should fail and the entire call should throw. If there is more supplied then needed, this contract keeps the rest. Does that make sense?

DavidKnott:
@fubuloubu I coded where the user could dynamically set the gas they wanted to spend and any leftover gas was forwarded back to them but when I looked back at it after I finished I felt that this example was complicated enough without dynamic gas.

fubuloubu:
@DavidKnott I agree with you, but I'm thinking this particular example is supposed to show (eventually) how much easier to read the code for a fully featured wallet would look in viper (as opposed to Solidity). That being said, I think it's a work in progress and doesn't necessary need to be finished now.

DavidKnott:
@fubuloubu I agree with you, and am happy to revisit this down the road.

PR #369, opened 2017-09-29
Adds viper serve

jacqueswww:
### - What I did

Allowed viper to sit as an HTTP server and compile code through an http interface. 
Later I want to integrate a version of remix or similar to post code to it, perhaps add autoreload for the compiler development side. But for now it's a nice minimal working unit ;)
### - How I did it

Used simple stdlib HTTPServer as a base and worked from there.

### - How to verify it
Run 
```
viper serve
```
And do a JSON HTTP POST to the `/compile' path:
```
curl -X POST \
  http://localhost:8000/compile \
  -H 'cache-control: no-cache' \
  -H 'content-type: application/json' \
  -d '{
	"code": "\ndef test() -> num256:\n    return 1\n"
}'
{"status": "failed", "message": "line 3: Unsupported type conversion: num(*) to num256\n    return 1\n-----------^"}
```

Adds viper serve.


fubuloubu:
I like the addition, looks great!

Does `server.py` need to live in the `viper/` source directory, or can it live in `bin/` instead?

jacqueswww:
Sure! I wasn't sure where to put it, will look at adding it to bin/.  Let me give it a shot, but it may not play nicely with setup.py.

fubuloubu:
Well, how were intending on using it? My thoughts are that an online service is not quite the core feature of the compiler specifically, but I understand how that function is important for the online IDE. If implementing into the core package makes it work for the Online IDE, I'm on board with that, but if it can be a separate script I believe that would be preferable. Even if that script is a demonstration for how to implement it for a website.

Just my 2 cents.

jacqueswww:
@fubuloubu hmmm, yes you are completely right! Let's make it a seperate bin/ script. How does `bin/viper_serve` sound ? This way the traditional compile stays seperate ;)

fubuloubu:
Sounds good to me, whatever you'd like to call it.

fubuloubu:
Nice!

DavidKnott:
@jacqueswww When you rebase the logging tests should pass.

jacqueswww:
@DavidKnott @fubuloubu I only saw that issue after this PR was created. Not sure if it really interferes, the backend part of it's basically the same (except this one can be packaged nicely with viper because it has no external dependencies).

setup.py
```diff
@@ -29,5 +29,5 @@
         'pytest-runner', # Must be after pytest-cov or it will not work
         #           due to https://github.com/pypa/setuptools/issues/196
     ],
-    scripts=['bin/viper']
+    scripts=['bin/viper', 'bin/viper-serve']
```
fubuloubu: Undo this change, and testing will pass

fubuloubu:
When you rebased, I think it removed your changes.

setup.py
jacqueswww: Ha, I see ! Will fix.

DavidKnott:
@jacqueswww I like the idea of having a built-in backend to facilitate people playing around with Viper in the browser.  Though I think down the road as Viper matures it may make sense to make it this as an add-on it's own repo.

jacqueswww:
@DavidKnott yes, I agree it does feel like it might belong in another repo. Here is my vision: we need to ship some type of server (that follows the same REST that the most popular online IDE uses at the time). The reason is simple, I would never trust an external service to compile my code, and I am sure a lot of people wouldn't.

So basically we can have 'browser-only' hosted online IDE (let's say viper-remix.ethereum.org) for viper (whichever IDE project it might be in the end). Which you can then select "Select Viper Compiler -> Local", http://localhost:8080/. You then pip install viper locally, and run viper-serve  and then point to locally running viper server.  In a similar way you then embed the deployment to testnet/mainnet in the IDE as well. What do you guys think ?

jacqueswww:
Been thinking of this on-and-off all day. On the other hand again, if we build an IDE that can be setup in a single pip install; and run we can drop this flow as well.

DavidKnott:
I think the eventual goal will be to create an online ide that's not reliant on a server.  I've looked into this a little and haven't found any easy ways to compile python in the browser.  

Depending how long that idea takes `pip install; idename runserver` seems like a good way to maintain modularity, simplicity (it's one line), and security (Etherem can release the idea.)

fubuloubu:
How about a smart contract that compiles smart contracts.... Isn't that the dream of all languages? Lol

jacqueswww:
@fubuloubu ah nooo ... inception! :laughing:

DavidKnott:
@jacqueswww LGTM

PR #372, opened 2017-10-02
Refactor parser parse_stmt method

DavidKnott:
### - What I did
I broke the `parse_stmt` method out into it's own class.

### - How I did it
1) Turned the `parse_stmt` conditional into a dictionary
2) Move `parser.py` into a parser folder
3) Create a seperate `stmt.py` file

### - How to verify it
Look through my PR to make sure my refactors increase simplicity and don't change any functionality.

None
### - Cute Animal Picture
image

### DIFF
--- a/viper/parser/stmt.py
+++ b/viper/parser/stmt.py
@@ -1,0 +1,269 @@
+import ast
+
+from viper.exceptions import (
+    ConstancyViolationException,
+    StructureException,
+    TypeMismatchException,
+    VariableDeclarationException,
+)
+from viper.functions import (
+    stmt_dispatch_table,
+)
+from .parser_utils import LLLnode
+from .parser_utils import (
+    getpos,
+    make_byte_array_copier,
+    base_type_conversion,
+    unwrap_location
+)
+from viper.types import (
+    BaseType,
+    ByteArrayType,
+    ListType,
+)
+from viper.types import (
+    get_size_of_type,
+    is_base_type,
+    parse_type,
+)
+from viper.types import (
+    are_units_compatible,
+    set_default_units,
+)
+
+
+class Stmt(object):
+    def __init__(self, stmt, context):
+        self.stmt = stmt
+        self.context = context
+        self.stmt_table = {
+            ast.Expr: self.expr,
+            ast.Pass: self.parse_pass,
+            ast.AnnAssign: self.ann_assign,
+            ast.Assign: self.assign,
+            ast.If: self.parse_if,
+            ast.Call: self.call,
+            ast.Assert: self.parse_assert,
+            ast.For: self.parse_for,
+            ast.AugAssign: self.aug_assign,
+            ast.Break: self.parse_break,
+            ast.Return: self.parse_return,
+        }
+        stmt_type = self.stmt.__class__
+        if stmt_type in self.stmt_table:
+            self.lll_node = self.stmt_table[stmt_type]()
+        elif isinstance(stmt, ast.Name) and stmt.id == "throw":
+            self.lll_node = LLLnode.from_list(['assert', 0], typ=None, pos=getpos(stmt))
+        else:
+            raise StructureException("Unsupported statement type", stmt)
+
+    def expr(self):
+        return Stmt(self.stmt.value, self.context).lll_node
+
+    def parse_pass(self):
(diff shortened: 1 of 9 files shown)


fubuloubu:
2 Failed tests in Travis. Why?

DavidKnott:
It's because i hardcoded the timestamps in for the logging tests and version 2.0.5 has diferent starting timetamps.  My other PR fixes this.

viper/parser/stmt.py
```diff
@@ -0,0 +1,269 @@
...
+            ast.AnnAssign: self.ann_assign,
+            ast.Assign: self.assign,
+            ast.If: self.parse_if,
+            ast.Call: self.call,
+            ast.Assert: self.parse_assert,
+            ast.For: self.parse_for,
+            ast.AugAssign: self.aug_assign,
+            ast.Break: self.parse_break,
+            ast.Return: self.parse_return,
+        }
```
jacqueswww: Just something small thing I would opt for readability. I like to have something like stmt_table on the class straight: 

```python

class Stmt():
   stmt_table = { ... }

File: viper/parser/stmt.py [hunk 2]
```diff
@@ -0,0 +1,269 @@
...
+        return LLLnode.from_list(['if', parse_value_expr(self.stmt.test, self.context), parse_body(self.stmt.body, self.context)] + add_on, typ=None, pos=getpos(self.stmt))
+
+    def call(self):
+        from .parser import (
+            parse_expr,
+            pack_arguments,
+            pack_logging_data,
+            pack_logging_topics,
+            external_contract_call_stmt,
+        )
```
jacqueswww: Coul we add # TODO here so we can evaluate the inline imports later  (I think once we factor everything from parse.py it should be doable) ;)

viper/parser/stmt.py
```diff
@@ -0,0 +1,269 @@
...
+                                        typ=None, pos=getpos(self.stmt))
+        elif isinstance(self.stmt.func, ast.Attribute) and isinstance(self.stmt.func.value, ast.Call):
+            return external_contract_call_stmt(self.stmt, self.context)
+        elif isinstance(self.stmt.func, ast.Attribute) and self.stmt.func.value.id == 'log':
+            if self.stmt.func.attr not in self.context.sigs['self']:
+                raise VariableDeclarationException("Event not declared yet: %s" % self.stmt.func.attr)
+            event = self.context.sigs['self'][self.stmt.func.attr]
+            topics, stored_topics, event, data = pack_logging_topics(event, self.stmt.args, self.context)
+            inargs, inargsize, inarg_start = pack_logging_data(event, data, self.context)
+            return LLLnode.from_list(['seq', inargs, stored_topics, ["log" + str(len(topics)), inarg_start, inargsize] + topics], typ=None, pos=getpos(self.stmt))
```
jacqueswww: Just for readability could we add some new lines after the return statements ;) I also had another idea, now that we have our own Stmt class we could perhaps replace all those isinstance() calls with self.stmt with `self.is_attribute` and `self.func_name` and `self.is_func` properties etc.. Of course this all to your liking, just a small style thing.

viper/parser/stmt.py
DavidKnott: I agree 100% the inline imports are only temporary.

viper/parser/stmt.py
DavidKnott: I like that a lot better, great suggestion!

viper/parser/stmt.py
DavidKnott: Could you write a quick example for how `self.is_attribute` would work?  Would it be syntactic sugar for `isinstance`?

viper/parser/stmt.py
jacqueswww: Sure, makes sense, since it's just another way to do a bunch if/elifs.

viper/parser/stmt.py
jacqueswww: So the idea was just as a bunch of handy aliases:

```

class Stmt():

    @property
    def is_attribute(self):
        return isinstance(self.stmt.func, ast.Attribute)

    @property
    def is_name(self):
        return isinstance(self.stmt, ast.Name) 

    @property
    def func_id(self):
        return getattr(self.stmt.func.value, 'id', None)
```

PR #380, opened 2017-10-05
Don't use fancy quotes

dvf:
### - What I did

Changed fancy quotes to normal quotes. If you copy paste, it will just export the variable with fancy quotes... and you won't know why `scrypt` is failing to install on Mac.

### - How I did it

`grep -r "”" . -l`

### - How to verify it

Just check... also if you're developing on a Mac, disable this by default under `Preferences > Keyboard > Text`. 

- Replaced fancy quotes with normal quotes in the installation documentation.

### DIFF
--- a/docs/installing-viper.rst
+++ b/docs/installing-viper.rst
@@ -91,8 +91,8 @@ install and test commands
 
     Use the following commands:
     ::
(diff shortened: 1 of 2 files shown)


DavidKnott:
LGTM

PR #384, opened 2017-10-07
Added safe remote purchase example

chgue:
### - What I did
Implemented and cleaned up the "Safe remote purchase" from the Solidity documentation (https://solidity.readthedocs.io/en/develop/solidity-by-example.html#safe-remote-purchase)
Edit: Fixed missing spaces in last line.
### - How I did it
Translated the code from Solidity to Viper.
Differences:
- Used selfdestruct() where applicable
- ~~Locked state is determined by the contract balance~~ Didn't pass tests and could be exploited.
- Comments explaining transaction details added
- Value does not need to be a multiple of two

### - How to verify it
Auditing code. Test cases could be implemented.

Added "Safe remote purchase" example

### - Cute Animal Picture
gorilla

### DIFF
--- a/examples/safe_remote_purchase/safe_remote_purchase.v.py
+++ b/examples/safe_remote_purchase/safe_remote_purchase.v.py
@@ -1,0 +1,38 @@
+#Safe Remote Purchase (https://github.com/ethereum/solidity/blob/develop/docs/solidity-by-example.rst) ported to viper and optimized
+
+#Rundown of the transaction:
+#1. Seller posts item for sale and posts safety deposit of double the item value. Balance is 2*value.
+#(1.1. Seller can reclaim deposit and close the sale as long as nothing was purchased.)
+#2. Buyer purchases item (value) plus posts an additional safety deposit (Item value). Balance is 4*value
+#3. Seller ships item
+#4. Buyer confirms receiving the item. Buyer's deposit (value) is returned. Seller's deposit (2*value) + items value is returned. Balance is 0.
+
+value: public(wei_value) #Value of the item
+seller: public(address)
+buyer: public(address)
+
+@constant
+def unlocked() -> bool: #Is a refund possible for the seller?
+    return (self.balance == self.value*2)
+    
+@payable
+def __init__():
+    self.value = msg.value / 2 #Seller initializes contract by posting a safety deposit of 2*value of the item up for sale
+    self.seller = msg.sender
+    
+def abort():
+    assert self.unlocked() == true #Is the contract still refundable
+    assert msg.sender == self.seller #Only seller can refund his deposit before any buyer purchases the item
+    selfdestruct(self.seller) #Refunds seller, deletes contract
+
+@payable
+def purchase():
+    assert self.unlocked() == true #Contract still open (item still up for sale)?
+    assert msg.value == (2*self.value) #Is the deposit of correct value?
+    self.buyer = msg.sender
+    
+def recieved():
+    assert self.unlocked() == false #Is the item already purchased and pending confirmation of buyer
+    assert msg.sender == self.buyer 
+    send(self.buyer, self.value) #Return deposit (=value) to buyer
+selfdestruct(self.seller) #Returns deposit (=2*value) and the purchase price (=value)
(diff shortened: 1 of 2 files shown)


examples/safe_remote_purchase/safe_remote_purchase.v.py
```diff
@@ -0,0 +1,38 @@
...
+    selfdestruct(self.seller) #Refunds seller, deletes contract
+
+@payable
+def purchase():
+    assert self.unlocked() == true #Contract still open (item still up for sale)?
+    assert msg.value == (2*self.value) #Is the deposit of correct value?
+    self.buyer = msg.sender
+    
+def recieved():
+    assert self.unlocked() == false #Is the item already purchased and pending confirmation of buyer
```
fubuloubu: `self.unlocked() == false` is the same as `not self.unlocked()`

examples/safe_remote_purchase/safe_remote_purchase.v.py
```diff
@@ -0,0 +1,38 @@
+#Safe Remote Purchase (https://github.com/ethereum/solidity/blob/develop/docs/solidity-by-example.rst) ported to viper and optimized
+
+#Rundown of the transaction:
+#1. Seller posts item for sale and posts safety deposit of double the item value. Balance is 2*value.
+#(1.1. Seller can reclaim deposit and close the sale as long as nothing was purchased.)
+#2. Buyer purchases item (value) plus posts an additional safety deposit (Item value). Balance is 4*value
+#3. Seller ships item
+#4. Buyer confirms receiving the item. Buyer's deposit (value) is returned. Seller's deposit (2*value) + items value is returned. Balance is 0.
+
+value: public(wei_value) #Value of the item
```
fubuloubu: @DavidKnott actually, since `wei_value` is the lowest divisible unit, perhaps it IS a good idea to ensure it is a multiple of 2 e.g. `assert (msg.value % 2) == 0` in `__init__()`. What do you think?

DavidKnott: You're right, even though Viper supports decimals the `wei_value` requires the `assert (msg.value % 2) == 0`.
My bad @chgue.

tests/examples/safe_remote_purchase/test_safe_remote_purchase.py
```diff
@@ -0,0 +1,64 @@
...
+#Rundown of the transaction:
+#1. Seller posts item for sale and posts safety deposit of double the item value. Balance is 2*value.
+#(1.1. Seller can reclaim deposit and close the sale as long as nothing was purchased.)
+#2. Buyer purchases item (value) plus posts an additional safety deposit (Item value). Balance is 4*value
+#3. Seller ships item
+#4. Buyer confirms receiving the item. Buyer's deposit (value) is returned. Seller's deposit (2*value) + items value is returned. Balance is 0.
+import pytest
+from ethereum.tools import tester 
+from ethereum import utils
+
```
DavidKnott: Use `tests/setup_transaction_tests.py` to help with deployment (it already has `assert_tx_failed`).  I'd consider using `get_contract` as well.

tests/examples/safe_remote_purchase/test_safe_remote_purchase.py
```diff
@@ -0,0 +1,64 @@
...
+    #Check if item value is set correctly (Half of deposit)
+    assert srp_tester.c.get_value() == 1
+    #Check if unlocked() works correctly after initialization
+    assert srp_tester.c.unlocked() == True
+"""
+def test_abort(srp_tester):
+    
+    #Check if unlocked
+    assert 
+    """
```
DavidKnott: Seems like this is the start of the testing the `abort()` function, just finish it.

tests/examples/safe_remote_purchase/test_safe_remote_purchase.py
```diff
@@ -0,0 +1,64 @@
...
+def test_purchase(srp_tester, assert_tx_failed):
+    srp_tester.c = srp_tester.s.contract(contract_code, language = "viper", args = [], value=2)
```
DavidKnott: I'd check the balances of the `buyer` and `seller` at the begining of the contract, once it's deployed, after it's aborted, and after `purchase()` is called to verify that the Ether is going where it should be.

chgue: Good idea, implemented this.

DavidKnott:
LGTM

PR #386, opened 2017-10-08
External contract calls cannot be used internally

DavidKnott:
### - What I did
I added a check to external contract calls so that they can't be used internally.

### - How I did it
I did this by checking that the `external_address != current_address`.

### - How to verify it
Look at the test I wrote.
None
### - Cute Animal Picture
image

### DIFF
--- a/tests/parser/features/test_external_contract_calls.py
+++ b/tests/parser/features/test_external_contract_calls.py
@@ -199,6 +199,34 @@ def __init__(arg1: address):
     print('Successfully executed a multiple external contract calls')
 
 
+def test_invalid_external_contract_call_to_the_same_contract(assert_tx_failed):
+    contract_1 = """
+def bar() -> num:
+    return 1
+    """
+    contract_2 = """
+class Bar():
+    def bar() -> num: pass
+
+def bar() -> num:
+    return 1
+
+def _stmt(x: address):
+    Bar(x).bar()
+
+def _expr(x: address) -> num:
+    return Bar(x).bar()
+    """
+    t.s = t.Chain()
+
+    c1 = get_contract(contract_1)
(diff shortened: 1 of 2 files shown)


tests/parser/features/test_external_contract_calls.py
```diff
@@ -199,6 +199,28 @@ def __init__(arg1: address):
...
 
+def test_invalid_external_contract_call_to_the_same_contract(assert_tx_failed):
+    contract = """
+class Bar():
+    def bar() -> num: pass
+
+def bar() -> num:
+    return 1
+
+def _expr(x: address) -> num:
```
fubuloubu: What's the difference between `_stmt` and `_expr`?

tests/parser/features/test_external_contract_calls.py
```diff
@@ -199,6 +199,28 @@ def __init__(arg1: address):
...
+    return Bar(x).bar()
+
+def _stmt(x: address) -> num:
+    return Bar(x).bar()
+    """
+    t.s = t.Chain()
+    c = get_contract(contract)
+    c._expr(t.a1)
```
fubuloubu: This should fail because we passed an account address instead of a contract address IMO.
Maybe insert `assert x.is_contract` in the underlying module for using external contracts?

DavidKnott: I agree with you, this pr is just to make sure external contract calls to the same contract don't work, I'll make another one that throws if there's no contract.

tests/parser/features/test_external_contract_calls.py
DavidKnott: I fixed this.

PR #387, opened 2017-10-08
Fix file pathing to parser.py

DavidKnott:
### - What I did
Fix fails paths so that `viper` and `viper-serv` commands work.
### - How I did it
I did this by looking at where `parser.py` was located.
### - How to verify it
Look at my PR
None
### - Cute Animal Picture
image


jacqueswww:
LGTM, I did the exact some thing here ;)

PR #388, opened 2017-10-09
Fix parse_to_lll import path.

jacqueswww:
### - What I did

bin/viper needed the full path to find the parse_to_lll function, I think this happened after the parser.py refactoring.

### - How to verify it
Install and run viper.

N/A


jacqueswww:
Closing as @DavidKnott already did this ;)

PR #391, opened 2017-10-10
ERC20 num-style example

jacqueswww:
### - What I did

- Copied a ERC20 example that we now can implement num-style :)

### - How I did it

- Looked at various examples on wiki pages.

### - How to verify it

 Read tests, or you can take the bytecode + abi and deploy it to geth, and view in ethereumwallet or similar.

Adds num ERC-20 example.


tests/examples/token/test_vipercoin.py
```diff
@@ -0,0 +1,177 @@
...
+
+
+@pytest.fixture
+def assert_tx_failed():
+    def assert_tx_failed(tester, function_to_test, exception=tester.TransactionFailed):
+        initial_state = tester.s.snapshot()
+        with pytest.raises(exception):
+            function_to_test()
+        tester.s.revert(initial_state)
+    return assert_tx_failed
```
fubuloubu: `assert_tx_failed()` exists in `setup_transaction_tests.py` Why duplicate it here?

tests/examples/token/test_vipercoin.py
```diff
@@ -0,0 +1,177 @@
...
+    a0 = token_tester.accounts[0]
+    a1 = token_tester.accounts[1]
+    k1 = token_tester.k1
+    a2 = token_tester.accounts[2]
+    k2 = token_tester.k2
+    contract = token_tester.c
+
+    # Allow 10 token transfers. Account a1 is allowed to spend 10 tokens of a0's account.
+    ALLOWANCE = 10
+    assert contract.approve(a1, 10) is True
```
fubuloubu: Use `ALLOWANCE` instead of 10 here

tests/examples/token/test_vipercoin.py
```diff
@@ -0,0 +1,177 @@
...
+
+    logs = chain.head_state.receipts[-1].logs[-1]
+    event_id = utils.bytes_to_int(utils.sha3(bytes('Transfer(address,address,uint256)', 'utf-8')))
+
+    assert logs.topics[0] == event_id
+    assert contract.translator.event_data[event_id] == {
+        'types': ['address', 'address', 'uint256'],
+        'name': 'Transfer', 'names': ['_from', '_to', '_value'],
+        'indexed': [True, True, False], 'anonymous': False
+    }
```
fubuloubu: @DavidKnott this seems crazy to me, it would be nice to have some event helpers in the viper module to decode all of this nonsense more easily.

DavidKnott: @fubuloubu If I'm understanding you correctly, you're saying you'd like to have test helper to test logs in a clearer way (hiding the specific checks within the helper).

tests/examples/token/test_vipercoin.py
```diff
@@ -0,0 +1,177 @@
...
+        with pytest.raises(exception):
+            function_to_test()
+        tester.s.revert(initial_state)
+    return assert_tx_failed
+
+
+def test_initial_state(token_tester):
+    assert token_tester.c.totalSupply() == TOKEN_TOTAL_SUPPLY == token_tester.c.balanceOf(token_tester.accounts[0])
+    assert token_tester.c.balanceOf(token_tester.accounts[1]) == 0
+
```
DavidKnott: I'd test for all inputs (symbol, name, etc).  It's trivial but I think it provides a bit of clarity.

examples/token/vipercoin.v.py
```diff
@@ -0,0 +1,95 @@
+# Viper Port of MyToken
```
DavidKnott: Can you change the file path to `examples/token/vipercoin.v.py` as I have a feeling we'll have multiple tokens examples down the road.

examples/token/vipercoin.v.py
```diff
@@ -0,0 +1,95 @@
...
+        log.Transfer(_from, _to, as_num256(_value))  # log transfer event.
+
+        return True
+    else:
+        return False
+
+
+# Allow _spender to withdraw from your account, multiple times, up to the _value amount.
+# If this function is called again it overwrites the current allowance with _value.
+def approve(_spender: address, _amount: num(num256)) -> bool:
```
DavidKnott: I'd love it if you included the below warning in the comments above approve:
NOTE: To prevent attack vectors like the one described here and discussed here, clients SHOULD make sure to create user interfaces in such a way that they set the allowance first to 0 before setting it to another value for the same spender. THOUGH The contract itself shouldn't enforce it, to allow backwards compatilibilty with contracts deployed before.

tests/examples/token/test_vipercoin.py
fubuloubu: Yeah, I guess Web3 handles that for any production uses, a test helper here would be adequate.

tests/examples/token/test_vipercoin.py
jacqueswww: :+1:

tests/examples/token/test_vipercoin.py
jacqueswww: :+1: good catch.

examples/token/vipercoin.v.py
jacqueswww: Wait you mean `/tokens` right ? ;)

examples/token/vipercoin.v.py
```diff
@@ -0,0 +1,95 @@
...
+    self.decimals = _decimals
+    self.totalSupply = _initialSupply * 10 ** _decimals
+    self.balances[msg.sender] = self.totalSupply
+
+
+@constant
+def symbol() -> bytes32:
+
+    return self.symbol
+
```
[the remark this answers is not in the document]
jacqueswww: Unfortunately it's not the same on the ABI level, that would add a function singature of `get_symbol` not `symbol`. (also I am using bytes32 for now, until we do a string type)

examples/token/vipercoin.v.py
jacqueswww: :+1:

tests/examples/token/test_vipercoin.py
jacqueswww: :+1:

jacqueswww:
@DavidKnott @fubuloubu I made all the changes mentioned. Thanks for the valuable feedback! The only one I left out was the log helpers, I wasn't exactly sure how we wanted this util function look like. My 2c worth would be something like 'get_last_event(chain, event_signature)` which would return the first decoded event that matches the event signature. Either way figure this probably a scope of another PR :)

fubuloubu:
@jacqueswww For sure another PR, was just a thought I had looking at the 8 lines of code it took you do something that should only take one for testing purposes.

tests/examples/tokens/test_vipercoin.py
```diff
@@ -0,0 +1,177 @@
...
+    a1 = token_tester.accounts[1]
+    k1 = token_tester.k1
+    a2 = token_tester.accounts[2]
+
+    contract = token_tester.c
+    chain = token_tester.s
+
+    assert contract.transfer(a1, 1) is True
+
+    logs = chain.head_state.receipts[-1].logs[-1]
```
fubuloubu: I made a log helper fixture that you should use here: #394

jacqueswww: @fubuloubu Done, thanks! Looks much better now ;)

examples/token/vipercoin.v.py
DavidKnott: Exactly :)

examples/token/vipercoin.v.py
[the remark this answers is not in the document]
DavidKnott: @jacqueswww You're 100% right :)

DavidKnott:
@jacqueswww Thanks for making the changes, as soon as it's passing I'll merge.

jacqueswww:
@DavidKnott done :D :man_dancing:

DavidKnott:
Thanks, merging it in.

PR #392, opened 2017-10-10
Maxmemory constants

jacqueswww:
### - What I did

Created 3 constants structures to make handling of the clamp-addresses easier.
Adding additional limits should be much easier now, but most importantly this helps in proper handling of 2** limits, for example optimizer.py never got updated as you can see in the diff.

### - How I did it

Created MemoryPositions,SizeLimits and LOADED_LIMIT_MAP. 

### - How to verify it

Tests verify it. Reading the changes in viper/utils.py especially.

N/A

### DIFF
--- a/viper/parser/parser.py
+++ b/viper/parser/parser.py
@@ -51,12 +51,8 @@ from viper.types import (
 )
 from viper.utils import (
     DECIMAL_DIVISOR,
-    RESERVED_MEMORY,
-    ADDRSIZE_POS,
-    MAXNUM_POS,
-    MINNUM_POS,
-    MAXDECIMAL_POS,
-    MINDECIMAL_POS,
+    MemoryPositions,
+    LOADED_LIMIT_MAP
 )
 from viper.utils import (
     bytes_to_int,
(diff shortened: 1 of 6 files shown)


viper/parser/parser.py
```diff
@@ -225,22 +221,17 @@ def get_contracts_and_defs_and_globals(code):
...
-                                        ['mstore', 28, ['calldataload', 0]],
-                                        ['mstore', ADDRSIZE_POS, 2**160],
-                                        ['mstore', MAXNUM_POS, 2**127 - 1],
-                                        ['mstore', MINNUM_POS, -2**127],
-                                        ['mstore', MAXDECIMAL_POS, (2**127 - 1) * DECIMAL_DIVISOR],
-                                        ['mstore', MINDECIMAL_POS, (-2**127) * DECIMAL_DIVISOR],
-                                    ], typ=None)
+initialiser_list = ['seq', ['mstore', 28, ['calldataload', 0]]]
+initialiser_list += [['mstore', pos, limit_size] for pos, limit_size in LOADED_LIMIT_MAP.items()]
+initializer_lll = LLLnode.from_list(initialiser_list, typ=None)
```
DavidKnott: I think this should be `initializer_list`.  Also it's easier for me to read if it's written like this:
```
initializer_lll = LLLnode.from_list(['seq', 
                                        ['mstore', 28, ['calldataload', 0],
                                        [['mstore', pos, limit_size] for pos, limit_size in LOADED_LIMIT_MAP.items()],
                                      ], typ=None)

jacqueswww: @DavidKnott sure! That was clearly my EN-GB kicking in :smile_cat:

jacqueswww: For the second part, unfortunately that would create a second sub-list, which isn't the same as the original sequence. I wasn't able to fine a more readable way to append the 'dynamic' LOADED_LIMIT_MAP list into the 'seq'.

DavidKnott:
LGTM

PR #394, opened 2017-10-12
Added fixture for helping get logs

fubuloubu:
### - What I did

Added a fixture for testing logs

### - How to verify it

Use it. Let me know if there are problems

NOTE: @DavidKnott, I thought about re-arranging all the code in 'parser/features/test_logging.py' but then I realized you would want to test more of the internals in there. I believe this fixture should probably not be used there, but basically anywhere else logs are used.

### - Cute Animal Picture

> Hooray

### DIFF
--- a/tests/setup_transaction_tests.py
+++ b/tests/setup_transaction_tests.py
@@ -105,7 +105,7 @@ curve_order = 218882428718392752222464057452572750885483644004160343436982041865
 
 @pytest.fixture
 def get_log():
-    def get_log(chain, contract, event_name):
+    def get_log(tester, contract, event_name):
         event_ids_w_name = [k for k, v in \
                 contract.translator.event_data.items() if v["name"] == event_name]
         assert len(event_ids_w_name) == 1, \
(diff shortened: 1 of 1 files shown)


DavidKnott:
@fubuloubu I think not using the helper for `parser/features/test_logging.py` was a good call.  What do you think about making it so that the developer can get whatever log they want as opposed to only the most recent one?

fubuloubu:
Well, for testing purposes you should be checking immediately after the log call was made. Although, I could see a scenario where two logs fire off, and now you can't access one. Maybe a slight modification where it just gets the last log available matching the name you're looking for?

DavidKnott:
@fubuloubu Fair point, LGTM

fubuloubu:
I'll keep this branch and think of some alternative ways that lets you get multiple logs, if desired (e.g. if you specify a log name, you get a list of logs with that name and do `logs[-1]` to get the last). If you don't specify a name, you get all logs, grouped by name (e.g. `logs = {'Log1' : [list...], 'Log2' : [list...], etc.]`)

PR #396, opened 2017-10-12
Upgraded the required version of pyethereum to 2.1.0

hwwhww:
### - What I did
Upgraded the required version of `pyethereum` from `2.0.5` to `2.1.0`.

### - How I did it
Simply updated `setup.py`

### - How to verify it
Passed CI

None

### - Cute Animal Picture

viper

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -23,7 +23,7 @@ setup(
     license=license,
     packages=find_packages(exclude=('tests', 'docs')),
     install_requires=[
-        'ethereum==2.0.5',
+        'ethereum==2.1.0',
         'bumpversion',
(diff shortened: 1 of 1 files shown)


fubuloubu:
@DavidKnott do you think we should do `ethereum >=2.1.0, < 2.2.0` instead? e.g. have to determine upgrades work for each minor upgrade (instead of changing for each patch version?)

DavidKnott:
@fubuloubu Normally I'd want to do `ethereum >=2.1.0, < 2.2.0` but because pyethereum is always evolving I think it's safer to stick with`ethereum==2.1.0` for now.

PR #399, opened 2017-10-12
Update Dockerfile

amitkumarj441:
### - What I did
         The updated version of the dockerfile reduces the size of the docker image which is created. This would drastically reduce the size of the docker image.
### - How I did it
         Each command adds a layer and takes up cumulative space.
### - How to verify it
         Build by running "docker build -t viper ."


DavidKnott:
@amitkumarj441 LGTM

PR #405, opened 2017-10-17
Return tuple support

jacqueswww:
### - What I did

Contracts can return tuple types now.
*Note* this doesn't cover tuple assignment yet.

### - How I did it

- Patched function signature.
- Altered parse_return in stmt.py, lots of work to handle dynamic type in the tuple (ByteArray only for now).
- More details have be placed in the commits ;)

### - How to verify it

Check the tests.

Added support for return a tuple from a contract.

### DIFF
--- a/tests/parser/functions/test_return_tuple.py
+++ b/tests/parser/functions/test_return_tuple.py
@@ -54,3 +54,13 @@ def out_very_long_bytes() -> (num, bytes <= 1024, num, address):
     assert c.four() == [1234, b"bytes", b"test", 4321]
     assert c.out_chunk() == [b"hello", 5678, b"world"]
     assert c.out_very_long_bytes() == [5555, long_string.encode(), 6666, "0x0000000000000000000000000000000000001234"]
+
+
+def test_return_type_signatures():
+    code = """
+def out_literals() -> (num, address, bytes <= 4):
+    return 1, 0x0000000000000000000000000000000000000000, "random"
+    """
+
+    c = get_contract(code)
+    assert c.translator.function_data['out_literals']['decode_types'] == ['int128', 'address', 'bytes']
--- a/tests/parser/functions/test_return_tuple.py
+++ b/tests/parser/functions/test_return_tuple.py
@@ -1,0 +1,56 @@
+import pytest
+from tests.setup_transaction_tests import chain as s, tester as t, ethereum_utils as u, check_gas, \
+    get_contract_with_gas_estimation, get_contract
+
+
+def test_return_type():
+    long_string = 35 * "test"
+
+    code = """
+chunk: {
+    a: bytes <= 8,
+    b: bytes <= 8,
+    c: num
+}
+
+def __init__():
+    self.chunk.a = "hello"
+    self.chunk.b = "world"
+    self.chunk.c = 5678
+
+def out() -> (num, address):
+    return 3333, 0x0000000000000000000000000000000000000001
+
+def out_literals() -> (num, address, bytes <= 4):
+    return 1, 0x0000000000000000000000000000000000000000, "random"
+
+def out_bytes_first() -> (bytes <= 4, num):
+    return "test", 1234
+
+def out_bytes_a(x: num, y: bytes <= 4) -> (num, bytes <= 4):
+    return x, y
+
+def out_bytes_b(x: num, y: bytes <= 4) -> (bytes <= 4, num, bytes <= 4):
+    return y, x, y
+
+def four() -> (num, bytes <= 8, bytes <= 8, num):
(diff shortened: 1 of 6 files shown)


viper/parser/stmt.py
```diff
@@ -275,5 +276,51 @@ def parse_return(self):
...
+                # Get value of dynamic offset counter.
+                return ['mload', dynamic_offset_counter]
+
+            def increment_dynamic_offset(dynamic_spot):
+                # Increment dyanmic offset counter in memory.
+                return ['mstore', dynamic_offset_counter,
+                                 ['add',
+                                        ['add', ['ceil32', ['mload', dynamic_spot]], 32],
+                                        ['mload', dynamic_offset_counter]]]
+
```
DavidKnott: Do you think splitting the `LLL` into functions makes the code easier to read?  I personally like it together in chunks, but may be biased because that's what I'm used to doing.

viper/types.py
```diff
@@ -311,19 +317,20 @@ def ceil32(x):
...
 # Gets the number of memory or storage keys needed to represent a given type
-def get_size_of_type(typ):
+def get_size_of_type(typ, in_return=False):
     if isinstance(typ, BaseType):
         return 1
     elif isinstance(typ, ByteArrayType):
-        return ceil32(typ.maxlen) // 32 + 2
+        return_offset = 1 if in_return else 0
+        return ceil32(typ.maxlen) // 32 + 2 + return_offset
     elif isinstance(typ, ListType):
```
DavidKnott: I think you could take out line 324 `return_offset = 1 if in_return else 0` if you want to.

tests/parser/functions/test_return_tuple.py
```diff
@@ -0,0 +1,56 @@
...
+    c = get_contract(code)
+
+    assert c.out() == [3333, "0x0000000000000000000000000000000000000001"]
+    assert c.out_literals() == [1, "0x0000000000000000000000000000000000000000", b"random"]
+    assert c.out_bytes_first() == [b"test", 1234]
+    assert c.out_bytes_a(5555555, "test") == [5555555, b"test"]
+    assert c.out_bytes_b(5555555, "test") == [b"test", 5555555, b"test"]
+    assert c.four() == [1234, b"bytes", b"test", 4321]
+    assert c.out_chunk() == [b"hello", 5678, b"world"]
+    assert c.out_very_long_bytes() == [5555, long_string.encode(), 6666, "0x0000000000000000000000000000000000001234"]
```
DavidKnott: Could you test the abi output as well?

DavidKnott: I'd like to see a test or two with more returns.  Is the maximum amount of returns uncapped now?  If so I think we should add a maximum (maybe 7?).  What're your thoughts?

jacqueswww: :+1: ah yes, 'course :)

jacqueswww: @DavidKnott So currently yes it's uncapped, I can add a maximum definitely however not sure it's really needed? If we do cap it I would suggest capping it at 10 at the very least.

On similar note I added an StructureException if the length of the two tuples are not the same. So this way if you wanted to you'd have to put a very long `-> (bytes, bytes, bytes, ...)` signature on the method, hehehe.

viper/parser/stmt.py
jacqueswww: @DavidKnott in this specific case splitting out which each one does and giving it a name helped me a lot. Usually I use `annotation=` but couldn't really here because we are building the with statement. Simple cases using a bunch of LLL definitely work for me, but I like describing complex cases like the dynamic output here as much as possible (mostly for my feature self).

viper/types.py
jacqueswww: @DavidKnott ha, thanks! Good catch - yes this isn't need at all anymore. I needed it in a previos iteration of the work.

tests/parser/functions/test_return_tuple.py
DavidKnott: Good call on the exception.  I think it's important to set a returns limit because if we don't it'll be limited by the call stack which could cause unexpected (and possibly unsafe) behavior to occur.  I'd be fine with 10.

viper/parser/stmt.py
DavidKnott: Thanks for sharing your thoughts, I'll play around more with how I write my `LLL`.

jacqueswww:
@DavidKnott grab a cup of coffee, and peruse the latest changes (after I found them on the other branch hahaha!) :coffee:

DavidKnott:
@jacqueswww LGTM

PR #406, opened 2017-10-18
BUG: Log exceptions have error and don't post the right message

fubuloubu:
### - What I did

I was fooling around with logs and noticed the exceptions for reserved and duplicate wouldn't post because of `arg.arg` where `arg` is a string (and therefore doesn't have an inner member arg)

### - How to verify it

Run a test raising these two exceptions and verify that they now show you what specific log argument is reserved or duplicated

Bug with Log Exceptions


DavidKnott:
@fubuloubu Thanks a lot!  LGTM

PR #408, opened 2017-10-18
Added method to get list of logs with optional filtering

fubuloubu:
### - What I did

Added `get_logs()` function to retrieve the list of logs produced by a contract in a given receipt, optional filtering argument to only collect logs of a specific type
Changed `get_log` fixture to use this new function (returning the last log in the filtered list)

### - How to verify it

Tests all work, let me know if there are any concerns

Added get_logs(), redefined get_log fixture to use get_logs()

### - Cute Animal Picture

turtle

### DIFF
--- a/tests/setup_transaction_tests.py
+++ b/tests/setup_transaction_tests.py
@@ -103,26 +103,25 @@ negative_G1 = [
 
 curve_order = 21888242871839275222246405745257275088548364400416034343698204186575808495617
 
+def get_logs(receipt, contract, event_name=None):
+    contract_log_ids = contract.translator.event_data.keys() # All the log ids contract has
+    # All logs originating from contract, and matching event_name (if specified)
+    logs = [log for log in receipt.logs \
+        if log.topics[0] in contract_log_ids and \
+            log.address == contract.address and \
+            (not event_name or \
+                    contract.translator.event_data[log.topics[0]]['name'] == event_name)]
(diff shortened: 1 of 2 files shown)


tests/setup_transaction_tests.py
```diff
@@ -103,26 +103,25 @@ def get_contract(source_code, *args, **kwargs):
 
 curve_order = 21888242871839275222246405745257275088548364400416034343698204186575808495617
 
+def get_logs(receipt, contract, event_name=None):
+    contract_log_ids = contract.translator.event_data.keys() # All the log ids contract has
```
DavidKnott: Why does `event_name` default to `None` when it's required in `get_log`?

jacqueswww: @DavidKnott @fubuloubu Hmm good point, considering this change. Can we rename get_log to get_last_log() and make `event_name=None` within `get_last_log` this way it's clear we want the last entry in the log (with optional even_name filtering).

fubuloubu: `event_name` is a filter, `None` means no filter applied
I like the suggestion about `get_last_log()` but that requires updates to the places it was used.

fubuloubu:
Finally got around to making the suggested changes

fubuloubu:
@DavidKnott any other issues?

DavidKnott:
Nope, LGTM.  Thanks for the contribution!

PR #420, opened 2017-10-20
Update travis

amitkumarj441:
### - What I did
         Added installation script.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,7 @@ install:
   - pip install .
 
 script:
+  - python setup.py install
   - python setup.py test
   - flake8 viper
 


DavidKnott:
Why is this necessary?  It's my understanding that everything that needs to be installed gets installed in `install:`

amitkumarj441:
Actually separate `setup.py` runs on travis to lower memory usage and I think travis needs a fix by adding `dev_requirements.txt`

DavidKnott:
@amitkumarj441 So by including `setup.py` in the scrips it prevents travis from doing it in the background and this ends up using less memory?

.travis.yml
```diff
@@ -12,6 +12,7 @@ install:
   - pip install .
 
 script:
+  - python setup.py install
```
jacqueswww: There is no point in this because we already do the `pip install .` as part of the install section.

jacqueswww:
Closing as it seems all our CI is still working fine.

Issue #413, opened 2017-10-19
Exception to fix for logging events


jacqueswww:
### What's your issue about?

There are 2 (perhaps boils down to one) bugs in relation to logging events, that are throwing exceptions:

1.)
```python
MyLog: __log__({arg1: bytes <= 3})

def foo(x: bytes <= 3):
    log.MyLog(x)
```
```
Traceback (most recent call last):
  File "./viper_bin", line 25, in <module>
    print('0x' + compiler.compile(code).hex())
  File "/home/jacques/projects/viper/viper/compiler.py", line 11, in compile
    lll = optimizer.optimize(parser.parse_tree_to_lll(parser.parse(code), code))
  File "/home/jacques/projects/viper/viper/parser/parser.py", line 336, in parse_tree_to_lll
    sub.append(parse_func(_def, _globals, {**{'self': sigs}, **contracts}, origcode))
  File "/home/jacques/projects/viper/viper/parser/parser.py", line 426, in parse_func
    ['seq'] + clampers + [parse_body(c, context) for c in code.body] + ['stop']
  File "/home/jacques/projects/viper/viper/parser/parser.py", line 426, in <listcomp>
    ['seq'] + clampers + [parse_body(c, context) for c in code.body] + ['stop']
  File "/home/jacques/projects/viper/viper/parser/parser.py", line 437, in parse_body
    return parse_stmt(code, context)
  File "/home/jacques/projects/viper/viper/parser/parser.py", line 966, in parse_stmt
    return Stmt(stmt, context).lll_node
  File "/home/jacques/projects/viper/viper/parser/stmt.py", line 56, in __init__
    self.lll_node = self.stmt_table[stmt_type]()
  File "/home/jacques/projects/viper/viper/parser/stmt.py", line 63, in expr
    return Stmt(self.stmt.value, self.context).lll_node
  File "/home/jacques/projects/viper/viper/parser/stmt.py", line 56, in __init__
    self.lll_node = self.stmt_table[stmt_type]()
  File "/home/jacques/projects/viper/viper/parser/stmt.py", line 144, in call
    inargs, inargsize, inarg_start = pack_logging_data(data_types, data, self.context)
  File "/home/jacques/projects/viper/viper/parser/parser.py", line 1018, in pack_logging_data
    holder, maxlen = pack_args_by_32(holder, maxlen, arg, typ, context, context.new_placeholder(BaseType(32)))
[...]

PR #424, opened 2017-10-22
Fix logging

DavidKnott:
### - What I did
Fix logging so that bytes variables can be logged.
Added more tests.
In response to #413 @jacqueswww 

### - How I did it
I added parsing for variables so that they can be logged.
I came up with tests that should be invalid and made sure they were / fixed them if they weren't.

### - How to verify it
Look over the tests I added and make sure they're working as expected.

None
### - Cute Animal Picture
image

### DIFF
--- a/tests/parser/features/test_logging.py
+++ b/tests/parser/features/test_logging.py
@@ -156,7 +156,7 @@ def foo():
     assert c.translator.decode_event(logs.topics, logs.data) == {'arg1': [1, 2], 'arg2': [timestamp, timestamp + 1, timestamp + 2], 'arg3': [[1, 2], [1, 2]], '_event_type': b'MyLog'}
 
 
-def test_logging_with_input_bytes():
+def test_logging_with_input_bytes(bytes_helper):
     loggy_code = """
 MyLog: __log__({arg1: indexed(bytes <= 4), arg2: indexed(bytes <= 29), arg3: bytes<=31})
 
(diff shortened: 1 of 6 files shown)


tests/parser/features/test_logging.py
```diff
@@ -156,12 +156,35 @@ def foo():
...
+    logs = s.head_state.receipts[-1].logs[-1]
+    event_id = u.bytes_to_int(u.sha3(bytes('MyLog(bytes4,bytes29,bytes31)', 'utf-8')))
+    # # Event id is always the first topic
+    assert logs.topics[0] == event_id
+    # # Event id is calculated correctly
+    assert c.translator.event_data[event_id]
+    # # Event abi is created correctly
+    assert c.translator.event_data[event_id] == {'types': ['bytes4', 'bytes29', 'bytes31'], 'name': 'MyLog', 'names': ['arg1', 'arg2', 'arg3'], 'indexed': [True, True, False], 'anonymous': False}
+    # Event is decoded correctly
+    assert c.translator.decode_event(logs.topics, logs.data) ==  {'arg1': b'bar\x00', 'arg2': b'bar\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00', 'arg3': b'foo\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00', '_event_type': b'MyLog'}
```
fubuloubu: Is there some sort of helper we can give when working with bytes? Like `b'bar\x00'` I really only care that the first 3 is `bar` and the remaining entries are asserted to be empty. Maybe something like `bytes_helper = lambda str, length: s.decode('utf-8') + bytearray(length-len(str))` so I can write:
```python
assert c.translator.decode_event(logs.topics, logs.data) ==  \
        {'arg1': b'bar\x00', 'arg2': bytes_helper('bar', 29), 'arg3': bytes_helper('foo', 31), '_event_type': b'MyLog'}
```

fubuloubu: Is there some sort of helper we can give when working with bytes? Like `b'bar\x00'` I really only care that the first 3 is `bar` and the remaining entries are asserted to be empty. Maybe something like `bytes_helper = lambda str, length: s.decode('utf-8') + bytearray(length-len(str))` so I can write:
```python
assert c.translator.decode_event(logs.topics, logs.data) ==  \
        {'arg1': b'bar\x00', 'arg2': bytes_helper('bar', 29), 'arg3': bytes_helper('foo', 31), '_event_type': b'MyLog'}
```

jacqueswww: @fubuloubu I think a custom assert is what I have used for similar functionality in the past (this was with standard python testcases). https://pytest.readthedocs.io/en/2.1.0/assert.html#defining-your-own-assertion-comparison (this seems to be the pytest way :smile:  )

DavidKnott: Okay, I'll do it that way.

fubuloubu: @jacqueswww Not sure I see how a custom assert would help, but a fixture that lets us do this testing easier is what I'm aiming for

jacqueswww: @fubuloubu fixture might be an option too yes, using custom assert will allow you to do something like fuzzy match 2 dicts, only comparing the first 3 chars of the values.

DavidKnott:
@jacqueswww I added a failing test because you can't index `>bytes32` and although you can log  `>bytes32` with data neither Solidity nor Viper currently support it.

Issue #415, opened 2017-10-20
Add line numbers to basetype exceptions


jacqueswww:
### What's your issue about?

`base_type_conversion` and `makesetter` throw relevant Structure & Type Exceptions without line numbers.

### How can it be fixed?

Just do it @jacqueswww :package: 

#### Cute Animal Picture

![](http://www.50-best.com/images/cute_animal_pictures/pdog_with_glasses.jpg)

fubuloubu:
Great catch!

jacqueswww:
Fixed ;)

PR #426, opened 2017-10-23
Add extra line numbers to exceptions.

jacqueswww:
### - What I did

- Added optional pos parameters, which can be used by various TypeMismatchException 's to show the programmer where the error occurred.
- #415 

### - How I did it
Looked at TypeMismatchException ;)

### - How to verify it
Example program:
```python
x: num

def __init__():
    self.x = "test"
```
Will generate a more useful exception now:
```
viper.exceptions.TypeMismatchException: line 5:4 Base type conversion from or to non-base type: bytes <= 4 num
```

### DIFF
--- a/tests/parser/exceptions/test_parser_exception_pos.py
+++ b/tests/parser/exceptions/test_parser_exception_pos.py
@@ -1,0 +1,14 @@
+from pytest import raises
+
+from viper.exceptions import ParserException
+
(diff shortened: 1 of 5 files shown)


fubuloubu:
Love it!

DavidKnott:
LGTM

PR #427, opened 2017-10-23
Docs: Overhaul of types

chgue:
### - What I did
Completely reworked most of the types documentation.
### - How to verify it
Read it :)
Overhaul of types

### DIFF
--- a/docs/types.rst
+++ b/docs/types.rst
@@ -48,15 +48,9 @@ Operator              Description
 ``x or y``            Logical disjunction  
 ``x == y``            Equality             
 ``x != y``            Inequality
-===================  ===================  
+====================  ===================
 
-The operators ``or`` and ``and`` apply the common short-circuiting rules:
-::
-    #Short-circuiting
-    return false and foo()
-    #Returns false without calling foo() since it is not necessary for the result
-    return true or bar()
-    #Returns true without calling bar() since it is not necessary for the result 
+The operators ``or`` and ``and`` apply the common short-circuiting rules.
 
 .. index:: ! num, ! int, ! integer
 Signed Integer (128 bit)
@@ -245,7 +239,7 @@ Syntax as follows: ``_address.<member>``, where ``_address`` is of the type ``ad
 
 Unit Types
 ==========
-Viper allows the definition of types with a discrete unit such as e.g. meters, seconds, wei, ... . These types may only be based on either ``num`` or ``decimal``.
+Viper allows the definition of types with discrete units e.g. meters, seconds, wei, ... . These types may only be based on either ``num`` or ``decimal``.
 Viper has multiple unit types built in, which are the following:
 
 =============  =====  =========  ==========================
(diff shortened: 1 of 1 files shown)


fubuloubu:
flake8 problems

docs/types.rst
```diff
@@ -14,90 +14,279 @@ to form complex types.
...
+------
+The only possible values are the constants ``true`` and ``false``.
+
+Operators
+---------
+
+====================  ===================  
+Operator              Description
+====================  ===================  
+``f(x) not g(y)``     Logical negation
```
fubuloubu: Please use `x not y` instead of `f(x)` notation. It is more confusing, and we're not describing functional flows or anything like that.

docs/types.rst
```diff
@@ -14,90 +14,279 @@ to form complex types.
...
+
+===================  ===========  =========  ====================================================================================
+Currency
+---------------------------------------------------------------------------------------------------------------------------------
+Keyword              Unit         Base type  Description
+===================  ===========  =========  ====================================================================================
+``wei_value``        1 wei        ``num``    An amount of `Ether <http://ethdocs.org/en/latest/ether.html#denominations>`_ in wei
+``currency_value``   1 currency   ``num``    An amount of currency
+``currency1_value``  1 currency1  ``num``    An amount of currency1
+``currency2_value``  1 currency2  ``num``    An amount of currency2
```
fubuloubu: @DavidKnott What the heck are `currency1_value` and `currency2_value` supposed to be?

docs/types.rst
chgue: @fubuloubu Will change with the next push when I finish the rest! Was lazy and didn't change that from the previous version :P 
You mentioned "some issues": Apart from this issue (and the weird currency1 stuff) is there anything else or just some proof reading? If it's the latter: I will do that when the rest is finished.
Thanks for the feedback!

docs/types.rst
```diff
@@ -14,90 +14,279 @@ to form complex types.
...
+``f(x) and g(y)``     Logical conjunction  
+``f(x) or g(y)``      Logical disjunction  
+``f(x) == g(y)``      Equality             
+``f(x) != g(y)``      Inequality
+====================  ===================  
+
+The operators ``or`` and ``and`` apply the common short-circuiting rules:
+::
+    #Short-circuiting
+    return false and foo()
```
fubuloubu: @vbuterin @DavidKnott  were we disallowing this (compound calling) in order to make sure the gas estimation was computable?

docs/types.rst
DavidKnott: @fubuloubu `currency1_value` is a type to describe what the variable is and is checked on assignment so you can do something like:
```
total_cost: currency1_value
def foo(price: currenc1_value):
    self.total_cost = price
```
But not:
```
total_cost: currency2_value
def foo(price: currenc1_value):
    self.total_cost = price
```

chgue:
@DavidKnott @fubuloubu Thanks for the feedback! Implemented your changes.

DavidKnott:
LGTM, merging it in.

Issue #428, opened 2017-10-24
If in range support


jacqueswww:
### What's your issue about?

Allow Ast.In syntax, such as:

```python
def in_test() -> bool:
    if x in [1, 2, 3, 4]:
        return True
    return False
```

Part of #390  :)

### How can it be fixed?

Just do it :heavy_check_mark: 

#### Cute Animal Picture

![](https://i.ytimg.com/vi/Ayp2YeIfbq4/maxresdefault.jpg)

jacqueswww:
I'm on it! :smile:

jacqueswww:
@fubuloubu Definitely, however I don't think I have actually been able to create list without fixed length (I tried the other day it's super difficult). Do you know how - so I can put the test case ?

fubuloubu:
Oh, yeah I mean we were talking about adding a simple `push()`/`pop()` type list a while back. Would be very useful for creating stacks and queues, as such uses can be accomplished with constant gas estimation. Not sure if it was still on the table, but we were talking about it.

jacqueswww:
Interesting! Yeah that would mean we'd need an in memory counter for the lists' size extra ;) can be done - ping me when the discussion is brought up again :)

jacqueswww:
This has been merged :smiley_cat:

PR #429, opened 2017-10-25
If in a list support

jacqueswww:
### - What I did
Implements #428, we can now use `x in [1, 2, 3, 4]` syntax :smiley_cat: 

### - How I did it

- Added handling for the specific case in expr.py/Expr.compare function.

### - How to verify it
Check tests, and write some code:

```python
allowed: num[10]


def set():
    self.allowed = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9]


def in_test(x: num) -> bool:
    if x in self.allowed:
            return True
    return False
```

### DIFF
--- a/tests/parser/features/iteration/test_range_in.py
+++ b/tests/parser/features/iteration/test_range_in.py
@@ -1,6 +1,7 @@
 import pytest
 from tests.setup_transaction_tests import chain as s, tester as t, ethereum_utils as u, check_gas, \
     get_contract_with_gas_estimation, get_contract
+from viper.exceptions import TypeMismatchException
 
 
 def test_basic_in_list():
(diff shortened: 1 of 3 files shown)


fubuloubu:
@DavidKnott Just checking, `list` types can only contain 1 type of element, right?

@jacqueswww I don't think the updated parser code checks that the type of the left hand side of the operation matches the containing type of the list. Needs a test to that effect at least.

fubuloubu:
Otherwise, nice job!

DavidKnott:
@fubuloubu I think lists only containing 1 data type makes the most sense.  It'd get messy/confusing really fast with multiple types.

fubuloubu:
Oh, I wasn't suggesting otherwise. Just confirming that assumption with you ;)

DavidKnott:
@jacqueswww Could you add some tests with different data types?

jacqueswww:
Will do :smile_cat:

jacqueswww:
@DavidKnott added a specific exception message ;)

DavidKnott:
@jacqueswww Thanks a lot, if I'm understanding correctly this pr is only meant to support checking for `num`'s in lists?

jacqueswww:
Yes, only numeric base32 types.

Issue #370, opened 2017-09-29
Fix voting example


DustinWehr:
### What's your issue about?
The problem is described here https://github.com/ethereum/viper/pull/302#issuecomment-326341764

### How can it be fixed?

I'm happy to help if @DavidKnott and @fubuloubu don't have the time.

3 options, all of which still need some scrutiny:
* @fubuloubu sketched a fix in two comments: [this change](https://github.com/ethereum/viper/pull/302#issuecomment-326353960) alone seemed like it'd work, but the resulting program is more restricted than the Solidity example.  [The second comment](https://github.com/ethereum/viper/pull/302#issuecomment-326357970) is about addressing that restriction. 
* [I wrote one](https://gist.github.com/DustinWehr/b4f3e0d3b1d29a55f4c3042e0485bd69) that differs more from @DavidKnott's.
* [More complex version of the previous](https://gist.github.com/DustinWehr/d178b3a0b50dcb217a26ad1f26be82e1), including a cycle breaking capability.

fubuloubu:
I think in general the examples don't need to be fully realized, production code. Their point is more to show how much simpler and clearer Viper syntax is as examples on how to get started.

That being said, replicating some Solidity examples is important as a comparison.

I didn't get a chance to look too deeply at what you did, but keep that in mind.

DavidKnott:
@DustinWehr It'd be great if you made a PR.  I second what @fubuloubu mentioned, it's a lot more important for people to understand examples and the viper syntax used then to have them be fully functionally.  Maybe you could make a pr with 2 voting examples a simple one and your more complex /  feature complete one?

DavidKnott:
@DustinWehr Are you working on this, if not I can do it this week.

DustinWehr:
I never did get running tests working on my system... Can you run them (the existing test) on this before I submit a PR? 
https://gist.github.com/DustinWehr/98c9310afab793588aa8e031f1d79aec

fubuloubu:
@DustinWehr I think you need a clearer example of what you're suggesting for `assert`, probably in a different issue. I sort of like the `require`/`assert` thing in solidity (like maybe 80% like it lol). In solidity, `require` means front end assumptions about inputs, and `assert` means the backend of calculations executed correctly. Your suggestion of `should` actually sounds like a great candidate for what you are describing in (1), but I don't really follow exactly what you mean there (e.g. `should` describes a scenario that can potentially be optimized away assuming formal correctness is verified for that contract).

[...]

PR #431, opened 2017-10-26
fix voting with delegation example

DustinWehr:
### - What I did
#370 

### - How I did it
with input from @DavidKnott and @fubuloubu 

### - How to verify it
run the tests, read the comments

fix voting with delegation example

### - Cute Animal Picture
image

### DIFF
--- a/examples/voting/ballot.v.py
+++ b/examples/voting/ballot.v.py
@@ -102,7 +102,7 @@ def delegate(to: address):
     self.voters[msg.sender].voted = True
     self.voters[msg.sender].delegate = to
 
-    # this call will throw iff this delegation would cause a loop of length <= 5.
+    # this call will throw if and only if this delegation would cause a loop of length <= 5.
     self.forward_weight(msg.sender)
 
 # Give your vote (including votes delegated to you)
--- a/examples/voting/ballot.v.py
+++ b/examples/voting/ballot.v.py
@@ -26,11 +26,13 @@ num_proposals: public(num)
 
 @constant
 def delegated(addr: address) -> bool:
-    return self.voters[addr].delegate != 0x0000000000000000000000000000000000000000
+    # equivalent to self.voters[addr].delegate != 0x0000000000000000000000000000000000000000
+    return not not self.voters[addr].delegate
 
 @constant
 def directly_voted(addr: address) -> bool:
-    return self.voters[addr].voted and self.voters[addr].delegate == 0x0000000000000000000000000000000000000000
+    # not <address> equivalent to <address> == 0x0000000000000000000000000000000000000000
+    return self.voters[addr].voted and not self.voters[addr].delegate
 
 # Setup global variables
 def __init__(_proposalNames: bytes32[2]):
@@ -95,7 +97,7 @@ def delegate(to: address):
     # Throws if sender tries to delegate their vote to themselves or to
     # the default address value of 0x0000000000000000000000000000000000000000
     # (the latter might not be problematic, but I don't want to think about it)
-    assert to != msg.sender and to != 0x0000000000000000000000000000000000000000
+    assert to != msg.sender and not not to
 
     self.voters[msg.sender].voted = True
     self.voters[msg.sender].delegate = to
--- a/examples/voting/ballot.v.py
+++ b/examples/voting/ballot.v.py
@@ -4,11 +4,11 @@
 voters: public({
     # weight is accumulated by delegation
     weight: num,
-    # if true, that person already voted
+    # if true, that person already voted (includes voting by delegating)
     voted: bool,
     # person delegated to
     delegate: address,
-    # index of the voted proposal
+    # index of the voted proposal. not meaningful unless `voted` is True.
     vote: num
 }[address])
 
@@ -22,6 +22,15 @@ proposals: public({
 
 voter_count: public(num)
 chairperson: public(address)
+num_proposals: public(num)
+
+@constant
+def delegated(addr: address) -> bool:
+    return self.voters[addr].delegate != 0x0000000000000000000000000000000000000000
+
+@constant
+def directly_voted(addr: address) -> bool:
+    return self.voters[addr].voted and self.voters[addr].delegate == 0x0000000000000000000000000000000000000000
 
 # Setup global variables
 def __init__(_proposalNames: bytes32[2]):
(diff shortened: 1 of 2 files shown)


DustinWehr:
This doesn't seem to be true anymore? 
```
# If `proposal` is out of the range of the array,		
# this will throw automatically and revert all	
# changes.
self.proposals[proposal].vote_count += self.voters[msg.sender].weight
```

examples/voting/ballot.v.py
```diff
@@ -22,6 +22,15 @@
 
 voter_count: public(num)
 chairperson: public(address)
+num_proposals: public(num)
+
+@constant
+def delegated(addr: address) -> bool:
+    return self.voters[addr].delegate != 0x0000000000000000000000000000000000000000
```
fubuloubu: @DavidKnott is there like a `null_addr` or just `null` (which figures out what the null value is for the given type being compared) so we don't ever have to write something like this? Counting the number of 0's will get tedius lol

Wait, does `0x0000000000000000000000000000000000000000 == 0x0`?

DustinWehr: I asked somewhere else if it's the unique falsey address, in which case wouldn't need to write it explicitly.

DustinWehr: But this test does pass as you suggested:
```
def test_zero_address(self):
        self.assertEqual(0x0000000000000000000000000000000000000000, 0x0)
```

DustinWehr: So does this though:
```
self.assertEqual(0x0000000000000000000000000000000000000000, 0)
```

fubuloubu: Well, if that's true you can probably write `assert not self.voters[addr].delegate` relying on the "falsey-ness" principle and it should work.

examples/voting/ballot.v.py
```diff
@@ -32,61 +41,88 @@ def __init__(_proposalNames: bytes32[2]):
...
-        # If the delegate already voted,
-        # directly add to the number of votes
-        self.proposals[self.voters[to].vote].vote_count += self.voters[msg.sender].weight
-    else:
-        # If the delegate did not vote yet,
-        # add to her weight.
-        self.voters[to].weight += self.voters[msg.sender].weight
+
+    # this call will throw iff this delegation would cause a loop of length <= 5.
+    self.forward_weight(msg.sender)
```
DavidKnott: `iff` should be `if`

examples/voting/ballot.v.py
DavidKnott: @fubuloubu Yup `assert not self.voters[addr].delegate` works

examples/voting/ballot.v.py
DustinWehr: I meant `iff`. The two asserts at the beginning of `forward_weight` always evaluate to true when `forward_weight` is called by `delegate`. I mentioned this as a potential optimization enabled by the `should` keyword [here](https://github.com/ethereum/viper/issues/397#issuecomment-336508695).

examples/voting/ballot.v.py
DustinWehr: Hmm I just got 
```
viper.exceptions.InvalidLiteralException: line 35: Cannot read 0x value with length 3. Expecting 42 (address incl 0x) or 66 (bytes32 incl 0x)
```
when I tried to use 0x0 in place of 0x0000000000000000000000000000000000000000. Maybe there was something wrong with my test case above?

examples/voting/ballot.v.py
fubuloubu: mathiness ftw! @DustinWehr maybe just spell out "if and only if"

examples/voting/ballot.v.py
```diff
@@ -32,61 +43,88 @@ def __init__(_proposalNames: bytes32[2]):
...
-        if self.voters[to].delegate:
-        # Because there are not while loops, use recursion to forward the delegation
-        # self.delegate(self.voters[to].delegate)
-            assert self.voters[to].delegate != msg.sender 
-            to = self.voters[to].delegate
+    assert not self.voters[msg.sender].voted
+    # Throws if sender tries to delegate their vote to themselves or to
+    # the default address value of 0x0000000000000000000000000000000000000000
+    # (the latter might not be problematic, but I don't want to think about it)
+    assert to != msg.sender and not not to
```
DavidKnott: Is there a reason you're using `not not ` instead of just checking for a `truthy` value?

examples/voting/ballot.v.py
```diff
@@ -22,6 +22,17 @@
 
 voter_count: public(num)
 chairperson: public(address)
+num_proposals: public(num)
+
+@constant
+def delegated(addr: address) -> bool:
+    # equivalent to self.voters[addr].delegate != 0x0000000000000000000000000000000000000000
+    return not not self.voters[addr].delegate
```
DavidKnott: Same question about `not not` here.

DavidKnott:
@DustinWehr Is there a reason you're using `not not ` instead of just checking for a `truthy` value?

DustinWehr:
The compiler complained without it. Apparently `not` converts non-boolean values, but `and` doesn't.

DavidKnott:
@DustinWehr I've thought about this a bit and think promoting `not not` in the examples isn't great. Would you mind switching it to `self.voters[addr].delegate == True`?

DustinWehr:
`self.voters[addr].delegate == True` is a syntax error:
```
viper.exceptions.TypeMismatchException: line 30: Unsupported types for comparison: 'address' 'bool'
    return self.voters[addr].delegate == True
```

I think both the current and the `== True` style are bad. The right thing to do is either make `0x0` a valid literal, or be ok with `0x0000000000000000000000000000000000000000`. Anything else is a hack. It's too early in the language's life to be resorting to hacks.

DavidKnott:
@DustinWehr I agree, I'm going to make 0x0 a valid address so you can do something like
assert self.voters[addr].delegate !=  0x0

DustinWehr:
Great!
Unfortunately I can no longer push to my remote fork after pulling the commits from the last couple weeks into my local copy and making the change from 0x0000000000000000000000000000000000000000 to 0x0 in these two files.

DavidKnott:
@DustinWehr I'm just gonna merge this in for now and create a VIP for the `0x0` change

DustinWehr:
what's a VIP?

DavidKnott:
Viper Improvement Proposal :)

PR #433, opened 2017-10-26
Add erc20 abi functionality

DavidKnott:
### - What I did
Goal: reduce repetition
Added `on_chain_market_maker.py` (created by @vbuterin) and tested it.
Added built-in erc20 abi to allow the following functionality:
```
token: address(ERC20)

def __init__(token_address: address):
    self.token = token_address

def transfer(_to: address, _value: num256) -> bool:
    return self.token.transfer(_to, _value)

/*** Includes built-in abi for all ERC20 functions ***/
```
### - How I did it
Created a precompiled contracts file where we can add standardized contracts and added the ERC20's abi.
Added `on_chain_market_maker.py` which utilizes the added functionality.
### - How to verify it
Look at the tests in `test_erc20_abi.py` and `test_on_chain_market_maker.py`

Add built-in erc20 abi functionality
### - Cute Animal Picture
image


fubuloubu:
So, this is somewhat touching the concept of imports or file references. I know we want to avoid inheritance between files, but I think being able to split up code files as a long-term goal is very desirable.

I think the best example of this is something like a contract factory, e.g.:

owned_item.v.py:
```python
owner: address
def __init__():
     self.owner = msg.sender
```

factory.v.py:
```python
import owned_item as item

def __init__():
     self.owner = msg.sender

def deploy_item():
     assert self.owner == msg.sender
     deploy(item)
```

Also, I don't think we have a method for deploying a contract from inside another contract either. Food for thought. The gas cost should be fixed, I can't think of any other implementation choices that would prohibit this.

PR #435, opened 2017-10-28
Added 'is_varname_valid()' check for globals

chgue:
### - What I did
Globals could contain reserved words. I added a check for "is_varname_valid()".
### - How I did it
See "parser.py".
Additionally, "test_setters.py" had to be changed, since it contained the opcode "pop" as a variable.
### - How to verify it
See "test_variable_declaration_exception.py".
Globals can not have invalid names anymore.


DavidKnott:
@chgue Thanks a lot!  LGTM

Issue #438, opened 2017-10-30
For in list support


jacqueswww:
### What's your issue about?

Allow iterating over a list in a for loop;
```
def data() -> num:
    s = [1, 2, 3, 4, 5, 6]
    for i in s:
        if i >= 3:
            return i
    return -1
```

### How can it be fixed?

Fill this in if you know how to fix it.

#### Cute Animal Picture
![](http://longwallpapers.com/Desktop-Wallpaper/cute-animal-wallpapers-full-hd-For-Desktop-Wallpaper.jpg)

fubuloubu:
As long as the size of the list is known at compile time, we can probably do a replacement with `for I in range(0, len(s)):` or something

fubuloubu:
Although the value of `i` needs to be the value in `s`, not the index as `range` provides....

Perhaps the syntax translates:
```python
for i in s:
   #stuff
```

To
```python
for _index_s in range(0, len(s)-1):
    i = s[_index_s]
    #stuff
```

jacqueswww:
@fubuloubu the latter is exactly what the LLL comes down to  with what I have so far ;) https://github.com/jacqueswww/viper/blob/438_for_i_in_list/viper/parser/stmt.py#L208

fubuloubu:
Nice. Personally, I would love to see some of these shortcuts rewrote as viper code substitutions in a pre-processor. It would make it clearer later on. LLL works for now though

jacqueswww:
Jip, definitely ;) I think token(ERC20) is the first use I have seen that could actually be seen as preprocessor.

PR #440, opened 2017-10-31
For in list support

jacqueswww:
### - What I did

#438 
- Adds support for, `for i in list` syntax. Allowing you to iterate over lists.

### - How I did it
- Alter Stmt.parse_for, and added Stmt.parse_for_list function.

### - How to verify it

Check tests ;)

Something like this will work now:

```python
def data() -> num:
    s = [1, 2, 3, 4, 5, 6]
    for i in s:
        if i >= 3:
            return i
    return -1
```

Adds support foreach like for statements.

### DIFF
--- a/viper/parser/stmt.py
+++ b/viper/parser/stmt.py
@@ -222,7 +222,7 @@ class Stmt(object):
         subtype = iter_list_node.typ.subtype.typ
         varname = self.stmt.target.id
         value_pos = self.context.new_variable(varname, BaseType(subtype))
-        i_pos = self.context.new_variable('_i', BaseType(subtype))
+        i_pos = self.context.new_variable('_index_for_' + varname, BaseType(subtype))
 
         if iter_var_type:  # Is a list that is already allocated to memory.
             iter_var = self.context.vars.get(self.stmt.iter.id)
--- a/viper/parser/stmt.py
+++ b/viper/parser/stmt.py
@@ -78,6 +78,22 @@ class Stmt(object):
         from .parser import (
             make_setter,
         )
+
+        if isinstance(self.stmt.targets[0], ast.Subscript):  # Check if we are doing assignment of an iteration loop.
+            raise_exception = False
+            if isinstance(self.stmt.targets[0].value, ast.Attribute):
+                list_name = "%s.%s" % (self.stmt.targets[0].value.value.id, self.stmt.targets[0].value.attr)
+                if list_name in self.context.in_for_loop:
+                    raise_exception = True
+
+            if isinstance(self.stmt.targets[0].value, ast.Name) and \
+               self.stmt.targets[0].value.id in self.context.in_for_loop:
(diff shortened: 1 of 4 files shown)


viper/parser/stmt.py
```diff
@@ -185,6 +191,83 @@ def parse_for(self):
...
+            if iter_var_type and isinstance(iter_var_type.typ, ListType):
+                return True
+
+        return False
+
+    def parse_for_list(self):
+        from .parser import (
+            parse_body,
+            make_setter
+        )
```
fubuloubu: Bad form to have imports inside a function, as that import will be executed every time the function is called

viper/parser/stmt.py
```diff
@@ -185,6 +191,83 @@ def parse_for(self):
...
+
+        iter_list_node = Expr(self.stmt.iter, self.context).lll_node
+        if not isinstance(iter_list_node.typ.subtype, BaseType):  # Sanity check on list subtype.
+            raise StructureException('For loops allowed only on basetype lists.', self.stmt.iter)
+        iter_var_type = self.context.vars.get(self.stmt.iter.id).typ if isinstance(self.stmt.iter, ast.Name) else None
+        subtype = iter_list_node.typ.subtype.typ
+        varname = self.stmt.target.id
+        value_pos = self.context.new_variable(varname, BaseType(subtype))
+        i_pos = self.context.new_variable('_i', BaseType(subtype))
```
fubuloubu: `_i` is bad form, would probably cause an error in doublely nested lists or something. Either deterministic name-mangling (`indexVar = 'i_' + namehash(varname)` where `namehash` is a hashing function with small output e.g. 8 bytes), or name extension (`indexVar = 'index_' + varname`) should be employed to avoid potential collisions

viper/parser/stmt.py
DavidKnott: I agree 100%, it needs to be done now but as we refactor `parser.py` I think it'll be easier to remove the imports inside functions.

DavidKnott:
What happens if someone tries to change the list while it's being iterated over?

fubuloubu:
@DavidKnott I was just thinking the same thing. I think as he currently has it, the list is copied into memory temporarily, which means that modifications to the existing list take place outside of the iterator. I think however that the compile should throw and Error if the base list (the one used for the iterator) is modified during the loop.

viper/parser/stmt.py
jacqueswww: Currently we don't support iterating over nested lists, because we don't allow comparison of lists either, I wrote a specific exception test for it. But will definitely make the indexVar change thanks ;)

jacqueswww:
Will add a check :+1:

jacqueswww:
@DavidKnott @fubuloubu added the necessary changes :smile:

DavidKnott:
LGTM

PR #442, opened 2017-11-02
fix the problems when install viper in Ubuntu16.04

kaikai1024:
### - What I did
little doc fix
### - How I did it
find it when i install viper
### - How to verify it
i don't know if it is ok while os is not Ubuntu16.04
fix it when install python3.6 and use virtualenv
### - Cute Animal Picture
: -D

### DIFF
--- a/docs/installing-viper.rst
+++ b/docs/installing-viper.rst
@@ -37,7 +37,7 @@ Install Python 3.6 and some necessary packages:
     wget https://www.python.org/ftp/python/3.6.2/Python-3.6.2.tgz
     tar xfz Python-3.6.2.tgz
     cd Python-3.6.2/
-    ./configure –prefix /usr/local/lib/python3.6
+    ./configure --prefix /usr/local/lib/python3.6
     sudo make
     sudo make install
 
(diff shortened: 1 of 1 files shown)


docs/installing-viper.rst
```diff
@@ -75,7 +75,7 @@ other development environment set-up.
 
 To create a new virtual environment for Viper run the following commands:
 ::
-    virtualenv -p /usr/local/lib/python3.6/bin/python --no-site-packages ~/viper-venv
+    virtualenv -p /usr/local/lib/python3.6/bin/python3 --no-site-packages ~/viper-venv
```
fubuloubu: I think this is my bad, I created a symlink `python` pointing to `python3` I think got copied when I showed the original author my install instructions for 16.04.LGTM

kaikai1024: And i wonder if it's better to add some examples just like the doc of the solidity
[https://solidity.readthedocs.io/en/develop/solidity-by-example.html](example of the solidity)

fubuloubu: We have examples in a separate section we are working on integrating into the docs

kaikai1024: cool

DavidKnott:
LGTM

PR #443, opened 2017-11-02
Any length bytes as map keys.

jacqueswww:
### - What I did
Implements #477 
Add support for using any length bytes as keys for maps.

```python
mapped_bytes: num[bytes <= 5]

def set(k: bytes <= 5, v: num):
    self.mapped_bytes[k] = v


def get(k: bytes <= 5) -> num:
    return self.mapped_bytes[k]
```

### - How I did it
Added necessary handling in add_variable_offset function.

### - How to verify it

Check the tests ;)


fubuloubu:
Whoa nice @jacqueswww!

P.S. I didn't check the tests yet

fubuloubu:
@jacqueswww Actually a good potential usecase for Hypothesis might be to rewrite these very similar bytes tests using `@given(integers(1,32))` so it will test contracts constructed using a random set of integers in the acceptable range of bytes so we can try all possibilities of bytearray lengths!

Also, maybe we can extrapolate that testing further with a randomized list of bytearray values (up to the given length) for testing.

Also also, maybe we can create a set of test utilities for testing bytearrays that we can deploy more widely.

jacqueswww:
@fubuloubu Awesome examples fof using something like hypothesis :+1:

tests/parser/features/test_bytes_map_keys.py
```diff
@@ -0,0 +1,73 @@
...
+
+def test_mismatched_byte_length():
+    code = """
+mapped_bytes: num[bytes <= 34]
+
+def set(k: bytes <= 35, v: num):
+    self.mapped_bytes[k] = v
+    """
+
+    with pytest.raises(TypeMismatchException):
```
DavidKnott: Can you add a test with a huge number of bytes?

viper/parser/parser_utils.py
```diff
@@ -378,6 +379,18 @@ def add_variable_offset(parent, key):
...
+        elif isinstance(typ, MappingType) and isinstance(key.typ, ByteArrayType):
+            if not isinstance(typ.keytype, ByteArrayType) or (typ.keytype.maxlen < key.typ.maxlen):
+                raise TypeMismatchException('Mapping keys of bytes cannot be cast, use exact same bytes type of: %s', str(typ.keytype))
+            subtype = typ.valuetype
+            if len(key.args[0].args) == 3:  # handle bytes literal.
+                sub = LLLnode.from_list([
+                    'seq',
+                    key,
+                    ['sha3', ['add', key.args[0].args[-1], 32], ceil32(key.typ.maxlen)]
+                ])
```
DavidKnott: Eventually I'd like to see byte literals being handled off chain as much as possible, but that can be another PR.

tests/parser/features/test_bytes_map_keys.py
jacqueswww: @DavidKnott sure thing, uhm what's a huge number ? 4096 bytes?

viper/parser/parser_utils.py
jacqueswww: +1

tests/parser/features/test_bytes_map_keys.py
fubuloubu: 4KB I'd veay is pretty large lol. What's the block storage limit anyways?

fubuloubu: Average block is like 15KB, so I think 4KB is more than enough

jacqueswww:
@DavidKnott added test ;)

DavidKnott:
LGTM

PR #446, opened 2017-11-07
Remove tester ref from assert_tx_failed()

fubuloubu:
### - What I did
I noticed that `assert_tx_failed()` fixture required a reference to `tester` that could just be referenced internally, fixing some usage issues people have been having. I wanted to also get rid of needing `lambda: ` but couldn't at this point

### - How I did it
Greped for all uses in `tests/` and updated `setup_transaction_tests.py`

### - How to verify it
Check out what tests I changed
Run pytest

Modified assert_tx_failed() to remove reference to tester as tester reference was already provided internally

### - Cute Animal Picture
puma

### DIFF
--- a/examples/stock/test_company.py
+++ b/examples/stock/test_company.py
@@ -19,7 +19,7 @@ def tester():
     return tester
 
 @pytest.fixture
-def assert_tx_failed(tester, function_to_test, exception = t.TransactionFailed):
+def assert_tx_failed(function_to_test, exception = t.TransactionFailed):
     initial_state = tester.s.snapshot()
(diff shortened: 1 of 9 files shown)


jacqueswww:
I like it! One less parameter makes it even easier to use :)

fubuloubu:
Yeah. The next step is that it should be a pytest fixture inside `conftest.py`, simplifying what's going on inside `setup_transaction_tests.py` and making it so we don't have to import it everywhere it's used

DavidKnott:
LGTM

Issue #448, opened 2017-11-09
compile_lll.py: buggy translation of LLL operator `sha3_32`


daejunpark:
### What's your issue about?

The translation of LLL operator `sha3_32` in `compile_lll.py` seems to be buggy.

For example, an LLL expression `[sha3_32 0]` is translated to EVM opcodes, `PUSH1, 0, PUSH1, 192, MSTORE, PUSH1, 192, PUSH1, 32, SHA3`, which perhaps should not. AFAIU, the translation logic should be that it first stores 0 in a predefined memory location (here 192), and then computes SHA3 hash of memory range of starting location 192 and width 32 (i.e., MEM[192...223]). But in the generated EVM code, it computes SHA3 hash of memory range of starting location **32** and width **192** (i.e., MEM[32...223]):
https://github.com/ethereum/viper/blob/b1fe12ddd8706ef3340a103916e3d3f2cc95ceef/viper/compile_lll.py#L203

Here is why it works until now. Fortunately (or unfortunately) the unintended memory range MEM[32...191] that is prepended to the value to be hashed (i.e., MEM[192...223]) is fixed over all Viper programs for now. They are pre-occupied to store the five size limits used to clamp values: `ADDRSIZE`, `MAXNUM`, `MINNUM`, `MAXDECIMAL`, `MINDECIMAL`:
https://github.com/ethereum/viper/blob/b1fe12ddd8706ef3340a103916e3d3f2cc95ceef/viper/utils.py#L76-L80

Here is a minimal example to reproduce. For the following Viper program:
```
balance: num[num]
def foo():
    self.balance[0] = 1
```
The intermediate LLL program is:
```
[seq,   
  [return,
    0,    
    [lll, 
      [seq, 
        [seq,
          [mstore, 28, [calldataload, 0]],
          [mstore, 32, 1461501637330902918203684832716283019655932542976],
          [mstore, 64, 170141183460469231731687303715884105727],
          [mstore, 96, -170141183460469231731687303715884105728],
          [mstore, 128, 1701411834604692317316873037158841057270000000000],
          [mstore, 160, -1701411834604692317316873037158841057280000000000]],
        # Line 3
        [if,
          [eq, [mload, 0], 3264763256 <foo>],
          [seq,
            pass,
            [assert, [iszero, callvalue]],
            # Line 4
            [sstore, [add, [sha3_32, 0 <self.balance>], 0], 1], 
            # Line 3
            stop]]],
      0]]]
```
[...]

PR #452, opened 2017-11-09
fix issue #448: buggy translation of LLL operator `sha3_32`

daejunpark:
### - What I did

Fix issue #448: compile_lll.py: buggy translation of LLL operator `sha3_32`

### - How I did it

As suggested in #448.

### - How to verify it

Look over the change made and verify that all tests are still passing.

### - Description for the change log

None

### - Cute Animal Picture

```
               __   __
              __ \ / __
             /  \ | /  \
                 \|/
            _,.---v---._
   /\__/\  /            \
   \_  _/ /              \
     \ \_|           @ __|
  hjw \                \_
  `97  \     ,__/       /
     ~~~`~~~~~~~~~~~~~~/~~~~
```

### DIFF
--- a/tests/compiler/test_sha3_32.py
+++ b/tests/compiler/test_sha3_32.py
@@ -1,0 +1,9 @@
+import pytest
+from viper.parser.parser_utils import LLLnode
+from viper import compile_lll, optimizer
+
+def test_sha3_32():
+    lll = ['sha3_32', 0]
+    evm = ['PUSH1', 0, 'PUSH1', 192, 'MSTORE', 'PUSH1', 32, 'PUSH1', 192, 'SHA3']
+    assert compile_lll.compile_to_assembly(LLLnode.from_list(lll)) == evm
+    assert compile_lll.compile_to_assembly(optimizer.optimize(LLLnode.from_list(lll))) == evm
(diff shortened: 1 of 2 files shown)


fubuloubu:
Add your test from #448

daejunpark:
@fubuloubu I added a quick test. Please let me know if you have a better idea to test this.

fubuloubu:
I haven't done much with LLL. The underlying function just computes the hash correct?

For any issue that didn't fail testing I look for a new test because obviously it's a problem. I'm not 100% familiar with the function, but just looking at your test you basically verify a translation of lll to opcodes. I think it would preferable to write the test as a snippet of viper and verify the output matches a Python reference implementation of SHA3 across a few examples/edge cases, showing end-to-end implementation is correct (versus a single translation stage). This way we test the whole function.

That's my 2 cents.

DavidKnott:
@fubuloubu I think it'd be hard to test this in viper because `sha3_32` is not callable directly in viper and is only used on the backend.
That being said, it'd be possible to test this using viper by declaring a mapping in viper and then checking the bytecode to see what locations in memory are hashed though as of right now I think that's probably more trouble then it's worth.

fubuloubu:
Oh, so this is how the mapping type finds the memory location for a specific key? I think at some point it would be good to show that it works as advertised like you said, I was more pointing out that the test that was added doesn't really accomplish much as is, it's basically an audit of the translation vs a proof of any underlying functionality.

DavidKnott:
@fubuloubu Exactly, I think there's room for both.  But for this PR, because the translation is what was incorrect I think the test added is fine.  Though down the road we could and probably should add a mapping to bytecode integration test.

DavidKnott:
LGTM, @daejunpark thanks a lot, merging this in.

Issue #453, opened 2017-11-10
Inconsistency of the value range of type `num`


daejunpark:
* viper Version: 0.0.2
* OS: linux
* Python Version (python --version): 3.6.1

### What's your issue about?

The following three do *not* agree on the value range of type`num`.

1. The [doc](https://viper.readthedocs.io/en/latest/types.html) says: `num` = [-2\*\*127+1, 2\*\*127-2]
> ``num``:  equivalent to``int128``, a signed integer strictly between -2\*\*127 and 2\*\*127-1.

2. `clamp` accepts [-2\*\*127, 2\*\*127-1]

```
def foo():
    one = 1
    x =  170141183460469231731687303715884105726 + one #  2**127 - 1
    z = -170141183460469231731687303715884105727 - one # -2**127
```

3. Viper compiler accepts [-2\*\*127+1, 2\*\*127-1]

For the following program:
```
def foo():
    x =  170141183460469231731687303715884105727 #  2**127 - 1
    y = -170141183460469231731687303715884105727 # -2**127 + 1
    z = -170141183460469231731687303715884105728 # -2**127
```
Viper compiler complained:
```
line 4: Number out of range: 170141183460469231731687303715884105728
    z = -170141183460469231731687303715884105728 # -2**127
---------^
```
### How can it be fixed?

Identify the correct behavior out of the three, and fix the other two.

#### Cute Animal Picture

```
o                  o             o                  o
  o ___/|__          o            o   o              o
o _/       \  /| o              o     ___/|__      o
 /  @ \\    \/ |    o            o  _/       \  /|   o  ___/|__
 \_   //    /\ |  o  ___/|__       /  @ \\    \/ | o  _/       \  /|
   \_______/  \| o _/       \  /|  \_   //    /\ |  o/  @ \\    \/ |
                  /  @ \\    \/ |    \_______/  \|   \_   //    /\ |
                  \_   //    /\ |                      \_______/  \|
                    \_______/  \| 
```

fubuloubu:
Number 2. is correct, 1. and 3. are wrong

Do we not test the extreme limits of num in those tests?

PR #457, opened 2017-11-10
fix issue #453

daejunpark:
### - What I did

 * ~Partially~ fix the issue #453.
 * ~Still the following are rejected by the compiler due to negative literal parse issue~
 * __NOTE__: It may be needed to fix the similar issue for `decimal` type, but I haven't looked into that yet.

### - How I did it

 * As suggested in https://github.com/ethereum/viper/issues/453#issuecomment-343348646

#### How I fixed the negative literal parsing issue

The Python parser generates an AST where a negative literal is represented as a unary minus expression, i.e., `UnaryOp(USub, Num(N))` instead of `Num(-N)`. Thus, literal `-170141183460469231731687303715884105728` (i.e., MINNUM) is represented as `UnaryOp(USub, Num(170141183460469231731687303715884105728))`, and the sub-node `Num(170141183460469231731687303715884105728)` is rejected by `expr.py`, since `170141183460469231731687303715884105728` is greater than MAXNUM.

But I couldn't find a clean way of modifying `expr.py` logic to consider the specific case, so I just added a pass that modifies the parse tree in advance by replacing `UnaryOp(USub, Num(N))`  with  `Num(-N)`, which allows to avoid modifying `expr.py` logic and I think is a much cleaner solution.

### - How to verify it

 * Check if the added test faithfully represents the semantics of `num` arithmetic

### - Cute Animal Picture

```
          __     __
         /  \~~~/  \
   ,----(     ..    )
  /      \__     __/
 /|         (\  |(
^ \   /___\  /\ |   hjw
   |__|   |__|-"
```


viper/parser/expr.py
```diff
@@ -78,7 +78,7 @@ def get_expr(self):
     def number(self):
         orignum = get_original_if_0x_prefixed(self.expr, self.context)
         if orignum is None and isinstance(self.expr.n, int):
-            if not (-2**127 + 1 <= self.expr.n <= 2**127 - 1):
```
jacqueswww: @fubuloubu @DavidKnott I believe using the clamp values here as well for literals is correct?

jacqueswww: @daejunpark  please use MemoryPositions.MINNUM and MemoryPositions.MAXNUM. As we are really striving to use constant from now on ;)

fubuloubu: Not sure I understand the comment, but the low end should just be `-2**127`.

Int N is always defined as the range [-2**(N-1), 2**(N-1)-1]. One bit goes towards the sign bit (hence N-1), and you can think of the non-negative end of the interval additionally containing the 0 value, which is why the non-negative side is always one less in value than the negative side (both contain equal numbers of entries that way). Think of it as negative and non-negative (literally sign bit or no sign bit) vs. positive/negative. I think that's where people get confused.

jacqueswww: @fubuloubu yes I think that's what we have here:

fubuloubu: Yeah, I love the collections of constants. I wish we used more constants instead of literals everywhere.

If your comment about was to just use these constants, then I 100% agree.

jacqueswww: @daejunpark just use the  MINNUM / MAXNUM constants thanks. And thanks for finding it ;)

daejunpark: @jacqueswww thanks for the suggestion. I fixed it.

daejunpark:
I also fixed the negative literal parsing issue. Now this PR addressed #453 completely. Please review it again.

DavidKnott:
@daejunpark Please rebase with master

viper/functions.py
```diff
@@ -185,8 +185,6 @@ def as_num256(expr, args, kwargs, context):
             raise InvalidLiteralException("Number out of range: " + str(expr.args[0].n), expr.args[0])
         return LLLnode.from_list(args[0], typ=BaseType('num256', None), pos=getpos(expr))
     elif isinstance(args[0], LLLnode):
-        if args[0].value == "sub" and args[0].args[0].value == 0 and args[0].args[1].value > 0:
-            raise InvalidLiteralException("Negative numbers cannot be num256 literals")
```
daejunpark: Now it is not reachable.

daejunpark: Note that it doesn't help to detect https://github.com/ethereum/viper/issues/469

daejunpark:
@DavidKnott Sure, I rebased it on the master. Also, I removed the deadcode, which you may want to take a look.

DavidKnott:
@daejunpark Looks good to me, thanks for removing the dead code as well!

PR #461, opened 2017-11-11
Fix issue #451: buggy translation of `repeat` when #iteration <= 0

### LINKED ISSUE #451: compile_lll.py: buggy translation of `repeat` when #iteration <= 0
### What's your issue about?

If the #iteration of `repeat` loop is less than or equal to 0, it replaces it by `2`, which perhaps should not, while `1` would be more reasonable:
https://github.com/ethereum/viper/blob/b1fe12ddd8706ef3340a103916e3d3f2cc95ceef/viper/compile_lll.py#L96

Fortunately, for now, the `repeat` loops will be rejected in the preprocess phase if the #iteration is less than or equal to 0, but it would be good to fix for later:
https://github.com/ethereum/viper/blob/b1fe12ddd8706ef3340a103916e3d3f2cc95ceef/viper/parser/parser_utils.py#L113-L114

### How can it be fixed?

Change the following line:
https://github.com/ethereum/viper/blob/b1fe12ddd8706ef3340a103916e3d3f2cc95ceef/viper/compile_lll.py#L96
to something like:
```
loops = num_to_bytearray(code.args[2].value) or [1]
```
or
```
loops = num_to_bytearray(code.args[2].value) or raise Exception(...)
```

#### Cute Animal Picture

```
                        _.-..
                      ,'9 )\)`-.,.--.
                      `-.|           `.
                         \,      ,    \)
                          `.  )._\   (\
                            |//   `-,//
                            ]||    //"
                     hjw    ""    ""
```

DavidKnott:
### - What I did
Fix `repeat` LLL in `compile_lll.py` so that it throws if `iteration_count <= 0`
Fixes #451, thanks @daejunpark
### - How I did it
Added a conditional the checks for invalid iteration counts
### - How to verify it
Look at the code I changed
None
### - Cute Animal Picture
image

### DIFF
--- a/viper/compile_lll.py
+++ b/viper/compile_lll.py
@@ -93,7 +93,9 @@ def compile_to_assembly(code, withargs=None, break_dest=None, height=0):
     # Repeat(memloc, start, rounds, body)
     elif code.value == 'repeat':
         o = []
-        loops = num_to_bytearray(code.args[2].value) or [2]
+        loops = num_to_bytearray(code.args[2].value)
+        if not loops:
+            raise Exception("Number of times repeated must be a constant nonzero positive integer: %r" % loops)
         start, end = mksymbol(), mksymbol()
         o.extend(compile_to_assembly(code.args[0], withargs, break_dest, height))
         o.extend(compile_to_assembly(code.args[1], withargs, break_dest, height + 1))
--- a/tests/compiler/LLL/test_repeat.py
+++ b/tests/compiler/LLL/test_repeat.py
@@ -1,0 +1,9 @@
+import pytest
+
+def test_repeat(t, get_contract_from_lll, assert_compile_failed):
+    good_lll = ['repeat', 0, 0, 1, ['seq']]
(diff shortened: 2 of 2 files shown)


viper/compile_lll.py
```diff
@@ -93,7 +93,9 @@ def compile_to_assembly(code, withargs=None, break_dest=None, height=0):
     # Repeat(memloc, start, rounds, body)
     elif code.value == 'repeat':
         o = []
-        loops = num_to_bytearray(code.args[2].value) or [2]
+        loops = num_to_bytearray(code.args[2].value)
+        if not loops:
+            raise Exception("Number of times repeated must be a constant nonzero positive integer: %r" % loops)
```
jacqueswww: Works for me, however as I see it - we need a way to test this. Perhaps using your new get_contract_from_lll fixture ;)

DavidKnott:
@jacqueswww Added a test for this, merging it in.

Issue #449, opened 2017-11-09
compile_lll.py: inconsistency in checking invalid cases


daejunpark:
### What's your issue about?

Inconsistency in checking invalid cases in `compile_lll.py`.

For the `with` variable lookup, it rejects immediately when the `with` statement is too deep:
https://github.com/ethereum/viper/blob/b1fe12ddd8706ef3340a103916e3d3f2cc95ceef/viper/compile_lll.py#L50-L54
but for the `with` variable update, it does not check at all, which may lead to generate an invalid/wrong opcode (e.g., if the depth is 17, then internally it will generate `SWAP17` (doesn't exist) that will end up being encoded as `0xa0`, which is `LOG0`):
https://github.com/ethereum/viper/blob/b1fe12ddd8706ef3340a103916e3d3f2cc95ceef/viper/compile_lll.py#L55-L60

However, I couldn't find an example program that leads to such a deep `with` statement. It might not be possible at this moment, but it would be good to fix for later.

### How can it be fixed?

Add the following check in the case of `code.value == "set"`:
```
     if height - withargs[code.args[0].value] > 16: 
         raise Exception("With statement too deep") 
```

#### Cute Animal Picture

```
       \/)/)
     _'  oo(_.-. 
   /'.     .---'
 /'-./    (
 )     ; __\
 \_.'\ : __|
      )  _/
     (  (,.
   mrf'-.-'
```

DavidKnott:
@daejunpark Thanks for brining this up, I'll make a pr to fix this.

PR #462, opened 2017-11-11
Fix issue #449: inconsistency in checking invalid cases

DavidKnott:
### - What I did
Fix `with` statement error in `compile_lll.py` which led to `SWAP17` turning into `LOG0` (because `SWAP17` doesn't exist)
Fixes issue #449, thanks @daejunpark 
### - How I did it
Added a conditional that throws with a with statement that's too deep.
### - How to verify it
Look at the code I added
None
### - Cute Animal Picture
image


DavidKnott:
@jacqueswww Thanks for the review.  I added tests, merging it in.

PR #465, opened 2017-11-13
Improve rlp list gas estimation

DavidKnott:
### - What I did
Get `RLPlist` function's `gas_estimate` closer to the actual amount of gas used.
### - How I did it
Kept reducing the `gas_estimate` without allowing it to go under the actual amount of gas used.

### - How to verify it
Look over the gas numbers I changed for `_RLPlist` in `functions.py`
None
### - Cute Animal Picture
image

### DIFF
--- a/viper/functions.py
+++ b/viper/functions.py
@@ -597,7 +597,7 @@ def _RLPlist(expr, args, kwargs, context):
         ['seq',
             ['with', '_sub', variable_pointer,
                 ['pop', ['call',
-                         10000 + 500 * len(_format) + 10 * len(args),
+                         1500 + 400 * len(_format) + 10 * len(args),
                          LLLnode.from_list(RLP_DECODER_ADDRESS, annotation='RLP decoder'),
                          0,
                          ['add', '_sub', 32],


viper/functions.py
```diff
@@ -597,7 +597,7 @@ def _RLPlist(expr, args, kwargs, context):
         ['seq',
             ['with', '_sub', variable_pointer,
                 ['pop', ['call',
-                         10000 + 500 * len(_format) + 10 * len(args),
+                         1500 + 400 * len(_format) + 10 * len(args),
```
jacqueswww: :+1:

tests/parser/functions/test_rlp_list.py
```diff
@@ -0,0 +1,111 @@
...
+    x = RLPList('\xf6\x9455555555555555555555\xa0GGGGGGGGGGGGGGGGGGGGGGGGGGGGGGGG', [address, bytes32])
+    return x[1]
+
+def foq() -> bytes <= 100:
+    x = RLPList('\xc5\x83cow\x03', [bytes, num])
+    return x[0]
+
+def fos() -> num:
+    x = RLPList('\xc5\x83cow\x03', [bytes, num])
+    return x[1]
```
jacqueswww: @DavidKnott this is the one scenario where we actually support mixed types in a list ;)

tests/parser/functions/test_rlp_list.py
```diff
@@ -0,0 +1,111 @@
+import pytest
+from tests.setup_transaction_tests import chain as s, tester as t, ethereum_utils as u, check_gas, \
+    get_contract_with_gas_estimation
+
+def test_rlp_decoder_code():
+    import rlp
```
jacqueswww: why the inline import? Isn't rlp a package ?

tests/parser/functions/test_rlp_list.py
```diff
@@ -0,0 +1,111 @@
...
+        success = True
+    except:
+        success = False
+    assert not success
+    try:
+        c.qov(rlp.encode([b'\x03', b'\x00\x01']))
+        success = True
+    except:
+        success = False
+    assert not success
```
jacqueswww: All these naked try / excepts can be replaced by assert_tx_failed now ;)

tests/parser/functions/test_rlp_list.py
DavidKnott: @jacqueswww Yup, we should probably add a nice section on RLPList to the docs, and then link to the Ethereum wiki.

tests/parser/functions/test_rlp_list.py
DavidKnott: You're right, I think there's no benefit to having it inline.

tests/parser/functions/test_rlp_list.py
fubuloubu: Palpatine says "Do eet"

DavidKnott:
@jacqueswww Took out the `with pytest.raises` in another pr merging this in.

Issue #407, opened 2017-10-18
Make the default function visibility internal


maraoz:
Currently, functions are public by default in Viper. 
This has security implications, especially when refactoring code (see for example, [the parity multisig wallet vulnerability](https://blog.zeppelin.solutions/on-the-parity-wallet-multisig-hack-405a8c12e8f7))

My suggestion is to make the default `internal`, and require a `@public` annotation to make the function `public`.

Thoughts?

fubuloubu:
Solidity defaults to public, I think this is an okay as is since we have `@internal` and `@constant`. Most functions you'd design will probably be publicly available.

@DavidKnott do you agree?

DavidKnott:
@fubuloubu I prefer having `internal` as the default function visibility for viper because it seems more in line with viper's core values.  Clearly it might come as a surprise to some developers but I think it's well worth it for the added safety it will provide.

seed:
@DavidKnott I agree. Does that mean getting rid of @internal and introducing @public?

maraoz:
@vbuterin thoughts? I feel like having internal visibility by default could have prevented hacks like Rubixi or even Parity Wallet.

DavidKnott:
@maraoz We decided to not have defaults and require explicit visibility on all methods.  Thanks a lot for creating this issue 👍

PR #468, opened 2017-11-14
Require function visibility declaration

DavidKnott:
### - Motivation
 Vitalik in recent Gitter convo -> "(imo mandating having one or the other is the right way to do it; there's so much of an established culture of "functions are public by default" that simply switching the default to private for functions would lead to even more confusion and bugs)"
@jacqueswww and I agree
Also covers #407

### - What I did
Require all function declarations to have either an `@public` or `@internal` decorator.

### - How I did it
Added a check for public decorators in `function_signature.py`.
Changed all functions that relied on the old default public visibility and declared them explicitly public

### - How to verify it
Look at the changes I made to `function_signature.py`.
I had to make changes to almost all function declarations in Viper, but not tests were added / lost.

Require function visibility to be declared
### - Cute Animal Picture
image

### DIFF
--- a/viper/function_signature.py
+++ b/viper/function_signature.py
@@ -70,7 +70,7 @@ class FunctionSignature():
                 pos += get_size_of_type(parsed_type) * 32
 
         # Apply decorators
-        const, payable, internal = False, False, False
+        const, payable, internal, public = False, False, False, False
         for dec in code.decorator_list:
             if isinstance(dec, ast.Name) and dec.id == "constant":
                 const = True
@@ -78,8 +78,15 @@ class FunctionSignature():
                 payable = True
             elif isinstance(dec, ast.Name) and dec.id == "internal":
                 internal = True
+            elif isinstance(dec, ast.Name) and dec.id == "public":
+                public = True
             else:
                 raise StructureException("Bad decorator", dec)
+        if public and internal:
+            raise StructureException("Cannot use public and internal decorators on the same function")
+        if not public and not internal and not isinstance(code.body[0], ast.Pass):
+            # import pdb; pdb.set_trace()
+            raise StructureException("Function visibility must be declared")
         # Determine the return type and whether or not it's constant. Expects something
         # of the form:
         # def foo(): ...
--- a/viper/function_signature.py
+++ b/viper/function_signature.py
@@ -85,7 +85,6 @@ class FunctionSignature():
         if public and internal:
             raise StructureException("Cannot use public and internal decorators on the same function")
         if not public and not internal and not isinstance(code.body[0], ast.Pass):
-            # import pdb; pdb.set_trace()
             raise StructureException("Function visibility must be declared")
         # Determine the return type and whether or not it's constant. Expects something
         # of the form:
--- a/viper/parser/parser.py
+++ b/viper/parser/parser.py
@@ -155,7 +155,7 @@ def _mk_getter_helper(typ, depth=0):
 # Make a list of getters for a given variable name with a given type
 def mk_getter(varname, typ):
     funs = _mk_getter_helper(typ)
-    return ['@constant\ndef get_%s%s(%s) -> %s: return self.%s%s' % (varname, funname, head.rstrip(', '), base, varname, tail)
+    return ['@public\n@constant\ndef get_%s%s(%s) -> %s: return self.%s%s' % (varname, funname, head.rstrip(', '), base, varname, tail)
             for (funname, head, tail, base) in funs]
 
 
--- a/viper/parser/parser.py
+++ b/viper/parser/parser.py
@@ -155,7 +155,7 @@ def _mk_getter_helper(typ, depth=0):
 # Make a list of getters for a given variable name with a given type
 def mk_getter(varname, typ):
     funs = _mk_getter_helper(typ)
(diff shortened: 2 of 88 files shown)


viper/function_signature.py
```diff
@@ -70,16 +70,23 @@ def from_definition(cls, code):
...
                 internal = True
+            elif isinstance(dec, ast.Name) and dec.id == "public":
+                public = True
             else:
                 raise StructureException("Bad decorator", dec)
+        if public and internal:
+            raise StructureException("Cannot use public and internal decorators on the same function")
+        if not public and not internal and not isinstance(code.body[0], ast.Pass):
+            # import pdb; pdb.set_trace()
+            raise StructureException("Function visibility must be declared")
```
fubuloubu: So, this will throw if neither @public nor @internal is set? Am I reading this right?

I thought the idea was to remove internal by defaulting to public?

viper/parser/parser.py
```diff
@@ -140,7 +140,7 @@ def _mk_getter_helper(typ, depth=0):
 # Make a list of getters for a given variable name with a given type
 def mk_getter(varname, typ):
     funs = _mk_getter_helper(typ)
-    return ['@constant\ndef get_%s%s(%s) -> %s: return self.%s%s' % (varname, funname, head.rstrip(', '), base, varname, tail)
+    return ['@public\n@constant\ndef get_%s%s(%s) -> %s: return self.%s%s' % (varname, funname, head.rstrip(', '), base, varname, tail)
```
fubuloubu: Please for the love of all things Python use the triple quotes here:
```python
return """
@public
@constant
def ....
""".format(...)

DavidKnott: Will do

viper/function_signature.py
fubuloubu: I just worry we end up with too many decorators, which ends up looking like solidity's modifiers (although it can't be user-defined for viper).

jacqueswww: @DavidKnott I also see a stuck pdb there, please remove :stuck_out_tongue:

seed:
Had made a very similar change locally :)

Does it make sense to allow internal payable functions? 
Should payable functions explicitly require a public decorator? Or should they become public automatically.

fubuloubu:
I think we can be smart about it such that things like `@payable` are explicitly public. I was also thinking if `@constant` would make sense and be safe as assumed public as well. The less decorators on a function the clearer it will be

seed:
@fubuloubu I tend to agree with @payable being public by default. Forbidding the combination of both @internal and @payable seems more important though.

However, making @constant public by default is definitely not a good idea IMO. I believe it would increase the deployment cost as the compiler will export the function in the ABI and generate dispatching code for it.

DavidKnott:
Although having `@payable` also mean public would be efficient I think it would hurt Viper's readability.  I'd rather have developers have to type more than force them to remember that `@payable` is a valid way to declare visibility. 

Also if we choose to add an `@external` visibility decorator down the road then `@payable` will no long be sufficient to determine visibility (as there will be 2 options).

@fubuloubu I'm very scared of too many decorators decreasing Viper's readability as the language develops and I think that's something we're going to need to pay close attention to as time goes on.

jacqueswww:
Looks good to me ;) yes looking at too many stacked decorators could harm readibility, however I don't think we will need vast amounts more than the ones we have discussed now.

fubuloubu:
I keep thinking about how (most) decorators can be basically shortcuts for commonly used preconditions (similar to modifers). It might actually help auditability if they were written as preprocessor macros, and then had an output mode that showed the viper code with the macro substitutions applied for auditing purposes. Something like:
`@payable: assert msg.value > 0`
`@external: assert msg.sender is not self`
`@public`: Basically anyone can call it, exported in about (what if ABI were a generated function, then this would simply be adding to ABI definitions)
`@internal: assert msg.sender is self`
`@constant`: not sure about this one...

fubuloubu:
It would make it easier to write new decorators like the `@updatable(owner)` one we were discussing. I think a good goal of a programming language is to write as little as possible in the IR (lll here) as required, because updates to lll are incompatible breaking changes

jacqueswww:
The problem with adding support for the macro/assert syntax is that we will replacing statically defined decorators with ones that can be implemented by the programmer. Which as far as I know is against the philosophy of viper. As it introduces another dimension for the reader to check, ie. will have to check the all the unique implementations of updatable, internal etc.

jacqueswww:
Using # pragma syntax would allow us to add more 'breaking' changes in the future (however I don't think that many decorators are needed in the first place).

jacqueswww:
Ha I see what you mean now! Yes aliasing those decorator with viper-internal could work ;)

DavidKnott:
@fubuloubu I think you already made a feature about macros and I'd like to go more in depth with it.  Would you mind creating a VIP for it using the VIP template?

fubuloubu:
@jacqueswww Yeah, only decorators that we build into the compiler would be recognized by the compiler, not suggesting custom decorators.

The idea is that this reduces compiler code complexity as we come up with new features, reducing the chances we introduce errors into the compiler.

fubuloubu:
See #472

DavidKnott:
Merging this in.

PR #470, opened 2017-11-15
Add checks preventing assigning values on defining types.

### LINKED ISSUE #410: Assignment allowed to invalid variables in contract body
* viper Version: 0.0.2
* pyethereum Version: 2.1.0
* OS: linux
* Python Version (python --version): 3.6

### What's your issue about?

It appears that some programs parse which perhaps should not, including assigning to a global-style declaration in function bodies.  These programs then behave unexpectedly.  Below is a minimal example:

```
from ethereum.tools import tester as t
from ethereum.slogging import configure_logging

#configure_logging(':trace')
s = t.Chain()
from viper import com
piler

code = """
def foo() -> num:
    q:num = 1
    return q

def foo2() -> num:
    q = 1
    return q
"""

t.languages['viper'] = compiler.Compiler() 

c = s.contract(code, language='viper')

print(c.foo())  # Should not parse? returns 0
print(c.foo2()) # Should return 1; returns 1
```

**This bug was discovered by Yi Zhang of the [KEVM project](https://github.com/kframework/evm-semantics). (@yzhang90)**

### How can it be fixed?

Make parsing consistent with what is specified in the old README.

Specifically, the offending statement can parse under these top-level rules:
```
 <varname> = <type>
        OR <var> = <expr>
```

with "q:num" being neither a valid varname or var (assuming varnames cannot contain colons, which would be sane).

Parsing should therefore likely fail in this case.

#### Cute Animal Picture

```
             ,,,,,,,,
           ,|||````||||
     ,,,,|||||       ||,
  ,||||```````       `||
,|||`                 |||,
||`     ....,          `|||
||     ::::::::          |||,
||     :::::::'     ||    ``|||,
||,     :::::'               `|||
`||,                           |||
 `|||,       ||          ||    ,||
   `||                        |||`
    ||                   ,,,||||
    ||              ,||||||```
   ,||         ,,|||||`
  ,||`   ||   |||`
 |||`         ||
,||           ||
||`           ||
|||,         |||
 `|||,,    ,|||
   ``||||||||`
```

jacqueswww:
### - What I did

Fixes #410 adds tests.

### - How I did it

Figured out it has to do with `.value` attribute on annotate assign.

### - How to verify it
The following will fail to compile now:
```python
def foo() -> num:
    q:num = 111
    return q
```

Fixes ambiguous assign on annotate syntax.

### DIFF
--- a/tests/parser/exceptions/test_structure_exception.py
+++ b/tests/parser/exceptions/test_structure_exception.py
@@ -131,6 +131,16 @@ def foo():
     """
 def foo():
     x = y = 3
+    """,
+    """
+def foo() -> num:
+    q:num = 111
+    return q
+    """,
+    """
+q:num = 111
+def foo() -> num:
+    return self.q
     """
 ]
 
--- a/viper/parser/parser.py
+++ b/viper/parser/parser.py
@@ -171,7 +171,9 @@ def add_contract(code):
 
 
 def add_globals_and_events(_defs, _events, _getters, _globals, item):
-    if isinstance(item.annotation, ast.Call) and item.annotation.func.id == "__log__":
+    if item.value is not None:
+        raise StructureException('May not assign value whilst defining type', item)
+    elif isinstance(item.annotation, ast.Call) and item.annotation.func.id == "__log__":
         if _globals or len(_defs):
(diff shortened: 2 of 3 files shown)


DavidKnott:
LGTM

PR #479, opened 2017-11-17
Fix wrong indentation in docs

adhikasp:
### - What I did

Indenting the warning blocks and move word around because its breaking 80 char limit.

### - How I did it

By... editing it?

### - How to verify it

See the rendered text. The `warning` is not separated like before.

Fix minor indentation problem in docs.

### - Cute Animal Picture

viper?

### DIFF
--- a/docs/viper-by-example.rst
+++ b/docs/viper-by-example.rst
@@ -120,9 +120,9 @@ And of course, no smart contract tutorial is complete without a note on
 security.
 
 .. note:: It's always important to keep security in mind when designing a smart
-contract. As any application becomes more complex, the greater the potential for
-introducing new risks. Thus, it's always good practice to keep contracts as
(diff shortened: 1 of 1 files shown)


DavidKnott:
@adhikasp I'm not too worried about the 80 char limit being broken in the docs, but can you change all `Notes` in `viper-by-example.rst` to be in the `Note` box (right now it cuts off after the first line)?

adhikasp:
Of course, I haven't finish the tutorial so I doesn't notice the remainder 😄 
Anyway updated.

DavidKnott:
@adhikasp I just opened the docs with your changes locally and the notes are still cut off after the first line, would you mind fixing that?

adhikasp:
My bad, I fix it and now properly check it by building locally. All seems good.

Issue #478, opened 2017-11-17
VIP: Use REVERT opcode for assert statement


jacqueswww:
This is the suggested template for new VIPs.

## Preamble

    VIP: 478
    Title:  REVERT for assert
    Author: Jacques Wagener
    Type: Standard Track
    Status: Draft
    Created: 2017-11-17

## Simple Summary

The `assert` statement should be converted to use REVERT opcode.

## Abstract

Suggests changing the `assert` statement to using REVERT opcode underlying, allowing gas to be remaining gas be returned.

## Motivation
Most scenarios one would prefer remaing gas to be sent back, and not completely absorbed. Currently the assert statement in Viper produces a an invalid jump, draining all gas.

## Specification
Syntax wise assert statys exactly the same:
```python
def test(x: num) -> bool:                                                                                                                                           
    assert x > 0                                                                                                                                                    
    return True 
```

## Backwards Compatibility
Previous Dapps or contracts that assumed gas will completely spent after a failed assert, will now receive gas back. It doesn't effect compatibility much, as viper isn't used in production yet.

## References
https://github.com/ethereum/eips/issues/140

## Copyright
Copyright and related rights waived via [CC0](https://creativecommons.org/publicdomain/zero/1.0/)

jacqueswww:
Wether a secondary keyword for 0xfd (INVALID) asserts are necessary, should be discussed - however I much prefer only having `assert`, having multiple require/assert/require keywords tends to just be confusing. And I don't really see the need for a throw/assert keyword that drains all gas ?

fubuloubu:
Yeah, I agree. I don't think there is a need to use the opcode that drains all gas. Perhaps this original opcode might not be as useful as originally intended? Unless it is there as a means to punish abuse of the system, like in overflows, recursion issues, loop issues, etc.

PR #492, opened 2017-11-22
Use revert opcode for assert statements

DavidKnott:
### - What I did
Assert now uses the `revert` opcode behind the scenes so gas is refunded when it's used. 
This PR is in response to VIP #478 
### - How I did it
I added in the revert opcode and changed `assert` in `compile_lll.py` to use `REVERT` instead of `PC, JUMPI`.
### - How to verify it
Look at `test_assert.py` to verify that unused gas is actually being refunded.
Modify assert to use revert opcode
### - Cute Animal Picture
image

### DIFF
--- a/tests/parser/features/test_assert.py
+++ b/tests/parser/features/test_assert.py
@@ -1,0 +1,18 @@
+import pytest
+from tests.setup_transaction_tests import chain as s, tester as t, ethereum_utils as u, check_gas, \
+    get_contract_with_gas_estimation, get_contract
+
+def test_assert_refund(t):
+    code = """
+@public
+def foo():
+    assert 1 == 2
+"""
+    c = get_contract_with_gas_estimation(code)
+    pre_balance = t.s.head_state.get_balance(t.a0)
+    with pytest.raises(t.TransactionFailed):
+        c.foo(startgas=10**6, gasprice=10)
+    post_balance = t.s.head_state.get_balance(t.a0)
+    # Checks for gas refund from revert
+    # 10**5 is added to account for gas used before the transactions fails
+    assert pre_balance < post_balance + 10**5
--- a/viper/compile_lll.py
+++ b/viper/compile_lll.py
@@ -155,7 +155,10 @@ def compile_to_assembly(code, withargs=None, break_dest=None, height=0):
     # Assert (if false, exit)
     elif code.value == 'assert':
         o = compile_to_assembly(code.args[0], withargs, break_dest, height)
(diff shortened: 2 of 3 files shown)


tests/parser/features/test_assert.py
```diff
@@ -0,0 +1,17 @@
...
+def foo():
+    assert 1 == 2
+"""
+    c = get_contract_with_gas_estimation(code)
+    pre_balance = t.s.head_state.get_balance(t.a0)
+    with pytest.raises(t.TransactionFailed):
+        c.foo(startgas=1000000, gasprice=10)
+    post_balance = t.s.head_state.get_balance(t.a0)
+    # Checks for gas refund from revert
+    assert pre_balance < post_balance + 100000
```
fubuloubu: Missing a zero. Can you use `1e6` instead?

DavidKnott: It's supposed to be 100,000 to account for the initial gas used in the function call before the revert, but I'll change it to `10**5`

fubuloubu: But you supply 1e6... I guess it's not the clearest why those two numbers work. Can you add a comment like "this makes sense because about X gas will be used and 1e6 - X is around 1e5 left over". Something like that.

DavidKnott:
@fubuloubu Made the changes you suggested

fubuloubu:
Lgtm

viper/compile_lll.py
```diff
@@ -173,21 +176,22 @@ def compile_to_assembly(code, withargs=None, break_dest=None, height=0):
...
         elif code.value == "clampgt":
-            o.extend(["SGT", 'ISZERO', 'PC', 'JUMPI'])
+            o.extend(["SGT", 'ISZERO'])
         elif code.value == "uclampge":
-            o.extend(["LT", 'PC', 'JUMPI'])
+            o.extend(["LT"])
         elif code.value == "clampge":
-            o.extend(["SLT", 'PC', 'JUMPI'])
+            o.extend(["SLT"])
+        o.extend(['PC', 'JUMPI'])
```
jacqueswww: @DavidKnott by extension shouldn't this also be using REVERT instead?

DavidKnott: Eventually yes, trying to keep the PRs small 😄

jacqueswww: Sure, makes sense. Created https://github.com/ethereum/viper/issues/527 so we don't loose track.

PR #497, opened 2017-11-23
Adds -f bytecode_runtime support.

jacqueswww:
### - What I did

Adds -f bytecode_runtime support. For populus integration the runtime bytecode is required, this adds a format to the viper binary. The solc equivalent would be: `--bin-runtime`.
### - How I did it

Came up for the populus integration. Traced the lll output from -f ir.

### - How to verify it

Hmm not sure, place a debugger in `def compile` and check the lll value.

Adds support for -f bytecode_runtime.

### DIFF
--- a/bin/viper
+++ b/bin/viper
@@ -11,7 +11,7 @@ from viper.parser import parser_utils
 
 parser = argparse.ArgumentParser(description='Viper {0} programming language for Ethereum'.format(viper.__version__))
 parser.add_argument('input_file', help='Viper sourcecode to compile')
-parser.add_argument('-f', help='Format to print', choices=['abi', 'json', 'bytecode', 'ir'], default='bytecode', dest='format')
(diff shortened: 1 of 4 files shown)


fubuloubu:
Nice! Coverage decreased from the added statements... Do we have open loop cases just showing that some bytecode is produced for all of these outputs? It doesn't necessarily need to match anything, just produce something other than empty. That will gain coverage for these circumstances

jacqueswww:
Ah yes good idea, I was contemplating how to write a test for this change 😋

jacqueswww:
@DavidKnott @fubuloubu I added a ttest the -0.03% are the extra lines in the viper bin. Which we just have to live with for now :)

fubuloubu:
Love it

DavidKnott:
LGTM

Issue #469, opened 2017-11-15
as_num256 failed to reject negative arguments


daejunpark:
* viper Version: 0.0.2
* pyethereum Version: 2.1.0
* OS: linux
* Python Version (python --version): 3.6

### What's your issue about?

AFAIU, `as_num256(x)` should not be allowed when `x < 0`. If so, the compiler should reject at compile-time as much as possible, or insert runtime check to throw a runtime exception. However, the current compiler does not generate the runtime check even if it cannot prove `x >= 0`.

For example, `foo()` returns `2**256 - 1`:
```
def foo() -> num256:
    return as_num256(0 - 1)
```
while this:
```
def foo() -> num256:
    return as_num256(-1)
```
failed to be parsed:
```
viper.exceptions.InvalidLiteralException: line 2: Number out of range: -1
    return as_num256(-1)
----------------------^
```

Similarly, `bar(-1)` returns `2**256 - 1`:
```
def bar(x: num) -> num256:
    return as_num256(x)
```

### How can it be fixed?

For each `as_num256(x)`, generate runtime-check `assert(x >= 0)` when `x >= 0` cannot be proved at compile-time.

#### Cute Animal Picture

```
                              ___     _,.--.,_
                           .-~   ~--"~-.   ._ "-.
                          /      ./_    Y    "-. \
                         Y       :~     !         Y
                         lq p    |     /         .|
                      _   \. .-, l    /          |j
                     ()\___) |/   \_/";          !
                      \._____.-~\  .  ~\.      ./
                                 Y_ Y_. "vr"~  T
                                 (  (    |L    j   -Row
                                 [nn[nn..][nn..]
                             ~~~~~~~~~~~~~~~~~~~~~~~
```

fubuloubu:
Totally on board with this. Does a runtime error get generated when `x: num256 - y: num256 > x`?

daejunpark:
@fubuloubu Yep,
```
from ethereum.tools import tester as t
from ethereum.slogging import configure_logging

#configure_logging(':trace')
s = t.Chain()
from viper import compiler

code = """
def foo(x: num256, y: num256) -> num256:
    return num256_sub(x, y)
"""

t.languages['viper'] = compiler.Compiler()

c = s.contract(code, language='viper')

print(c.foo(1,2)) # TransactionFailed
```

error:
```
Traceback (most recent call last):
  File "run.py", line 17, in <module>
    print(c.foo(1,2))
[...]

PR #506, opened 2017-11-25
Add check to as_num256 to make sure number is greater than 0

DavidKnott:
### - What I did
Add a signed check to `as_num256` that throws if `as_num256` is called on a negative input (`num256 is unsigned).
This PR is in response to #469

### - How I did it
I created a test that called `as_num256` on a negative value and then I made it fail.
### - How to verify it
Look at the test I created.
None
### - Cute Animal Picture
image

### DIFF
--- a/tests/parser/functions/test_as_num256.py
+++ b/tests/parser/functions/test_as_num256.py
@@ -1,0 +1,21 @@
+import pytest
+from tests.setup_transaction_tests import chain as s, tester as t, ethereum_utils as u, check_gas, \
+    get_contract_with_gas_estimation, get_contract
+
+def test_as_num256_with_negative_num(assert_compile_failed):
+    code = """
+@public
+def foo() -> num256:
+    return as_num256(1-2)
+"""
+    assert_compile_failed(lambda: get_contract_with_gas_estimation(code), Exception)
+
+
+def test_as_num256_with_negative_input(assert_tx_failed):
+    code = """
+@public
+def foo(x: num) -> num256:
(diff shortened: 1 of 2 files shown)


tests/parser/functions/test_as_num256.py
```diff
@@ -0,0 +1,21 @@
...
+from tests.setup_transaction_tests import chain as s, tester as t, ethereum_utils as u, check_gas, \
+    get_contract_with_gas_estimation, get_contract
+
+def test_as_num256_with_negative_num(assert_tx_failed):
+    code = """
+@public
+def foo() -> num256:
+    return as_num256(1-2)
+"""
+    assert_tx_failed(lambda: get_contract_with_gas_estimation(code), Exception)
```
fubuloubu: This doesn't have a custom Exception?

DavidKnott: Nope, it doesn't

fubuloubu: Would it be possible to let this be a different kind of exception versus the standard TransitionFailed exception?

DavidKnott: I switched it to compile failed.  Eventually I think we should create custom exceptions for all of `compile_lll.py`

fubuloubu: For sure, at least something that differentiates lll exceptions from viper (or solidity) ones

DavidKnott:
Merging

Issue #509, opened 2017-11-25
The scope of index variable is different in "for ... in ... range()" and "for ... in ... list()"


yzhang90:
### What's your issue about?
There are two programs:
```
@public
def foo(x:num): 
    p = 0 
    for i in range(3): 
        p += i
    for i in range(4): 
        p += i
```
```
@public
def foo(x:num): 
    p = 0 
    for i in range(3): 
        p += i
    for i in [1, 2, 3, 4]: 
        p += i
```
The compiler accepts the first program but rejects the second program.

### How can it be fixed?

I'm not sure whether this behavior is intended or not. In python, it accepts both program. In my opinion, at least it should  both accepts OR both rejects the program OR document this behavior properly.

#### Cute Animal Picture

![image](https://user-images.githubusercontent.com/5641590/33235384-42ac55c6-d1fd-11e7-9e8f-ed0385a42e60.png)

fubuloubu:
@jacqueswww you implemented the `i in list` feature right? Is `i` implemented as a variable in the calling scope? It should only exist in the `for` loop scope (i.e. destroyed after usage)

jacqueswww:
@DavidKnott can be close now ;)

PR #513, opened 2017-11-26
Fix augassign and assign

### LINKED ISSUE #508: Augassign can bypass function parameters mutability check
### What's your issue about?
The function parameter should be immutable and should not be modified during the function. The direct assignment to a function parameter is forbidden.  However, there is no such check in the Augassign. For example, the following program can compile:
```
@public
def foo(x:num):
    x += 5     
```

### How can it be fixed?

In the stmt.py, add mutability check to function parameters.

#### Cute Animal Picture

image

DavidKnott:
### - What I did
Refactor `stmt.py` `aug_assign` and `assign functions so that they check for the proper invalids.
This fixes #508 and #509
@yzhang90 Thanks again!
### - How I did it
I created invalid `assignment` and  `for_in` tests that should've been failing and then got them to fail.
### - How to verify it
Look at the tests I added in `test_assignment.py` and `test_for_in_list.py`
None
### - Cute Animal Picture
image

### DIFF
--- a/viper/parser/stmt.py
+++ b/viper/parser/stmt.py
@@ -188,7 +188,7 @@ class Stmt(object):
         self.context.forvars[varname] = True
         o = LLLnode.from_list(['repeat', pos, start, rounds, parse_body(self.stmt.body, self.context)], typ=None, pos=getpos(self.stmt))
         del self.context.vars[varname]
-
+        del self.context.forvars[varname]
         return o
 
     def _is_list_iter(self):
@@ -270,13 +270,12 @@ class Stmt(object):
             self.context.remove_in_for_loop(iter_list_node.annotation)
         del self.context.vars[varname]
         del self.context.vars['_index_for_' + varname]
+        del self.context.forvars[varname]
         return o
 
     def aug_assign(self):
         target = self.get_target(self.stmt.target)
         sub = Expr.parse_value_expr(self.stmt.value, self.context)
-        if isinstance(self.stmt.target, ast.Name) and self.stmt.target.id in self.context.forvars:
-            raise StructureException("Altering iterator '%s' which is in use!" % target.annotation, self.stmt)
         if not isinstance(self.stmt.op, (ast.Add, ast.Sub, ast.Mult, ast.Div, ast.Mod)):
             raise Exception("Unsupported operator for augassign")
         if not isinstance(target.typ, BaseType):
@@ -425,5 +424,5 @@ class Stmt(object):
         if target.location == 'storage' and self.context.is_constant:
             raise ConstancyViolationException("Cannot modify storage inside a constant function: %s" % target.annotation)
         if not target.mutable:
-            raise ConstancyViolationException("Cannot modify function argument: %s", target.annotation)
+            raise ConstancyViolationException("Cannot modify function argument: %s" % target.annotation)
         return target
--- a/viper/parser/stmt.py
+++ b/viper/parser/stmt.py
@@ -80,22 +80,6 @@ class Stmt(object):
         from .parser import (
             make_setter,
         )
-
-        if isinstance(self.stmt.targets[0], ast.Subscript) and self.context.in_for_loop:  # Check if we are doing assignment of an iteration loop.
-            raise_exception = False
-            if isinstance(self.stmt.targets[0].value, ast.Attribute):
-                list_name = "%s.%s" % (self.stmt.targets[0].value.value.id, self.stmt.targets[0].value.attr)
-                if list_name in self.context.in_for_loop:
-                    raise_exception = True
-
-            if isinstance(self.stmt.targets[0].value, ast.Name) and \
-               self.stmt.targets[0].value.id in self.context.in_for_loop:
-                list_name = self.stmt.targets[0].value.id
-                raise_exception = True
-
-            if raise_exception:
-                raise StructureException("Altering list '%s' which is being iterated!" % list_name, self.stmt)
-
         # Assignment (eg. x[4] = y)
         if len(self.stmt.targets) != 1:
             raise StructureException("Assignment statement must have one target", self.stmt)
@@ -105,11 +89,8 @@ class Stmt(object):
             variable_loc = LLLnode.from_list(pos, typ=sub.typ, location='memory', pos=getpos(self.stmt), annotation=self.stmt.targets[0].id)
             o = make_setter(variable_loc, sub, 'memory', pos=getpos(self.stmt))
         else:
-            target = Expr.parse_variable_location(self.stmt.targets[0], self.context)
-            if target.location == 'storage' and self.context.is_constant:
-                raise ConstancyViolationException("Cannot modify storage inside a constant function!", self.stmt.targets[0])
-            if not target.mutable:
-                raise ConstancyViolationException("Cannot modify function argument", self.stmt.targets[0])
+            # Checks to see if assignment is valid
+            target = self.get_target(self.stmt.targets[0])
             o = make_setter(target, sub, target.location, pos=getpos(self.stmt))
         o.pos = getpos(self.stmt)
         return o
@@ -170,7 +151,6 @@ class Stmt(object):
         from .parser import (
             parse_body,
         )
-
         # Type 0 for, eg. for i in list(): ...
         if self._is_list_iter():
             return self.parse_for_list()
@@ -204,9 +184,11 @@ class Stmt(object):
             start = Expr.parse_value_expr(self.stmt.iter.args[0], self.context)
             rounds = self.stmt.iter.args[1].right.n
         varname = self.stmt.target.id
-        pos = self.context.vars[varname].pos if varname in self.context.forvars else self.context.new_variable(varname, BaseType('num'))
-        o = LLLnode.from_list(['repeat', pos, start, rounds, parse_body(self.stmt.body, self.context)], typ=None, pos=getpos(self.stmt))
+        pos = self.context.new_variable(varname, BaseType('num'))
         self.context.forvars[varname] = True
+        o = LLLnode.from_list(['repeat', pos, start, rounds, parse_body(self.stmt.body, self.context)], typ=None, pos=getpos(self.stmt))
+        del self.context.vars[varname]
+
         return o
 
     def _is_list_iter(self):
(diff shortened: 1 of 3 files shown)


viper/parser/stmt.py
```diff
@@ -204,9 +184,11 @@ def parse_for(self):
             start = Expr.parse_value_expr(self.stmt.iter.args[0], self.context)
             rounds = self.stmt.iter.args[1].right.n
         varname = self.stmt.target.id
-        pos = self.context.vars[varname].pos if varname in self.context.forvars else self.context.new_variable(varname, BaseType('num'))
-        o = LLLnode.from_list(['repeat', pos, start, rounds, parse_body(self.stmt.body, self.context)], typ=None, pos=getpos(self.stmt))
+        pos = self.context.new_variable(varname, BaseType('num'))
```
yzhang90: if you create new variable for each index variable, it will reject program like:
```
def foo():
    for i in range(10):
        pass
    for i in range(20):
        pass
```
Is it intended?

DavidKnott: It doesn't reject it because although I'm creating a new index variable every time I'm also deleting it at the end of the for loop, see [here](https://github.com/DavidKnott/viper/blob/b163146e5171b181b716a7681f903cf6562d014e/viper/parser/stmt.py#L271).

yzhang90: i see. I didn't read the code carefully. Now, the scope of the index variable is the loop block instead of the whole function.

viper/parser/stmt.py
```diff
@@ -288,18 +268,20 @@ def parse_for_list(self):
...
-            return o
+        del self.context.vars[varname]
+        del self.context.vars['_index_for_' + varname]
+        return o
 
     def aug_assign(self):
-        target = Expr.parse_variable_location(self.stmt.target, self.context)
+        target = self.get_target(self.stmt.target)
         sub = Expr.parse_value_expr(self.stmt.value, self.context)
+        if isinstance(self.stmt.target, ast.Name) and self.stmt.target.id in self.context.forvars:
```
yzhang90: This check is included in self.get_target(self.stmt.target)?

viper/parser/stmt.py
```diff
@@ -420,3 +402,28 @@ def increment_dynamic_offset(dynamic_spot):
...
+            if raise_exception:
+                raise StructureException("Altering list '%s' which is being iterated!" % list_name, self.stmt)
+
+        if isinstance(target, ast.Name) and target.id in self.context.forvars:
+            raise StructureException("Altering iterator '%s' which is in use!" % target.id, self.stmt)
+        target = Expr.parse_variable_location(target, self.context)
+        if target.location == 'storage' and self.context.is_constant:
+            raise ConstancyViolationException("Cannot modify storage inside a constant function: %s" % target.annotation)
+        if not target.mutable:
+            raise ConstancyViolationException("Cannot modify function argument: %s", target.annotation)
```
yzhang90: you want "Cannot modify function argument: %s" %  target.annotation ('%' instead of ',')

viper/parser/stmt.py
yzhang90: wait. You may also want to remove the index variable from self.context.forvars. Otherwise, the following program will report weird error:
```
@public
def foo():
    for i in range(10):
        pass
    i = 100  // create new variable i
    i = 200  // look up the variable i and check whether it is in forvars
```

DavidKnott:
@yzhang90 Made the changes you suggested, thanks for the review!

yzhang90:
In C-like language, all variables are block-scoped. In Python, variables are method-scoped. After this change, index variable would be block-scoped and all other variable would be method-scoped. I hope you're aware of this and properly document this behavior somewhere.

DavidKnott:
To me the change I made seems cleaner / more intuitive, though it wouldn't be to hard to do it another way (keeping track of the index variables currently in use with forvar).  What do you think?

fubuloubu:
@yzhang90 That's a good point, although I believe this is the better way to go as it reduces potential user bugs, which was C's reason for scoping in that way. Documentation :+1:

grosu:
@DavidKnott C is not the only language with block-scoped variables (and C is certainly not to be taken as an example when talking about safe/secure languages).  After teaching programming languages and formal verification for more than 15 years, I can tell you for a fact that it is easier to formally analyze/verify programs with block-scoped variables.  That's because you do not need to modify the program in order to analyze it compositionally (i.e., moving the variable declarations to the beginning of the method, in this case).  Compositional verification is very important in practice, because it gives you the divide-and-conquer needed to do proofs in parallel (e.g., it takes about 1h currently to automatically verify the `transferFrom` ERC20 function of the HKG implementation).  If you have to modify the program in order to do compositional verification, then you have to write properties/specifications (e.g. loop invariants) for a modified program instead of the original one, which make everything harder to understand and more error prone when programs get larger.

Juts my 2c, from a formal verification perspective.

yzhang90:
In addition to what @grosu said, viper's variable scoping(other than index variable) is also different from both python and C. For example, considering the following program:
```
@public
def foo(x:num):
    if  x > 0:
        y:num
        y = 10
    else:
        y:bool
        y = false
   // both C and python would accept the program up to this point. (ignore the syntax).
   // however viper compiler would reject it because y is declared twice.
    print(y) // C would reject it because y is out of scope, but Python would accept it and print y depends on the run time value of x, which is bad in my point of view. 
```

DavidKnott:
@yzhang90 @fubuloubu @grosu Thank you so much for the feedback, I think it makes sense to stick with block-scoping here.  Merging it in.

yzhang90:
@DavidKnott I'm totally for the block-scoping. What @grosu and I said is to *enforce block-scoping* to *all the variables* (not only just index variables). Currently, if you annassign a variable in a block(i.e if statement),  it is still visible outside the if block.

fubuloubu:
Well, if blocks are different than for loops

It might make sense to do it this way, it is easier to verify formally if there isn't conditional variable creation inside program blocks

(Just realizing blockchain and program language both use blocks as a technical term ughhhh)

yzhang90:
I totally agree with you @fubuloubu. But currently, neither syntax nor semantics will reject the following confusing program:
```
@public
def foo():
    for i in range(10):
        a:num
        a += 1
```
What do you think `a` should be after running the `for` loop? In the current Viper, it prints `10`. But a programmer may get confused since he/she thinks a fresh variable is declared every time in the loop.

fubuloubu:
Ah, yeah that should be something like "uninitialized local" as we allow uninitialized globals to be used in that way. I'd rather it raise a compiler error in that specific instance instead of returning `a=1`

fubuloubu:
And by that I mean catch the inner loop def of `a` as being initialized in the loop. I think that `if` statements don't necessarily need to follow this rule however

yzhang90:
That would complicate the semantics of variable lookup. If you enforce block-scoping to every variable, that will make things a lot easier. If you want `a` to be visible outside the loop, you should declare it outside the loop. If you declare it inside a loop and you use it outside, the compiler simply reports an error. The same rule applies to if as well. In viper, only functions definition/ if /for can create block. The implementation of the block-scoping is **very easy** as well. You just copy the context first when parse if/for instead of directly modifying the original context.

DavidKnott:
@yzhang90 The more you explained block scoping the more I like it!  Seems super in line with Vipers principles.  I'll get working on implementing it.

fubuloubu:
I'm still weary about `if` block scoping, but you're definitely right in that the pros (less bugs, easier to formally verify) far outweigh the cons of limiting expressiveness even a bit.

For testing, I need one test that show the following:

```python
def foo(choice: bool) -> num:
     if choice:
         a: num
         a = 1
         return a
      else:
         b: num
         b = 2
         return b
```

I'm worried that block scoping the if might have an issue here, but viper is pass by value so I don't think it's really a problem. Perhaps try one with lists too

yzhang90:
Regarding your example, I don't see any problem with block-scoping. `a` lives in `then scope` and `return a` also happens before `then scope` ends. So does `b` in `else scope`. `a` and `b` are protected by their own scope and can not see each other. 
With the current approach, your program is *almost* equivalent to:
```
def foo(choice: bool) -> num:
     a:num
     b:num
     if choice:
         a = 1
         return a
      else:
         b = 2
         return b
```
The only difference is that you can not use `b` in the else branch, otherwise it will throw "Duplicate variable b".

yzhang90:
@DavidKnott Thats' great! One thing I need to point out is that it is not trivial to implement `block scoping` as I described previously. You need to do some extra work depending on whether you want to reuse the memory. If you want to reuse the memory, you basically need to *free* the memory you declared in the block.
For example:
```
def foo(choice: bool):
     if choice:
         a: num
         a = 1
     b: num
     b = b + 1
```
Suppose `a` takes the memory starting from 320. Since `a` is dead outside the if block, `b` can reuse the memory starting from 320 with one caution. when you declare `b:num`, you need to clean the memory with mstore(320, 0) to clean the memory used by a. Otherwise, b will *accidentally* read the memory written by a.

fubuloubu:
@yzhang90 Yeah, you're right. My concern was there would be some issue returning a variable inside that inner scope, but I think that were unfounded. :+1:

grosu:
Kept an eye on this interesting discussion.   Would like to add that, in the world of formal verification, languages that do not stick to block scoping are notoriously hard to verify, reason about, or transform/refactor.  That also means that programmers may have a harder time to make sense of their code in some situations.  I agree with @DavidKnott that block scoping is in line with Viper's principles.

Disclaimer: @yzhang90 and myself are biased here, because we intend to verify (hopefully lots of) Viper programs, so we want to make our job easier. :)

fubuloubu:
@grosu I'm totally on board with that, making formal verification of contracts as easy as possible is what we want for sure. I would even go a step further in advocating for easy-to-use formal tools where one can specify simple behaviorial models and then generate a set of test cases showing your program implements that model to a high degree of fidelity, and to the maximum extent possible that the bytecode coverage doesn't have holes (aiming 100% MC/DC ideally).

Making functional testing easy and comprehensive and coverage metrics clear and consistent is my wet dream.

DavidKnott:
@yzhang90 Would you mind creating a VIP for changing Viper to be entirely block scoped?  I see it being implemented in two steps.  First without memory reuse and then down the road with.

yzhang90:
Sure. I'll do that.

PR #517, opened 2017-11-27
Update the hex of `rlp_decoder_tx` in comment

mhchia:
### - What I did
The `rlp_decoder_tx` changed and the tx hex and contract address in the instruction in comment is therefore stale. So I just update them.

### - How I did it

Karl told me about this modification. The hex also used in `./tests/setup_transaction_tests.py`. Besides, I verified the address of the tx created from the hex.

### - How to verify it

Create a tx with the hex using `rlp.decode` and check if its address is equaled to the address `RLP_DECODER_ADDRESS`

### - Cute Animal Picture
hqdefault


mhchia:
Besides, is it a good idea to have an API like `get_rlp_decoder_tx_hex` in viper, which can avoid hardcoded tx hex for the viper users?

DavidKnott:
@mhchia I like that idea a lot!  Would you mind making an issue for it?

fubuloubu:
Thanks!

But also @DavidKnott would we need that? I don't like having static addresses in the code/tests unless there's a very good reason

DavidKnott:
@fubuloubu Yes we do because `RLP_DECODER` is a separate contract that needs to be deployed every time.  Down the road we'll be deploying it on the main chain / testnets and then all viper contracts will use the address of those already deployed contracts.

mhchia:
@DavidKnott Definitely don't mind it. I'm working on it.:)

mhchia:
https://github.com/ethereum/viper/issues/522

DavidKnott:
Thanks a lot!

jacqueswww:
Closing as this is just too old (spring cleaning), please an open issue if this is still relevant.

PR #524, opened 2017-11-29
Add logging error to argument number mismatch and improve bytes support

### LINKED ISSUE #519: Compiler doesn't check the number of event arguments consistent with event declarations.
### What's your issue about?
```
Transfer: __log__({_from: indexed(num), _to: indexed(num), _value: num})

@public
def foo():
    log.Transfer(1, 2, 3, 4, 5, 6, 7, 8, 9 ,10)
```
The compiler compiles the above program (but should reject as the number of arguments doesn't match the number of parameters in the declaration).

### How can it be fixed?

In stmt.py, add proper check in the `def call(self)` method.

#### Cute Animal Picture

image

### LINKED ISSUE #520: Logging events with bytes array do not work as expected
When declaring an event with a bytes array parameter, like so;
```
Newgreeting: __log__({_sender: indexed(address), _newgreeting: bytes <= 20})
```

One would expect that when you would log a bytes array value into the log line;
```
@public
@payable
def setGreeting(_greeting: bytes <= 20) -> bool:
    self.greeting = _greeting
    log.Newgreeting( msg.sender, _greeting)
    return True
```

So with ```_greeting = "Hello"``` the expected payload of the event call back argument should be:
```
0x48656c6c6f000000000000000000000000000000
```

The code however returns
```
0x0000000000000000000000000000000000000000
```

Setting the log line with a string literal "Hello", so;
```
@public
@payable
def setGreeting(_greeting: bytes <= 20) -> bool:
    self.greeting = _greeting
    log.Newgreeting( msg.sender, "Hello")
    return True
```

results in the right string in the payload of the event call back argument.

Diving a bit deeper into this issue with a ```viper -f ir <filename.v.py>``` call shows that the compiler tries to convert the bytes array into a number of some kind.  This is not expected, one would expect a simple byte-4-byte memory copy from one bytes array to the other.

```
           # Line 16
            [mstore,
              448,
              /* bytearray to number, verify no leading zbytes */
              [with,
                _sub,
                352 <_greeting>,
                [with,
                  _el1,
                  [mload, [add, 32, _sub]],
                  [with,
                    _len,
                    [clamp, 0, [mload, _sub], 32],
                    [seq,
                      [assert, [or, [iszero, _len], [div, _el1, [exp, 256, 31]]]],
                      [div, _el1, [exp, 256, [sub, 3, _len]]]]]]]],
```

>  I guess this is where you insert a cute animal picture ;-)
Cute Cat picture

DavidKnott:
### - What I did
Add a check to make sure the number of logging arguments inputed match the number of  logging arguments declared.
Fixes #519 
Add support for data byte arrays of different sizes.
Fixes #520

### - How I did it
Added a conditional that checks to make sure a log is receiving the correct number of arguments.
Change the offset of `bytearray_to_num` to be the `maxlen` of a bytearray though this needs to be refined further (I'm thinking in a future PR) because the current bytes output padding is reversed (see below):
```
MyLog: __log({arg1: bytes <= 20})

def foo(_arg1: bytes <= 20):
    log.MyLog(_arg1)
```
```
c.foo('hello')
outputs => b'\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00hello'
should be => b'hello\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00'
```
### - How to verify it
Look at the tests I added
None
### - Cute Animal Picture

image

### DIFF
--- a/tests/parser/features/test_logging.py
+++ b/tests/parser/features/test_logging.py
@@ -1,6 +1,6 @@
 import pytest
 from tests.setup_transaction_tests import chain as s, tester as t, ethereum_utils as u, check_gas, \
-    get_contract_with_gas_estimation, get_contract, assert_tx_failed
+    get_contract_with_gas_estimation, get_contract
 from viper.exceptions import VariableDeclarationException, TypeMismatchException, StructureException
 
 
@@ -107,7 +107,11 @@ def test_event_logging_cannot_have_more_than_three_topics(assert_tx_failed):
     loggy_code = """
 MyLog: __log__({arg1: indexed(bytes <= 3), arg2: indexed(bytes <= 4), arg3: indexed(address), arg4: indexed(num)})
 
+Transfer: __log__({_from: indexed(num), _to: indexed(num), _value: num})
+
 @public
+def foo():
+    log.Transfer(1, 2, 3, 4, 5, 6, 7, 8, 9 ,10)@public
 def foo():
     log.MyLog('bar', 'home', self)
     """
@@ -163,7 +167,7 @@ def foo():
     assert c.translator.decode_event(logs.topics, logs.data) == {'arg1': [1, 2], 'arg2': [timestamp, timestamp + 1, timestamp + 2], 'arg3': [[1, 2], [1, 2]], '_event_type': b'MyLog'}
 
 
-def test_logging_with_input_bytes(bytes_helper):
+def test_logging_with_input_bytes_1(bytes_helper):
     loggy_code = """
 MyLog: __log__({arg1: indexed(bytes <= 4), arg2: indexed(bytes <= 29), arg3: bytes<=31})
 
@@ -171,6 +175,7 @@ MyLog: __log__({arg1: indexed(bytes <= 4), arg2: indexed(bytes <= 29), arg3: byt
 def foo(arg1: bytes <= 29, arg2: bytes <= 31):
     log.MyLog('bar', arg1, arg2)
 """
+
     c = get_contract_with_gas_estimation(loggy_code)
     c.foo('bar', 'foo')
     logs = s.head_state.receipts[-1].logs[-1]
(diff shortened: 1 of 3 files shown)


tests/parser/features/test_logging.py
```diff
@@ -107,7 +107,11 @@ def test_event_logging_cannot_have_more_than_three_topics(assert_tx_failed):
     loggy_code = """
 MyLog: __log__({arg1: indexed(bytes <= 3), arg2: indexed(bytes <= 4), arg3: indexed(address), arg4: indexed(num)})
 
+Transfer: __log__({_from: indexed(num), _to: indexed(num), _value: num})
+
 @public
+def foo():
+    log.Transfer(1, 2, 3, 4, 5, 6, 7, 8, 9 ,10)@public
 def foo():
     log.MyLog('bar', 'home', self)
```
fubuloubu: If `foo()` defined twice here? How does this work? Also `@public` for the second `foo()` is messed up

tests/parser/features/test_logging.py
```diff
@@ -182,7 +187,53 @@ def foo(arg1: bytes <= 29, arg2: bytes <= 31):
     # # Event abi is created correctly
     assert c.translator.event_data[event_id] == {'types': ['bytes4', 'bytes29', 'bytes31'], 'name': 'MyLog', 'names': ['arg1', 'arg2', 'arg3'], 'indexed': [True, True, False], 'anonymous': False}
     # Event is decoded correctly
-    assert c.translator.decode_event(logs.topics, logs.data) ==  {'arg1': b'bar\x00', 'arg2': bytes_helper('bar', 29), 'arg3': bytes_helper('foo', 31), '_event_type': b'MyLog'}
+    assert c.translator.decode_event(logs.topics, logs.data) ==  {'arg1': b'bar\x00', 'arg2': bytes_helper('bar', 29), 'arg3': bytes_helper('', 28) + b'foo', '_event_type': b'MyLog'}
```
fubuloubu: Add a `#TODO` comment to remember to fix this once you figure out the issue with the padding

tests/parser/features/test_logging.py
```diff
@@ -404,17 +455,37 @@ def test_logging_fails_after_a_global_declaration(assert_tx_failed):
...
 
+def test_logging_fails_when_number_of_arguments_is_greater_than_declaration(assert_tx_failed):
+    loggy_code = """
+MyLog: __log__({arg1: num})
+
+@public
+def foo():
+    log.MyLog(1, 2)
+"""
+    assert_tx_failed(lambda: get_contract_with_gas_estimation(loggy_code), VariableDeclarationException)
```
fubuloubu: Does it really matter to use `get_contract_with_gas_estimation()` when it's going to fail anyways?

Unless `get_contract()` isn't imported, then disregard

jacqueswww: @fubuloubu I think we just wrap everything with `get_contract_with_gas_estimation` now so we have code to test the estimation on :)

viper/parser/parser.py
```diff
@@ -693,7 +693,7 @@ def pack_args_by_32(holder, maxlen, arg, typ, context, placeholder):
             if input.typ.maxlen > typ.maxlen:
                 raise TypeMismatchException("Data input bytes are to big: %r %r" % (input.typ, typ))
             if arg.id in context.vars:
-                size = context.vars[arg.id].size
+                size = input.typ.maxlen
```
jacqueswww: we should really rename input actually, since it's a python built-in function.

tests/parser/features/test_logging.py
fubuloubu: Then maybe we can rename all to `get_contract`, keeping the gas estimation inside. No since in having two fixtures in that case

viper/parser/parser.py
fubuloubu: 💯

tests/parser/features/test_logging.py
jacqueswww: True, and of we want to switch estimates off we just have parameter for get_contract ;)

fubuloubu: Yeah `def get_contract(code, check_gas=True)`

tests/parser/features/test_logging.py
DavidKnott: @fubuloubu I didn't wanna add a TODO so I just figured out the issues with padding 👍

DavidKnott:
@fubuloubu I made the changes you requested and fixed the issue with padding.

PR #526, opened 2017-11-29
Change @internal decorator to @private decorator.

jacqueswww:
### - What I did

Implements https://github.com/ethereum/viper/issues/471.

### - How I did it

Was quite simple, just renamed internal in function_signature.py to private, and then replaced all other occurrences.

### - How to verify it
Compile something like:
```python

@private
def _add(a: num, b: num) -> num:
    return a + b


@public
def add(a: num, b: num) -> num:
    self._add(a, b)

```
And check/ttest the abi and contract output.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -48,10 +48,11 @@ Following the principles and goals, Viper **does not** provide the following fea
 Compatibility-breaking Changelog
 ********************************
 
+* **2017.11.29**: ``@internal`` renamed to ``@private```.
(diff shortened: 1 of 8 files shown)


fubuloubu:
Nice. You `grep`'d for all the uses right? I love grep

jacqueswww:
@fubuloubu I did, looks good to me :smile_cat: 
image

DavidKnott:
LGTM

jacqueswww:
Cool, thanks :smile:

PR #532, opened 2017-11-30
Fix variable list logging.

### LINKED ISSUE #521: Event logging doesn't handle the situation when the type of a variable ListType.
### What's your issue about?
Two programs:
```
Bar: __log__({_value: num[2]})

@public
def foo():
    log.Bar([1, 2])
```
```
Bar: __log__({_value: num[2]})

@public
def foo():
    a = [1, 2]
    log.Bar(a)
```
The first program compiles but the second program throws python exception "AttributeError: 'Name' object has no attribute 'elts'" because `pack_args_by_32` only handles list literals.

### How can it be fixed?

In `pack_args_by_32`, add logic to handle variable of ListType.

#### Cute Animal Picture

image

jacqueswww:
### - What I did

Fix #521, allowing logging using a variable of list type.

### - How I did it

Added section that copies list memory section into placeholder, see changes in `parser.py`.

### DIFF
--- a/tests/parser/features/test_logging.py
+++ b/tests/parser/features/test_logging.py
@@ -1,6 +1,6 @@
 import pytest
 from tests.setup_transaction_tests import chain as s, tester as t, ethereum_utils as u, check_gas, \
-    get_contract_with_gas_estimation, get_contract, assert_tx_failed
+    get_contract_with_gas_estimation, get_contract, assert_tx_failed, get_last_log
 from viper.exceptions import VariableDeclarationException, TypeMismatchException, StructureException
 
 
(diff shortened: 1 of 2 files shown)


tests/parser/features/test_logging.py
```diff
@@ -448,3 +448,32 @@ def ioo(inp: bytes <= 100):
     c.ioo(b"moo4")
     assert s.head_state.receipts[-1].logs[0].data == b'moo4'
     print("Passed raw log tests")
+
+
```
DavidKnott: I'd love to see a test with a list as function input.

DavidKnott: Can you add a test with a list as function input?  Can you also add a test with a list of a different type?

jacqueswww:
@DavidKnott thanks for the feedback. Also just a note, #524 might conflict with this one being in the same function, I will adapt this one on as necessary if that one merges to master.

DavidKnott:
@jacqueswww Could you quick rebase this?

DavidKnott:
LGTM, thanks a lot!

PR #533, opened 2017-12-01
 Check return type of lists match.

### LINKED ISSUE #531: Compiler doesn't check the actual return type (when it is a ListType)
### What's your issue about?
```
@public
def foo()->bool[2]:
    a: decimal[2]
    a[0] = 1
    return a
```
```
@public
def foo()->bool[2]:
    a: bool[1000]
    a[0] = 1
    return a
```
Viper accepts the above programs which don't seem to be correct. The compiler will neither check the list subtype nor the number of elements in the list.

### How can it be fixed?

add checking in `def parse_return`

#### Cute Animal Picture

image

jacqueswww:
### - What I did

Adds Fix #531.

### - How I did it

Was a bit trickier to detect the subtypes, I opted using split() because it interprets nested lists as well, without having to chain down to subtype.typ.subtype scenario, using `re` was actually the cleaner option! :smile:  

### - How to verify it

Check the issue #531, basically return a list of a different type.

N/A

### DIFF
--- a/tests/parser/syntax/test_list.py
+++ b/tests/parser/syntax/test_list.py
@@ -109,6 +109,20 @@ b: num[5]
 def foo():
     x = self.b[0].cow
     """,
+    """
+@public
+def foo()->bool[2]:
+    a: decimal[2]
+    a[0] = 1
+    return a
+    """,
+    """
+@public
+def foo()->bool[2]:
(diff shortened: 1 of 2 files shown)


DavidKnott:
LGTM

Issue #504, opened 2017-11-24
decimal number range check is not consistent with the document


yzhang90:
### What's your issue about?
In the types doc, it says `decimal` is a value with a precision of 10 decimal places between `-2**127` and `2**127 - 1`. In the following program 
```
a:decimal
@public
def foo():
    self.a = 170141183460469231731687303715884105727.888
```
a is greater than 2**127 -1 but it still compiles.

### How can it be f**ixe**d?

In expr.py:
```
def number(self):
        ...
        elif isinstance(self.expr.n, float):
            numstring, num, den = get_number_as_fraction(self.expr, self.context)
            if not (-2**127 * den < num < 2**127 * den):
                ...

```
change from `-2**127 * den < num < 2**127 * den` to `SizeLimits.MINNUM * den <= num <= SizeLimits.MAXNUM *den`

#### Cute Animal Picture

![image](https://user-images.githubusercontent.com/5641590/33224087-79ff227e-d12b-11e7-9be3-d94f29ffa9bc.png)

fubuloubu:
Typically the decimal counts towards the total, but I agree with the clamping here. Otherwise we would have information loss when converting to `num` when the value of decimal is above `2**127 - 1 + 0.5`

jacqueswww:
On it :) thanks @yzhang90 !

PR #534, opened 2017-12-01
Clamping decimal literals correctly.

jacqueswww:
### - What I did

Fix for #504 clamping of decimal literal.

### - How I did it
Just used suggested limit of den * MAXNUM/MINNUM & added test.

### - How to verify it
Check test.


DavidKnott:
LGTM

PR #537, opened 2017-12-03
Create contract data type

DavidKnott:
### - What I did
Make it possible to declare contracts as discussed on the bimonthly Viper call:
The following syntax can now be used to make external contract calls:
```
    contract_1 = """
@public
def bar() -> num:
    return 1
"""

    contract_2 = """
class Bar():
    def bar() -> num: pass

bar_contract: Bar

@public
def foo(contract_address: contract(Bar)) -> num:
    self.bar_contract = contract_address
    return self.bar_contract.bar()
    """

    c1 = get_contract(contract_1)
    c2 = get_contract(contract_2)
    assert c2.foo(c1.address) == 1
```
### - How I did it
 1) Allow for external contracts to be used when declaring variables
 2) Create a special contract data type that can be used to save the address of an external contract
### - How to verify it
Look at the tests I added and make sure that I didn't break any existing functionality
Add contract data type
### - Cute Animal Picture
image

### DIFF
--- a/tests/parser/features/external_contracts/test_external_contract_calls.py
+++ b/tests/parser/features/external_contracts/test_external_contract_calls.py
@@ -3,6 +3,7 @@ from tests.setup_transaction_tests import chain as s, tester as t, ethereum_util
     get_contract_with_gas_estimation, get_contract
 from viper.exceptions import StructureException, VariableDeclarationException, InvalidTypeException
 
+
 def test_external_contract_calls():
     contract_1 = """
 @public
@@ -161,8 +162,9 @@ def set_lucky(_lucky: num):
 lucky: public(num)
 
 @public
-def set_lucky(_lucky: num):
+def set_lucky(_lucky: num) -> num:
     self.lucky = _lucky
+    return self.lucky
     """
 
     lucky_number_2 = 3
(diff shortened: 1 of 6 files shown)


fubuloubu:
I like it! We need to change the `class` keyword when defining the contract interface to something else though, either `contract` or `interface`. I understand this would need to come after we've defined our own front-end parser instead of using Python's

Example:
```python
contract Bar(): # a keyword that doesn't imply inheritance
    bar() -> num # def and pass not really necessary

bar: Bar

@public
def foo(contract_address: Bar) -> num: # Bar is already a contract interface type
    self.bar = contract_address
    return self.bar.bar()
```

fubuloubu:
And then maybe in the future we can define relative references like so:

Bar.vy:
```python
@public
def bar() -> num:
    return 1
```
Foo.vy
```python
# Use a keyword that doesn't imply inheritance
Bar: contract("./Bar.vy")
bar: Bar

@public
def foo(contract_address: Bar) -> num: # Bar is already a contract interface type
    self.bar = contract_address
    return self.bar.bar()
```

This would find and parse the relevant file, and get all the method signatures for that file. We could probably extend this in some way to be inter-operable with solidity and other languages (for working with already deployed contracts)

tests/parser/features/external_contracts/test_external_contract_calls.py
```diff
@@ -375,3 +377,130 @@ class Foo():
...
+
+@public
+def set_lucky(contract_address: contract(Bar)):
+    self.bar_contract = contract_address
+    self.bar_contract.set_lucky(1)
+
+@public
+def get_lucky(contract_address: contract(Bar)) -> num:
+    self.bar_contract = contract_address
+    return self.bar_contract.get_lucky()
```
jacqueswww: @DavidKnott I would love one more example/test here:

1.a) set the address of bar_contact to one contract (with something like, set_contract).
1.b) call get_lucky() and returns let's say 1.

2.a) then set the address to a secondary contract.
2.b) call get_luck() again an returns let's say 2.

It's basically just a slightly more elaborate version of this contract, but it steps over multiple blocks  ;)

viper/parser/stmt.py
```diff
@@ -129,7 +129,14 @@ def call(self):
...
             contract_name = self.stmt.func.value.attr
-            contract_address = Expr.parse_value_expr(ast.parse('self.token_address').body[0].value, self.context)
+            var = self.context.globals[self.stmt.func.value.attr]
+            contract_address = unwrap_location(LLLnode.from_list(var.pos, typ=var.typ, location='storage', pos=getpos(self.stmt), annotation='self.' + self.stmt.func.value.attr))
+            return external_contract_call_stmt(self.stmt, self.context, contract_name, contract_address)
+        elif isinstance(self.stmt.func.value, ast.Attribute) and self.stmt.func.value.attr in self.context.globals:
+            contract_name = self.context.globals[self.stmt.func.value.attr].typ.unit
+            var = self.context.globals[self.stmt.func.value.attr]
+            contract_address = unwrap_location(LLLnode.from_list(var.pos, typ=var.typ, location='storage', pos=getpos(self.stmt), annotation='self.' + self.stmt.func.value.attr))
+            # import pdb; pdb.set_trace()
```
jacqueswww: remove pdb ;)

DavidKnott:
@jacqueswww I added the tests you suggested, merging this in.  Now I'll get started on removing the old way to make external contract calls.

Issue #548, opened 2017-12-06
Duplication semantics between "suicide" and "selfdestruct"


yzhang90:
### What's your issue about?

Viper has support for `suicide` and  `selfdestruct`.  Their semantics are the same. To avoid confusion, it's better to keep only one statement.
 
### How can it be fixed?

Since `selfdestruct` is more self-explained than `suicide`, just keep `selfdestruct` and remove `suicide`.

#### Cute Animal Picture

![image](https://user-images.githubusercontent.com/5641590/33680637-e21f2c68-da87-11e7-8fef-8ca2b2498c78.png)

fubuloubu:
:+1:

fubuloubu:
just checked, no usage in any examples or anything

fubuloubu:
@DavidKnott `SUICIDE` opcode has a different number at the end inside `viper/opcodes.py` Whats up with that?

jacqueswww:
+1 let's remove `suicide` key word.

DavidKnott:
@fubuloubu Not sure, `SUCIDE` might've had a different cost before.  Won't look into it too much given that we're taking it out.

fubuloubu:
Lol, well if they do the same thing but have a different cost can we use the one that has the least cost behind the scenes?

DavidKnott:
@yzhang90 Closing this now that we've removed `sucide`

PR #549, opened 2017-12-06
Removed suicide keyword

fubuloubu:
### - What I did
Removed suicide per #548

### - How I did it
Grep magic

### - How to verify it
Tests fine!

Removed suicide keyword and opcode

### - Cute Animal Picture

Don't club baby seals

### DIFF
--- a/docs/old_readme.txt
+++ b/docs/old_readme.txt
@@ -62,7 +62,7 @@ Note that not all programs that satisfy the following are valid; for example, th
         OR break
         OR return <expr>
         OR send(<expr>, <expr>)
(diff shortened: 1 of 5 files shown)


fubuloubu:
Basically as a reminder that we made this decision? Could probably just modify one of the cases that checks for any unspecified function with a malformed keyword. I can do that, just trying to understand the utility.

DavidKnott:
+1 Adding an invalid test with `sucide`, other then that LGTM

fubuloubu:
Alright, here you go!

DavidKnott:
@fubuloubu Thanks a lot, LGTM

PR #550, opened 2017-12-07
Internal contract call gas estimates

jacqueswww:
### - What I did

- Improve internal contract call gas estimates.
- Added annotations so the LLL repr is easier to read.

### - How I did it

Found places in expr.py & stmt.py were internal calls are handed, then looked up the gas of those calls.

### - How to verify it

Check tests ;)


DavidKnott:
LGTM

PR #556, opened 2017-12-08
fix-return-list-type

### LINKED ISSUE #553: Return ListType is not properly implememted
### What's your issue about?
```
@public
def foo() -> decimal[2][2]:
    a = [[1,2],[3,4]]
    return a
```
will be rejected by the compiler, but
```
@public
def foo() -> decimal[2][2]:
    return [[1,2],[3,4]]
```
is accepted. 

### How can it be fixed?
In `parse_return` function, first remove the check `if sub_base_type != ret_base_type and sub.value != 'multi': ...` And you can directly return a memory address only when `sub_base_type == return_base_type` and `sub.value != "multi"`. Otherwise, it needs to go through make_setter function to do the conversion.

#### Cute Animal Picture

image

DavidKnott:
### - What I did
Fix returns so that lists as input/assigned value can be returned.
Fixes #553 
### - How I did it
Changed a check for returns in `stmt.py` to check that return types are matching.

### - How to verify it
Look at the tests I added
None
### - Cute Animal Picture
image

### DIFF
--- a/viper/parser/stmt.py
+++ b/viper/parser/stmt.py
@@ -363,14 +363,7 @@ class Stmt(object):
         elif isinstance(sub.typ, ListType):
             sub_base_type = re.split(r'\(|\[', str(sub.typ.subtype))[0]
             ret_base_type = re.split(r'\(|\[', str(self.context.return_type.subtype))[0]
-            if sub_base_type != ret_base_type and sub.value != 'multi':
-                raise TypeMismatchException(
-                    "List return type %r does not match specified return type, expecting %r" % (
-                        sub_base_type, ret_base_type
-                    ),
-                    self.stmt
-                )
-            if sub.location == "memory" and sub.value != "multi":
+            if sub_base_type == ret_base_type and sub.location == "memory" and sub.value != "multi":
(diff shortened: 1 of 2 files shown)


viper/parser/stmt.py
```diff
@@ -363,14 +363,7 @@ def parse_return(self):
...
             ret_base_type = re.split(r'\(|\[', str(self.context.return_type.subtype))[0]
-            if sub_base_type != ret_base_type and sub.value != 'multi':
-                raise TypeMismatchException(
-                    "List return type %r does not match specified return type, expecting %r" % (
-                        sub_base_type, ret_base_type
-                    ),
-                    self.stmt
-                )
-            if sub.location == "memory" and sub.value != "multi":
+            if sub_base_type == ret_base_type and sub.location == "memory" and sub.value != "multi":
```
jacqueswww: I think having an exception here will still be useful?

fubuloubu: Agree, is there a way that this Exception no longer matters? Sort of unexplained.

jacqueswww:
Updated in #719

PR #565, opened 2017-12-09
Split comments over two lines and minor edits: safe remote purchases

jamesray1:
Split comments over two lines where necessary to fit on the page on readthedocs in the safe remote purchases example Python code. Update Safe Remote Purchases doc: line numbers, minor edits.
screenshot from 2017-12-09 17-20-03


DavidKnott:
LGTM

PR #567, opened 2017-12-10
Update to use conftest; fix some incorrect tests

mslipper:
@DavidKnott - like we talked about over breakfast the other day, I've updated the tests to use `conftest` files. I also discovered a few tests that looked broken (i.e., no assertions or invalid assertions) so I took the liberty of fixing those as well. I've placed review comments on all those places 😄 

Otherwise, all the actual testing logic should be the same as it was before, just moved into `conftest` and refactored to use fixtures.

Cute animal image:

image


tests/parser/features/external_contracts/test_external_contract_calls.py
```diff
@@ -168,7 +165,7 @@ def set_lucky(_lucky: num) -> num:
     """
 
     lucky_number_2 = 3
-    c2 = get_contract(contract_1)
+    c2 = get_contract(contract_2)
```
mslipper: This looked broken.

tests/parser/types/test_bytes.py
```diff
@@ -196,9 +191,11 @@ def foo(x: bytes <= 32) -> num:
         success = True
     except:
         success = False
+    assert not success
```
mslipper: There were no assertions for these cases. I assumed that we're expecting `not success` here.

fubuloubu: These cases should actually be using `assert_tx_failed` fixture instead

.gitignore
```diff
@@ -31,3 +31,4 @@ docs/modules.rst
 
 # IDEs
 .idea/
+*.iml
```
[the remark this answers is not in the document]
fubuloubu: Also, in general I disagree with putting your personal choice of IDE-generated files into our project gitignore. The proper process is to edit the `.gitignore_global` in your home directory or startup directory to contain whatever editors you use (I use vim, so all my `.swp` files get ignored from the globals file). Git picks up these global ignores from that file and prevents it from getting into your projects. If everyone listed their IDE files in our project, the gitignore would be like hundreds of lines long.

The only exception to this would be something the project uses like the py.test `.coverage/`directory, etc.

tests/parser/features/external_contracts/test_external_contract_calls.py
DavidKnott: Good catch

tests/conftest.py
```diff
@@ -19,6 +60,16 @@ def t():
     tester.s = tester.Chain()
     return tester
 
+@pytest.fixture(scope="module")
+def chain():
+    s = tester.Chain()
+    s.head_state.gas_limit = 10**9
+    return s
+
```
DavidKnott: Eventually I think we can remove the `t.s = chain` patter from our test files and either use chain itself or set `t.s = chain` in the `conftest.py`, that can be another PR though.

.gitignore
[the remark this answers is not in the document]
DavidKnott: @fubuloubu You bing up a good point.  I hadn't thought about minimizing the `.gitignore` before by not customizing to specific IDE's.  Do any `IDE`'s use the local `.gitignore` (vs `.gitignore_global`) for something special?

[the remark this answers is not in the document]
fubuloubu: It works like a Cascade, git first checks out your global and then appends the entries from the local git repo root, and finally from the folder down, if you only need it in a specific directory.

You can have your settings live in either place, but the best rule of thumb is to have your personal preferences stored in your global, and project specific entries live in the repo root .gitignore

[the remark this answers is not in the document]
mslipper: I had no idea about `.gitignore_global`! I'll remove that change and update my global gitignore.

tests/conftest.py
mslipper: Will do that separately - I think that might break some existing tests that rely on shared state between test cases, so I want to be sure that I get it right.

tests/parser/functions/rlp/test_block_number.py
```diff
@@ -0,0 +1,10 @@
+def test_block_number(get_contract_with_gas_estimation, fake_tx):
+    fake_tx()
+
+    block_number_code = """
+@public
+def block_number() -> num:
+    return block.number
+"""
+    c = get_contract_with_gas_estimation(block_number_code)
+    assert c.block_number() == 2
```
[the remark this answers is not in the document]
mslipper: That's fair. I can refactor.

tests/parser/types/test_bytes.py
mslipper: Will update.

mslipper:
@fubuloubu @DavidKnott updates are in. I'm also happy to take a stab at creating the `eth-tester` plugin - any recommended starting points?

.gitignore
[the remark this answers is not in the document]
fubuloubu: Awesome! If you could also remove that one for `.idea/` and the comment above, that would be great. No idea why I didn't reject that addition at the time.

fubuloubu:
@mslipper `eth-tester` is actually a separate repository in this organization, checkout https://github.com/ethereum/eth-tester

There is plans to convert `eth-tester` to a py.test plugin, after we complete `pyethereum20` integration and also maybe `py-evm` is completed. We can start the process of moving our Viper conftest fixtures to use `eth-tester`, which may require some PRs to eth-tester itself to improve what it has to provide for use in our own test cases. Eventually `eth-tester` will get remade as a py.test plugin and be called `pytest-ethereum`

This is why it's a long term plan lol.

fubuloubu:
LGTM by the way!

DavidKnott:
@mslipper Thanks a lot for the contribution!  Merging it in.

jacqueswww:
@mslipper Very cool! :) Saw this in master with my last PR, much appreciated!

PR #568, opened 2017-12-11
Improve for loop error

### LINKED ISSUE #562: Failed to compile a for loop with a variable as a range() argument
This code compilation is failing with the next error message: _**"Repeat must have a nonzero positive integral number of rounds"**_.

`some_value = 4`
`for i in range(some_value):` 
`// do something`

The same situation is present if check that the value is grater than 0:

`some_value = 4`
`if some_value > 0:` 
`for i in range(some_value):` 
`// do something`

* viper Version: 0.0.2

DavidKnott:
### - What I did
Improve for loop error message. @khotkevych is is this error clearer?
Fixes #562 
### - How I did it
Looked at what caused the error and changed the error message to match it. 
### - How to verify it
Look at the error I changed.
None
### - Cute Animal Picture
image

### DIFF
--- a/viper/parser/stmt.py
+++ b/viper/parser/stmt.py
@@ -182,7 +182,7 @@ class Stmt(object):
         # Type 1 for, eg. for i in range(10): ...
         if len(self.stmt.iter.args) == 1:
             if not isinstance(self.stmt.iter.args[0], ast.Num):
-                raise StructureException("Repeat must have a nonzero positive integral number of rounds", self.stmt.iter)
+                raise StructureException("Range only accepts literal values", self.stmt.iter)
             start = LLLnode.from_list(0, typ='num', pos=getpos(self.stmt))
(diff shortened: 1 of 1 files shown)


fubuloubu:
I think @khotkevych's example actually still satisfies what the error message says, but it doesn't compile (and I don't think it should). The error message should probably say something more like "range only accepts positive, non-zero integers as arguments"

DavidKnott:
@fubuloubu It does satisfy the error message but the error that triggers the ones I changed is only when the number of iterations isn't fixed as different errors are triggered when `num_iterations <= 0`.

fubuloubu:
Ah, maybe something like "range only accepts literal values"

jacqueswww:
+1 for "range only accepts literal values"

DavidKnott:
Changed it to "range only accepts literal values," merging it in.

PR #577, opened 2017-12-12
Removed "..note :: Unlike the existence of the function ``public()``,…

jamesray1:
… there is no equivalent ``private()`` function. Variables simply default to private if initiated without the ``public()`` function."

### - Cute Animal Picture

screenshot from 2017-12-12 14-50-51

### DIFF
--- a/docs/viper-by-example.rst
+++ b/docs/viper-by-example.rst
@@ -261,11 +261,6 @@ Like other examples, we begin by initiating our variables - except this time,
 we're not calling them with the ``public`` function. Variables initiated this
 way are, by default, private.
 
-..note ::
-Unlike the existence of the function ``public()``, there is no equivalent
-``private()`` function. Variables simply default to private if initiated
-without the ``public()`` function.
-
 The ``funders`` variable is initiated as a mapping where the key is a number,
 and the value is a struct representing the contribution of each participant.
 This struct contains each participant's public address and their respective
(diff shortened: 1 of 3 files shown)


docs/viper-by-example.rst
```diff
@@ -255,16 +255,10 @@ previous examples. Let's dive right in.
...
 
 Like other examples, we begin by initiating our variables - except this time,
-we're not calling them with the ``public`` function. Variables initiated this
-way are, by default, private.
-
-..note ::
-Unlike the existence of the function ``public()``, there is no equivalent
-``private()`` function. Variables simply default to private if initiated
-without the ``public()`` function.
+we're not calling them with the ``public`` function. Instead, we call them with the ``private`` function.
```
jacqueswww: Just this one should be altered tho state that it's implicit private.

jamesray1: OK, done.

fubuloubu:
@jacqueswww I think the fact we re-use `public` to construct a getter for a given global is a little confusing. I wonder if there is a better or clearer term where we're not reusing something, like maybe a decorator `@getter` instead of this thing that goes around a variable's signature

jamesray1:
From the simple open auction section:

> You may notice all of the variables being passed into the ``public``
function. By declaring the variable *public*, the variable is
callable by external contracts. Initializing the variables without the  ``public``
function defaults to a private declaration and thus only accessible to methods
within the same contract. The ``public`` function additionally creates a
‘getter’ function for the variable, accessible with a call such as
``self.get_beneficiary(some_address)``.

DavidKnott:
LGTM, merging.

PR #580, opened 2017-12-12
Minor edits

jamesray1:
with ``num`` type > with the ``num`` type
crowdfunding period is over - as determined> crowdfunding period is over—as determined

### - Cute Animal Picture

screenshot from 2017-12-12 15-39-37


DavidKnott:
LGTM

PR #588, opened 2017-12-14
Packing list arguments

### LINKED ISSUE #554: Need support for packing List Type arugment
### What's your issue about?
```
@public
def foo(x:num[2]): 
    pass 
@public 
def bar(): 
    x:num[2] 
    self.foo(x)
```
The program fails to compile because packing List Type argument is not supported.

#### Cute Animal Picture

image

jacqueswww:
### - What I did

Fixes #554  - allow list to passed as function arguments.

### - How I did it
Looked at pack_arguments function.
I had a big struggle actually figuring out that with static arrays you pack them right in the parameter area. Unlike dynamic lists and dynamic byte arrays, that use a dynamic offset/pointer to the dynamic section after the parameter area.

### - How to verify it

Check the tests, I added a few to be extra sure ;)


DavidKnott:
Can you test making a call with a list from storage?

jacqueswww:
@DavidKnott will do thanks ;)

jacqueswww:
@DavidKnott I think this one ready now :smiley_cat:

DavidKnott:
Oops, I'm sorry.  Can you rebase this one more time? Merged something else in first.

jacqueswww:
@DavidKnott done ;)

PR #592, opened 2017-12-19
Check for missing return statement

jacqueswww:
### - What I did

Partial fix for #590.
Check for a minimum of 1 return statement if specified.

### - How I did it
Added a count in the function context.

### - How to verify it
```python
@public
def foo() -> num:
        pass
```

### DIFF
--- a/tests/parser/exceptions/test_invalid_payable.py
+++ b/tests/parser/exceptions/test_invalid_payable.py
@@ -27,6 +27,7 @@ x: num
 @payable
 def foo() -> num:
     self.x = 5
+    return self.x
     """,
     """
 @public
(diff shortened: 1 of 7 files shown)


fubuloubu:
LGTM for now, obviously will have to rewrite it when we do the branching check instead.

Actually, can you add a TODO note to that effect in the parser code?

DavidKnott:
Looks like a good first step, particularly because making sure every endpoint has the correct return is significantly more logic.  I'd love to see a test or two with conditional returns to make the limits of this fix clearer.

DavidKnott:
LGTM, merging it in.

fubuloubu:
Since we have the correct algorithm in #590 according to @yzhang90, perhaps we can upgrade the check now?

Issue #545, opened 2017-12-05
VIP: Implement lexical block level scoping in Viper


yzhang90:
## Preamble

    VIP: <to be assigned>
    Title: Implement lexical block-level scoping in Viper
    Author: @yzhang90 @daejunpark @pdaian @grosu
    Type: Standard Track
    Status: Draft
    Created: 2017-12-1

## Simple Summary
Viper should correctly implement block-level scoping semantics while fully disallowing shadowing to avoid misleading and error-prone code.

## Abstract
The semantics of variable scoping is very important for a programming language because the scoping determines how a (variable) name binds to the actual object. The current Viper compiler mixes different scoping approaches and makes it very hard to reason about programs. We propose uniform use of lexical block-level scoping.

## Motivation
The principles behind Viper are security, simplicity and auditability.  However, the current scoping in Viper is not clear and can make the code misleading and error-prone.

In Viper, a local variable is accessible throughout the whole function after it is declared.
For example, in the following program, `b` is equal to `1` if `choice` is `true` and `0` otherwise. Even though `a` is declared and assigned inside `if`, it is still accessible after the `if` statement finishes:
```
def foo(choice: bool):
    if (choice):
        a = 1
    b = a
```
However, index variables are treated differently from other local variables.
The scope of index variables is the `for` loop and the index variables are not accessible outside the loop. For example, Viper rejects the following program:
```
def foo():
    for i in range(10):
        pass
    i += 1
```
What's more, an annotated assignment `a:num`, which normally also initializes `a` to `0`, is only executed once in the loop and thus `b` equals `10` in the end:
```
def foo():
    for i in range(10):
        a:num
        a += 1
    b = a
```
With the current scoping rules, a program which declares a variable inside the `if...then...else` branch or the `for` loop is **almost** equivalent to declaring the variable before the `if` statement or the `for` loop.  For instance,
```
def foo(choice: bool) -> num:
     if choice:
         a = 1
         return a
     else:
         b = 2
         return a + b
```
is equivalent to
```
def foo(choice: bool) -> num:
     a: num
     b: num
     if choice:
         a = 1
         return a
     else:
         b = 2
         return a + b
```
The difference is that for the second program, one can negate the `choice` variable and swap the `then` and `else` branch, but you could not do the same thing for the first program (because `a` is not defined before expression `a + b`!).

Situations like the above make it unnecessarily difficult to reason about programs and program equivalence.
[...]

PR #601, opened 2017-12-22
Block scoping

jacqueswww:
### - What I did
Partial PR for #545. I chose to separate this part for easier review.

The `a:num =1` annotate assign will probably be a big onslaught of changes which I would rather keep separate from block scoping.

### - How I did it

Add a `blockscopes` list to context. Using the `start_blockscope` and `end_blockscope`method to set the start and end of a blockscope for `for` and `if` blocks.

### - How to verify it

Check the tests ;) (`test_blockscope.py`).

### - Cute Animal Picture

69f2c95efe348b7cefa11314c50da19b--cute-baby-bunnies-cutest-bunnies


viper/parser/parser.py
```diff
@@ -277,13 +277,31 @@ def __init__(self, vars=None, globals=None, sigs=None, forvars=None, return_type
...
     def remove_in_for_loop(self, name_of_list):
         self.in_for_loop.remove(name_of_list)
 
+    def start_blockscope(self, blockscope_id):
+        if blockscope_id not in self.blockscopes:
+            self.blockscopes.append(blockscope_id)
+
+    def end_blockscope(self, blockscope_id):
+        if blockscope_id not in self.blockscopes:
+            return
```
fubuloubu: Shouldn't this raise an error?

viper/parser/parser.py
```diff
@@ -277,13 +277,31 @@ def __init__(self, vars=None, globals=None, sigs=None, forvars=None, return_type
...
+        # Remove all variables that have specific blockscope_id attached.
+        poplist = []
+        for name, var_record in self.vars.items():
+            if blockscope_id in var_record.blockscopes:
+                poplist.append(name)
+        for name in poplist: self.vars.pop(name)
+        # Remove blockscopes
+        self.blockscopes = self.blockscopes[self.blockscopes.index(blockscope_id):]
```
fubuloubu: Simplify to `self.blockscopes.pop(blockscope_id)`

jacqueswww: Pop would give the desired effect. If you have nested scopes, you want to slice awat all nested scopes as well.

viper/parser/parser.py
jacqueswww: I could probably remove that check completely?

fubuloubu: Maybe, I mean this is an internal thing you added to keep track of blockscopes, which is something Python already does with the AST module to prevent uses from changing it arbitrarily.

Wait, does AST provide any information for this?

viper/parser/parser.py
fubuloubu: ah, that's an interesting thought. but the blockscopes should still both appear in order so I don't think this would happen e.g.

```python
for i in range(5): # "for" block started
    # "for" body
    if i == 3: # "if block started
         # "if" body
         return i
    # "if block ended
# "for" block ended
```

jacqueswww: Looking at it now pop would be 100%, was just my thought process at that moment... hahaha :smile_cat:

viper/parser/parser.py
jacqueswww: Not that I am aware no, since python doesn't use block scoping like we want to. I'll remove it - just a bit of 'defensive coding' ;)

DavidKnott: I don't think AST will help here as it only parses with python scoping.  I think this check can be removed given that we're only calling it internally and inputting `blockscope_id`'s that will be in `self.blockscopes`

viper/parser/stmt.py
```diff
@@ -100,11 +100,23 @@ def parse_if(self):
...
             add_on = [parse_body(self.stmt.orelse, self.context)]
+            self.context.end_blockscope(block_scope_id)
         else:
             add_on = []
-        return LLLnode.from_list(['if', Expr.parse_value_expr(self.stmt.test, self.context), parse_body(self.stmt.body, self.context)] + add_on, typ=None, pos=getpos(self.stmt))
+
+        block_scope_id = id(self.stmt)
+        self.context.start_blockscope(block_scope_id)
+        o = LLLnode.from_list(
+            ['if', Expr.parse_value_expr(self.stmt.test, self.context), parse_body(self.stmt.body, self.context)] + add_on,
```
DavidKnott: What does `self.stmt.test` do?

DavidKnott:
@jacqueswww Tests look good!

viper/parser/parser.py
fubuloubu: Python does scoping, at least behind the scenes. All languages do some aspect of scoping. Python specifically tracks indenting to figure out block scoping. This might be handled in the lexer and not made present in the AST, but it's definitely in the Python front-end parser somewhere.

fubuloubu: You need to pop scopes from the list to prevent this from compiling:
```python
def foo(b: bool):
    if b:
        a: num = 1
    a += 2
```

DavidKnott: @jacqueswww I don't think the above example will compile.  Does it currently?

viper/parser/parser.py
```diff
@@ -277,13 +277,31 @@ def __init__(self, vars=None, globals=None, sigs=None, forvars=None, return_type
...
+        self.blockscopes = []
 
     def set_in_for_loop(self, name_of_list):
         self.in_for_loop.add(name_of_list)
 
     def remove_in_for_loop(self, name_of_list):
         self.in_for_loop.remove(name_of_list)
 
+    def start_blockscope(self, blockscope_id):
+        if blockscope_id not in self.blockscopes:
```
OktayGardener: Will blockscopes be distinct? If so, then consider changing this into a set for O(log(n)) lookup time for the "not in" lookups (this expression degrades into a O(n) operation). More on this:
https://juliank.wordpress.com/2008/04/29/python-speed-x-in-list-vs-x-in-set/

jacqueswww: 👍

viper/parser/parser.py
jacqueswww: Did for me, will be reworking this PR today ;)

viper/parser/stmt.py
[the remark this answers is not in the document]
jacqueswww: @DavidKnott I believe it's basically the comparison expression or statement part: `if (<test>)`

jacqueswww:
@DavidKnott I believe this one is ready now too :)

PR #608, opened 2017-12-25
Implement continue

DavidKnott:
### - What I did
Implement `continue` which skips to the end of an iteration in a `for` loop.

### - How I did it
Looked at how `break` works and added `continue`.
Added a `JUMPDEST` to where `continue` skips to.

### - How to verify it
Look at the tests I created.

Implement `continue`
### - Cute Animal Picture
image

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -49,6 +49,7 @@ Compatibility-breaking Changelog
 ********************************
 
 * **2017.12.25**: Change name from Viper to Vyper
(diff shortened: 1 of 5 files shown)


fubuloubu:
Should add an `if` statement test e.g.
```python
for i in range(5):
   if i % 2 == 0:
       continue
   n += i
return n # should be 9
```

DavidKnott:
@fubuloubu I added the test you suggested, thanks for the review

PR #610, opened 2017-12-25
Change badges from viper to vyper

DavidKnott:
### - What I did
Change Vypers badges to match its new name.
### - How to verify it
Make sure the links I changed worke
None
### - Cute Animal Picture
image


fubuloubu:
Can we move the chat room history for the old one?

DavidKnott:
@fubuloubu We've can but we've gotta email gitter about it.

PR #612, opened 2017-12-25
Fix rlp decoding loop memory

DavidKnott:
### - What I did
Fix `rlp_decoder.py` so that `repeat` memory doesn't overlap with the output.

### - How I did it
Looked for where memory was being overwritten and fixed it.
### - How to verify it
Look at the tests I added.
None
### - Cute Animal Picture
image


tests/parser/functions/rlp/test_rlp_list.py
```diff
@@ -64,6 +64,16 @@ def too(inp: bytes <= 100) -> bool:
 def voo(inp: bytes <= 1024) -> num:
     x = RLPList(inp, [num, num, bytes32, num, bytes32, bytes])
     return x[1]
+
+@public
+def loo(inp: bytes <= 1024) -> num:
+    x = RLPList(inp, [num, num, num, num, num, num, num, num, num, num])
+    return x[0] + x[1] + x[2] + x[3] + x[4] + x[5] + x[6] + x[7] + x[8] + x[9]
```
OktayGardener: Haha, nice. Isn't this a pain coding tho?
``` sum(x[:9])```

File: tests/parser/functions/rlp/test_rlp_list.py [hunk 2]
```diff
@@ -64,6 +64,16 @@ def too(inp: bytes <= 100) -> bool:
 def voo(inp: bytes <= 1024) -> num:
     x = RLPList(inp, [num, num, bytes32, num, bytes32, bytes])
     return x[1]
+
+@public
+def loo(inp: bytes <= 1024) -> num:
+    x = RLPList(inp, [num, num, num, num, num, num, num, num, num, num])
```
OktayGardener: You can also do ```[num]*n```, ```[num]*10``` in your case.

tests/parser/functions/rlp/test_rlp_list.py
DavidKnott: @OktayGardener Thanks for your suggestions, but Viper doesn't support `sum` or 
`[num] * n

OktayGardener: Ah, I see, too bad.

PR #614, opened 2018-01-01
Implement floor divide

DavidKnott:
### - What I did
Implement `//` which is pythonic syntax for divide and floor

### - How I did it
Added support for `//` in `expr.py :: def arithmetic`

### - How to verify it
Look at the tests I added to `test_num.py` and `test_decimal.py`
Add support for divide and floor
### - Cute Animal Picture
image

### DIFF
--- a/viper/parser/expr.py
+++ b/viper/parser/expr.py
@@ -292,6 +292,17 @@ class Expr(object):
                 o = LLLnode.from_list(['exp', left, right], typ=BaseType('num', new_unit), pos=getpos(self.expr))
             else:
                 raise TypeMismatchException('Only whole number exponents are supported', self.expr)
+        elif isinstance(self.expr.op, ast.FloorDiv):
+            if left.typ.positional or right.typ.positional:
+                raise TypeMismatchException("Cannot use divide and floor positional values!", self.expr)
+            new_unit = combine_units(left.typ.unit, right.typ.unit, div=True)
+            if ltyp == rtyp == 'num' or ltyp == rtyp == 'decimal':
+                o = LLLnode.from_list(['sdiv', left, ['clamp_nonzero', right]], typ=BaseType('num', new_unit), pos=getpos(self.expr))
+            elif ltyp == 'decimal' and rtyp == 'num':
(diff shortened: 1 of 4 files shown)


fubuloubu:
Does `num / num` produce a decimal currently?

DavidKnott:
No it doesn't

viper/parser/expr.py
```diff
@@ -292,6 +292,22 @@ def arithmetic(self):
...
+        elif isinstance(self.expr.op, ast.FloorDiv):
+            if left.typ.positional or right.typ.positional:
+                raise TypeMismatchException("Cannot use divide and floor positional values!", self.expr)
+            new_unit = combine_units(left.typ.unit, right.typ.unit, div=True)
+            if ltyp == rtyp == 'num':
+                o = LLLnode.from_list(['sdiv', left, ['clamp_nonzero', right]], typ=BaseType('num', new_unit), pos=getpos(self.expr))
+            elif ltyp == rtyp == 'decimal':
+                o = LLLnode.from_list(['with', 'l', left, ['with', 'r', ['clamp_nonzero', right],
+                                            ['sdiv', 'l', 'r']]],
+                                            # ['mul', 'l', DECIMAL_DIVISOR]
```
adhikasp: Why not just `['sdiv', left, ['clamp_nonzero', right]]` like when dividing `num` with `num`?

DavidKnott: @adhikasp Great catch, simplifying it.

fubuloubu:
@DavidKnott I think this might actually be a bit confusing since regular num-num division produces floor division by default, and only decimal division produces non-floor division. I agree there is a need to do floor division when dealing with decimals, I think this could be better accomplished by re-casting to a num, or otherwise adding a `floor` function that keeps the decimal type but removed the decimal value. The `floor` (and `ceil`) functions would be much more broadly useful, and less confusing in practice.

Either that, or require floor division for num-num division to make it clear that the outcome is floored division.

jacqueswww:
Closing, because https://github.com/ethereum/vyper/pull/677 was the solution we decided on instead ;)

PR #618, opened 2018-01-02
Fix in range with different types

### LINKED ISSUE #581: "in" list doesn't check whether variable type matches the list subtype
### What's your issue about?
```
@public
def testin(x: address) -> bool:
    y = 1
    s = [1, 2, 3, 4]
    if x in s:
        return True
    return False
```
This program is allowed by the Viper compiler. However the type of the varialbe `x` doesn't match the list subtype `num`.

### How can it be fixed?

Add type check in `build_in_comparator`. Pay attention to the situation when list is of MixedType.

#### Cute Animal Picture

image

DavidKnott:
### - What I did
Add a check to `in_range` so it fails if there's a type mismatch.
Fixes #581 
### - How I did it
Added a types check
### - How to verify it
Look at the test I added
None

### - Cute Animal Picture
image

### DIFF
--- a/tests/parser/features/iteration/test_range_in.py
+++ b/tests/parser/features/iteration/test_range_in.py
@@ -107,3 +107,15 @@ def is_owner() -> bool:
     # Owner in place 0 can be replaced.
     c.set_owner(0, t.a1)
     assert c.is_owner() is False
+
+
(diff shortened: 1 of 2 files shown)


jacqueswww:
Looks good to me :)

PR #619, opened 2018-01-02
Force type on assignment

### LINKED ISSUE #545: VIP: Implement lexical block level scoping in Viper

jacqueswww:
### - What I did

Partially fixes #545 - by forcing setting type on assignment.

### - How I did it
- Altered ann_assign and assign in `stmt.py`.
- Fixed a ton of tests.

### - How to verify it
Try a simple:

```python
def a ():
    c = 10
```


fubuloubu:
Okay yes, #545 does specify type specifications and we all agreed on moving forward with that so it's in the clear. Just wanted to make sure this pretty large change was understood by the community via VIP.

P.S. We should probably rev the build number and put this into the change log

jacqueswww:
@fubuloubu good idea. Yes I completely agree with it being a big change - when I started just before christmas, I even went and doubled checked on gitter to make sure it was intended ;)

fubuloubu:
@jacqueswww I'm thinking that all PRs which address/link to VIPs should have a mandated update to the change log or something (maybe including VIP info). And when there's a batch of them ready, @DavidKnott can do the rev number update when they are all merged in, updating the middle number if there is a breaking change (like this one), and the last number if it is a feature addition.

As much as I hate using the "p" word, I think we need to start following a process for this lol

jacqueswww:
@DavidKnott @fubuloubu I fixed checking on annotated assign, added a few more test cases and did a simple changelog entry. Not sure about the revision steps? As far as I know we don't have a version number anywhere yet?

fubuloubu:
It should be just typing `bumpversion` at the root folder

But I think @DavidKnott should handle that as the repo maintainer, he might want to batch multiple commits to upgrade the version at the same time

DavidKnott:
@jacqueswww Rebased and merged this in.

jacqueswww:
@DavidKnott excellent!

PR #620, opened 2018-01-03
Fix unit exponent expressions

### LINKED ISSUE #547: Exponent with units is not correct
### What's your issue about?

The following program is rejected by the compiler:
```
@public
def foo():
    a : num(wei)
    b : num(wei**2)
    a = 2
    b = a**2
```

### How can it be fixed?

In arithmetic pow, it calls `combine_units` which only handles multiplication and division, in my point of view.

#### Cute Animal Picture

image

jacqueswww:
### - What I did

- Fixes #547.
- Correct builds unit type when using ` ** `.

### - How I did it

See changes on `Expr.arithmetic`.

### - How to verify it
```python
@public
def foo():
    a : num(wei)
    b : num(wei**2)
    a = 2
    b = a**2
````

N/A

### DIFF
--- a/tests/parser/syntax/test_unit_exponents.py
+++ b/tests/parser/syntax/test_unit_exponents.py
@@ -1,0 +1,44 @@
+import pytest
+from pytest import raises
+
+from viper import compiler
+from viper.exceptions import TypeMismatchException
+
+fail_list = [
+    """
+@public
+def baa() -> decimal:
+    return 2.0 ** 2
+    """,
+    """
+@public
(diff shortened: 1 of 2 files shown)


fubuloubu:
LGTM, although your PR description above has a couple of mistakes I think. I was like "WTF?" when I saw `**@public` lol

jacqueswww:
Haha ** slipped on there 😂

DavidKnott:
LGTM, merging

PR #623, opened 2018-01-04
Fix err msg for functions without enough arguments

michaelsproul:
### - What I did

Fixed the error message that is displayed when a Vyper function is called without enough arguments.

### - How I did it

Added a `%` so that the string formatting actually happens, and pulled the function name from the `ast.Call` object using `element.func.id` (assuming the inner part of the `Call` is a `Name(id=...)`).

### - How to verify it

Run the included test case in `tests/parser/exceptions/test_insufficient_arguments.py`, or compile the example Vyper program therein.

Fix the error message for functions called without enough arguments

### - Cute Animal Picture

A pademelon, from Australia :D

Pademelon

### DIFF
--- a/viper/functions.py
+++ b/viper/functions.py
@@ -113,7 +113,10 @@ def signature(*argz, **kwargz):
                 elif isinstance(expected_arg, Optional):
                     subs.append(expected_arg.default)
                 else:
-                    raise StructureException("Not enough arguments for function %s", element)
+                    raise StructureException(
+                        "Not enough arguments for function: {}".format(element.func.id),
+                        element
+                    )
             kwsubs = {}
             element_kw = {k.arg: k.value for k in element.keywords}
             for k, expected_arg in kwargz.items():
--- a/tests/parser/exceptions/test_insufficient_arguments.py
+++ b/tests/parser/exceptions/test_insufficient_arguments.py
@@ -1,0 +1,20 @@
+import pytest
+from pytest import raises
+
+from viper import compiler
+from viper.exceptions import StructureException
+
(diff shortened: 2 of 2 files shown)


michaelsproul:
Oops, committed a cardinal sin by pushing before testing locally... (was having issues with `pycryptodome` on OS X at my end). Should be OK now...

michaelsproul:
... Third time lucky... mucked up Pep8...

viper/functions.py
```diff
@@ -113,7 +113,7 @@ def g(element, context):
                 elif isinstance(expected_arg, Optional):
                     subs.append(expected_arg.default)
                 else:
-                    raise StructureException("Not enough arguments for function %s", element)
+                    raise StructureException("Not enough arguments for function %s" % element.func.id)
```
jacqueswww: I think the even better fix would be: 

StructureException("Not enough arguments for function %s" % element.func.id, element)

By passing element as second parameter, the exception will report the exact the error has occured :smile: 

```
viper.exceptions.StructureException: line 3: Not enough arguments for function %s
    return as_wei_value(10)
-----------^

```

fubuloubu:
The `%` operator is really python2, use [`format` ](https://docs.python.org/3.6/library/string.html#format-examples)

michaelsproul:
Ok, all updated now.

I also added a semicolon to the message, but can take it out again if you guys have a strong opposition to it :)

DavidKnott:
@michaelsproul LGTM

michaelsproul:
Thanks! This is fun :D

PR #624, opened 2018-01-07
Add an error message for calls to undefined top-level functions

michaelsproul:
### - What I did

Previously, the compiler would error ungracefully if an undefined top-level function was called, due to an unhandled case in `Expr.call`. I filled in the missing case so that an appropriate error is raised instead.

I also made it so that the compiler offers suggestions when a non-existent top-level function
is mistakenly called instead of a contract method, e.g. `bar(..)` instead of `self.bar(..)`.

### - How I did it

I modified the existing handling of top-level functions in `Expr.call` so that the case where the function name is _not_ in the `dispatch_table` is handled correctly.

### - How to verify it

Run the tests in `tests/parser/exceptions/test_undef_function.py`.

Add an error message for calls to undefined top-level functions

### - Cute Animal Picture

Wombat Image

### DIFF
--- a/tests/parser/exceptions/test_undef_function.py
+++ b/tests/parser/exceptions/test_undef_function.py
@@ -1,0 +1,31 @@
+from pytest import raises
+
+from viper import compiler
+from viper.exceptions import StructureException
+
+
+def test_undef_toplevel():
+    code = """
+@public
+def foo():
+    x = bar(55)
+    """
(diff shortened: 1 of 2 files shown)


fubuloubu:
I like that! Lgtm

DavidKnott:
Thanks a lot for the contribution, merging.

PR #626, opened 2018-01-08
Prevent mixed lists

jacqueswww:
### - What I did

Removes the possibility to create mixed lists.

### - How I did it

Removed MixedType and added an exception where it would usually be created.

### - How to verify it
```python
@public
def test():
    a = [1, 1.33]
```
should throw necessary exception.

Mixed list support removed.


DavidKnott:
LGTM, gonna merge when I bump the version 👍

jacqueswww:
Perfect :)

Issue #501, opened 2017-11-24
Change project name


fubuloubu:
### What's your issue about?
ETH Zurich, a university in Switzerland, already has a programming language named [Viper](http://viper.ethz.ch/). On October 30th I was contacted by a member of their faculty with a request for us to change our name in order avoid potential confusion with the myriad of tools available for that project. Their project has the potential to target the Ethereum ecosystem for formal verification purposes, so a name change was important to them.

Over the past month we have discussed many possible name changes, including (for posterity):
* Adder/Blackadder
* Anaconda
* Serpent 3.0
* Ophidian/Ophidiam

In the end, `vyper` was chosen as our desired project name, in order to disambiguate with the other Viper project and match our `.vy` file extension choice.

### How can it be fixed?
In order to accomplish the name change, we need to do the following
* Acquire project names at the various services we use e.g. PyPI, ReadTheDocs, snapcraft, etc. (Please list all below)
* Change the name on GitHub
* Communicate to the development community the name change, how to update git remotes, etc.
* Communicate to the wider Ethereum community about the name change, and market the project in order to build recognition

Please list below if there are further concerns or comments/questions. This must be a concerted effort by our entire community in order to be successful.

jacqueswww:
* Change the python module/binary and all import statements to vyper ?

fubuloubu:
@jacqueswww honestly completely forgot about that lol

DavidKnott:
I have `eth-vyper` pypi and `vyper` read the docs.

fubuloubu:
@DavidKnott @jacqueswww @vbuterin What do you think the best way is to explain and market the name change to the wider community? And at what point in time? (first beta, first release, etc)

DavidKnott:
For explaining the name change I think we should just tell the community what happened.  In terms of releases I think Vyper is ready beta nowish.

fubuloubu:
@jacqueswww @DavidKnott I think we have all our ducks in a row on this, just need to finish #627 and #630. Can you guys think of anything else that needs to get done?

fubuloubu:
Lol, who marked this "Easy Pickings"?

jacqueswww:
This has been done. Wooooo :champagne: :fireworks:

PR #630, opened 2018-01-12
Rename python module to Vyper.

jacqueswww:
### - What I did

Part of #501.

OK here goes :drum: 
Renamed viper module to vyper. This should be merged right after the docs merge.

### - How I did it
n/a

### - How to verify it
I guess if tests pass we are OK ? haha

Renamed viper to vyper.


jacqueswww:
@fubuloubu @DavidKnott get's tricky now if you look at CI output, I also created the following PR at pyethereum:  https://github.com/ethereum/pyethereum/pull/854

fubuloubu:
That is quite a chicken and egg problem lol.

I think the best case scenario is that someone for pyethereum adds a new `vyper` option first, releases that to pypi, to which you upgrade the version of pyethereum we use in our package, THEN  allowing us to make the switch under the hood in this PR. After that is complete, cycle back to pyethereum and remove the old `viper` language option, which can be released at your leisure.

Eventually, we should move our tests to a testing framework built as a `py.test` plugin using `eth-tester` with `py-evm` as a backbone, but there is some additional stuff that is needed IMO before we can do this relating to how the compilers are handled, as well as adding some useful features for measuring coverage and profiling gas costs.

fubuloubu:
@DavidKnott needs a `pre beta` tag

DavidKnott:
@jacqueswww Can you rebase this?

jacqueswww:
@DavidKnott done :)

fubuloubu:
Soooo many changes. Thanks for your effort @jacqueswww!

jacqueswww:
Sure thing! Glad to see the rename is almost done 😁

PR #633, opened 2018-01-16
Fully clean project

fubuloubu:
Noticed some extra junk that wasn't getting cleaned out


DavidKnott:
LGTM, thanks for the contribution!

PR #639, opened 2018-01-22
Any length logging of data memory

jacqueswww:
### - What I did
Fixes #613.

- Data memory section of logs can now return any length bytes.
- Uses 'bytes' on abi level (no change needed on pyethereum side. Topics  (indexed) still use bytes[x] format, as topics are limited to 32bytes.

### - How I did it
Reworked the way `pack_args_by_32` and `pack_logging_data` work. Allowing for the extra required dynamic section (similar to a return statement).

### - How to verify it

Check the tests in `test_logging_bytes_extended.py`.


viper/parser/parser.py
```diff
@@ -700,71 +701,86 @@ def pack_logging_topics(event_id, args, expected_topics, context):
...
-            holder.append(LLLnode.from_list(['mstore', placeholder, bytes_to_int(bytez + b'\x00' * (32 - bytez_length))], typ=typ, location='memory'))
-        # Variables
-        else:
-            value = parse_expr(arg, context)
-            if value.typ.maxlen > typ.maxlen:
-                raise TypeMismatchException("Data input bytes are to big: %r %r" % (value.typ, typ))
-            holder.append(LLLnode.from_list(['mstore', placeholder, ['mload', ['add', value, 32]]], typ=typ, location='memory'))
+
+        source_expr = Expr(arg, context)
+        if isinstance(arg, ast.Str):
```
DavidKnott: Does `if isinstance(arg, ast.Str)` need to be in there twice? (see 725)

viper/parser/parser.py
```diff
@@ -773,13 +789,51 @@ def check_list_type_match(provided):  # Check list types match.
...
+        placeholder = context.new_placeholder(BaseType(32))
+        placeholder_map[i] = placeholder
+        if not isinstance(typ, ByteArrayType):
+            holder, maxlen = pack_args_by_32(holder, maxlen, arg, typ, context, placeholder)
+
+    # Dynamic position starts right after the static args.
+    holder.append(LLLnode.from_list(['mstore', dynamic_offset_counter, maxlen]))
+
+    # Calculate maximum dynamic offset placeholders, used for gas estimation.
+    for i, (arg, data) in enumerate(zip(args, expected_data)):
```
DavidKnott: `for i, (arg, data) in enumerate(zip(args, expected_data)):`is the same as line 800, can we calculate the dynamic offset placeholders in the loop above?

jacqueswww: Because of the way maxlen possibly gets altered when we pack lists, the dynamic offset (maxlen right after the first loop) is required to start calculating the dynamic section, also if we put it one for loop I'd also be scared of perhaps not understanding what we are doing later on ?

viper/parser/parser.py
jacqueswww: Will have a lookg thanks

viper/parser/parser.py
DavidKnott: I don't see this as a compiler bottleneck and combining loops would make it less readable; let's leave it as it.  Thanks for the detailed explanation.

jacqueswww:
@DavidKnott fixed that duplicate if statement. Thanks for the review :smile:

DavidKnott:
LGTM, merging.

Issue #361, opened 2017-09-26
Invalid cast results in runtime error rather than compile error/warning


ftramer:
* viper Version: 0.0.1
* pyethereum Version: 2.0.4
* OS: osx
* Python Version (python --version): 3.6.2
* Environment (output of `pip freeze`):
bumpversion==0.5.3
certifi==2017.7.27.1
chardet==3.0.4
coverage==4.4.1
cycler==0.10.0
cytoolz==0.8.2
ethereum==2.0.4
ethereum-abi-utils==0.4.0
ethereum-serpent==2.0.2
ethereum-utils==0.4.0
frozendict==1.2
idna==2.5
Jinja2==2.9.6
MarkupSafe==1.0
matplotlib==2.0.2
Naked==0.1.31
numpy==1.13.1
pbkdf2==1.3
py==1.4.34
py-ecc==1.1.3
pycryptodome==3.4.6
pyethash==0.1.27
pylru==1.0.9
pyparsing==2.2.0
pysha3==1.0.2
pytest==3.2.2
pytest-cov==2.5.1
pytest-runner==2.12.1
python-dateutil==2.6.1
pytz==2017.2
PyYAML==3.12
repoze.lru==0.7
requests==2.18.3
rlp==0.6.0
scipy==0.19.1
scrypt==0.8.0
shellescape==3.4.1
six==1.10.0
toolz==0.8.2
urllib3==1.22
viper==0.0.1
web3==3.13.5

### What's your issue about?

The following codes compile but throw exceptions when called:

```
def test1():
    y = 0x1000000000000000000000000000000000000000000000000000000000000000
    x = as_num128(y)
```

```
def test2():
  x = as_num128(msg.sender)
```

### How can it be fixed?
In the case of `test1` where the value of y is known at compile time, it seems that a compile error should be raised as the code will always fail. For `test2` the code could potentially run error free if `msg.sender` fits in a num128 at runtime. In both cases, I believe a compile-time warning would be useful to inform the programmer that the cast may fail at runtime.

![cat](https://user-images.githubusercontent.com/13002470/30878385-67f909da-a2b1-11e7-9c2d-ba8c674470f0.jpg)

DavidKnott:
@ftramer Thanks for reporting these errors, I'll make a PR to fix them asap.

fubuloubu:
If you make a test for this, change it to this instead:
```python
def test1():
    y = 2**127 # 1 over what int128 can support
    x = as_num128(y)
```

[...]

PR #643, opened 2018-01-23
Remove address as an input option from as_num128

jacqueswww:
### - What I did

As discussed in #361 .

### - How I did it

See functions.py

### - How to verify it
See added test.


fubuloubu:
There has to be a way to determine that first one as a compile-time error.

DavidKnott:
@fubuloubu There is if it's a literal, but for `memory` and `storage` values there's no way.  I think it's a much bigger change than this PR to have Vyper look for literals declared in memory and then use their values to move runtime errors to compile time.

fubuloubu:
Yup, agreed. That's what constant folding and dead code optimization can do, and that's definitely a larger issue

PR #645, opened 2018-01-24
Underdefined Paths in compiler, added Exceptions

fubuloubu:
### - What I did
Added some extra exceptions for paths in the compiler that would not produce a result when more types are added.

### - How I did it
Looked for underdefined paths in this specific function

### - How to verify it
No change to testing, would only appear when a new type is added. If not applied, then in the future when that type is added, compiler would give a very cryptic error that would be difficult to debug

Please feel free to change the exception types or messages as necessary

### - Cute Animal Picture
sleeping giraffe


DavidKnott:
LGTM

PR #646, opened 2018-01-24
WIP: Add full num256 support

fubuloubu:
### - What I did
Was working with a token contract that used full num256 values and noticed it would not compile. Started investigating, and there seems to be a very easy inroads to be made to adding it.

Not sure if we want to go down this path though, please comment.

### - Cute Animal Picture
sleeping whales

### DIFF
--- a/viper/parser/expr.py
+++ b/viper/parser/expr.py
@@ -228,6 +228,9 @@ class Expr(object):
             op = 'add' if isinstance(self.expr.op, ast.Add) else 'sub'
             if ltyp == rtyp:
                 o = LLLnode.from_list([op, left, right], typ=BaseType(ltyp, new_unit, new_positional), pos=getpos(self.expr))
+            elif ltyp == 'num256' and rtyp == 'num' or ltyp == 'num' and rtyp == 'num256':
+                o = LLLnode.from_list([op, left, right],
+                                      typ=BaseType('num256', new_unit, new_positional), pos=getpos(self.expr))
             elif ltyp == 'num' and rtyp == 'decimal':
                 o = LLLnode.from_list([op, ['mul', left, DECIMAL_DIVISOR], right],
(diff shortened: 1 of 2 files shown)


fubuloubu:
Note:
Fails one test that would get changed when support for modulo is added for `num256`.

Major additional testing required

viper/parser/expr.py
```diff
@@ -301,6 +306,9 @@ def arithmetic(self):
             return LLLnode.from_list(['clamp', ['mload', MemoryPositions.MINNUM], o, ['mload', MemoryPositions.MAXNUM]], typ=o.typ, pos=getpos(self.expr))
         elif o.typ.typ == 'decimal':
             return LLLnode.from_list(['clamp', ['mload', MemoryPositions.MINDECIMAL], o, ['mload', MemoryPositions.MAXDECIMAL]], typ=o.typ, pos=getpos(self.expr))
+        elif o.typ.typ == 'num256':
+            # No clamping required, takes up full register
```
jacqueswww: So this is where it gets funky with taking up the full 32byte word. We still actually have to protect against overflow hey, clamping makes that easy to fail with `num` type. With this scenario we'd have to find each operator '* ^ +' and make sure there is no overflow. Just something to keep in mind ;)

fubuloubu: Oh you mean like specialty code? (like safemath) Yeah that makes sense. Weren't we looking to implement safemath more generally for all types? I remember @DavidKnott talking about this.

fubuloubu: With that approach, much later down the line we can do a model checker on the code and see if it can violate the clamp before we add it as an optimization step, for things like tokens (unless it's mintable, and even then we only need to ensure the total supply never overflows) since most tokens can never overflow.

Way future thoughts though.

jacqueswww:
Closing for now, we can always re-asses this later.

Issue #637, opened 2018-01-19
Limit map key size 


yzhang90:
### What's your issue about?
1. In vyper compiler, the key_type of map can only be base type and byte array type, which is inconsistent with the doc(https://github.com/ethereum/vyper/blob/master/docs/types.rst) : "Here _KeyType can be almost any type except for mappings, a contract, or a struct".

2. Assume key type can only be base type and byte array type. I can have the following program:
```
a:num[bytes<=100000]
@public
def __init__():
    self.a["absssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssss"] = 1
    pass
```
will cause the compiler to throw an error "Can't have a zerovalent argument to an opcode or a pseudo-opcode! seq", which is weird.

### How can it be fixed?

Since the storage is limited, it's better to restrict the size of the key of the map.

#### Cute Animal Picture

![image](https://user-images.githubusercontent.com/5641590/35161380-a92d46f6-fd05-11e7-9148-846a24becb14.png)

jacqueswww:
@yzhang90 thanks for the report.
Having a look now.

jacqueswww:
@DavidKnott @fubuloubu @yzhang90 Ok I have investigated this, there was a fault in the way we did the sha3 accross the key bytearray. I have now fixed it that any size can work (any size larger than 32 would've  caused trouble).

Now the question:
Do we want to limit the key size?

DavidKnott:
I think we should let people use any sized key.  The downside is gas costs and people should be able to choose whether or not those are worth it themselves.  Can you guys think of situations where someone would need to use more than 32 bytes as a key?

fubuloubu:
Not offhand. Mapping structs maybe.

I was discussing with @jacqueswww in the Gitter, I think it we should not enable larger keys by default. The writer has the option of using it via `my_mapping[sha3(A, B, ...)]`, which would be accepting the costs of this decision explicitly. Personally, I also don't like extended key sizes as it it increases the vector for a key collision attack (finding another set of objects `A, B, ...` that overwrite the key in the hashmap), which is a possibility[1](https://eprint.iacr.org/2012/672.pdf), and the risk increases when data over the key size is used. This might be paranoia, because the attack is still very difficult, but it seems like a lot of concerns for what I currently believe has little utility in practice.

Anyways, we'd have to come up with a practical limit for the amount of data that can be used to generate a key in the hashmap. How would we choose that limit?

DavidKnott:
[...]

PR #647, opened 2018-01-25
Proper retrieval of map values from storage, with bytes key

### LINKED ISSUE #631: Allow data part of log opcode to be any length of bytes
### What's your issue about?

image

If `log4(p, s, t1, t2, t3, t4)` defines the  log opcode, memory from p to p + s represents the data section of the log[0-4] instruction. t1-t4 are calls for topics (or indexed).  Since the opcode support that to be any length. Currently the un-indexed "data" parameters ( p & s parameters) of a logging operation is limited to 32 bytes in vyper.

The following should work after this fix:

```python
MyLog: __log__({arg1: indexed(bytes <= 3), arg2: bytes <= 64})

@public
def foo():
    log.MyLog('bar', 'stuuuuff')
```
### How can it be fixed?

Add support for it in vyper. Also look at updating pyethereum's `abi.py`  to interpret `bytes[n]` notation on events.

jacqueswww:
### - What I did

Fix for #631.  Allow retrieving values from storage, with bytearrays as keys.

### - How I did it

Struggled a bit, but then managed to use the len part of the bytearray in memory to use the proper ranges to sha3 everytime.

### - How to verify it

See test.
```python
a:num[bytes<=100000]

@public
def __init__():
    self.a["aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"] = 1069
```


DavidKnott:
@jacqueswww I think you meant to reference issue  #637 ?

viper/parser/parser_utils.py
```diff
@@ -402,18 +402,19 @@ def add_variable_offset(parent, key):
...
+                    ['sha3', ['add', key.args[0].args[-1], 32], ['mload', key.args[0].args[-1]]]
                 ])
             else:
-                sub = LLLnode.from_list(['sha3', ['add', key.args[0].value, 32], ceil32(key.typ.maxlen)])
+                sub = LLLnode.from_list(['sha3', ['add', key.args[0].value, 32], ['mload', key.args[0].value]])
         else:
             subtype = typ.valuetype
             sub = base_type_conversion(key, key.typ, typ.keytype)
         if location == 'storage':
+            # import ipdb; ipdb.set_trace()
```
DavidKnott: Stray pdb 😄

tests/parser/features/test_bytes_map_keys.py
```diff
@@ -93,3 +93,33 @@ def set(k: bytes <= 35, v: num):
...
+def get_it3(key: bytes<=100000) -> num:
+    return self.a[key]
+    """
+
+    c = get_contract(code)
+
+    assert c.get_it2() == 1069
+    assert c.get_it2() == 1069
+    assert c.get_it3(b"a" * 33) == 1069
+    assert c.get_it3(b"test") == 0
```
DavidKnott: Can you add a test making sure it fails when declaration has `key: bytes  <= 50` and input is `inp: bytes <= 51`?

viper/parser/parser_utils.py
jacqueswww: ha, thanks!

tests/parser/features/test_bytes_map_keys.py
jacqueswww: This one already exists in line 85 ;)

jacqueswww:
@DavidKnott thanks for the review, I made the necessary changes.

jacqueswww:
@DavidKnott can finally be merged :smile:

PR #655, opened 2018-01-30
Remove unreachable line

fubuloubu:
### - What I did
Was looking at coverage and this line jumped out at me

### - How I did it
Removed it

### - How to verify it
Verify I didn't make a mistake


DavidKnott:
LGTM

Issue #516, opened 2017-11-27
Compiler allows a global variable name to be the same as a function name


yzhang90:
### What's your issue about?
The compiler should rejects the following program:
```
foo: decimal[3] 
@public def foo(): 
    self.foo = [1, 2.1, 3]
```
because the global variable and the function have the same name and it is very confusing to write such program.

### How can it be fixed?

Checking name duplication in parser.py

#### Cute Animal Picture

![image](https://user-images.githubusercontent.com/5641590/33246550-5bf5ccf4-d2db-11e7-9c3e-c7bde400face.png)

DavidKnott:
@yzhang90 On it 👍

DavidKnott:
@fubuloubu I'm working on this.  Do you think the following should be allowed?
```
def foo():
    foo: num
```
I'm leaning towards yes, as I'm thinking that uniforming naming can increase readability.

fubuloubu:
I'm not a big fan of that personally, but I think it would work because `foo()` is in the `self` "namespace" so that's fine

jakerockland:
@yzhang90 @DavidKnott was poking around at bugs that could be addresses and ended up here, but it looks like this issue can be closed with #659 merged?

jacqueswww:
@jakerockland that is correct ;) sorry for the later reply.

PR #659, opened 2018-02-02
Fix function name global variable shadowing

DavidKnott:
### - What I did
Make it illegal for function names to match global variable names.
Response to #516 

### - How I did it
Add a check when functions are being created again global variable names.

### - How to verify it
Look at the tests I changed.

Functions cannot have the same name as globals

### - Cute Animal Picture
image

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -48,6 +48,7 @@ Following the principles and goals, Vyper **does not** provide the following fea
 Compatibility-breaking Changelog
 ********************************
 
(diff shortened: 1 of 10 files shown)


fubuloubu:
LGTM.

Question about the struct: does Solidity do the same `struct__member()` getter syntax?

fubuloubu:
I think the answer is no, and now that we have tuples maybe we should return it that way?

jacqueswww:
LGTM :+1:

PR #660, opened 2018-02-05
Disallow calling non-constant function in a constant function

### LINKED ISSUE #638: constant function allows to call non-constant local function
### What's your issue about?

```
f:num

@public
def a (x:num)->num:
    self.f = 100
    return x+5

@constant
@public
def b():
    p = self.a(10)
```
This program should be rejected since `b` is a constant function. It should not call `a` which modifies the storage.

#### Cute Animal Picture

image

jacqueswww:
…ction.

### - What I did

Fixes #638. 

### - How I did it

Added checks in `expr.py` and `stmt.py`.

### - How to verify it
```python
f:num

@public
def a (x:num)->num:
    self.f = 100
    return x+5

@constant
@public
def b():
    p: num = self.a(10)
```
Should throw exception.

### DIFF
--- a/tests/parser/exceptions/test_constancy_exception.py
+++ b/tests/parser/exceptions/test_constancy_exception.py
@@ -52,6 +52,31 @@ def foo() -> num:
 @public
 def foo(x: num):
     x = 5
+    """,
+    """
+f:num
+
+@public
+def a (x:num)->num:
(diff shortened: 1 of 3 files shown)


DavidKnott:
LGTM

PR #662, opened 2018-02-07
Viper renaming, the end is in sight.

jacqueswww:
### - What I did

I picked a number of places we still use viper instead of vyper.
The vyper binary was a really obvious one we missed.

### - How I did it

ctrl+f

### - How to verify it

grep iper -R ./


fubuloubu:
Nice!

DavidKnott:
LGTM, merging!

fubuloubu:
It never ends! #664

PR #663, opened 2018-02-07
Test coverage improvements.

### LINKED ISSUE #654: Coverage Gaps
### What's your issue about?
We are at 93% coverage, so very close to pretty good on that. I've been looking at the gaps and identifying easy pickings to increase statement coverage.

### How can it be fixed?
Check these out:
* https://coveralls.io/builds/15294210/source?filename=viper%2Futils.py#L133
* https://coveralls.io/builds/15294210/source?filename=viper%2Ftypes.py (may need to skip exceptions)
* https://coveralls.io/builds/15294210/source?filename=viper%2Fparser%2Fparser_utils.py#L90
* https://coveralls.io/builds/15294210/source?filename=viper%2Fparser%2Fparser_utils.py#L327
* https://coveralls.io/builds/15294210/source?filename=viper%2Fcompiler.py
* https://coveralls.io/builds/15294210/source?filename=viper%2Fcompile_lll.py#L74
* https://coveralls.io/builds/15294210/source?filename=viper%2Fcompile_lll.py#L190
* https://coveralls.io/builds/15294210/source?filename=viper%2Ffunctions.py#L114
* https://coveralls.io/builds/15294210/source?filename=viper%2Ffunctions.py#L163
* https://coveralls.io/builds/15294210/source?filename=viper%2Ffunctions.py#L517

---

I skipped most exceptions because they require a little more introspection to understand if they are possible.

jacqueswww:
### - What I did

Partially fixes #654. Improves coverage to ~95%. 
I rather submitted PR now, as it was getting quite lengthy.

### - How I did it

Reading htmlcov and #654.

### - How to verify it

Pytest and check coverage report.


tests/parser/types/test_node_types.py
```diff
@@ -0,0 +1,90 @@
...
+    NullType,
+    StructType,
+    TupleType,
+    canonicalize_type,
+    get_size_of_type,
+)
+
+
+def test_null_type():
+    node1 = NullType()
```
fubuloubu: I'd prefer if this stuff was done in Vyper code instead of leveraging internal types

vyper/functions.py
```diff
@@ -156,15 +156,12 @@ def floor(expr, args, kwargs, context):
     )
 
 
-@signature(('num', 'decimal'))
```
fubuloubu: What was behind this change?

vyper/types.py
```diff
@@ -128,7 +128,7 @@ def __init__(self, members):
         self.members = copy.copy(members)
 
     def __eq__(self, other):
-        return other.__class__ == StructType and other.members == self.members
```
fubuloubu: What was behind this change?

vyper/types.py
```diff
@@ -180,22 +180,6 @@ def canonicalize_type(t, is_indexed=False):
     raise Exception("Invalid or unsupported type: " + repr(t))
 
 
-def parse_abi_type(t):
```
fubuloubu: Why remove this?

tests/parser/types/test_node_types.py
jacqueswww: Most of these scenarios test the Type classes directly - and couldn't be reached by writing vyper code. See this as a type of compiler-writer future proofing.

vyper/functions.py
jacqueswww: There is no reason to do decimal(decimal), especially because we are statically typing now.

vyper/types.py
jacqueswww: This was clearly a copy paste mistake, there is no way you'd want equalise tuple to struct, especially because structtype looks different internally ;)

vyper/types.py
jacqueswww: Doesn't get used anywhere, and would be super easy to add if we ever need it.

vyper/utils.py
```diff
@@ -138,8 +138,6 @@ def is_varname_valid(varname):
         return False
     if varname.lower() in reserved_words:
         return False
-    if varname[0] == '~':
```
DavidKnott: Why are you removing this check?

fubuloubu: Nice

fubuloubu: No need to test illegal syntax! Python core already does it for us!

fubuloubu: Wait, that `ast` error isn't quite what I expected. Can you potentially use it outside of instantiation (where type hinting doesn't come into play), e.g. in a function name or something?

DavidKnott:
@jacqueswww Quick rebase and then LGTM

jacqueswww:
@DavidKnott done :smile:

jacqueswww:
Closing because branch pt1 is included in pt2: #679

PR #664, opened 2018-02-09
More name change stuff

fubuloubu:
### - What I did
Finished grep foo'ing for `viper`

### - How I did it
I'll never tell!

### - How to verify it
More grep foo!

Also verify I didn't break `readthedocs` or test/build process somehow

### - Cute Animal Picture
ninja cat


fubuloubu:
Note:
1. https://viper.tools/ website still uses `viper`. Who owns the website?
2. `docs/old_readme.txt` still says `viper`. Do we need this file anymore?

jacqueswww:
Nice!

Issue #629, opened 2018-01-11
VIP: Standardize type conversions


DavidKnott:
## Preamble

    VIP: <to be assigned>
    Title: Standardize type conversions
    Author: David Knott
    Type: Standard Track
    Status: Draft
    Created: 2018-01-11

## Simple Summary
Right now Vyper type conversions are not uniform, they are performed by built in functions with different syntax:
`decimal` 
`bytes_to_num`
`as_num128` (int128
`as_num256` (uint256)

## Abstract
I propose standardizing how type conversions are performed in Viper.

## Motivation
Standardizing type conversion will simplify both the developer experience and the compiler by  decreasing the amount of build in functions.

## Specification
```
convert(input, new_typ)
decimal => convert(1, decimal) => 1.0
bytes_to_num =? convert('\x00', num) => 0
```

## Backwards Compatibility
This change won't be backwards compatible, if it's implemented it should be done in the next release.

## Copyright
Copyright and related rights waived via [CC0](https://creativecommons.org/publicdomain/zero/1.0/)

fubuloubu:
I like it v v much

fubuloubu:
@jacqueswww I think this one is complete?

jacqueswww:
Yes it is :) Marking it as done and closing the issue.

PR #668, opened 2018-02-09
Standardize type conversions

DavidKnott:
### - What I did
Standardize type conversions, implementing VIP #629 

### - How I did it
Broke out type conversion into its own file and combined the different ways to convert between types into one:
```
convert(input, 'output_type')
x: num
y: decimal = convert(x, 'decimal`)
```
### - How to verify it
Look at `test_convert.py`
Standardize type conversions.
### - Cute Animal Picture
image


fubuloubu:
Love it! Reviewing...

fubuloubu:
I remember discussing putting the type in quotes, but I just had a conflicting thought (perhaps for later on down the road). When syntax highlighting is developed for vyper, it would actually be more apparent to look at `y: num256 = convert(x, num256)` than to have `num256` in quotes, since there will be syntax highlighting for built-in types

DavidKnott:
Merging this.

vyper/functions/functions.py
```diff
@@ -156,47 +71,62 @@ def floor(expr, args, kwargs, context):
...
+    #             ['mload', ['add', args[0], 32]], typ=BaseType('bytes32')
+    #             )
+    #         elif args[0].location == "storage":
+    #             return LLLnode.from_list(
+    #                 ['sload', ['add', ['sha3_32', args[0]], 1]], typ=BaseType('bytes32')
+    #             )
+    #     else:
+    #         return LLLnode(value=args[0].value, args=args[0].args, typ=BaseType('bytes32'), pos=getpos(expr))
+    # else:
+    #     raise Exception("Conversion of {} to {} is not allowed.".format(typ, args[1]))
```
jacqueswww: @DavidKnott I think we should remove this comment block now?

Issue #542, opened 2017-12-04
VIP: Safe External Contract Calls


DavidKnott:
This is the suggested template for new VIPs.

## Preamble

    VIP: <to be assigned>
    Title: Safe External Contract Calls
    Author: David
    Type: Standard Track
    Status: Draft
    Created: 2017-12-4

## Simple Summary
We can use the `STATICCALL` opcode to make external contract calls safer.

## Abstract
External contract calls (primarily those that change state) are dangerous but we can make them safer by having developers think about how they external contract calls and wether or not they want to trust the contract they're calling to.

## Motivation
Many attacks have involved external contract calls so it's important for us to think carefully about how we can make them as safe as possible

## Specification
The `static` and `modifiable` keywords will allow developers to declare addresses that state changing external contract calls can always be made to.  If an address is `static` no state changes may be made (`STATICCALL` opcode is used), if an address is `modifiable` state changes may be made( `CALL` opcode is used).
```
    contract_1 = """
lucky: public(num)
@public
def set_lucky(_lucky: num):
    self.lucky = _lucky
"""

    contract_2 = """
class Bar():
    def set_lucky(_lucky: num): pass
modifiable_bar_contract: modifiable(Bar)
static_bar_contract: static(Bar)
@public
def __init__(contract_address: contract(Bar)):
    self.modifiable_bar_contract = contract_address
    self.static_bar_contract = contract_address

@public
def trusted_set_lucky(_lucky: num):
    self.modifiable_bar_contract.set_lucky(_lucky)

@public
def untrusted_set_lucky(_lucky: num):
    self.static_bar_contract.set_lucky(_lucky)
    """

    c1 = get_contract(contract_1)
    c2 = get_contract(contract_2, args=[c1.address])
    c2.trusted_set_lucky(7)
    assert c1.get_lucky() == 7
    # Fails after attempting a state change after a call to an untrusted address
    assert_tx_failed(lambda: c2.untrusted_set_lucky(5))
```

## Backwards Compatibility
This VIP is not backwards compatible as it will make multiple state changing external contract calls to untrusted addresses invalid.

## Copyright
Copyright and related rights waived via [CC0](https://creativecommons.org/publicdomain/zero/1.0/)

DavidKnott:
@fubuloubu Made the `static` and `modifiable` updates to the example.  Let me know if you've got any questions

fubuloubu:
[...]

PR #673, opened 2018-02-13
Add static and modifiable keywords

DavidKnott:
### - What I did
Add `STATICALL` opcode protections for external contract calls.
Implements VIP #542 
@jacqueswww 
### - How I did it
Added support for `modifiable` and `static` wrappers when declaring contract variables.
### - How to verify it
Look at `test_modifiable_external_contract_calls.py`

Add modifiable and static keywords for external contract calls.
### - Cute Animal Picture
image

### DIFF
--- a/tests/parser/features/external_contracts/test_external_contract_calls.py
+++ b/tests/parser/features/external_contracts/test_external_contract_calls.py
@@ -1,7 +1,7 @@
 from vyper.exceptions import StructureException, VariableDeclarationException, InvalidTypeException
 
 
-def test_external_contract_calls(get_contract_with_gas_estimation):
+def test_external_contract_calls(get_contract, get_contract_with_gas_estimation):
     contract_1 = """
 @public
 def foo(arg1: num) -> num:
@@ -18,13 +18,13 @@ class Foo():
 def bar(arg1: address, arg2: num) -> num:
     return Foo(arg1).foo(arg2)
     """
-    c2 = get_contract_with_gas_estimation(contract_2)
+    c2 = get_contract(contract_2)
 
     assert c2.bar(c.address, 1) == 1
     print('Successfully executed an external contract call')
 
 
-def test_complicated_external_contract_calls(get_contract_with_gas_estimation):
+def test_complicated_external_contract_calls(get_contract, get_contract_with_gas_estimation):
     contract_1 = """
 lucky: public(num)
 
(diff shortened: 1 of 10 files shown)


fubuloubu:
I think there is a bit of a misunderstanding here, I thought the static/mutable call was supposed to be defined for each *method* not the whole contract. The idea is to define which calls are constant e.g. static.

Check out this comment:
https://github.com/ethereum/vyper/issues/542#issuecomment-351950814

fubuloubu:
We should also perhaps consider using `constant` instead of `static`, to match up with the use of `@constant` as a decorator for static methods

DavidKnott:
@fubuloubu I chatted with @jacqueswww about this after our call today.  After thinking about it more marking a contract address as static or modifiable makes more sense to me then marking each function because as soon as someone makes a modifiable call to a contract it essentially makes the entire contract modifiable.

fubuloubu:
Okay, I guess that makes sense. Basically any function call has the ability to make state changes to itself as soon as you call it.

vyper/parser/context.py
```diff
@@ -0,0 +1,87 @@
+from vyper.utils import (
```
jacqueswww: Cool :smile_cat:

tests/parser/features/external_contracts/test_external_contract_calls.py
```diff
@@ -53,13 +53,13 @@ def array() -> bytes <= 3: pass
 def bar(arg1: address) -> num:
     return Foo(arg1).foo()
```
jacqueswww: For `Foo(arg1).foo()` type calls. Should we consider enforcing modifiable and static keywords perhaps?
Because it might seem inconsintent to the 'declared address on storage approach'.
`Foo(arg1, 'static').foo()` or `Foo(arg1, type='static').foo()` would be my syntax suggestions.

vyper/utils.py
```diff
@@ -119,6 +119,9 @@ class SizeLimits:
 # Valid base units
 valid_units = ['currency', 'wei', 'currency1', 'currency2', 'sec', 'm', 'kg']
 
+# Valid attributes for global variables
+valid_global_keywords = ['public', 'modifiable', 'static', 'address', 'num']
+
```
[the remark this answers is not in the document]
fubuloubu: Absolutely not allow those lol. They're allowed currently?

tests/parser/features/external_contracts/test_external_contract_calls.py
fubuloubu: Neither of those is great. What about `modifiable(Foo(arg1).Foo())`? This is a sort of secondary syntax (preferring to explicitly set), so I'm okay with it being uglier

jacqueswww:
@DavidKnott will you rebase and add those above tests? Otherwise just let me know then I will fork your branch and add them there :smile:

jacqueswww:
Updated branch in https://github.com/ethereum/vyper/pull/713

PR #677, opened 2018-02-13
Fix num division

DavidKnott:
### - What I did
Change division by `num` to output a `decimal` instead of a `num`. 
Add `ceil` built in function for rounding up.
Remove support for:
```
x: num = 7
x /= 5
```
### - How I did it
Changed `def arithmetic` in `expr.py` to return `decimal` instead of a `num` when dividing by a `num`.
### - How to verify it
Look at `test_ceil.py`, `test_num.py`, and all the other tests I changed.
Division by num returns decimal type.
### - Cute Animal Picture
image

### DIFF
--- a/examples/market_maker/on_chain_market_maker.v.py
+++ b/examples/market_maker/on_chain_market_maker.v.py
@@ -24,10 +24,10 @@ def initiate(token_addr: address, token_quantity: currency_value):
 @public
 @payable
 def eth_to_tokens():
-    fee: wei_value = msg.value / 500
+    fee: wei_value = floor(msg.value / 500)
     eth_in_purchase: wei_value = msg.value - fee
     new_total_eth: wei_value = self.total_eth_qty + eth_in_purchase
-    new_total_tokens: currency_value = self.invariant / new_total_eth
+    new_total_tokens: currency_value = floor(self.invariant / new_total_eth)
     self.token_address.transfer(msg.sender,
                                 convert(self.total_token_qty - new_total_tokens, 'num256'))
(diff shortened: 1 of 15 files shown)


examples/market_maker/on_chain_market_maker.v.py
```diff
@@ -24,10 +24,10 @@ def initiate(token_addr: address, token_quantity: currency_value):
 @public
 @payable
 def eth_to_tokens():
-    fee: wei_value = msg.value / 500
+    fee: wei_value = floor(msg.value / 500)
```
jacqueswww: @DavidKnott I think we chose the write option in specifying floor or ceil - much clearer which operation we are using now.

tests/parser/functions/test_ceil.py
```diff
@@ -0,0 +1,18 @@
...
+    return ceil(.0000000001)
+
+@public
+def foq() -> num:
+    return ceil(170141183460469231731687303715884105723.0000000001)
+"""
+    c = get_contract_with_gas_estimation(code)
+    assert c.foo() == 1
+    assert c.fop() == 1
+    assert c.foq() == 170141183460469231731687303715884105724
```
jacqueswww: @DavidKnott for consistency maybe add some floor tests ;)

fubuloubu:
LGTM basically what @jacqueswww said

tests/parser/functions/test_ceil.py
DavidKnott: Will do.

DavidKnott:
@jacqueswww Added tests for floor, merging

Issue #675, opened 2018-02-13
Payable should not be constant


fubuloubu:
### What's your issue about?
This compiles:
```python
@public
@constant
@payable
def get_real() -> decimal:
    return 0.0
```
I don't think `@payable` and `@constant` should be allowed to be together because changing the balance of ether of the smart contract is modifying the underlying state.

### How can it be fixed?
Add an exception if they're both present

#### Cute Animal Picture
![baby rhino](https://www.gannett-cdn.com/-mm-/2029718f4118781dc0c3283af0e2d030be823a9d/c=140-0-3371-2429&r=x404&c=534x401/local/-/media/2016/10/17/IAGroup/DesMoines/636123108878450705-RRW85964.JPG)

fubuloubu:
So we have something in the compiler that checks for state-changing opcodes in `@constant` functions?

maurelian:
That would be good. Ideally, constant would also use the `STATICCALL` opcode.

maurelian:
Oh, you're right. 🤦‍♂️

jacqueswww:
I agree we should probably prevent that ;)

jacqueswww:
Closing as PR is merged.

PR #678, opened 2018-02-14
Ban functions with payable and constant decorators

DavidKnott:
### - What I did
Ban functions with both `payable` and `constant` decorators because `constant` functions have not need to accept value.
In response to #675
### - How I did it
Add a check for `payable` and `constant` decorators to `function_signature.py`
### - How to verify it
Look at the test I added to `test_constant.py`
Ban functions with payable and constant decorators
### - Cute Animal Picture
image

### DIFF
--- a/tests/parser/features/decorators/test_constant.py
+++ b/tests/parser/features/decorators/test_constant.py
@@ -1,3 +1,6 @@
+from vyper.exceptions import StructureException
+
+
 def test_constant_test(get_contract_with_gas_estimation_for_constants):
     constant_test = """
 @public
@@ -10,3 +13,14 @@ def foo() -> int128:
     assert c.foo() == 5
 
     print("Passed constant function test")
+
+
+def test_invalid_constant_and_payable(get_contract_with_gas_estimation_for_constants, assert_compile_failed):
+    code = """
+@public
+@payable
+@constant
+def foo() -> num:
+    return 5
+"""
+    assert_compile_failed(lambda: get_contract_with_gas_estimation_for_constants(code), StructureException)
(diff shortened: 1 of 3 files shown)


jacqueswww:
LGTM, ship it! :ship:

vyper/function_signature.py
```diff
@@ -84,7 +84,9 @@ def from_definition(cls, code, _sigs=None):
             else:
                 raise StructureException("Bad decorator", dec)
         if public and private:
-            raise StructureException("Cannot use public and private decorators on the same function", code)
+            raise StructureException("Cannot use public and private decorators on the same function: {}".format(name))
+        if payable and const:
+            raise StructureException("Function `{}` cannot have a constant function that's payable.".format(name))
```
fubuloubu: "Function `{}` cannot have a constant function that's payable." should be "Function `{}` cannot be both constant and payable." for clarities' sake

tests/parser/features/decorators/test_constant.py
```diff
@@ -1,3 +1,6 @@
+from vyper.exceptions import StructureException
```
fubuloubu: Out of curiousity, why `StructureException`?

DavidKnott: `StructureException` is what we've been using as a catch all for exceptions involving anything to do with how the contract and functions are formed.

fubuloubu: Makes me think of the `struct` type. Maybe something like `UsageException` or `ContextException` might be more appropiate

DavidKnott: @fubuloubu I'm open changing `StructureException` to something clearer but if we do I think we should change it in another PR and all at once.

fubuloubu: :+1:

PR #679, opened 2018-02-14
More Test Coverage Improvements

jacqueswww:
### - What I did

Continuation of #654.

### - How I did it

By running a lot of pytest commands ;)

### - How to verify it

Just review the changes, and coverage report.

Test coverage improvements.


vyper/__init__.py
```diff
@@ -2,7 +2,7 @@
 
 from pkg_resources import get_distribution
 if (sys.version_info.major, sys.version_info.minor) < (3, 6):
-    raise Exception("Requires python3.6+")
+    raise Exception("Requires python3.6+")  # pragma: no cover
```
fubuloubu: Add a short comment "why"

vyper/compile_lll.py
```diff
@@ -63,10 +63,10 @@ def compile_to_assembly(code, withargs=None, break_dest=None, height=0):
         return ['DUP' + str(height - withargs[code.value])]
     # Setting variables connected to with statements
     elif code.value == "set":
-        if height - withargs[code.args[0].value] > 16:
-            raise Exception("With statement too deep")
         if len(code.args) != 2 or code.args[0].value not in withargs:
             raise Exception("Set expects two arguments, the first being a stack variable")
+        if height - withargs[code.args[0].value] > 16:
+            raise Exception("With statement too deep")
```
fubuloubu: What did shifting this statement do?

vyper/compile_lll.py
```diff
@@ -105,8 +105,6 @@ def compile_to_assembly(code, withargs=None, break_dest=None, height=0):
     elif code.value == 'repeat':
         o = []
         loops = num_to_bytearray(code.args[2].value)
-        if not loops:
-            raise Exception("Number of times repeated must be a constant nonzero positive integer: %r" % loops)
```
fubuloubu: You were able to find that there was no scenario where `loops` would not be set?

vyper/compile_lll.py
```diff
@@ -297,6 +295,8 @@ def assembly_to_evm(assembly):
                     o += codes[i]
                     break
         else:
-            raise Exception("Weird symbol in assembly: " + str(item))
+            # Should never reach because, assembly is create in compile_to_assembly.
+            raise Exception("Weird symbol in assembly: " + str(item)) # pragma: no cover
```
fubuloubu: Awesome!

vyper/compiler.py
```diff
@@ -20,8 +16,7 @@ def gas_estimate(origcode, *args, **kwargs):
     if code.value == 'seq':
         if code.args[-1].value == 'return':
             code = code.args[-1].args[1].args[0]
-    else:
-        code = code.args[1].args[0]
```
fubuloubu: No other LLL statement ends up here except `seq`?

vyper/compiler.py
```diff
@@ -36,7 +31,7 @@ def mk_full_signature(code, *args, **kwargs):
     for idx, func in enumerate(abi):
         func_name = func['name'].split('(')[0]
         # Skip __init__, has no estimate
-        if func_name in abi and func_name != '__init__':
+        if func_name in gas_estimates and func_name != '__init__':
```
fubuloubu: What was the benefit of checking in `gas_estimates` versus `abi`? Is `gas_estimates` a filtered version of `abi`?

vyper/compiler.py
```diff
@@ -51,4 +46,4 @@ def mk_full_signature(self, code, *args, **kwargs):
         return mk_full_signature(code, *args, **kwargs)
 
     def gas_estimate(self, code, *args, **kwargs):
```
fubuloubu: This method in the compiler class never gets tested? Why?

vyper/parser/stmt.py
```diff
@@ -114,7 +113,7 @@ def assign(self):
         sub = Expr(self.stmt.value, self.context).lll_node
         # Determine if it's an RLPList assignment.
         if isinstance(self.stmt.value, ast.Call) and getattr(self.stmt.value.func, 'id', '') is 'RLPList':
-            pos = self.context.new_variable(self.stmt.targets[0].id, set_default_units(sub.typ))
+            pos = self.context.new_variable(self.stmt.targets[0].id, sub.typ)
```
fubuloubu: Does `set_default_units` happen automagically now?

vyper/types.py
```diff
@@ -128,7 +128,7 @@ def __init__(self, members):
         self.members = copy.copy(members)
 
     def __eq__(self, other):
-        return other.__class__ == StructType and other.members == self.members
+        return other.__class__ == TupleType and other.members == self.members
```
fubuloubu: Was this a mistake in the code?

vyper/types.py
```diff
@@ -154,7 +154,7 @@ def canonicalize_type(t, is_indexed=False):
         return canonicalize_type(t.subtype) + "[%d]" % t.count
     if isinstance(t, TupleType):
         return "({})".format(
-            ",".join(canonicalize_type(x) for x in t.subtypes)
+            ",".join(canonicalize_type(x) for x in t.members)
```
fubuloubu: Was this a mistake in the code?

vyper/parser/stmt.py
fubuloubu: Or wait, the function doesn't exist anymore, right?

vyper/utils.py
```diff
@@ -138,8 +138,6 @@ def is_varname_valid(varname):
         return False
     if varname.lower() in reserved_words:
         return False
-    if varname[0] == '~':
-        return False
```
fubuloubu: This wasn't possible? @DavidKnott said something about this before

vyper/parser/stmt.py
jacqueswww: Correct that's the one I mentioned on #654 ;) where it all boils down to return typ

vyper/compile_lll.py
jacqueswww: If you check the one if statement checks len(code.args), where the other assumes it's already there. That's why I swapped the two around ;)

fubuloubu: Nice gotcha

vyper/compiler.py
jacqueswww: I believe it's used outside our scope, I believe as part of the necessary methods of pyethereum Compiler class.

vyper/types.py
jacqueswww: Yes, as we discussed on the pt1 PR ;) basically a mistake in the code (was copied from StructType definition).

vyper/types.py
jacqueswww: Yes same copy-paste-from-StructType problem.

vyper/utils.py
jacqueswww: Yes, also discussed in pt1. Python ast prevents this. With tokenizer PR we can also consider completely removing the '~' operator now as well :smile:

fubuloubu: What does that operator do? I've never used it

jacqueswww: It's a boolean operator for complement.  Never used before myself either:
```
~ x
Returns the complement of x - the number you get by switching each 1 for a 0 and each 0 for a 1. This is the same as -x - 1.
```

jacqueswww: Really interesting :thinking: 
```
>>> a = 3
>>> "{0:b}".format(a)
'11'
>>> "{0:b}".format(~a)
'-100'
>>> ~a
-4
```

vyper/compile_lll.py
jacqueswww: It's a duplicate exception, so never gets reached ;) in parser_utils.py line 121 the case is already cover.

vyper/compiler.py
jacqueswww: Correct, I think it was some legacy case that needed it. There was no scenario that I could find that needed it.

jacqueswww:
@fubuloubu @DavidKnott I think I have addressed all questions/changes now  :smile: It's probably ready to merge now :fireworks: :champagne:

vyper/compiler.py
fubuloubu: @jacqueswww just answer this one an it'll be LGTM

jacqueswww: @fubuloubu Ha, I remember now. If you look at abi it's a list of dictionaries, func_name is a string. So obviously the a string can never be in a list of dicts :smile_cat:  So made sense to check gas_estimates which is a dict of function_names to gas usage.
Crazy one, I know!

jacqueswww: Also made sense to check gas_estimates considering the statement under the if.

PR #683, opened 2018-02-17
Semicolon statement check

jacqueswww:
### - What I did
So I found out how super easy it is to use the stdlib tokenize library: https://docs.python.org/3/library/tokenize.html
Solves these two perfectly:

- Prevent multi-statements in a single line.
- Alias 'contract' to 'class' (we can obviously remove this, is here for demonstration purposes).

### - How I did it

Played with tokenize library.

### - How to verify it
Check tests.

This
```python
@public
def test() -> num:
    a: num = 1; b: num = 2
    return a + b
```
fails, and
```python
contract Bar():
    def bar() -> num: pass
``` 
is alias to 
```python
class Bar():
    def bar() -> num: pass
```


fubuloubu:
I don't not like it, but it definitely solves the problem and seems *somewhat* easy to use.
Also, if we want "contract" to be the preferred syntax, we should raise if "class" is used.

jacqueswww:
@fubuloubu well I like the fact that it uses the stdlib, that a massive plus for me, also the fact that tokenizer is already built for the python syntax :smile_cat: The problem with any other solutions is that one kind of has to step out of the Python.AST realm. The renaming of contract was just a demo, because I found that it was quite easy to do in the end ;)

jacqueswww:
@fubuloubu I see :)  I can change it that contract  _has_ to be used instead of class. Definitely well be clearer.

I don't think so. @DavidKnott  any you remember?

jacqueswww:
@DavidKnott @fubuloubu I now also realise we could be banning a long list of operators for example boolean shift operators '>>' etc.

fubuloubu:
I think that would be best done handling through parsing the AST e.g. "Unsupported operator ast.BoolShift"

jacqueswww:
@DavidKnott done.

PR #686, opened 2018-02-21
Prevent non-constant function call on assign

### LINKED ISSUE #671: Only allow one function call in assignment expressions
### What's your issue about?

An expression should only allow constant expression to be used.

```python
x: decimal

@constant
@private
def a () -> num:
    return 1

@private
def b () -> num:
    self.x = 333.33
    return 2

@public
def test():
    a: num = 3 * self.a()  # allowed
    b: num = self.a() or 4 * self.b()  # not allowed 
```

### How can it be fixed?

Let's discuss any downsides, and then implement it.

jacqueswww:
### - What I did

Fixes #671 Disallows  non-constant function in assignments.

### - How I did it

Added a context state `in_assignment`, that gets toggled on assignment, then throws necessary exceptino on call.

### - How to verify it
Check tests.


fubuloubu:
Does ERC20 now fail? Is there a way to bypass this?

jacqueswww:
@fubuloubu ERC20 is still fine, this only prevents a non-constant call on assignment.

jacqueswww:
@fubuloubu  @DavidKnott Thinking of this use case some more, I think I might have considered this incorrectly. We should prevent usage of non-constant functions in expressions like
`a = self.a() or self.b()`
And still allow assignment of results of non function calls?

fubuloubu:
Yeah, your example (depending on short circuiting is done) potentially does not execute the given statement `self.b()`. I think that is the idea behind the issue, and clearly it's a bit non-obvious if allowed.

Doing `assert erc20.transfer(address, 20)` isn't as big a deal, because the statement has a clear execution flow.

jacqueswww:
@fubuloubu updated, as we discussed :)

tests/parser/exceptions/test_constancy_exception.py
```diff
@@ -70,7 +70,7 @@ def b():
 f:int128
 
 @public
-def a(x:num):
+def a (x:int128):
```
fubuloubu: Why is there a space between `a` and the parens?

jacqueswww: must've been a typo that slipped in, wil fix.

DavidKnott:
LGTM, merging

PR #690, opened 2018-03-02
Some more Viper -> Vyper renamings

holgerd77:


fubuloubu:
Thanks! LGTM

It never ends....

Issue #579, opened 2017-12-12
Line numbers on readthedocs don't match up with the code as on Github


jamesray1:
![screenshot from 2017-12-12 15-14-51](https://user-images.githubusercontent.com/16969914/33867218-382d76f2-df4f-11e7-8307-e79d6a8edbfd.png)

![screenshot from 2017-12-12 15-18-23](https://user-images.githubusercontent.com/16969914/33867276-b866305c-df4f-11e7-8740-243060f02905.png)

#### Cute Animal Picture

![screenshot from 2017-12-12 15-16-31](https://user-images.githubusercontent.com/16969914/33867239-742c8792-df4f-11e7-9c2b-fbd59baee847.png)

blaisep:
@jamesray1 I think you might be looking at a different commit from the one used on readthedocs.
readthedocs used commit `9be6289d733e940c6552cb1352d316a479a392f0`
and in _that_commit `def finalize():` appears on line 28.
The code in `docs/viper-by-example.rst` specifies:
```
.. literalinclude:: ../examples/crowdfund.v.py
  :language: python
  :lines: 24-27
```
So a quick fix would be to update  `docs/viper-by-example.rst` to say:
```
.. literalinclude:: ../examples/crowdfund.v.py
  :language: python
  :lines: 28-31
```

fubuloubu:
@blaisep can you explain "include text in docstrings" for the uninitiated? (e.g. myself lol)

blaisep:
@fubuloubu For example, if we look at the rst source for:
http://docs.python-requests.org/en/master/api/#requests.request
in the file:
https://github.com/requests/requests/blob/master/docs/api.rst , notice that the section describing the `requests.request` function is not filled out in detail. All that appears in the rST source is: 
```
.. autofunction:: request
```

So the description of `requests.request` appears in https://github.com/requests/requests/blob/master/requests/api.py#L17 . Also, the HTML output makes use of the Sphinx `viewcode` extension : http://www.sphinx-doc.org/en/stable/ext/viewcode.html 
So if you're reading the doc and you want to read the python code, click on the link called _source_ and it displays the relevant section of the `.py` file. Sphinx resolves the line numbers and HTML refs when the docs get built.

sguberman:
I think there are two issues here:

- The original issue raised by @jamesray1
    - When viewed on readthedocs, the line *spacing* does not match between the line numbers on the left and the lines of code on the right.
- Another issue found by @blaisep
[...]

PR #692, opened 2018-03-07
change lines included in code examples to match discussion text

sguberman:
### - What I did
The example code sections did not match the discussion text. I made them match. Relates to #579 

### - How I did it
Rather than including specific line numbers, where possible I used the `pyobject` option to match the function being discussed. This way, if the source code file is updated, the docs will follow automatically once rebuilt. I learned about this by reading the Sphinx [docs](http://www.sphinx-doc.org/en/master/markup/code.html#includes) related to showing code examples.

### - How to verify it
Read the docs! Specifically, any of the Vyper By Example sample code in line with the explanatory paragraphs.

Fix example code shown shown in line with Vyper By Example docs.

### - Cute Animal Picture
A very cute albino Tanuki!
image

### DIFF
--- a/docs/vyper-by-example.rst
+++ b/docs/vyper-by-example.rst
@@ -55,7 +55,7 @@ Now, the constructor.
 
 .. literalinclude:: ../examples/auctions/simple_open_auction.v.py
   :language: python
-  :lines: 16-22
+  :pyobject: __init__
 
 The contract is initialized with two arguments: ``_beneficiary`` of type
 ``address`` and ``bidding_time`` with type ``timedelta``, the time difference
@@ -72,7 +72,7 @@ With initial setup out of the way, lets look at how our users can make bids.
 
 .. literalinclude:: ../examples/auctions/simple_open_auction.v.py
   :language: python
-  :lines: 24-38
+  :pyobject: bid
 
 The ``@payable`` decorator will allow a user to send some ether to the
 contract in order to call the decorated method. In this case, a user wanting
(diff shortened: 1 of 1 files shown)


fubuloubu:
Fantastic! Do you know if there is a way to view this on read the docs, or does it need to be done locally to verify?

sguberman:
Good point @fubuloubu that did not occur to me at all, and I should have explained in my "how to verify it" section that I built locally to confirm the correct changes to the documentation. However, your question gave me an idea. I set up my own account at readthedocs and connected it to my fork and this branch, so you should be able to view the changes yourself [here](https://sguberman-vyper.readthedocs.io/en/latest/vyper-by-example.html).

fubuloubu:
Awesome! Looks great!

Is it a bug that so much extra space appears at the end of the full file inclusions? That bothers me lol

fubuloubu:
I remember seeing it before, so I don't think you caused it.

sguberman:
Yeah that bothers me too, and it's what tipped me off to the line spacing being different between the line numbers and the code. I didn't try to fix that at all this time, just wanted to make sure the correct shorter code sections were being shown with the text. A quick fix would be to just turn off line numbering for those full file inclusions, but I thought it would be better to separate the two concerns into different PRs.

You shouldn't be seeing any blank space after the shorter inclusions because I left the line numbers off (as was done in the original).

fubuloubu:
LGTM!

PR #694, opened 2018-03-08
Fix for slice.

### LINKED ISSUE #693: Fix slice function
### What's your issue about?
Found a case where slice function breaks.

```python
def test_conv2(get_contract):
	code = """
@public
def ret10_slice() -> bytes <= 10:
	b: bytes <= 32 = concat(convert(65, 'bytes32'), '')
	`# return b
	c: bytes <= 10 = slice(b, start=31, len=1)
	return c
	"""

	c = get_contract(code)
	assert c.ret10_slice() == b'A'
```

jacqueswww:
### - What I did

Fixes #693.

### - How to verify 

Just check tests.


tests/parser/functions/test_slice.py
```diff
@@ -83,12 +83,12 @@ def foo(inp: bytes <= 10, start: int128, len: int128) -> bytes <= 10:
...
-    assert c.foo(b"badminton", 0, 9) == b"badminton"
-    assert c.foo(b"badminton", 1, 8) == b"adminton"
-    assert c.foo(b"badminton", 1, 7) == b"adminto"
-    assert c.foo(b"badminton", 1, 0) == b""
-    assert c.foo(b"badminton", 9, 0) == b""
+    # assert c.foo(b"badminton", 3, 3) == b"min"
+    # assert c.foo(b"badminton", 0, 9) == b"badminton"
+    # assert c.foo(b"badminton", 1, 8) == b"adminton"
+    # assert c.foo(b"badminton", 1, 7) == b"adminto"
+    # assert c.foo(b"badminton", 1, 0) == b""
```
fubuloubu: Why are these commented out?

jacqueswww: Ha, thanks! That's not supposed to be there :see_no_evil:

PR #696, opened 2018-03-09
Add a test for parsing a null byte

dshulyak:
### - What I did
Trying to convert last byte of the signature into an integer

### - How I did it
slice start with 64 with length of 1, and then convert to int128

### DIFF
--- a/tests/parser/functions/test_convert.py
+++ b/tests/parser/functions/test_convert.py
@@ -1,4 +1,4 @@
-def test_convert_to_num(chain, get_contract_with_gas_estimation, assert_tx_failed):
+def test_convert_to_num(chain, get_contract_with_gas_estimation, assert_tx_failed, bytes_helper):
     code = """
 a: int128
 b: uint256
@@ -33,13 +33,18 @@ def bytes_to_num() -> (int128, int128):
     literal: int128 = convert('a', 'int128')
     storage: int128 = convert(self.d, 'int128')
     return literal, storage
+
+@public
+def zero_bytes(inp: bytes <= 1) -> int128:
+    return convert(inp, 'int128')
 """
     c = get_contract_with_gas_estimation(code)
     assert c.int128_to_num(1) == [1, 1, 1]
     assert c.uint256_to_num(1) == [1, 1]
     assert c.bytes32_to_num() == [1, 1]
     assert c.bytes_to_num() == [97, 97]
-
+    assert c.zero_bytes(b'\x01') == 1
+    assert c.zero_bytes(b'\x00') == 0
 
 def test_convert_to_uint256(t, chain, get_contract, utils, assert_tx_failed):
     code = """


dshulyak:
@jacqueswww @DavidKnott could you guys take a look at this problem? I am sorry if it is not a bug in vyper. perhaps you can suggest a workaround then?

fubuloubu:
You're trying to convert a two byte zero (sliced from the end of a longer bytestring) into a `int128`?

dshulyak:
@fubuloubu why two bytes? I have 65 bytes ecdsa signature, if last byte is zero contract fails. I nailed it down to the converter.

```python
slice(zero, start=64, len=1)
```
returns '\x00' correctly, but converter fails

The limit of 66 in the signature is not necessary, I used this https://github.com/ethereum/vyper/issues/495 as an example.
I also tried bytes_to_num from older vyper, but it fails exactly the same way

dshulyak:
I updated tests, they are now more explicit I guess. Input bytes can be anything, the only byte that matters is a last byte

dshulyak:
I found a crazy workaround for what I need to achieve

```python
extract32(concat(slice(concat(convert(0, 'bytes32'), ''), start=0, len=31), inp), 0, type=uint256)
```

fubuloubu:
What the heck, `convert` works with no type?? Definitely have ourselves a few bugs here.

Next time would prefer this as an issue instead of a breaking PR. Sometimes the problems aren't what you think, it's better to just note the issue and work on the exact problem causing it. It might even be something downstream of what you'd expect.

dshulyak:
i am somewhat confused by your statements
> What the heck, convert works with no type??

it is passed as the second argument, what do you mean?
> Next time would prefer this as an issue instead of a breaking PR

i can follow a workflow of course, but how it would be better? the test does exactly what i do in the contract and confirms that it actually fails on ci. 

i expect that this
```python
convert('\x00', 'int128') 
```
will return 0
the same way as this
```python
convert('\x01', 'int128') 
```
will return 1

am i wrong and it is not expected?

fubuloubu:
> it is passed as the second argument, what do you mean?

Oh sorry, I read that statement wrong. You did `concat` with a blank second argument, I thought the second argument was for `convert`

 > the test does exactly what i do in the contract and confirms that it actually fails on ci.

Yes but your actual example isn't as simple. I'm thinking perhaps `slice` is the problem, not `convert`. It could be `convert` as you say, it could be that the combination doesn't play nicely, who knows! The point is until you find the part of the compiler that breaks, and can prove a fix for it, you don't really know if it's a true fix or that you even have the problem nailed down.

PRs should be actionable improvements that suggest fixes to issues (test cases alonh with accompanying fixes), instead of breaking changes (that add new tests). We definitely appreciate the additional test for this issue, but we can't merge this PR until an accompanying fix is discovered, which means we need to juggle 2 PRs for a single issue which is a pain.

fubuloubu:
Looks like you changed the test case I was referring to, so it looks like you are correct that `convert` is the problem. I still think we shouldn't introduce new test cases until we know why it's failing and how to fix it.

jacqueswww:
@dshulyak @fubuloubu I will not be able to look at this until the week after next. However I do believe it has todo with the fact that we can't convert arbitrary length byte arrays e.g." convert('\x00', 'int128')" but only 32 byte types.

a: byte32 = "\x01"
Convert(a, 'int128') 

Might do what you had in mind depending on the string packing.
Keep in mind that a single byte string to int128 conversion like above can be very ambiguous. Because to the reader can't be sure if last or first byte is being set.

dshulyak:
>  However I do believe it has todo with the fact that we can't convert arbitrary length byte arrays e.g." convert('\x00', 'int128')" but only 32 byte types.

this is not the case because `convert('\x01', 'int128')` works as expected and returns 1. i highlighted it in the test. problem appears only when a single byte is`\x00`. 

```
a: byte32 = "\x01"
Convert(a, 'int128')
```
This actually fails with `AttributeError: 'ByteArrayType' object has no attribute 'typ'` 

> Keep in mind that a single byte string to int128 conversion like above can be very ambiguous. Because to the reader can't be sure if last or first byte is being set.

i am not sure that we are speaking about same thing, please take a look at the test. it is exactly one byte (0 or 1). what can be ambiguous here?

dshulyak:
I made some tests and it is obvious that converter expects bytes in big endian (there is nothing ambiguous it simply expects bytes in one order always). But it fails every time i insert leading null bytes, lets say i will pass 100321 in big endian , the representation will be `\x01\x87\xe1`, converter will parse it correctly. However, if I will add null bytes e.g `\x00\x01\x87\xe1` converter will fail. You can try any other language, it will parse these bytes with or without null bytes correctly.

jacqueswww:
Created issue https://github.com/ethereum/vyper/issues/705 for this.

jacqueswww:
Closing as this works now.

PR #702, opened 2018-03-15
 Fix floor and ceil for negative inputs

michaelsproul:
### - What I did

Corrected the implementations of `floor` and `ceil` so that they behave correctly for negative inputs.

### - How I did it

Modified both `floor` and `ceil` to conditionally branch, and use an appropriate signed division to get the right rounding behaviour.

### - How to verify it

Read the code and run the included tests.

Make floor and ceil round towards negative and positive infinity respectively, rather than 0

### - Cute Animal Picture

https://www.flickr.com/photos/ekilby/4844114344

### DIFF
--- a/vyper/optimizer.py
+++ b/vyper/optimizer.py
@@ -11,7 +11,7 @@ def get_int_at(args, pos, signed=False):
         o = LOADED_LIMIT_MAP[args[pos].args[0].value]
     else:
         return None
-    if signed:
+    if signed or o < 0:
         return ((o + 2**255) % 2**256) - 2**255
     else:
         return o % 2**256
--- a/vyper/parser/parser_utils.py
+++ b/vyper/parser/parser_utils.py
@@ -231,6 +231,10 @@ def get_number_as_fraction(expr, context):
         t += 1
     top = int(context_slice[:t].replace('.', ''))
     bottom = 1 if '.' not in context_slice[:t] else 10**(t - context_slice[:t].index('.') - 1)
+
+    if expr.n < 0:
+        top *= -1
+
     return context_slice[:t], top, bottom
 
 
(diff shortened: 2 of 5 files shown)


vyper/optimizer.py
```diff
@@ -11,7 +11,7 @@ def get_int_at(args, pos, signed=False):
         o = LOADED_LIMIT_MAP[args[pos].args[0].value]
     else:
         return None
-    if signed:
+    if signed or o < 0:
```
michaelsproul: This is the hackiest line in this PR. I had to do this to prevent `-1` getting turned into `2**256 - 1 = 115792089237316195423570985008687907853269984665640564039457584007913129639935`, which would then get added to `999999999` and overflow when optimising `ceil(-0.0000000001)`.

The other solution I considered was making the additions (and other operations) carried out by the optimiser modulo `2**256` like their equivalent EVM instructions, but I figured this was sub-optimal as it silently hides the overflow, rather than preventing it.

vyper/parser/parser_utils.py
```diff
@@ -231,6 +231,10 @@ def get_number_as_fraction(expr, context):
         t += 1
     top = int(context_slice[:t].replace('.', ''))
     bottom = 1 if '.' not in context_slice[:t] else 10**(t - context_slice[:t].index('.') - 1)
+
+    if expr.n < 0:
+        top *= -1
```
michaelsproul: This makes `x: decimal = -0.05` parse as -0.05 rather than _positive_ 0.05

michaelsproul:
Thanks, there's no rush :)

michaelsproul:
Added some new tests, let me know if they need further refinement

Issue #708, opened 2018-03-21
ERC20 decimals is not public


BenKaufman10:
### What's your issue about?
Is there any reason in the ERC20 standard in Vyper the ``decimals`` is not public?
Like here: https://github.com/ethereum/vyper/blob/master/examples/tokens/ERC20.v.py
In line 17 ``decimals: int128`` is not public.
Also decimals is not accessible from the premade ERC20 contract.
However, in the vypercoin implementation the decimals is public.
If there is no reason for that I will add it to the code and submit a PR.

### How can it be fixed?

``decimals: public(int128)``
And adding it to the premade contract as well.

#### Cute Animal Picture
![image](https://user-images.githubusercontent.com/10667901/37700397-09d31d6e-2cf4-11e8-8da5-8aa233bbbf49.png)

jacqueswww:
Agreed, these should be added as they seem to be part of the standard.

PR #709, opened 2018-03-21
Added getters for name and decimal ERC20

BenKaufman10:
PR for issue #708 
Added name and getters for ERC20 name and decimals.
Made ERC20 decimals public

### - Cute Animal Picture
image

### DIFF
--- a/vyper/premade_contracts.py
+++ b/vyper/premade_contracts.py
@@ -2,7 +2,9 @@ import ast
 
 erc20 = """
 class ERC20():
+    def name() -> bytes32: pass
     def symbol() -> bytes32: pass
+    def decimals() -> uint256: pass
     def balanceOf(_owner: address) -> uint256: pass
     def totalSupply() -> uint256: pass
     def transfer(_to: address, _amount: uint256) -> bool: pass
--- a/examples/tokens/ERC20.v.py
+++ b/examples/tokens/ERC20.v.py
@@ -14,7 +14,7 @@ Approval: event({_owner: indexed(address), _spender: indexed(address), _value: u
 name: public(bytes32)
 symbol: public(bytes32)
 totalSupply: public(uint256)
-decimals: int128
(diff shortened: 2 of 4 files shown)


vyper/premade_contracts.py
```diff
@@ -2,7 +2,9 @@
 
 erc20 = """
 class ERC20():
+    def name() -> bytes32: pass
     def symbol() -> bytes32: pass
+    def decimals() -> uint256: pass
```
jacqueswww: @BenKaufman10 please add tests fore the extra two functions.

jacqueswww: I think you can just add some asserts in test_initiat in test_on_chain_market_maker.py

fubuloubu:
I think it's import to note that several of these methods are optional per the spec, although I don't think that matters currently because there should be no harm in adding additional functions to a contract interface as nothing is enforced until the call is made. If the call fails, it's because the contract does not implement the full ERC20 interface.

jacqueswww:
@fubuloubu yes exactly, IMHO the ERC20 shortcut interface should have all the optional functions exposed - for those that want to use it.

vyper/premade_contracts.py
BenKaufman10: Alright, I added it there.

BenKaufman10:
@jacqueswww I tried to fix it, you can see in the last commit. For some reason it looks like it didn't check the lates commit (6e17e598906c1f5195c46b753dffe25d75d3035f) which was supposed to fix it.
Can you please check it out and let me know if I missed something?

jacqueswww:
Try reverting your last commit and committing over it, this will trigger the CI again. I don't have permission to retrigger this.

BenKaufman10:
Turns out it was a bug with Travis-ci, just closed the PR and re-opened it and everything worked well.

PR #710, opened 2018-03-21
Fix assignment checks

### LINKED ISSUE #703: List assignment test with mismatched number of elements incorrectly passes
As the title says.

https://github.com/ethereum/vyper/blob/3eb961ecccb8ced76f19217843bb9ee88f6f8750/tests/parser/features/iteration/test_for_in_list.py#L8

jacqueswww:
### - What I did
Fixes #703 .

-  Structs and lists where not properly type-checked on assignment.

### - How I did it
Turns out the variable location the was allocated always assumed the the type of the right hand side. Instead of using the information from the stmt.annotation. So simple fix, added necessary fail tests so we can pick this up again.

### - How to verify it
Check tests.

N/A

### - Cute Animal Picture

094f3793044be99fbe20c769e36a23de

### DIFF
--- a/tests/parser/features/iteration/test_for_in_list.py
+++ b/tests/parser/features/iteration/test_for_in_list.py
@@ -5,7 +5,7 @@ def test_basic_for_in_list(get_contract_with_gas_estimation):
     code = """
 @public
 def data() -> int128:
-    s: int128[5] = [1, 2, 3, 4, 5, 6]
+    s: int128[5] = [1, 2, 3, 4, 5]
     for i in s:
         if i >= 3:
             return i
--- a/tests/parser/syntax/test_ann_assign.py
+++ b/tests/parser/syntax/test_ann_assign.py
@@ -47,7 +47,24 @@ def do_stuff() -> bool:
 @public
 def test():
     a: bool = False or self.do_stuff()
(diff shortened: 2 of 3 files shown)


fubuloubu:
There's a lot of "technical debt" like this in our code base, i.e. it is very hard to find errors between all the `typ` variables. We may want to refactor this at some point so things are clearer

jacqueswww:
Good point, `typ`-debt. We probably just need to choose better names for each `typ` instance right?

fubuloubu:
Well, that's one option. I think now that we have a good structure of what Vyper is, we can refactor in order to make it as easy as possible to contribute to the compiler. That's the whole reason we choose to do it in Python to begin with.

I was working on a bit of a refactor to streamline this approach into successive layers, basically each layer only having one complex activity e.g. dealing with `ast` names, creating a Vyper ast, converting to LLL IR, optimizing, dealing with LLL IR to bytecode conversion, etc. Check it out here: https://github.com/fubuloubu/blocktract. This would make a more structured environment to work in.

Once I figured that out, I wanted to work on type composition and some of the more advanced Vyper v2.0 updates we've been discussing, either as an update or as a compiler extension for more advanced users.

PR #718, opened 2018-03-23
Adds support deleting for nested map keys.

### LINKED ISSUE #715: Fix deleting a value from a map.
### What's your issue about?

It seems that `del` keyword doesn't work for a map.
```
E           vyper.exceptions.StructureException: line 14: Unsupported statement type
E               del self.big_storage[key]
E           ----^
```

Example test case:
```
def test_map_delete(get_contract):
    code = """
big_storage: bytes32[bytes32]

@public
def set(key: bytes32, value: bytes32):
    self.big_storage[key] = value

@public
def get(key: bytes32) -> bytes32:
    return self.big_storage[key]

@public
def delete(key: bytes32):
    del self.big_storage[key]
    """

    c = get_contract(code)

    assert c.get("test") == b""
    assert c.set("test", "value")
    assert c.get("test") == b"value"
    c.delete("test")
    assert c.get("test") == b""
```

This will be done

### How can it be fixed?

Fixed by using `SSTORE 0x0`on the sha32'ed key value.

jacqueswww:
### - What I did

Fixes #715.

### - How I did it

Added parse_delete to handle the specific case.

### - How to verify it

Check tests `del self.big_map["test"]` should now work.

Add deleting support for a map.

### DIFF
--- a/tests/parser/exceptions/test_structure_exception.py
+++ b/tests/parser/exceptions/test_structure_exception.py
@@ -165,6 +165,18 @@ def foo() -> int128:
 q:int128 = 111
 def foo() -> int128:
     return self.q
+    """,
+    """
+b: bytes32[int128]
+@public
+def foo():
+    del self.b[0], self.b[1]
+    """,
+    """
+@public
+def foo():
+    b: int128
+    del b
     """
 ]
 
--- a/tests/parser/features/test_map_delete.py
+++ b/tests/parser/features/test_map_delete.py
@@ -1,0 +1,52 @@
+
+
+def test_map_delete(get_contract_with_gas_estimation):
+    code = """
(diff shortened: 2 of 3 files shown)


fubuloubu:
Can you `del` a struct?

jacqueswww:
It can, but it seems to do it incorrectly. Will have a look.

Thrasymachus01:
Any update on this? I am getting an error when trying to use this keyword:
`vyper.exceptions.SyntaxException: Deleting is not supported`

fubuloubu:
@Thrasymachus01 this was removed in v0.2.0

To "delete" a value in a mapping, use `self.mapping[key] = empty(ValueType)` or similar

PR #729, opened 2018-03-29
2 tests exposing an issue: Events with list not 1st arg initialized from a variable doesn't work.

denis-bogdanas:
### - What I did
Added 2 tests for event logging.


jacqueswww:
@denis-bogdanas putting this in an issue as this doesn't provide a solution.

jacqueswww:
This has been fixed  ;)

PR #740, opened 2018-03-30
vyper-run

jacqueswww:
### - What I did
Adds a `vyper-run` command.

### - How I did it
Just added simple binary.

### - How to verify it
Basically allows you to very quickly prototype a contract/ call /syntax.

Given an example contract:
```python
testmap: {
    a: int128,
    b: int128
}[int128]

@public
def __init__(a: int128, b: int128):
    self.testmap[1] = {
        a: a,
        b: b
    }


@public
def test123() -> (int128, int128):
    self.testmap[1] = {
        a: 123,
        b: 456
    }

    return self.testmap[1].a, self.testmap[1].b


@public
def get() -> (int128, int128):
    return self.testmap[1].a, self.testmap[1].b
```

One can super quickly just run the methods:
```
$ vyper-run 733.v.py "get();test123()" -i 1,3
Initializing chain from provided state
Calling get()
[1, 3]

Calling test123()
[123, 456]
```


BenKaufman10:
This looks like a really great feature!
One improvement I think will be very good is to make it also print out events (or make a custom printing mechanism for that but I think it would be unnecessary).
IMO this can be super handy and make it more usable, what do you think?

jacqueswww:
I agree that would be an excellent addition.

PR #741, opened 2018-03-31
fixed #735, removed pyethereum dependency

### LINKED ISSUE #735: Remove dependency on pyethereum
### What's your issue about?

This is a blocker for pyethereum Casper branch.

Vyper depends on [pyethereum 2.1.3](https://github.com/ethereum/vyper/blob/master/setup.py#L26), this introduces errors to the installation.

```
error: ethereum 2.1.0 is installed but ethereum==2.1.3 is required by {'vyper'}
```

### How can it be fixed?

Move pyethereum to dev_requirements.txt.

The pyethereum is mainly imported in two places of the codebase. Mostly it is imported in `scripts/`, removing it should not affect the functionality of the package. It is also imported in [parser_utils](https://github.com/ethereum/vyper/blob/master/vyper/parser/parser_utils.py#L3), which should be replaceable with constants in `vyper/opcodes`

ChihChengLiang:
### - What I did

To address #735, 

- Removed pyethereum installation dependency
- Moved `bumpversion`, `pytest-cov`, and `pytest-runner` out of `install_requires`.

### - How I did it

Copy pasting from [py-evm constants](https://github.com/ethereum/py-evm/blob/master/evm/constants.py#L83-L84)

### - How to verify it

check if new constants in `vyper/opcodes.py` fit [py-evm constants](https://github.com/ethereum/py-evm/blob/master/evm/constants.py#L83-L84) and  [pyethereum opcodes](https://github.com/ethereum/pyethereum/blob/develop/ethereum/opcodes.py)

### - Cute Animal Picture

image

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,0 +1,1 @@
+bumpversion
--- a/setup.py
+++ b/setup.py
@@ -22,12 +22,8 @@ setup(
     url='https://github.com/ethereum/vyper',
     license=license,
     packages=find_packages(exclude=('tests', 'docs')),
-    install_requires=[
-        'ethereum==2.1.3',
-        'bumpversion',
-        'pytest-cov',
-        'pytest-runner',  # Must be after pytest-cov or it will not work
-                          # due to https://github.com/pypa/setuptools/issues/196
-    ],
+    install_requires=[],
+    setup_requires=['pytest-runner'],
+    tests_require=['pytest', 'pytest-cov', 'ethereum==2.1.3'],
     scripts=['bin/vyper', 'bin/vyper-serve']
 )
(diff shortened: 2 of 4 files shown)


fubuloubu:
I think it would be advisable to actually import the constants from py-evm directly (instead of copying)

ChihChengLiang:
I recommend maintaining constants here. For two rationales,
1. This makes Vyper as lightweight as possible. User won't need to install relatively heavy `py-evm` package (and its dependencies) when using Vyper.
2. Importing py-evm here will cause circular dependency issue when the implementation of Casper using Vyper in py-evm.

ChihChengLiang:
OK, constants are imported from `py-evm` now as advised. Also upgraded `pyethereum` from 2.1.3 to 2.3.1 in `test_require` to avoid version conflicts between `py-evm`(requires pycryptodome>=3.4.7) and `pyethereum` (requires pycryptodome==3.4.6).

jacqueswww:
LGTM, tested it in new virtualenv and seems to do the trick.

Issue #752, opened 2018-04-03
VIP: Allow declaring custom units


ben-kaufman:
## Preamble

    VIP: 752
    Title: Allow declaring custom units
    Author: Ben Kaufman ben.kaufman10@gmail.com
    Type: Standard Track
    Status: Draft
    Created: 2018-04-03

## Simple Summary
Add a new keyword ``unit`` to allow declaring custom units like so:

## Abstract
Adding a way to create custom units can make the code more readable and improve security (type-safety).

## Motivation
The motivation for this came as I saw that the Casper contract uses labels incorrectly. they use the ``m`` which stands for meter where they clearly didn't meant to represent meters. This make the code very confusing and might be a bigger issue in the future.
Currently, there are a few units Vyper support, but there are much more measurements Vyper doesn't take in count. For example, in the company.v.py the stocks are labeled as ``currency`` while it actually should be a ``stock``.

## Specification
The proposed ``unit`` declaration is as below:
```
stock: unit
stocks_count: int128(stock)
wei_per_stock: int128(wei / stock)
```
However, it may be declared in a different way such as with a dedicated function.

## Backwards Compatibility
This change is backward compatible

## Copyright
Copyright and related rights waived via [CC0](https://creativecommons.org/publicdomain/zero/1.0/)

#### Cute Animal Picture

![image](https://user-images.githubusercontent.com/10667901/38243786-4ce270da-3741-11e8-9ac9-e00a1c8d90b7.png)

jacqueswww:
I think this a decent idea. Perhaps we should consider sometype of predefined unit table ? 
Something like (don't worry about the syntax for now, but just the idea):
```python
units = (
    ('m', 'meter'),
    ('cm', 'centimeter')
)
```
This would force a description on each unit used in the contract.

BenKaufman10:
Sounds good, I think it will be better. I will update the VIP in accordance with a syntax for that. I will also add it to the list of subjects for the next meeting.

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

fubuloubu:
I wamted the unit labels to be double quoted so it's obvious that those are labels and not types or variables

fubuloubu:
Example:
```python
stocks_count: int128("stock")
wei_per_stock: int128("wei" / "stock")
```

I don't think we need an extra syntax to declare a unit, if you made a spelling mistake it will tell you when you compile, because that's the whole point of units.

fubuloubu:
[...]

PR #778, opened 2018-04-16
752 custom units

jacqueswww:
### - What I did
Implements #752, custom units.

### - How I did it
Added support for units as a custom global keyword.
```python
# specify units used in the contract.
units: {
    cm: "centimeter",
    km: "kilometer"
}

# global storage
a: int128(cm)


@public
def test() -> int128(km):
    b: int128(km)
    b = 100
    return b

```
### - How to verify it
See tests and exceptions in parser.py


fubuloubu:
Later on down the line, I wonder if we could define conversion tables for units like this, where a numeric conversion exists between different units (e.g. ether)

PR #791, opened 2018-04-23
Real abi type

axic:
### - What I did

Removes the obsolete `real` and `real128x128` ABI types (they have been replaced with `fixedMxN`). 

It also removes them from being restricted keywords, though I am not sure if that is desired.

### - Cute Animal Picture


fubuloubu:
I wish we still did `get_[var]` for the getter so we could generate a `get_real()` method

get real

PR #793, opened 2018-04-26
Improved storage variable layout

### LINKED ISSUE #769: VIP: Improve storage variable layout
### Preamble

```
VIP: #769
Title: Improve storage variable layout
Author: Daejun Park
Type: Standard Track 
Status: Draft
Created: 2018-04-09
```

### Simple Summary

Improve the layout of dynamically-sized state variables in the storage, adopting [that of Solidity](http://solidity.readthedocs.io/en/v0.4.21/miscellaneous.html#layout-of-state-variables-in-storage).

### Specification

A map entry `m[k]` is stored at the location `hash(k . index(m))`, where "`.`" is byte-concatenation.

A nested map entry's location can be defined recursively. For example, `m[k1][k2]` is stored at `hash(k2 . hash(k1 . index(m)))`

A formal specification can be found [here](https://github.com/kframework/evm-semantics/blob/master/edsl.md#hashed-location-for-storage) (in progress).

----

Below is a quick reference of `loc` that computes the storage location.

Lists:
```
loc(a[i])    =   #(a) + i
loc(a[i][j]) = #(#(a) + i) + j
```

Structs:
```
loc(s.x)   =   #(s) + x
loc(s.x.y) = #(#(s) + x) + y
```

Maps:
```
loc(m[k])    =   #(m . k)
loc(m[k][l]) = #(#(m . k) . l)
```

In general: (`E` is an arbitrary nested data structure, e.g., a list of structs of maps ...):
```
loc(E[i]) = #(loc(E)) + i    // list of ...
loc(E.x)  = #(loc(E)) + x    // struct of ...
loc(E[k]) = #(loc(E) . k)    // map of ...

loc(a) = a                   // list name
loc(s) = s                   // struct name
loc(m) = m                   // map name
```

NOTE:
 * `#` denotes the `keccak256` hash.
 * `.` is byte-concatenation.
 * Assume there is the implicit `index` conversion for the list name `a`, the struct name `s`, the struct field names `x` and `y`, and the map name `m`.

----

Rationale:

The number of elements (or fields) of a list (or a struct) is expected to be much smaller than the size of the storage address space. Thus, even if the elements (or the fields) are stored in a consecutive region of the storage, it is very unlikely the region is overlapped with the other regions, since the starting locations of the regions will be well distributed (well spread across the entire storage) by `keccak256`.

### Motivation

The main difference from the current scheme is the use of `.` instead of `+` to compute the offset. This change is critical for avoiding potential collisions, since `.` is injective, while `+` is not due to the modulo arithmetic. (Note that this is orthogonal to the potential hash collision of `keccak256`.)

jacqueswww:
### - What I did
Fixes #769


vyper/compile_lll.py
```diff
@@ -226,6 +226,20 @@ def compile_to_assembly(code, withargs=None, break_dest=None, height=0):
...
+            'PUSH1', MemoryPositions.FREE_VAR_SPACE,
+            'MSTORE'
+        ])
+        o.extend(compile_to_assembly(code.args[1], withargs, break_dest, height))
+        o.extend([
+            'PUSH1', MemoryPositions.FREE_VAR_SPACE2,
+            'MSTORE',
+            'PUSH1', 64, 'PUSH1', MemoryPositions.FREE_VAR_SPACE, 'SHA3'
+        ])
+        return o
```
jacqueswww: @daejunpark would be great if you could double check this for me :smile:

vyper/utils.py
```diff
@@ -72,15 +72,16 @@ def calc_mem_gas(memsize):
...
     MINNUM = 96
     MAXDECIMAL = 128
     MINDECIMAL = 160
     FREE_VAR_SPACE = 192
-    BLANK_SPACE = 224
-    FREE_LOOP_INDEX = 256
+    FREE_VAR_SPACE2 = 224
+    BLANK_SPACE = 256
+    FREE_LOOP_INDEX = 288
+    RESERVED_MEMORY = 320
```
daejunpark: @jacqueswww Is it intended to have the same size of the reserved memory even if we have the additional one (`FREE_VAR_SPACE2`)?

vyper/compile_lll.py
daejunpark: `sha3_64` looks good to me.

daejunpark:
@jacqueswww I'd like to take a closer look at `parser_utils.py` sometimes later this week (I'm traveling now), if it is not urgent to merge.

In the meantime, @yzhang90 could you take a look at this as well?

vyper/utils.py
jacqueswww: @daejunpark there seemed to have been an extra slot btween 288 and 320 that was never used. So I took the liberty of just shifting the addresses up.

vyper/compile_lll.py
jacqueswww: Great thanks ;)

vyper/parser/parser_utils.py
```diff
@@ -415,8 +416,24 @@ def add_variable_offset(parent, key, pos):
         else:
             subtype = typ.valuetype
             sub = base_type_conversion(key, key.typ, typ.keytype, pos=pos)
+
+        if location == 'storage':
+            return LLLnode.from_list(['sha3_64', parent, sub],
+                                     typ=subtype,
+                                     location='storage')
+        elif location == 'storage_prehashed':
```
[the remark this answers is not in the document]
yzhang90: Please confirm that `storage_prehashed` is only used in `List`, `Struct` and `Tuple`. You don't need to consider it when the type is `Map`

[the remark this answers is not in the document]
jacqueswww: Yes that is correct, prehashed should only occur with `!= Map` I will remove this section ;)

Issue #764, opened 2018-04-06
VIP: uint256 as a first class citizen


jacqueswww:
## Preamble

    VIP: #764
    Title: Uint256 as a first class citizen.
    Author: Jacques Wagener
    Type: Standard Track 
    Status: Draft
    Created: 2018-04-06
    Requires (*optional): <VIP number(s)>
    Replaces (*optional): <VIP number(s)>

## Simple Summary
Allow generic arithmetic operations to work instead of using specific functions.

## Abstract
Currently using uint256 can be very tedious and difficult because all arithmetic has to occur with 

## Motivation

Initially it was thought that working with unsigned integer are needed and the very large signed integer types could be used and converted. However time, and time again the need for `uint256` to exist as arisen. Using `uint256_*` functions do do any arbitrary calculation becomes a serious issue for readability.

## Specification

| function | operator|
| ---------- | ----------- |
| bitwise_and | remains |
| bitwise_or | remains |
| bitwise_xor | remains |
| bitwise_not | remains |
| uint256_add | + |
| uint256_sub | - |
| uint256_mul | * |
| uint256_div | / |
| uint256_exp | ** |
| uint256_mod | % |
| uint256_addmod | remains |
| uint256_mulmod | remains |
| uint256_gt | > |
| uint256_ge | >= |
| uint256_lt | < |
| uint256_le | <= |
| shift | remains |

*NB* comparison between uint256 and other types should not be implicit. 

It is important  double check SafeMath is inserted, however working of the functions templates should insert the existing SafeMath behaviour, as seen in:
https://github.com/ethereum/vyper/blob/master/vyper/functions/functions.py#L559

Literal integer assignment cases to handled on assignment (as discussed below):

```
Case 1. L < -2**127. Reject (preferably compile-time).
Case 2. -2**127 <= L < 0. L must be of int128. No conversion.
Case 3. 0 <= L < 2**127. The type of L will be identified based on its context.
Case 4. 2**127 <= L < 2**256. L must be of uint256. No conversion.
Case 5. L >= 2**256. Reject (preferably compile-time).

Now, in Case 3, the type of the literal can be identified as follows.

First, we reject an integer expression (arithmetic, comparison, etc.) whose sub-expressions are all integer literals. (This restriction should be acceptable by users, as @fubuloubu pointed out.)
```

## Backwards Compatibility

We can consider keeping `uint256_` functions. But I think removing those functions is best to avoid confusion for new comers.

## Copyright
[...]

PR #797, opened 2018-04-30
uint256 improvements

jacqueswww:
### - What I did

Implements #764. uint256 improvements.


vyper/utils.py
```diff
@@ -90,6 +90,7 @@ class SizeLimits:
     MINNUM = -2**127
     MAXDECIMAL = (2**127 - 1) * DECIMAL_DIVISOR
     MINDECIMAL = (-2**127) * DECIMAL_DIVISOR
+    MAX_UINT256 = 2**256
```
fubuloubu: Max is `2**256-1`

PR #814, opened 2018-05-10
cast `topics` to tuple

ldct:
### - What I did

In a new version of rlp these `topics` becomes a tuple. This is a backwards-compatible change to support that.

### - How to verify it

Unit tests

### - Cute Animal Picture

image


jacqueswww:
Merging, even though eth-tester probably override this.

PR #815, opened 2018-05-10
Use eth-tester.

pipermerriam:
Let's do this!


tests/conftest.py
```diff
@@ -1,16 +1,114 @@
...
+        value
+        for value
+        in self.get_all()
+        if get_key(value) not in values_to_remove
+    ]
+    for value in queued_values:
+        if get_key(value) in values_to_remove:
+            continue
+        self.queue.put_nowait(value)
+eth_tester.utils.filters.Filter.remove = Filter_remove
```
pipermerriam: This looks like a good case for using the [`monkeypatch`](https://docs.pytest.org/en/latest/monkeypatch.html) fixture that pytest provides assuming you are not aware of it.

tests/conftest.py
```diff
@@ -62,108 +160,104 @@ def bytes_helper(str, length):
...
     return get_contract_with_gas_estimation_for_constants
 
 
 @pytest.fixture
-def get_contract(chain):
-    def get_contract(source_code, *args, **kwargs):
-        return chain.contract(source_code, language="vyper", *args, **kwargs)
-    return get_contract
+def assert_tx_failed(tester):
+    def assert_tx_failed(function_to_test, exception=TransactionFailed):
```
pipermerriam: These would be pretty easy to implement using `contextlib.contextmanager` if you wanted to *niceness* of `pytest.raises(...)`

```python
@contextlib.contextmanager
def tx_fails(exception=...):
    try:
        yield
    except exception:
        pass
    else:
        raise AssertionError("Did not raise")
```

tests/conftest.py
```diff
@@ -62,108 +160,104 @@ def bytes_helper(str, length):
...
-                (not event_name or
-                 contract.translator.event_data[log.topics[0]]['name'] == event_name)]
-        assert len(logs) > 0, "No logs in last receipt"
-
-        # Return all events decoded in the receipt
-        return [contract.translator.decode_event(log.topics, log.data) for log in logs]
+def get_logs(w3):
+    def get_logs(tx_hash, c, event_name):
+        tx_receipt = w3.eth.getTransactionReceipt(tx_hash)
+        logs = getattr(c._classic_contract.events, event_name)().processReceipt(tx_receipt)
```
pipermerriam: latest web3 now supports `c._classic_contract.events[event_name]`

tests/examples/auctions/test_simple_open_auction.py
```diff
@@ -1,77 +1,76 @@
...
 
 
-def test_initial_state(auction_tester):
+def test_initial_state(w3, tester, auction_contract):
     # Check beneficiary is correct
-    assert utils.remove_0x_head(auction_tester.c.beneficiary()) == auction_tester.accounts[0].hex()
+    assert auction_contract.beneficiary() == w3.eth.accounts[0]
     # Check bidding time is 5 days
-    assert auction_tester.c.auction_end() == auction_tester.s.head_state.timestamp + 432000
+    assert auction_contract.auction_end() == tester.backend.chain.header.timestamp + EXPIRY - 1
```
pipermerriam: *if* you want to be future proof, I'd recommend changing this to not reach into the backend and to instead use` tester.get_block_by_number['latest']['timestamp']`.  In fact, now that I've seen this, I'm inclined to rename the location of backend to `tester._backend` so that it's clear that it's a private API.

tests/conftest.py
```diff
@@ -62,108 +160,104 @@ def bytes_helper(str, length):
...
+        lll = optimizer.optimize(LLLnode.from_list(lll))
+        bytecode = compile_lll.assembly_to_evm(compile_lll.compile_to_assembly(lll))
+        abi = []
+        contract = w3.eth.contract(bytecode=bytecode, abi=abi)
+        deploy_transaction = {
+            'data': contract._encode_constructor_data(args, kwargs)
+        }
+        tx = w3.eth.sendTransaction(deploy_transaction)
+        address = w3.eth.getTransactionReceipt(tx)['contractAddress']
+        contract = w3.eth.contract(address, abi=abi, bytecode=bytecode, ContractFactoryClass=ConciseContract)
```
pipermerriam: I see you are using `ConciseContract` copiously.  My first inclination was to push you away from it, but I think that's mostly my personal preference speaking.

Was the choice to use this happenstance?  Intentional over the more explicit call semantics for the default `Contract` class?

tests/examples/safe_remote_purchase/test_safe_remote_purchase.py
```diff
@@ -7,86 +7,93 @@
...
-contract_code = open("examples/safe_remote_purchase/safe_remote_purchase.v.py").read()
+
 # Inital balance of accounts
-INIT_BAL = 1000000000000000000000000
+INIT_BAL_a0 = 1000000000000000000000000
+INIT_BAL_a1 = 1000000000000000000000000
+
 
+def ETH(x):
+    return x * 10**18
```
pipermerriam: In case you aren't aware, there are nice APIs for this in both `eth-utils` and web3

- `eth_utils.to_wei` and `eth_utils.from_wei`
- `w3.toWei` and `w3.fromWei`

tests/examples/tokens/test_vypercoin.py
```diff
@@ -1,150 +1,127 @@
...
+            f.read(),
+            *[TOKEN_NAME, TOKEN_SYMBOL, TOKEN_DECIMALS, TOKEN_INITIAL_SUPPLY]
+        )
 
 
 def pad_bytes32(instr):
     """ Pad a string \x00 bytes to return correct bytes32 representation. """
-    bstr = instr.encode()
-    return bstr + (32 - len(bstr)) * b'\x00'
+    return instr + (32 - len(instr)) * b'\x00'
```
pipermerriam: *nitpick* The built-in `v.ljust(32, b'\x00')` is both faster and more concise.

tests/examples/wallet/test_wallet.py
```diff
@@ -1,30 +1,34 @@
...
-
-print([utils.encode_hex(a) for a in [t.a1, t.a2, t.a3, t.a4, t.a5]])
-
+@pytest.fixture
+def c(w3, tester, get_contract):
+    a0, a1, a2, a3, a4, a5, a6 = w3.eth.accounts[:7]
+    with open('examples/wallet/wallet.v.py') as f:
+        code = f.read()
+    # Sends wei to the contract for future transactions gas costs
+    import ipdb; ipdb.set_trace()
```
pipermerriam: I used to use `ipdb` but I've recently switched to `pdbpp` which allows you to just use the built-in `pdb` but to get a nice repl.  It also works with `pytest --pdb` natively.

tests/parser/features/test_assert.py
```diff
@@ -1,17 +1,17 @@
-import pytest
 
 
-def test_assert_refund(t, get_contract_with_gas_estimation):
+def test_assert_refund(w3, tester, get_contract_with_gas_estimation, assert_tx_failed):
```
pipermerriam: Curious of the choice to use both `w3` and `tester` here.  `w3.eth.getBalance(...)` should allow you to avoid direct `tester` interactions which is mostly just academic, but it does reduce how closely your API is tied to the `eth-tester` API (and the `w3` API is likely to be more stable).

tests/parser/features/test_gas.py
```diff
@@ -11,8 +11,9 @@ def foo() -> int128:
 
     c = get_contract_with_gas_estimation(gas_call)
 
-    assert c.foo(startgas=50000) < 50000
-    assert c.foo(startgas=50000) > 25000
+    assert c.foo(call={"gas": 50000}) < 50000
+    assert c.foo(call={"gas": 50000}) > 25000
+
     print('Passed gas test')
```
pipermerriam: total nitpick on my part, but converting these `print` statements to use `logging` allows you to get the same output in your terminal while still allowing that output to be easily silenced, or even tuned to only show a subset by logger or by log level.

tests/parser/functions/test_block_number.py
```diff
@@ -1,6 +1,6 @@
 
-def test_block_number(get_contract_with_gas_estimation, chain):
-    chain.mine(1)
+def test_block_number(get_contract_with_gas_estimation, tester):
+    tester.mine_blocks(1)
```
pipermerriam: Just in case you aren't aware, `w3.testing.mine(n)` can be used if you want to stick with web3 APIs

tests/parser/functions/test_sha3.py
```diff
@@ -11,9 +11,9 @@ def bar() -> bytes32:
 
     c = get_contract_with_gas_estimation(hash_code)
     for inp in (b"", b"cow", b"s" * 31, b"\xff" * 32, b"\n" * 33, b"g" * 64, b"h" * 65):
-        assert c.foo(inp) == utils.sha3(inp)
+        assert '0x' + c.foo(inp).hex() == sha3(inp).hex()
 
-    assert c.bar() == utils.sha3("inp")
+    assert '0x' + c.bar().hex() == sha3(b"inp").hex()
```
pipermerriam: Reduces readability on the assert statement but you could drop the `hex` overhead and do the direct assertion.

tests/conftest.py
```diff
@@ -1,16 +1,114 @@
...
+
+@pytest.fixture(scope="module")
+def w3(tester):
+    w3 = Web3(EthereumTesterProvider(tester))
+    w3.eth.setGasPriceStrategy(zero_gas_price_strategy)
+    return w3
+
+
+@pytest.fixture
+def sha3():
```
pipermerriam: I'd recommend renaming this to `keccak`.  (which makes me realize we should deprecate `Web3.sha3` and rename it to `keccak` as well.

tests/examples/safe_remote_purchase/test_safe_remote_purchase.py
jacqueswww: Ah yes, I remember now! :smile:

tests/parser/functions/test_block_number.py
jacqueswww: neat, yes I think having tests use `web3` as much as possible is the way to go.

tests/parser/features/test_assert.py
jacqueswww: I agree, for tests direct _tester_ acces should be minimised, and w3 optimised will change accordingly ;)

tests/examples/wallet/test_wallet.py
```diff
@@ -1,73 +1,106 @@
...
 
     # Turns the raw sigs into sigs
-    sigs = [(utils.big_endian_to_int(x[64:]), utils.big_endian_to_int(x[:32]), utils.big_endian_to_int(x[32:64])) for x in
-            map(lambda z: utils.decode_hex(z[2:]), raw_sigs)]
+    sigs = [
+        (
+            w3.toInt(x[64:]),  # v
+            w3.toInt(x[:32]),  # r
+            w3.toInt(x[32:64])  # s
+        ) for x in map(lambda z: w3.toBytes(hexstr=z[2:]), raw_sigs)
```
pipermerriam: The `[2:]` is no longer necessary here.  Same with all other calls with a `hexstr=thing[2:]`

tests/examples/wallet/test_wallet.py
```diff
@@ -1,73 +1,106 @@
...
+
+    h = w3.sha3((0).to_bytes(32, "big") + b'\x00' * 12 + w3.toBytes(hexstr=recipient[2:]) + (25).to_bytes(32, "big") + b'')
+    h2 = w3.sha3(b"\x19Ethereum Signed Message:\n32" + h)
 
-    h = utils.sha3(utils.encode_int32(0) + b'\x00' * 12 + utils.decode_hex(recipient[2:]) + utils.encode_int32(25) + b'')
-    h2 = utils.sha3(b"\x19Ethereum Signed Message:\n32" + h)
     # Check to make sure the signatures are valid
-    assert '0x' + utils.encode_hex(utils.sha3(utils.ecrecover_to_pub(h2, sigs[0][0], sigs[0][1], sigs[0][2]))[12:]) == accounts[0]
-    assert '0x' + utils.encode_hex(utils.sha3(utils.ecrecover_to_pub(h2, sigs[1][0], sigs[1][1], sigs[1][2]))[12:]) == accounts[1]
+    assert Account.recoverHash(h2, sigs[0]).lower() == accounts[0]
```
pipermerriam: If you want, you can use `eth_utils.is_same_address` which handles equality for addresses across checksum, normal hex, and bytes representations.

tests/examples/wallet/test_wallet.py
```diff
@@ -1,73 +1,106 @@
 import pytest
 
+from eth_keys import KeyAPI
```
pipermerriam: You can also use `from eth_keys import keys` and remove anywhere that you are doing `keys = KeysAPI()`

fubuloubu:
This PR proves the need for an opinionated testing library for python, something you can just import and know that there is always one preferred way of doing something for testing.

fubuloubu:
Woo!

PR #820, opened 2018-05-15
Modified the requirement of py-evm

hwwhww:
### - What I did
`sharding` repo has [a dependency hell of eth-tester/pyetherem/py-evm/vyper](https://github.com/ethereum/sharding/issues/86#issuecomment-388955286
). Since the only thing that Vyper "package" needs from `py-evm` is importing constants (`from evm.constants import GAS_IDENTITY, GAS_IDENTITYWORD`), the easy way to solve the problem is making Vyper package doesn't require a fixed `py-evm` version and only use fixed version for testing.

### - How I did it
Modified the requirement of `py-evm` from
```python
install_requires=['py-evm>=0.2.0a12]
```
to
```python
install_requires=['py-evm>=0.2.0a12,<1.0.0']
...
tests_require=[
    ...
    'py-evm==0.2.0a12',
],
```

### - How to verify it
CI passed.

Modified the requirement of py-evm

### - Cute Animal Picture
bat

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -22,8 +22,17 @@ setup(
     url='https://github.com/ethereum/vyper',
     license=license,
     packages=find_packages(exclude=('tests', 'docs')),
-    install_requires=['py-evm==0.2.0a16'],
+    install_requires=['py-evm>=0.2.0a16,<1.0.0'],
     setup_requires=['pytest-runner'],
-    tests_require=['pytest', 'pytest-cov', 'eth-tester==0.1.0b24'],
-    scripts=['bin/vyper', 'bin/vyper-serve', 'bin/vyper-run']
+    tests_require=[
+        'pytest',
+        'pytest-cov',
+        'eth-tester==0.1.0b24',
+        'py-evm==0.2.0a16',
+    ],
+    scripts=[
+        'bin/vyper',
+        'bin/vyper-serve',
+        'bin/vyper-run',
+    ]
 )


jacqueswww:
@hwwhww I think we might consider having to do multiple runs with the CI.
The CI obviously install py-evm. And that's why this wasn't pointed out:

https://github.com/ethereum/vyper/blob/0d92d86752bcfca875e042cec6048488db3d479c/vyper/parser/parser_utils.py#L3

carver:
I reserved a name in a few places, just in case we decide it's a good idea :)
http://eth-constants.readthedocs.io/en/latest/releases.html#v0-1-0-alpha-1
https://pypi.org/project/eth-constants/

fubuloubu:
@carver where'd you get that git repo template from? that's super useful!

hwwhww:
Sorry for the late reply!

@jacqueswww 
> I think we might consider having to do multiple runs with the CI.

I did some tests with `python setup.py install` and `python setup.py test` and checked the `py-evm` version. I can also test it by actually using this commit in `sharding` repo.

#### Copied constants or `eth-constants` packages
I considered simply copying constants in https://github.com/ethereum/sharding/issues/86#issuecomment-388955286 - option 1, which is the easiest way IMO.
`eth-constants ` idea sounds more structured, but after that, everytime we want to add more constants in `py-evm`, we need a new `eth-constants` release cut. Would that be more overhead for @carver? ;)

carver:
Absolutely, maintaining eth-constants separately would carry extra overhead. If there's a simpler solution, or we don't mind including all of py-evm into Vyper, then eth-constants is unnecessary, and I'm happy with that. :)

If we do want to use eth-constants. One approach would be a *lazy* inclusion of constants. So all constants would start simply in py-evm, and as soon as other libraries want it, we can port them over. Many of them would probably never need to be moved. It looks like Vyper is using a tiny subset of the constants.

BTW, I haven't actually started any work on eth-constants. Unless someone speaks up to say it would help them right now, I probably won't get to it until after the py-evm public alpha release and aftermath. So probably 1 month or more.

PR #826, opened 2018-05-18
Added support for uint256 custom units and removed currency1,2 

ben-kaufman:
### - What I did
Added support for uint256 custom units.
Removed currency1_value and currency2_value, will remove currency and replace it with custom units where needed in the next commit

Added support for uint256 custom units.
Removed currency1_value and currency2_value.

### - Cute Animal Picture

image


ben-kaufman:
@jacqueswww is this PR ok? It now has a conflict which I believe the version here is the correct one. Please let me know if you need me to change something here.

jacqueswww:
As part of the three VIPs, I am doing now I think this one might not be necessary anymore.
Let me finish the VIPs and then I wilk come and review ;)

ben-kaufman:
Closed in favour of #840

PR #828, opened 2018-05-19
#825 - removed uint256 arithmetic functions and fixed boolean return …

### LINKED ISSUE #825: Solidity Compatible ERC20 broken
The [Solidity compatible ERC20 token](https://github.com/ethereum/vyper/blob/master/examples/tokens/ERC20_solidity_compatible/ERC20.v.py) no longer compiles, since it was not updated after the removal of separate uint256 math functions. This is a super easy fix. I can do it later in the week if no one gets to it before then.

cmgannon26:
### - What I did
Fixes #825 

Removed deprecated uint256 arithmetic functions in favor of regular arithmetic operators. Also fixed casing in bool return values ("true" is now "True") in withdraw(), transfer(), transferFrom(), and approve().

### - Cute Animal Picture

whdq-512188356

### DIFF
--- a/examples/tokens/ERC20_solidity_compatible/ERC20.v.py
+++ b/examples/tokens/ERC20_solidity_compatible/ERC20.v.py
@@ -21,8 +21,8 @@ num_issued: uint256
 def deposit():
     _value: uint256 = convert(msg.value, 'uint256')
     _sender: address = msg.sender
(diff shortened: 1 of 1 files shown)


fubuloubu:
I don't see the casing updates

cmgannon26:
You couldn't see them because I didn't commit them. :) My bad.

Should be fixed now.

jacqueswww:
LGTM

PR #840, opened 2018-05-25
Removed currency1 and currency2 (and some more)

ben-kaufman:
### - What I did
Fixed an error message.
Fixed a typo.
Removed ``currency1`` and ``currency2`` types, as discussed after adding custom units.

### - Cute Animal Picture

image


fubuloubu:
Should we remove `currency_value` too under the same argument?

jacqueswww:
Thanks for the contribution :smile:

PR #847, opened 2018-05-29
Update ERC20 Compatible token 

maurelian:
### - What I did

Fix an issue compiling Phil Daian's ERC20 token impl

```
vyper.exceptions.TypeMismatchException: line 22: Expecting one of ('num_literal', 'int128', 'bytes32') for argument 1 of convert
    _value: uint256 = convert(msg.value, 'uint256')
```

### - How I did it

Converted `uint256` values to type `uint256(wei)`.

### - How to verify it

**Good question!** I ran the full test suite for the project, and noticed that it did not test this token. 
I also tried running this [test file](https://github.com/ethereum/vyper/blob/patch/updateERC20/tests/examples/tokens/ERC20_solidity_compatible/run_tests.py 
), but I got an error with the rlp lib: 

```
    from rlp.utils import decode_hex, encode_hex, ascii_chr, str_to_bytes
ImportError: cannot import name 'decode_hex'
```

Update ERC20 Compatible token 

### - Cute Animal Picture

image

### DIFF
--- a/examples/tokens/ERC20_solidity_compatible/ERC20.v.py
+++ b/examples/tokens/ERC20_solidity_compatible/ERC20.v.py
@@ -9,17 +9,17 @@
 # language interoperability and not for production use.
 
 # Events issued by the contract
-Transfer: event({_from: indexed(address), _to: indexed(address), _value: uint256})
-Approval: event({_owner: indexed(address), _spender: indexed(address), _value: uint256})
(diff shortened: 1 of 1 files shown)


fubuloubu:
Awesome! Glad the labeled units are working now.

Yeah, the RLP library just broke like that for almost everyone. Will have to fix that...

PR #851, opened 2018-05-30
Do not put out the full stack trace when using vyper compiler binary.

### LINKED ISSUE #821: Cleaner Exceptions
### What's your issue about?

Currently we output the whole stack trace when any vyper syntax error or exception occurs.
With vyper entering beta, the time has come to supress the trace and return cleaner error messages.

Full tracebacks should still be toggleable, vir the vyper binary. But by default the full trace (for the binary) should not be seen.

### How can it be fixed?

see `vyper/excpetions.py` and `bin/vyper`.

jacqueswww:
### - What I did
Fixes #821 .

### DIFF
--- a/bin/vyper
+++ b/bin/vyper
@@ -1,13 +1,15 @@
 #!/usr/bin/env python
 import argparse
 import json
-
+import sys
 import vyper
 
 from vyper import compiler, optimizer
 from vyper.parser.parser import parse_to_lll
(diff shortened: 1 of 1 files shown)


fubuloubu:
And this shows the source code where the violation occured, right?

LGTM if so!

jacqueswww:
Yes in most cases, if not return s a line number 😊

PR #855, opened 2018-05-31
Improve static call keywords

### LINKED ISSUE #770: VIP: Denote Static Call per Method in Contract Interface Definition
# Preamble
VIP: 770
Title: Denote Static Call per Method in Contract Interface Definition
Author: @fubuloubu
Type: Standard Track
Status: Draft
Created: 2018-04-09

## Simple Summary
Current method of defining contract interfaces decides whether to static call all methods based on the usage of the `modifiable`/`static` keyword. Suggestion would make this per-method.

## Abstract
Currently, to work with external contracts, one must supply a keyword denoting whether to make calls using `STATICCALL` opcode for ALL methods in a external contract variable, e.g.:
https://github.com/ethereum/vyper/blob/22be6dbae5a767a2d5525deb7cc93c45860e1982/tests/parser/features/external_contracts/test_modifiable_external_contract_calls.py#L14-L18

Current method assumes all methods are state modifiable if `modifiable` keyword is used when referring to a variable storing that contract's address. This means if you would like to safely interact with a contract that implements multiple methods, only some of which are state changing, you will need to define two variables to do so, or always call read-only methods with a regular `CALL` opcode.

Proposed method would make this per-method, as detailed in Specification.

## Motivation
This increases utility of the contract interface mechanism, and increases transparency into which calls should be made statically, and which are allowed to modify state.

Update per VIP #748 would make this syntax mesh much nicer than it currently will.

## Specification
Proposed syntax would be (using Basic ERC20 interface standard):
```python
contract ERC20Basic:
    def balanceOf(_owner: address) -> uint256: static
    def totalSupply() -> uint256: static
    def transfer(_to: address, _amt: uint256) -> bool: modifying
    # Other methods not implemented in interface are not callable

token: ERC20Basic

@constant
@public
def tokenBalance() -> uint256:
    return self.token.balanceOf(msg.sender)

@public
def giveMeTokens():
    self.token.transfer(msg.sender, self.tokenBalance())
```

We could also change the `contract` keyword (which is a search/replace for `class`) into `interface`, which I think makes the purpose clearer for these external contract interface types.

## Backwards Compatibility
Not backwards compatible, modifies syntax for interface definitions

## Copyright

jacqueswww:
### - What I did
Fixes #770.

### DIFF
--- a/tests/parser/features/external_contracts/test_modifiable_external_contract_calls.py
+++ b/tests/parser/features/external_contracts/test_modifiable_external_contract_calls.py
@@ -65,8 +65,8 @@ static_bar_contract: ConstBar
 
 @public
 def __init__(contract_address: address):
-    self.modifiable_bar_contract.address = contract_address
-    self.static_bar_contract.address = contract_address
+    self.modifiable_bar_contract = contract_address
+    self.static_bar_contract = contract_address
 
 @public
 def modifiable_set_lucky(_lucky: int128) -> int128:
--- a/tests/parser/features/external_contracts/test_modifiable_external_contract_calls.py
+++ b/tests/parser/features/external_contracts/test_modifiable_external_contract_calls.py
@@ -11,11 +11,14 @@ def set_lucky(_lucky: int128):
 """
 
     contract_2 = """
-class Bar():
-    def set_lucky(_lucky: int128): pass
+class ModBar():
+    def set_lucky(_lucky: int128): modifying
 
-modifiable_bar_contract: modifiable(Bar)
-static_bar_contract: static(Bar)
+class ConstBar():
+    def set_lucky(_lucky: int128): constant
+
+modifiable_bar_contract: ModBar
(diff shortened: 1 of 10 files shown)


fubuloubu:
Woo!

tests/parser/features/external_contracts/test_modifiable_external_contract_calls.py
```diff
@@ -51,20 +54,23 @@ def set_lucky(_lucky: int128) -> int128:
...
+
+modifiable_bar_contract: ModBar
+static_bar_contract: ConstBar
 
 @public
 def __init__(contract_address: address):
-    self.modifiable_bar_contract = contract_address
-    self.static_bar_contract = contract_address
+    self.modifiable_bar_contract.address = contract_address
+    self.static_bar_contract.address = contract_address
```
jacqueswww: Remove the `.address` syntax for now.
Then we can consider adding support for it later.

PR #856, opened 2018-06-01
Disallow variables with built in functions name

ben-kaufman:
Disallow variables to have the name of a built in function.

### - What I did
Didn't have the time to open an issue for that since I want it to be in the beta release.
What do you think?

### - Cute Animal Picture

image

### DIFF
--- a/vyper/utils.py
+++ b/vyper/utils.py
@@ -146,19 +146,12 @@ reserved_words = ['int128', 'int256', 'uint256', 'address', 'bytes32',
                   'wei', 'finney', 'szabo', 'shannon', 'lovelace', 'ada',
                   'babbage', 'gwei', 'kwei', 'mwei', 'twei', 'pwei', 'contract', 'units']
 
-built_in_functions = ['floor', 'ceil', 'as_unitless_number',
-                      'convert', 'slice', 'len', 'concat',
-                      'sha3', 'method_id', 'keccak256', 'ecrecover',
-                      'ecadd', 'ecmul', 'extract32', 'as_wei_value',
-                      'raw_call', 'RLPList', 'blockhash', 'bitwise_and',
-                      'bitwise_or', 'bitwise_xor', 'bitwise_not',
-                      'uint256_addmod', 'uint256_mulmod', 'shift',
-                      'create_with_code_of', 'min', 'max', 'send',
-                      'selfdestruct', 'raw_log']
-
 
 # Is a variable or member variable name valid?
 def is_varname_valid(varname, custom_units):
+    from vyper.functions import dispatch_table, stmt_dispatch_table
+    built_in_functions = [x for x in stmt_dispatch_table.keys()] + \
+      [x for x in dispatch_table.keys()]
     if custom_units is None:
         custom_units = []
     if varname.lower() in [cu.lower() for cu in custom_units]:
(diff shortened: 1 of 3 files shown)


fubuloubu:
Add a failing testcase (name appears in list).
Bonus points: add a case for a few of them.

LGTM otherwise

vyper/utils.py
```diff
@@ -146,6 +146,16 @@ def in_bounds(cls, type_str, value):
...
 
+built_in_functions = ['floor', 'ceil', 'as_unitless_number',
+                      'convert', 'slice', 'len', 'concat',
+                      'sha3', 'method_id', 'keccak256', 'ecrecover',
+                      'ecadd', 'ecmul', 'extract32', 'as_wei_value',
+                      'raw_call', 'RLPList', 'blockhash', 'bitwise_and',
+                      'bitwise_or', 'bitwise_xor', 'bitwise_not',
+                      'uint256_addmod', 'uint256_mulmod', 'shift',
+                      'create_with_code_of', 'min', 'max', 'send',
+                      'selfdestruct', 'raw_log']
```
jacqueswww: We should see if there is not a more dynamic way of pulling these from `functions.py` ?
I believe there is a dict there that could pull this from. Allowing us to not have to worry about changing it in more than one place later.

jacqueswww:
@ben-kaufman thanks for adding this! 

As mentioned above : instead of using a hard coded list we could probably do something like this instead:
```
builtin_functions = [x for x in stmt_dispatch_table.keys()] + [x for x in dispatch_table.keys()]
```

ben-kaufman:
@jacqueswww good idea.
I'm having a weird issue while trying to implement it.
As I do ``import vyper.functions`` the ``make test`` throws this error:
```
ImportError: cannot import name 'get_original_if_0_prefixed'
```
Any chance you know what I did wrong here?

jacqueswww:
Sounds like a recursive import perhaps?
Try importing `from vyper.functions import stmt_dispatch_table` instead of the whole file?

ben-kaufman:
Already tried it... Still the same problem

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

jacqueswww:
Hmmm that is strange. 🤔

jacqueswww:
How about making a function called `def is_builtin_function(name)` in functions.py and try using that?

ben-kaufman:
This might work, I'll try that later today. If it doesn't work I'll try to push the first way just to make sure that it's not some bug on my machine.

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

jacqueswww:
I just tested locally, and it is a recursive import. For the sake of completing this PR. Just do this:

```python
# Is a variable or member variable name valid?
def is_varname_valid(varname, custom_units):
    from vyper.functions import dispatch_table, stmt_dispatch_table
    built_in_functions = [x for x in stmt_dispatch_table.keys()] + \
        [x for x in dispatch_table.keys()]
    if custom_units is None:
        custom_units = []
    if varname.lower() in [cu.lower() for cu in custom_units]:
        return False
    if varname.lower() in base_types:
        return False
    if varname.lower() in valid_units:
        return False
        return False
    if varname.upper() in opcodes:
        return False
    if varname.lower() in built_in_functions:
        return False
    return True
```

ben-kaufman:
Will do that, thanks!

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

ben-kaufman:
Just fixed it, thanks for the help!

jacqueswww:
LGTM

PR #857, opened 2018-06-01
Improved built in functions documentation

ben-kaufman:
Improved the documentation for Vyper's built in functions.

### - Cute Animal Picture

image

### DIFF
--- a/docs/built-in-functions.rst
+++ b/docs/built-in-functions.rst
@@ -19,24 +19,54 @@ Functions
   def floor(a) -> b:
     """
     :param a: value to round down
-    :type a: either decimal or int128
+    :type a: decimal
 
-    :output b: integer
+    :output b: int128
     """
 Rounds a decimal down to the nearest integer.
 
-**decimal**
------------
+**ceil**
+---------
 ::
 
-  def decimal(a) -> b:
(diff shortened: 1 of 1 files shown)


ben-kaufman:
I just added documentation for Vyper's "low-level" functions. I still have a few questions in order to be able to finish this documentation. Any help with that will be really appreciated.

Regarding ``raw_call``:
1. What is the outsize?
2. Should I write num_literal in the docs as the argument type or should I say int128?
3. Is the function returns something? If so what?

What is ``raw_log``?

jacqueswww:
`raw_log` like `raw_call` allows you to emit a log statement without specifying the abi type, by just passing it a bunch of arguments.

jacqueswww:
`outsize` is the basically the size of the bytes that get returned from the `CALL` opcode. Basically the maxlen to the bytes output from `rawcall`.

jacqueswww:
I think just say something like "fixed literal value" ;)

fubuloubu:
maybe it can be called `bytes_size` to be more explicit?

ben-kaufman:
Thinking about it, why do we need to specify the variable names in the ``raw_call` function.
I think it would be better to remove it if there's no special reason for that.

PR #866, opened 2018-06-04
Support for fallback/default function

### LINKED ISSUE #781: No mention of fallback function anywhere
### What's your issue about?

Searches for "fallback" in Vyper docs and in the repository return no results. If fallback functions are intentionally not allowed for security reasons, this should be mentioned in the README and/or docs.

 If this feature exists or is planned, I would love to know! Its something I consider very useful in Solidity

#### Cute Animal Picture

image

jacqueswww:
### - What I did
Fixes #781.

### DIFF
--- a/examples/wallet/wallet.v.py
+++ b/examples/wallet/wallet.v.py
@@ -57,3 +57,9 @@ def approve(_seq: int128, to: address, value: wei_value, data: bytes[4096], sigd
     self.seq += 1
     # Use raw_call to send the transaction
     return raw_call(to, data, outsize=4096, gas=3000000, value=value)
+
+
+@public
+@payable
+def __default__():
+    pass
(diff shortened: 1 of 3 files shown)


examples/wallet/wallet.v.py
```diff
@@ -57,3 +57,9 @@ def approve(_seq: int128, to: address, value: wei_value, data: bytes[4096], sigd
     self.seq += 1
     # Use raw_call to send the transaction
     return raw_call(to, data, outsize=4096, gas=3000000, value=value)
+
+
+@public
+@payable
+def __default__():
+    pass
```
fubuloubu: Just confirming, this means it allows anyone to add funds?

jacqueswww: Yes that is correct.

jacqueswww:
Found it ;)

PR #870, opened 2018-06-05
Fix convert units

### LINKED ISSUE #865: Convert units should carry
### What's your issue about?

As @djrtwo found.

`convert` should preserve the unit that have been set, currently it doesn't:

```
b: uint256(meter) = 1234
a: int128 = convert(b, "int128") # should error
a: int128(meter) = convert(b, "int128") # should work
```

### How can it be fixed?

jacqueswww:
### - What I did
Fixes #865 .


djrtwo:
looks good to me

djrtwo:
Shouldn't `int128` -> `decimal` also keep unit? Should add test.

I guess any `convert` we support should keep units, no?

jacqueswww:
@djrtwo `int128` -> `decimal` already does so. But adding a test can not hurt :smile:

djrtwo:
I ❤️ Tests.

PR #876, opened 2018-06-06
Cleanup docs - literals are not always interpreted as int128's & correct typecasting example

rohithpr:
### - What I did
Updated docs as a deprecated function was still being recommended. Developers should now use `convert(_literal, 'uint256')` instead of `as_uint256(_literal)`.

### - How I did it
Doc change.

### - How to verify it
Visual verification should be sufficient.

### - Cute Animal Picture

baby_elephant_1024x768

### DIFF
--- a/docs/types.rst
+++ b/docs/types.rst
@@ -110,7 +110,7 @@ Values
 Integer values between 0 and (2\ :sup:`256`-1).
 
 .. note::
-    Integer literals are always interpreted as ``int128``. In order to assign a literal to a ``uint256`` use ``as_uint256(_literal)``.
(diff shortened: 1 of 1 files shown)


ben-kaufman:
As I wrote in the issue, the entire statement isn't correct anymore. Could you please change it in accordance?

rohithpr:
@ben-kaufman 

Got it! So integer literals can be treated as `uint256` while assigning them to a variable, if the variable is declared as `uint256`. However, in the case of `max(uint256_variable, 3)`: `3` is being treated as int128. Looks like there is some inconsistency in how numbers are treated. I'm not sure how to proceed here.

ben-kaufman:
@rohithpr yes, I believe that is because the default for literal as a function parameter is ``int128``. So I suggest writing something like: By default, literals will be interpreted as an ``int128``. In cases where appropriate (such as value assignment) num literal might be interpreted as a ``uint256`` e.g. ``a:  uint256 = 10``.

rohithpr:
@ben-kaufman - I've updated the note with the relevant info. The Jenkins build has passed as well.

PR #877, opened 2018-06-06
Support max(int128, uint256)

### LINKED ISSUE #873: Fix literal int for min/max.
### What's your issue about?

a: uint256 = 1
b: uint256 = max(2, a)

Should be allowed to compiled, as 2 is a valid uint256.

### How can it be fixed?

functions.py

#### Cute Animal Picture
^.^

rohithpr:
### - What I did
Added support for `max(int128, uint256)` which returns a `uint256`.

Closes #873.

### - How I did it
If one of the parameters passed is `int128` and the other one is `uint256` for the `max` method then allow comparison rather than throwing an error.

### - How to verify it
Call `max(convert(_literal, 'uint256), _literal))`.

Added support for `max(int128, uint256)` which returns a `uint256`.

### - Cute Animal Picture

36288655fbb0466af9b636b0e45514bf

### DIFF
--- a/tests/parser/functions/test_minmax.py
+++ b/tests/parser/functions/test_minmax.py
@@ -17,3 +17,158 @@ def goo() -> uint256:
     assert c.goo() == 83
 
     print("Passed min/max test")
+
+
+def test_max_var_uint256_literal_int128(get_contract_with_gas_estimation):
+    """Tests to verify that max works as expected when a variable/literal uint256 and a literal int128 are passed."""
+    code = """
+@public
+def foo() -> uint256:
+    a: uint256 = 2 ** 200
+    b: uint256 = 5
+    return max(a, 5) + max(b, 5)
+
+@public
+def goo() -> uint256:
+    a: uint256 = 2 ** 200
+    b: uint256 = 5
+    return max(5, a) + max(5, b)
+
+@public
+def bar() -> uint256:
+    a: uint256 = 2
+    b: uint256 = 5
+    return max(a, 5) + max(b, 5)
+
+@public
+def baz() -> uint256:
+    a: uint256 = 2
+    b: uint256 = 5
+    return max(5, a) + max(5, b)
+
+@public
+def both_literals() -> uint256:
+    return max(2 ** 200, 2)
+"""
+    c = get_contract_with_gas_estimation(code)
+    assert c.foo() == 2 ** 200 + 5
+    assert c.goo() == 2 ** 200 + 5
+    assert c.bar() == 5 + 5
+    assert c.baz() == 5 + 5
+    assert c.both_literals() == 2 ** 200
+
+
+def test_min_var_uint256_literal_int128(get_contract_with_gas_estimation):
+    """Tests to verify that max works as expected when a variable/literal uint256 and a literal int128 are passed."""
+    code = """
+@public
+def foo() -> uint256:
+    a: uint256 = 2 ** 200
+    b: uint256 = 5
+    return min(a, 5) + min(b, 5)
+
+@public
+def goo() -> uint256:
+    a: uint256 = 2 ** 200
+    b: uint256 = 5
+    return min(5, a) + min(5, b)
+
+@public
+def bar() -> uint256:
+    a: uint256 = 2
+    b: uint256 = 5
+    return min(a, 5) + min(b, 5)
+
+@public
+def baz() -> uint256:
+    a: uint256 = 2
+    b: uint256 = 5
+    return min(5, a) + min(5, b)
+
+@public
+def both_literals() -> uint256:
+    return min(2 ** 200, 2)
+"""
+    c = get_contract_with_gas_estimation(code)
+    assert c.foo() == 5 + 5
+    assert c.goo() == 5 + 5
+    assert c.bar() == 2 + 5
+    assert c.baz() == 2 + 5
+    assert c.both_literals() == 2
+
+
+def test_minmax_var_uint256_var_int128(get_contract_with_gas_estimation, assert_compile_failed):
+    """Tests to verify that max throws an error if a variable uint256 and a variable int128 are passed."""
+    from vyper.exceptions import TypeMismatchException
+    code_1 = """
+@public
+def foo() -> uint256:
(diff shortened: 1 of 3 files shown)


rohithpr:
@jacqueswww - I've made an attempt at resolving the issue 873. Please take a look at it and let me know what you think.

Couple of points I wanted to mention:

- As things stand, I'm not checking if the `int128` is a `x.typ.is_literal`. Should I check that? The issue says "literal" but if literals are supported then variables can be too, I guess.

Assume: `a_u256: uint256 = 10`

- The return type of `min(a_u256, -100)` would have to be `int128`.
- The return type of `min(a_u256, 100)` can be either `int128` or `uint256`.

The smaller of the two is the variable which originally had the type `uint256` but to maintain consistency with the first case we'd have to return a `int128`. It may cause issues if someone glosses over the fact that the method can take in a literal and a `uint256` only to give back a `int128`.

- If `min()` doesn't support mixed types then should `max()` support it?

jacqueswww:
@rohithpr We had some at length discussions on this matter in gitter, if you want to scroll back and check it out.

Some notes on this particular PR:

- This issue is for literals only (see the example in the issue). So should only address the item in the example (not variables).
- Hence the `max(convert(3, 'uint256'), convert(5, 'int128'))` syntax is exactly what we are trying to avoid.
- We can not allow mixed types without clamping -> run time error.
- Static checks have to occur on the int value -> compile time error.

vyper/functions/functions.py
```diff
@@ -632,9 +632,14 @@ def minmax(expr, args, kwargs, context, is_min):
         comparator = 'gt' if is_min else 'lt'
     else:
         comparator = 'sgt' if is_min else 'slt'
-    if left.typ.typ == right.typ.typ:
+    if left.typ.typ == right.typ.typ or \
+            (all([typ in ('int128', 'uint256') for typ in (left.typ.typ, right.typ.typ)]) and
```
jacqueswww: There is a much nicer pattern one can use for this type of check ;)

```python
or {left.typ.typ, right.typ.typ} == {'int128', 'uint256'}
```

rohithpr: This would be `False` if the inputs are `{uint256, uint256}`! :)

Looking at this as an option more closely led me to `{left.typ.typ, right.typ.typ} - {'int128', 'uint256'} == set()` which would work as expected. However, after considering your other comment this whole block might change and this mightn't be required!

jacqueswww: Right but the first part of or statement would catch that case.

rohithpr:
Note: Instead of adding the method `_can_compare_with_uint256` we can also make the check in line. However, in the interest of readability I moved the logic to a function. If it check was inline the new code block would look like this:

```python
    elif not is_min and \
            {left.typ.typ, right.typ.typ} - {'int128', 'uint256'} == set() and \
            ((left.typ.typ == 'int128' and left.typ.is_literal and left.value >= 0) or \
             (right.typ.typ == 'int128' and right.typ.is_literal and right.value >= 0)):
        o = ['if', [comparator, '_l', '_r'], '_r', '_l']
        if right.typ.typ == 'uint256':
            otyp = right.typ
        else:
            otyp = left.typ
        otyp.is_literal = False
```

^ Let me know if this is preferable to what I have currently pushed and I'll push that.

rohithpr:
@jacqueswww - I think I've addressed all your points, take a look at it when you can.

I didn't understand what you meant by clamping here:

> We can not allow mixed types without clamping -> run time error.

Could you please tell me what "clamping" is and how to do it?

jacqueswww:
Sure thing, I will look it over and show you ;)

vyper/functions/functions.py
rohithpr: Haha, true!

vyper/functions/functions.py
```diff
@@ -625,6 +625,13 @@ def _max(expr, args, kwargs, context):
 
 
 def minmax(expr, args, kwargs, context, is_min):
+    def _can_compare_with_uint256(operand):
+        if operand.typ.typ == 'uint256':
+            return True
+        elif operand.typ.typ == 'int128' and operand.typ.is_literal and operand.value >= 0:
```
jacqueswww: Replace operand.value > 0 with the utils function (there is obviously an upper bound to account for as well :smiling_imp: ):
```python
operand.typ.typ == 'int128' and operand.typ.is_literal and SizeLimits.in_band('uint256', operand.value):
```

vyper/functions/functions.py
```diff
@@ -636,6 +643,13 @@ def minmax(expr, args, kwargs, context, is_min):
...
         otyp = left.typ
         otyp.is_literal = False
+    elif not is_min and _can_compare_with_uint256(left) and _can_compare_with_uint256(right):
+        o = ['if', [comparator, '_l', '_r'], '_r', '_l']
+        if right.typ.typ == 'uint256':
+            otyp = right.typ
+        else:
+            otyp = left.typ
+        otyp.is_literal = False
     # elif left.typ.typ == 'int128' and right.typ.typ == 'decimal':
```
jacqueswww: This commented area can be taken out, for now I believe it's better we don't support implicit conversion.

tests/parser/functions/test_minmax.py
```diff
@@ -17,3 +17,84 @@ def goo() -> uint256:
...
+
+@public
+def baz() -> uint256:
+    a: uint256 = 2
+    b: uint256 = 5
+    return max(5, a) + max(5, b)
+
+@public
+def both_literals() -> uint256:
+    return max(2 ** 200, 2)
```
jacqueswww: Just so we cover all bases please add a `min()` in here somewhere.

jacqueswww:
@rohithpr Oh yes, clamps are runtime checks for overflowing numbers. Best example:
https://github.com/ethereum/vyper/blob/0fe9831a86833ead559c9d6403fe1e88edc069b9/vyper/parser/parser.py#L434
But for the above case it's not a problem anymore, because it's only literals. And we don't allow mixed types.

rohithpr:
@jacqueswww - I've resolved all the review comments.

PR #880, opened 2018-06-07
Version statement

### LINKED ISSUE #852: Add pragma tag
### What's your issue about?
We should add a simple pragma tag. That the current compiler will compare to, and warn the user if the version is not the same.

Format:
```
# pragma: x.y.z
```

Change the severity of the warning based on:
Given version format: x.y.z

z -> Patch, minor changes. As long x.y is the same, should compile the same.
x.y -> major changes, will most likely not compile.

### How can it be fixed?
Also discussed at length in https://github.com/ethereum/vyper/issues/456.

jacqueswww:
### - What I did
Fixes #852.

### DIFF
--- a/tests/compiler/test_pre_parser.py
+++ b/tests/compiler/test_pre_parser.py
@@ -52,3 +52,36 @@ def foo(contract_address: address) -> int128:
     c1 = get_contract(contract_1)
     c2 = get_contract(contract_2)
     assert c2.foo(c1.address) == 1
+
+
+def test_version_pragma(get_contract):
+    code = """
+# @version 0.0.4
+
+@public
+def test():
+    pass
+    """
+    assert get_contract(code)
+
+
+def test_version_empty_version(assert_compile_failed, get_contract):
+    code = """
+# @version
+
+@public
+def test():
+    pass
+    """
+    assert_compile_failed(lambda: get_contract(code))
+
+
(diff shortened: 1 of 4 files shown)


tests/compiler/test_pre_parser.py
```diff
@@ -52,3 +52,47 @@ def foo(contract_address: address) -> int128:
...
+@public
+def test():
+    pass
+    """
+    assert_compile_failed(lambda: get_contract(code))
+
+
+def test_version_empty_version_mismatch(assert_compile_failed, get_contract):
+    code = """
+# @version 0.1.0
```
fubuloubu: make this like `9.9.9` so we never run into this passing.

vyper/__init__.py
```diff
@@ -10,4 +10,4 @@
 try:
     __version__ = pkg_resources.get_distribution('vyper').version
 except pkg_resources.DistributionNotFound:
-    __version__ = 'development'
+    __version__ = '0.0.0development'
```
fubuloubu: Can't we get this from bumpversion?

jacqueswww: Yes bumpversion sets it in setup.py, and setup.py sets it in pkg_resources module ;)
So if not in development (locally in your own repo), but installed it pulls in the bumpversion/setup.py.

PR #882, opened 2018-06-07
Allow conversion of uint256 to uint256

### LINKED ISSUE #881: convert(_uint256, 'uint256') fails
* vyper Version: master
* pyethereum Version: x.x.x
* OS: osx/linux/win
* Python Version (python --version):

### What's your issue about?

Compiling either of the following pieces of code causes the error: `Expecting one of ('num_literal', 'int128', 'bytes32') for argument 1 of convert`

```python
@public
def foo() -> uint256:
    a: uint256 = 2 ** 20
    return convert(a, 'uint256')
```

```python
@public
def foo() -> uint256:
    return convert(2 ** 200, 'uint256')
```

If the code is as follows a run time error will occur:

```python
@public
def foo() -> uint256:
    i: uint256 = 200
    return convert(2 ** i, 'uint256')
```

`convert()` should be able to convert a `uint256` into `uint256` without any issues.

### How can it be fixed?

Add `uint256` to the list of tuple of expected arg types for `to_uint256()`.

#### Cute Animal Picture

elephant

rohithpr:
Closes #881

### - What I did
Allow `uint256` to be `convert`ed into `uint256`.

### - How I did it
Added `uint256` to the list of tuple of expected arg types for `to_uint256()`.

### - How to verify it
Added tests that ensure that the issue doesn't occur at both compile time and run time.

`uint256`s can be `convert`ed into `uint256`.

### - Cute Animal Picture

elephant

### DIFF
--- a/tests/parser/functions/test_convert_uint256.py
+++ b/tests/parser/functions/test_convert_uint256.py
@@ -26,3 +26,24 @@ def foo() -> uint256:
 
     c = get_contract_with_gas_estimation(code)
     assert c.foo() == 2 ** 256 - 1
+
+
+def test_convert_uint256_to_uint256(get_contract_with_gas_estimation):
+    code = """
+@public
+def foo() -> uint256:
+    return convert(convert(1, 'uint256'), 'uint256')
+
+@public
+def goo() -> uint256:
+    return convert(convert(2 ** 200, 'uint256'), 'uint256')
+
+@public
+def bar() -> uint256:
+    a: uint256 = 200
+    return convert(convert(2 ** a, 'uint256'), 'uint256')
+"""
+    c = get_contract_with_gas_estimation(code)
+    assert c.foo() == 1
+    assert c.goo() == 2 ** 200
+    assert c.bar() == 2 ** 200
--- a/vyper/types/convert.py
+++ b/vyper/types/convert.py
@@ -37,7 +37,7 @@ def to_int128(expr, args, kwargs, context):
         return byte_array_to_num(in_node, expr, 'int128')
 
 
-@signature(('num_literal', 'int128', 'bytes32'), 'str_literal')
+@signature(('num_literal', 'int128', 'bytes32', 'uint256'), 'str_literal')
 def to_uint256(expr, args, kwargs, context):
     input = args[0]
     typ, len = get_type(input)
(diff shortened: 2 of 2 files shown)


fubuloubu:
Can you give a good explanation why this is necessary?

rohithpr:
- Partly to do with how literals/expressions are treated as `int128`s or `uint256`s depending on the context and context.
- Partly for the sake of consistency and limiting gotcha's.

The following are acceptable currently:

- `convert(2, 'int128')`
- `convert(2 ** 127 - 1, 'int128')`
- `convert(2 ** 127 - 1, 'uint256')`

However, `convert(2 ** 127, 'uint256')` (or any number greater than that) throws an error saying `Expecting one of ('num_literal', 'int128', 'bytes32') for argument 1 of convert`.

Sure, it causes a compile time error and it isn't going to cause any issues in live DApp but I can't think of a good reason why `int`s can be converted into `int`s but `uint`s can't be converted into `uint`s.

PS: Python, JS etc. all allow one to do `num(num(x))`, `str(str(x))` etc. and these methods are idempotent (mostly). As it stands, Vyper allows this for one datatype but not for another and I'd prefer consistency over gas savings.

haydenadams:
> As it stands, Vyper allows this for one datatype but not for another i.e. convert(convert(2, 'int128'), 'int128') is okay but convert(convert(2, 'uint256'), 'uint256') is not. I'd prefer consistency over gas savings.

Agreed on consistency. Why not just disallow `convert(convert(2, 'int128'), 'int128') `?

fubuloubu:
Well, we want to create literals that will determine their own type in context (within reason). So that example is not even relevant.

`convert(convert(a_uint256, int128), int128)` would be closer to what is described, and it should indeed be disallowed. The first conversion to `int128` will work because it was a different type, but the second with throw an exemption during compilation because the types are the same.

My theory is if the types are the same for two converts, you made a mistake somewhere and should take this opportunity to re-evaluate why you wanted to do this.

rohithpr:
> The first conversion to int128 will work because it was a different type, but the second with throw an exemption during compilation because the types are the same.

`convert(convert(convert(2, 'int128'), 'int128'), 'int128')` compiles and gives the expected result as well.

As you can see from the `signature` decorator [here](https://github.com/ethereum/vyper/blob/master/vyper/types/convert.py#L26), `int128`s are a valid input when converting to `int128`.

But yeah, now that I am little more familiar with Vyper and after seeing the discussion here I understand that idempotency isn't required and `to_int128` is the odd method where `int128`s themselves are accepted. I should probably close this PR and open a different one to change the behavior of `to_int128`?

jacqueswww:
+1 for just dissallowing `convert(convert(2, 'int128'), 'int128')`. Will create issue.
The general rule of thumb I have for vyper is to minimise what you can do, and just-just allow you to write a contract :tongue:

PR #895, opened 2018-06-14
Fixed the assert documentation

ben-kaufman:
Mentioned the correct way to use the assert function in the docs.

I think we should consider removing one of the assert syntaxes.
Currently, both ``assert 1 < 2`` and ``assert(1 < 2)`` will work the same. We might want to remove one of them for uniformity.

### - Cute Animal Picture

image

### DIFF
--- a/docs/built-in-functions.rst
+++ b/docs/built-in-functions.rst
@@ -348,7 +348,9 @@ CAUTION! This method will delete the contract from the Ethereum blockchain. All 
     """
 
 Asserts the specified condition, if the condition is equals to true the code will continue to run.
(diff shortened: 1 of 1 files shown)


fubuloubu:
is that still considered pythonic? I mean, `print` in python 3 has parens now.

ben-kaufman:
Apparently, in Python 3 you can do both ways as well. Personally, this inconsistency is one of the things I don't really like in Python so I am still in favor of removing one of them but I don't think it matters that much. What do you think?

fubuloubu:
Streamlining approaches is important, but so is code complexity. We'd probably have to implement another check for this syntax to force developers to use one over the other, and at the end of the day I don't think this specific issue is that big of a deal. @jacqueswww what do you think?

jacqueswww:
I think for now we leave it as is. Because I quite like assert without the braces ;)
Therefore I am merging this documentation change.

Issue #892, opened 2018-06-13
Write default function documentation


jacqueswww:
### What's your issue about?

Now that we have fallback/default function support https://github.com/ethereum/vyper/pull/866. We need some solid documentation on using them :smile: 

- more than one example
- notes on not being able to parse parameters and why
- plus points for showing you can still pass calldata along (in proxy scenario).
- send having a gas limit
estimated word count: > 170

### How can it be fixed?
Inspiration:  
https://github.com/ethereum/vyper/blob/master/tests/parser/functions/test_default_function.py
https://solidity.readthedocs.io/en/latest/contracts.html?highlight=fallback#fallback-function

#### Cute Animal Picture
![](https://static.boredpanda.com/blog/wp-content/uuuploads/cute-baby-animals/cute-baby-animals-13.jpg)

status-open-bounty:
![Contract deploying](https://openbounty.status.im/img/deploying_contract.png)
Transaction: [0x0b0a1b181acf787695c91e88ee47a0bb446ce1189667297f275b8c79ed69a255](https://etherscan.io/tx/0x0b0a1b181acf787695c91e88ee47a0bb446ce1189667297f275b8c79ed69a255)

jakerockland:
@jacqueswww Looks like this issue probably can be closed?

PR #896, opened 2018-06-14
Documentation for default funcs

liamzebedee:
### - What I did

Documented new default function usage.

### - How I did it

Read the source changes, researched a bit about the EVM execution sematics to communicate a stronger understanding.

### - How to verify it

Generate docs and visit `structure-of-a-contract.html`

Documented default functions of contracts (fallback funcs).

### - Cute Animal Picture

image

### DIFF
--- a/docs/structure-of-a-contract.rst
+++ b/docs/structure-of-a-contract.rst
@@ -77,7 +77,7 @@ The following operations will consume more gas than the 2300 gas stipend:
 - Calling an external function which consumes a large amount of gas
 - Sending Ether
 
-Since global variables are still accessible, you can pass data to the default function through the raw byte buffer provided to EVM's CALL (``msg.data``).
+Although the default function receives no arguments, it can still access the ``msg`` global, including the address of who is interacting with the contract (``msg.sender``), the amount of ETH sent (``msg.value``), and the gas (``msg.gas``).
 
 .. _structure-events:
 
--- a/docs/structure-of-a-contract.rst
+++ b/docs/structure-of-a-contract.rst
@@ -55,12 +55,12 @@ If the function is annotated as `@payable`, this function is executed whenever t
 
 ::
 
-    Sent: event({sender: indexed(address)})
+    Payment: event({amount: int128, from: indexed(address)})
     
     @public
     @payable
     def __default__():
-        log.Sent(msg.sender)
+        log.Payment(msg.value, msg.sender)
 
 
 Considerations
@@ -68,7 +68,9 @@ Considerations
 
 Unlike Solidity, Vyper generates a default function is one isn't found, in the form of a REVERT call. Note that this still `generates an exception <https://github.com/ethereum/wiki/wiki/Subtleties>`_ and thus will not succeed in receiving funds. 
 
-Send calls to the contract give it a maximum stipend of 2300 gas, leaving not much room to perform other operations except basic logging. The following operations will consume more gas than the 2300 gas stipend:
+Ethereum specifies that the operations will be rolled back if the contract runs out of gas in execution. `send` calls to the contract come with a free stipend of 2300 gas, which does not leave much room to perform other operations except basic logging. **However**, if the sender includes a higher gas amount through a `call` instead of `send`, then more complex functionality can be run.
+
+The following operations will consume more gas than the 2300 gas stipend:
 
 - Writing to storage
 - Creating a contract
--- a/docs/structure-of-a-contract.rst
+++ b/docs/structure-of-a-contract.rst
@@ -42,6 +42,41 @@ Functions are the executable units of code within a contract.
 and have different levels of visibility (:ref:`visibility-and-getters`)
 towards other contracts. Functions must be decorated with either @public or @private.
 
+Default function
+----------------
+
+A contract can also have a default function, which is executed on a call to the contract if no other functions match the given function identifier (or if none was supplied at all, such as through someone sending it Eth). It is the same construct as fallback functions `in Solidity <https://solidity.readthedocs.io/en/latest/contracts.html?highlight=fallback#fallback-function>`_. 
+
+This function is always named `__default__` and must be annotated with `@public` and `@payable`. It cannot have arguments and cannot return anything.
+
+Furthermore, this function is executed whenever the contract receives plain Ether (without data). This is why the default function cannot accept arguments and return values - it is a design decision of Ethereum to make no differentiation between sending ether to a contract or a user address.
+
+**Example:**
+
+::
+
+    Sent: event({sender: indexed(address)})
+    
+    @public
+    @payable
+    def __default__():
+        log.Sent(msg.sender)
+
+
+Considerations
+~~~~~~~~~~~~~~
+
+Unlike Solidity, Vyper generates a default function is one isn't found, in the form of a REVERT call. Note that this still `generates an exception <https://github.com/ethereum/wiki/wiki/Subtleties>`_ and thus will not succeed in receiving funds. 
+
+Send calls to the contract give it a maximum stipend of 2300 gas, leaving not much room to perform other operations except basic logging. The following operations will consume more gas than the 2300 gas stipend:
+
+- Writing to storage
+- Creating a contract
+- Calling an external function which consumes a large amount of gas
+- Sending Ether
+
+Since global variables are still accessible, you can pass data to the default function through the raw byte buffer provided to EVM's CALL (`msg.data`).
+
 .. _structure-events:
 
 Events
--- a/docs/structure-of-a-contract.rst
+++ b/docs/structure-of-a-contract.rst
@@ -66,7 +66,7 @@ If the function is annotated as `@payable`, this function is executed whenever t
 Considerations
 ~~~~~~~~~~~~~~
 
-Unlike Solidity, Vyper generates a default function is one isn't found, in the form of a REVERT call. Note that this still `generates an exception <https://github.com/ethereum/wiki/wiki/Subtleties>`_ and thus will not succeed in receiving funds. 
+Unlike Solidity, Vyper generates a default function if one isn't found, in the form of a REVERT call. Note that this still `generates an exception <https://github.com/ethereum/wiki/wiki/Subtleties>`_ and thus will not succeed in receiving funds. 
 
 Ethereum specifies that the operations will be rolled back if the contract runs out of gas in execution. ``send`` calls to the contract come with a free stipend of 2300 gas, which does not leave much room to perform other operations except basic logging. **However**, if the sender includes a higher gas amount through a ``call`` instead of ``send``, then more complex functionality can be run.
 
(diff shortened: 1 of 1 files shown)


liamzebedee:
Fixes #892

docs/structure-of-a-contract.rst
```diff
@@ -42,6 +42,41 @@ Functions are the executable units of code within a contract.
 and have different levels of visibility (:ref:`visibility-and-getters`)
 towards other contracts. Functions must be decorated with either @public or @private.
 
+Default function
+----------------
+
+A contract can also have a default function, which is executed on a call to the contract if no other functions match the given function identifier (or if none was supplied at all, such as through someone sending it Eth). It is the same construct as fallback functions `in Solidity <https://solidity.readthedocs.io/en/latest/contracts.html?highlight=fallback#fallback-function>`_. 
+
+This function is always named `__default__` and must be annotated with `@public` and `@payable`. It cannot have arguments and cannot return anything.
```
fubuloubu: I think `@payable` is optional, confirm @jacqueswww?

docs/structure-of-a-contract.rst
```diff
@@ -42,6 +42,41 @@ Functions are the executable units of code within a contract.
 and have different levels of visibility (:ref:`visibility-and-getters`)
 towards other contracts. Functions must be decorated with either @public or @private.
 
+Default function
+----------------
+
+A contract can also have a default function, which is executed on a call to the contract if no other functions match the given function identifier (or if none was supplied at all, such as through someone sending it Eth). It is the same construct as fallback functions `in Solidity <https://solidity.readthedocs.io/en/latest/contracts.html?highlight=fallback#fallback-function>`_.
```
fubuloubu: "if no other functions match the given function identifier" I am pretty sure this is the behavior we have, but I don't actually like this behavior now that I think about it. The default function should not have any call data associated with it (aka fails if call data is provided) and I think in general if the function signature switch falls through, the contract should revert by default. We can keep this behavior in place for the fallback function, since that will check that call data is zero if no function signatures match.

@jacqueswww Please confirm current behavior matches this documentation, and also comment on my suggestion for modifying the behavior of the default function. This should be compatible with any contracts created by Solidity. In fact, do we have a test case for one contract sending another Eth internally?

docs/structure-of-a-contract.rst
```diff
@@ -42,6 +42,41 @@ Functions are the executable units of code within a contract.
...
+
+This function is always named `__default__` and must be annotated with `@public` and `@payable`. It cannot have arguments and cannot return anything.
+
+Furthermore, this function is executed whenever the contract receives plain Ether (without data). This is why the default function cannot accept arguments and return values - it is a design decision of Ethereum to make no differentiation between sending ether to a contract or a user address.
+
+**Example:**
+
+::
+
+    Sent: event({sender: indexed(address)})
```
fubuloubu: Suggestion: maybe add `amount` to this event, so if someone tries to run it you can see the amount you sent in the event.

docs/structure-of-a-contract.rst
```diff
@@ -42,6 +42,41 @@ Functions are the executable units of code within a contract.
...
+Unlike Solidity, Vyper generates a default function is one isn't found, in the form of a REVERT call. Note that this still `generates an exception <https://github.com/ethereum/wiki/wiki/Subtleties>`_ and thus will not succeed in receiving funds. 
+
+Send calls to the contract give it a maximum stipend of 2300 gas, leaving not much room to perform other operations except basic logging. The following operations will consume more gas than the 2300 gas stipend:
+
+- Writing to storage
+- Creating a contract
+- Calling an external function which consumes a large amount of gas
+- Sending Ether
+
+Since global variables are still accessible, you can pass data to the default function through the raw byte buffer provided to EVM's CALL (`msg.data`).
```
fubuloubu: See above, I think we should disallow this behavior

docs/structure-of-a-contract.rst
liamzebedee: Nope, documented in the unit test case.

fubuloubu: I'm not :100:, wait for confirmation from @jacqueswww unless you found in the code/tests where this is allowed.

liamzebedee: @fubuloubu Haha I found the test, I just misread the failure condition initially.

https://github.com/ethereum/vyper/blob/master/tests/parser/functions/test_default_function.py#L34

docs/structure-of-a-contract.rst
```diff
@@ -42,6 +42,41 @@ Functions are the executable units of code within a contract.
...
+    def __default__():
+        log.Sent(msg.sender)
+
+
+Considerations
+~~~~~~~~~~~~~~
+
+Unlike Solidity, Vyper generates a default function is one isn't found, in the form of a REVERT call. Note that this still `generates an exception <https://github.com/ethereum/wiki/wiki/Subtleties>`_ and thus will not succeed in receiving funds. 
+
+Send calls to the contract give it a maximum stipend of 2300 gas, leaving not much room to perform other operations except basic logging. The following operations will consume more gas than the 2300 gas stipend:
```
haydenadams: You can still choose to include a higher gas amount when you send ETH to an address, which allows more complex fallback functions that will automatically fail unless extra gas is included. At least thats how it is with Solidity fallbacks, hoping its the same here.

Many ICOs allowed purchasing by sending ETH to the fallback and including ~200,000 gas. I have even heard of MEW automatically s the higher gas amount when it detected a user sending to a known ICO with a fallback.

docs/structure-of-a-contract.rst
jacqueswww: Good idea.

docs/structure-of-a-contract.rst
jacqueswww: @fubuloubu this is the current behaviour.

And therefore I think it's fine to leave it as is (in the docs). However we can definitely scope a VIP for splitting default function into 2 cases:
1.) Receive funds 
2.) Function Not Found (404)
I saw this on solidity something similar is being discussed. https://github.com/ethereum/solidity/issues/3198#issuecomment-380493711

docs/structure-of-a-contract.rst
jacqueswww: Yes payable is optional, depending on the use case ;)

docs/structure-of-a-contract.rst
jacqueswww: @haydenadams yes it is I believe, created an issue to demonstrate this as a tests. 
@liamzebedee   please make it clear that one can still send more gas to a default function, if you don't use `send`. for example raw_call / CALL allows you to send as much gas as you like.

liamzebedee:
Ready for another review!

liamzebedee:
Ping @fubuloubu

docs/structure-of-a-contract.rst
```diff
@@ -42,6 +42,43 @@ Functions are the executable units of code within a contract.
...
+Ethereum specifies that the operations will be rolled back if the contract runs out of gas in execution. ``send`` calls to the contract come with a free stipend of 2300 gas, which does not leave much room to perform other operations except basic logging. **However**, if the sender includes a higher gas amount through a ``call`` instead of ``send``, then more complex functionality can be run.
+
+The following operations will consume more gas than the 2300 gas stipend:
+
+- Writing to storage
+- Creating a contract
+- Calling an external function which consumes a large amount of gas
+- Sending Ether
+
+Since global variables are still accessible, you can pass data to the default function through the raw byte buffer provided to EVM's CALL (``msg.data``).
```
fubuloubu: I think accessing `msg.data` is discouraged behavior. Please mention that, or skip this section. You could perhaps mention accessing `msg.sender` instead.

docs/structure-of-a-contract.rst
```diff
@@ -42,6 +42,43 @@ Functions are the executable units of code within a contract.
...
+    @public
+    @payable
+    def __default__():
+        log.Payment(msg.value, msg.sender)
+
+
+Considerations
+~~~~~~~~~~~~~~
+
+Unlike Solidity, Vyper generates a default function is one isn't found, in the form of a REVERT call. Note that this still `generates an exception <https://github.com/ethereum/wiki/wiki/Subtleties>`_ and thus will not succeed in receiving funds.
```
fubuloubu: Typo: "... generates a default function *if* one isn't found"

docs/structure-of-a-contract.rst
```diff
@@ -42,6 +42,43 @@ Functions are the executable units of code within a contract.
...
+    def __default__():
+        log.Payment(msg.value, msg.sender)
+
+
+Considerations
+~~~~~~~~~~~~~~
+
+Unlike Solidity, Vyper generates a default function is one isn't found, in the form of a REVERT call. Note that this still `generates an exception <https://github.com/ethereum/wiki/wiki/Subtleties>`_ and thus will not succeed in receiving funds. 
+
+Ethereum specifies that the operations will be rolled back if the contract runs out of gas in execution. ``send`` calls to the contract come with a free stipend of 2300 gas, which does not leave much room to perform other operations except basic logging. **However**, if the sender includes a higher gas amount through a ``call`` instead of ``send``, then more complex functionality can be run.
```
fubuloubu: This is more nuanced than your description. I can call a default/fallback directly from a raw transaction (in a contract or from an EOA) and the gas stipend is more than 2300. This stipend only holds when when an `address.transfer(uint)` is called, as implemented by Solidity and Vyper compilers as a mitigation against excessive gas usage.

It is therefore considered a *best practice* to ensure your payable default function is compatible with the 2300 gas stipend. If it is currently an error, then overusing the gas stipend should be a warning instead. There are legitimate use cases for a default that exceeds this amount. Lots of nuance here.

@jacqueswww, do you agree?

docs/structure-of-a-contract.rst
jacqueswww: +1 a block mentioning that when using `msg.data` care should be taking.

jacqueswww:
@liamzebedee looking good, make above changes and we are good to go!

liamzebedee:
👍 take a peek!

jacqueswww:
@liamzebedee sorry for only getting back to you now, slipped past my radar. I think it's ready to merge. Thanks :)

jacqueswww:
@liamzebedee do you have a payout address set in openbounty.status.im. There is a bounty on this issue, and I want to pay it out :smile:

liamzebedee:
@jacqueswww sure am :)

pablanopete:
@liamzebedee sorry for the slow response on bounty payout, could you drop your wallet address here so I can send you the bounty amount Out of Band while we wait on a dashboard fix in Status Open Bounty.

If you don't feel comfortable putting it here in the comment could you email a ERC20 compatible wallets address to me? chris.hutchinson@status.im

liamzebedee:
Cheers @pablanopete :)

0xf87de2add59a3c23ab6365f8addfc5db9c86b3b9

pablanopete:
Sent thanks! https://etherscan.io/tx/0xaecb732bfeaaa32adcf080038aafc5d8c60e11320ec520b102473eee74286833

PR #908, opened 2018-06-18
Add builtin constants

### LINKED ISSUE #843: VIP: Add constants for max in min values
## Preamble

    VIP: #843
    Title: Constants for max in min values
    Author: Jacques Wagener
    Type: Standard Track
    Status: Draft
    Created: <date created on, in ISO 8601 (yyyy-mm-dd) format>

## Simple Summary
Add 6 commonly used constants.

## Abstract
Adding some commonly used constants will improve the experience of porting over existing solidity contracts.

## Motivation
In most cases one should not actually have to use MIN/MAX constants when programming vyper. But there seem to be legacy cases were it's necessary (when MAX values are used for flags).

## Specification
Because we store the min max values for clamps in any case. Just mload the necessary values as appropriate.

```
ZERO_ADDRESS = 0x0000000000000000000000000000000000000000
MAX_INT128 = 2**127 - 1
MIN_INT128 = -2**127
MAX_DECIMAL = (2**127 - 1) * DECIMAL_DIVISOR
MIN_DECIMAL = (-2**127) * DECIMAL_DIVISOR
MAX_UINT256 = 2**256 - 1
```

## Backwards Compatibility
Forward compatible, as no old annotations will be altered.

## Copyright

jacqueswww:
### - What I did
Fixes #843. 
Adding bultin constants.

### DIFF
--- a/docs/types.rst
+++ b/docs/types.rst
@@ -255,7 +255,7 @@ Keyword        Unit   Base type  Description
     Two ``timedelta`` can be added together, as can a ``timedelta`` and a ``timestamp``, but not two ``timestamps``.
 
 ===================  ===========  =========  ====================================================================================
-Wei 
+Wei
 ---------------------------------------------------------------------------------------------------------------------------------
 Keyword              Unit         Base type  Description
 ===================  ===========  =========  ====================================================================================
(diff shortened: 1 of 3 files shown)


docs/types.rst
```diff
@@ -421,6 +421,23 @@ Here ``_KeyType`` can be almost any type except for mappings, a contract, or a s
...
+**********
+Builtin Constants
+**********
+
+Vyper has a few convenience constants builtin.
+
+======= ============ ==========================================
+Type    Name         Value
+======= ============ ==========================================
+address ZERO_ADDRESS 0x0000000000000000000000000000000000000000
```
maurelian: I noticed that Python's built in constants use CamelCase:
https://docs.python.org/3/library/constants.html

jacqueswww: We discussed this some more, and decided to leave as is.

The reason for this is because we want to reserve CamelCase names for contract interface names.

maurelian: interface names are `camelCase`, not `CamelCase`. My fault for not being more explicit about the leading capitalization.

jacqueswww: For now we will just go with the current "capital snake case". As this is not an uncommon convention for a python programmer ;)

PR #923, opened 2018-06-22
Allow docblocks to exist on functions.

### LINKED ISSUE #819: Address Literals should be force to be checksummed.
### What's your issue about?
As there is currently not a use case were one would *not* need an address to use upper/lower case checksum. We should enforce this.

Shout out to @pipermerriam ;)
### How can it be fixed?

See the `number` method.

jacqueswww:
### - What I did
Fixes #819 .

### DIFF
--- a/vyper/parser/parser.py
+++ b/vyper/parser/parser.py
@@ -573,8 +573,7 @@ def parse_body(code, context):
     o = []
     for stmt in code:
         lll = parse_stmt(stmt, context)
-        if lll:
-            o.append(lll)
+        o.append(lll)
     return LLLnode.from_list(['seq'] + o, pos=getpos(code[0]) if code else None)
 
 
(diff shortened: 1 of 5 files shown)


fubuloubu:
Woo!

vyper/parser/parser.py
```diff
@@ -572,7 +572,9 @@ def parse_body(code, context):
         return parse_stmt(code, context)
     o = []
     for stmt in code:
-        o.append(parse_stmt(stmt, context))
+        lll = parse_stmt(stmt, context)
+        if lll:
```
pipermerriam: More of a higher level suggestion.  Return value checking is error prone (easy to forget to do).  You might look for an alternate pattern using something like `try/except`.

jacqueswww: agreed

fubuloubu:
Woo!

PR #925, opened 2018-06-23
Update crowdfund.vy example contract

davesque:
### What I did

Improved whitespace in `crowdfund.vy` example contract.

### How I did it

Did it.

### How to verify it

Changes are only whitespace.

### Cute Animal Picture

Cute animal picture

### DIFF
--- a/examples/crowdfund.vy
+++ b/examples/crowdfund.vy
@@ -1,4 +1,4 @@
-#Setup private variables (only callable from within the contract)
+# Setup private variables (only callable from within the contract)
 funders: {sender: address, value: wei_value}[int128]
 nextFunderIndex: int128
 beneficiary: address
@@ -6,6 +6,7 @@ deadline: timestamp
 goal: wei_value
 refundIndex: int128
 timelimit: timedelta
+
 
 # Setup global variables
 @public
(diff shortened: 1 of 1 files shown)


davesque:
Hmm...it seems quite a shame to let Solidity dictate the conventions with so many things.  It's pretty straight forward to convert between snake casing and camel casing.  I wonder if it isn't reasonable to just do so even for standardized APIs.  There shouldn't ever be any naming conflicts since Solidity doesn't use snake casing at all.

It's also a bit confusing that every other example contract uses snake casing except for the ERC20 ones since they're standardized as you mention.

fubuloubu:
The problem is that `sha256("transferFrom(address,address,uint256)")` is `4b6685e7...` and `sha256("transfer_from(address,address,uint256)")` is `4c1fd07f4...`, and this signature is what the standard is enforcing so any client can call this code through the ABI without being aware of the implementation of the code. Snake case vs. Camel case

jacqueswww:
+1 for the whitespace change, I think this must also be in the forced lint checker ;)

fubuloubu:
I think camel case in general, otherwise styling is inconsistent. We could also do the underscore thing for private, and double underscore for internal?

davesque:
@fubuloubu Oops, of course.  I forgot about the method identifiers.

I agree that, if camelCasing is the style, then it should be consistent across the board.

But I don't really like the idea of underscores for private methods, etc.  If we can't innovate on code style, then it seems like we should just stick to what Solidity does to avoid confusion.

jacqueswww:
Agreed, I think this is perfect hash-out for having all method be camelcase.

davesque:
Just for my information, why are there so many other example contracts that use snake casing?  It's not clear to me why there's a difference there.

davesque:
I updated this branch by the way to only include those whitespace changes.

jacqueswww:
> Just for my information, why are there so many other example contracts that use snake casing? It's not clear to me why there's a difference there.

Pretty much just happened this way, most of the contract were written this way - because we all come from python ;)
Only later did we realise all the ERC require camelcase, and now we will work to make it part of our yet to be created style guide.

fubuloubu:
Having the underscores would be useful IMO, picture this:

```python
def __myInternalMethod():
    pass
#... Lots of code...
def _myPrivateMethod():
    pass
#... Lots of code...
def myPublicMethod():
    self._myPrivateMethod() # This is private!
    self.__myInternalMethod() # This is internal!
```

It becomes very obvious to me what is public vs. private vs. internal.

However, this should be a style guide suggestion, not a requirement.

PR #930, opened 2018-06-24
Remove preamble. We don't use it.

fubuloubu:
### - What I did
Removed the preamble

### - Why I did it
We don't use the preamble for anything right now.
This style was copied from the EIPs, which auto-build a webpage mirroring the current EIPs. We don't do that right now. I don't think we really need most of what's in the preamble.

### - Cute Animal Picture
tartan rabbit

### DIFF
--- a/.github/ISSUE_TEMPLATE/vip.md
+++ b/.github/ISSUE_TEMPLATE/vip.md
@@ -2,18 +2,6 @@
 name: Vyper Improvement Proposal (VIP)
 about: This is the suggested template for new VIPs.
 ---
-## Preamble
-
-    VIP: <to be assigned>
-    Title: <VIP title>
-    Author: <list of authors' names and optionally, email addresses>
-    Type: <Standard Track | Informational | Meta>
-    Status: Draft
-    Created: <date created on, in ISO 8601 (yyyy-mm-dd) format>
-    Requires (*optional): <VIP number(s)>
-    Replaces (*optional): <VIP number(s)>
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/vip.md
```diff
@@ -2,18 +2,6 @@
...
-## Preamble
-
-    VIP: <to be assigned>
-    Title: <VIP title>
-    Author: <list of authors' names and optionally, email addresses>
-    Type: <Standard Track | Informational | Meta>
-    Status: Draft
-    Created: <date created on, in ISO 8601 (yyyy-mm-dd) format>
-    Requires (*optional): <VIP number(s)>
-    Replaces (*optional): <VIP number(s)>
```
jacqueswww: Perhaps we should keep VIP#, Title, Created?

jacqueswww: Well at least Auther & Created, which is useful to. Title is obvious.

fubuloubu: VIP # is issue number, Auto generated.
Author is original submitter. Maybe a quick blurg about co-authors or original submitters.
Created is also in the issue metadata.

We don't use this preamble for anything outside of GitHub.

jacqueswww: Hmm... fair enough. Drop it like it's hot.

PR #939, opened 2018-07-06
Unicode in identifiers

### LINKED ISSUE #936: Unicode characters are allowed to be a part of identifiers
### Version Information

* vyper Version: Master
* pyethereum Version: x.x.x
* OS: osx/linux/win
* Python Version (python --version):

### What's your issue about?

In Vyper, Unicode characters are allowed to be a part of identifiers (variable names, method names, etc.). This could lead to [Homograph attacks](https://en.wikipedia.org/wiki/IDN_homograph_attack) where a piece of code appears to be acting on one variable but is actually acting on some other variable. Unicode characters should probably be disallowed from being a part of identifiers, especially since Vyper aims to be a language that is easy to audit.

---

FWIW, Solidity throws an error when I try to name a variable with a unicode character.

Code:

```solidity
pragma solidity ^0.4.21;

contract Foo {
    
    function x(int128 i) public view {
        uint256 Other_Balance = 3; // ASCII O
        
        // a bunch of code
        if (i>0) {
            uint256 Оther_Balance = 2; // Unicode О
        }
        
        // a bunch of code
        return Other_Balance; // ASCII O
    }
}
```

Error:
```
browser/Untitled5.sol:10:21: ParserError: Expected ';' but got 'ILLEGAL'
            uint256 Оther_Balance = 2; // Unicode О
                    ^
```

---

Similar code in Vyper, which compiles without throwing any errors:

```python
@public
def foО(i: int128) -> uint256:
    Other_Balance: uint256 = 3 # ASCII O
    
    # a whole bunch of code
    if i > 0:
        Оther_Balance: uint256 = 2 # Unicode О
    
    # a whole bunch of code

    return Other_Balance # ASCII O
```

---

### How can it be fixed?

Disallow Unicode characters in identifiers.

#### Cute Animal Picture

Baby Elephant

sushantkumr:
### - What I did

Disallowed Unicode characters in variables, function names and event names.
Closes #936. 

### - How I did it

* Using regex, added a check to allow only lowercase, uppercase alphabets, numbers and underscore. This check has been applied on variables, function names and event names.

### - How to verify it

Contract fails to compile if any identifier contains Unicode characters.

Unicode characters no longer allowed in identifiers.

### - Cute Animal Picture
piglet_dog


sushantkumr:
@fubuloubu I ve made an attempt to fix #936. Please review the changes.

Just wanted to point out a few things:

* Function and Event names did not have checks.
* Existing tests under test_crowdfund.py and test_uint256.py failed after running 'make test' because the function names were OPCODES in lowercase. Those have been fixed.
* utils.py did not have a tests. That has been created.

vyper/utils.py
```diff
@@ -166,4 +168,6 @@ def is_varname_valid(varname, custom_units):
         return False
     if varname.lower() in built_in_functions:
         return False
+    if not re.match('^[_a-z][a-zA-Z0-9_]*$', varname.lower()):
```
rohithpr: The regex should change to `re.match('^[_a-z][a-z0-9_]*$', varname.lower()` or `re.match('^[_a-zA-Z][_a-zA-Z0-9]*$', varname`. If there's a `.lower()` then there's no need to include `A-Z` in the regex. `A-Z` should be allowed as the first character of the identifier if `.lower()` is not being used.

Both the options look fine to me I recommend dropping `.lower()` and allowing `A-Z` to be the first char.

tests/parser/syntax/test_utils.py
```diff
@@ -0,0 +1,89 @@
...
+    """
+@public
+def foo(i: int128) -> int128:
+    ő1qwerty : int128 = i
+    return ő1qwerty
+    """,
+    """
+@public
+def foo(i: int128) -> int128:
+    fő1234 : int128 = i
```
rohithpr: the first four tests all check the same condition anyway, you can get rid of three of them!

vyper/exceptions.py
```diff
@@ -70,3 +70,11 @@ class InvalidTypeException(ParserException):
...
 class TypeMismatchException(ParserException):
     pass
+
+
+class FunctionDeclarationException(ParserException):
+    pass
+
+
+class EventDeclarationException(ParserException):
+    pass
```
jacqueswww: So was about to merge, but then occurred to me that if we introduce these two exceptions we have to use them more than just in the added unicode checks.
So please take the other VariableDeclarationException or similar exceptions that occur on on function and event declaration and use the above two exceptions. This way the use of the exceptions are consistent.

sushantkumr:
@jacqueswww i have made the changes you requested. Kindly review.

PR #958, opened 2018-07-18
Functions returning lists inside argument tuple's

### LINKED ISSUE #955: Functions, Exception: ("Can't return type %s as part of tuple", <class 'vyper.types.types.ListType'>)
I'm working on a sample game to demonstrate some proof of concepts. Including, showing both versions of the contract (Solidity & Vyper) so i can show a comparison of building and usage. 

### What's your issue about?
I'm trying to convert a solidity contract into Vyper. 

Can't seem to return the same bag of values. 

For the following vyper function : 

```
@public
@constant
def getStoryNode(id: uint256) -> (uint256, bytes[256], bytes[256], address, uint256[256]):
    return (self.stories_nodes[id].id, self.stories_nodes[id].title, self.stories_nodes[id].body, self.stories_nodes[id].owner, self.stories_nodes[id].childNodesIds)
```

I get the following error when compiling the contract: 
`Exception: ("Can't return type %s as part of tuple", <class 'vyper.types.types.ListType'>)`

Solidity Version : 
```
  /**
  * Returns node's id, title, body, owner's address
  */
  function getStoryNode(uint id) public view returns(uint, string, string, address, uint[]) {
      StoryNode storage node = storiesNodes[id];
      return (node.id, node.title, node.body, node.owner, node.childNodesIds);
  }
```

### How can it be fixed?

Can stmt.py::parse_return be improved to handle this? Maybe i can do do this myself and submit PR. 

### Extra related questions:

Unless i missed something in the docs... strings & dynamically sized mappings aren't a thing, right? 

#### Cute Animal Picture

lmgsz8bj28a11

ehanoc:
### - What I did
Trying to fix #955 by myself as i need support for a project of mine.  

### - How I did it
A look through how statements are built for everything else. A bit of research of the state machine instructions and LLL code. 

Added to stmt.py the check for ListType for when building the instructions of Tuple's arguments

### - How to verify it
First wrote a test to trigger the problem. 

Added as types/test_function_returns.py

Functions bag of values can now hold inner lists

### - Cute Animal Picture

u4i5nl3bloa11

### DIFF
--- a/tests/parser/types/test_function_returns.py
+++ b/tests/parser/types/test_function_returns.py
@@ -1,0 +1,13 @@
+def test_function_return_list_in_tuple(get_contract_with_gas_estimation):
+    list_tester_code = """
+@public
+def foo() -> (uint256, uint256[2]):
+    return (0, [0, 1])
+    """
+
+    c = get_contract_with_gas_estimation(list_tester_code)
+    assert isinstance(c.foo()[1], list)
+    assert len(c.foo()[1]) == 2
+    assert c.foo()[0] == 0
+    assert c.foo()[1].sort() == [0, 1].sort()
+    print("Passed functions return as tuple")
--- a/vyper/parser/stmt.py
+++ b/vyper/parser/stmt.py
@@ -534,7 +534,7 @@ class Stmt(object):
                     # list lefthand side types
                     sub_list_types = LLLnode.from_list(self.context.new_placeholder(arg.typ), typ=arg.typ, location='memory')
                     # left k & right values
-                    setter = make_setter(sub_list_types, arg, location = 'memory', pos=getpos(self.stmt))
+                    setter = make_setter(sub_list_types, arg, location='memory', pos=getpos(self.stmt))
                     # build inner list argument
                     list_arg = LLLnode.from_list(['seq', setter, ['return', sub_list_types, get_size_of_type(self.context.return_type) * 32]], typ=None, pos=getpos(self.stmt))
                     # append to tuples list args
--- a/vyper/parser/stmt.py
+++ b/vyper/parser/stmt.py
@@ -530,6 +530,15 @@ class Stmt(object):
 
                 elif isinstance(arg.typ, BaseType):
                     subs.append(make_setter(variable_offset, arg, "memory", pos=getpos(self.stmt)))
+                elif isinstance(arg.typ, ListType):
+                    # list lefthand side types
+                    sub_list_types = LLLnode.from_list(self.context.new_placeholder(arg.typ), typ=arg.typ, location='memory')
+                    # left k & right values
+                    setter = make_setter(sub_list_types, arg, location = 'memory', pos=getpos(self.stmt))
+                    # build inner list argument
+                    list_arg = LLLnode.from_list(['seq', setter, ['return', sub_list_types, get_size_of_type(self.context.return_type) * 32]], typ=None, pos=getpos(self.stmt))
+                    # append to tuples list args
+                    subs.append(list_arg)
                 else:
                     raise Exception("Can't return type %s as part of tuple", type(arg.typ))
 


tests/parser/types/test_function_returns.py
```diff
@@ -0,0 +1,13 @@
+def test_function_return_list_in_tuple(get_contract_with_gas_estimation):
+    list_tester_code = """
+@public
+def foo() -> (uint256, uint256[2]):
+    return (0, [0, 1])
```
jacqueswww: Please add tests for these type of cases as well:

```
def foo() -> ( bytes[10], uint256, uint256[2]) # dynamic mixed with static.
def foo() -> (int128, int128[10], uint256)  # mixed base types
```

jacqueswww: We will also require test list-in-tuple returns for variables from memory and storage, such as:

```
a: int128[10]
...
return (self.a, 1)
```
and
```
def foo() ...
   a: int128[3] = [1, 2, 3]
   return (a, 1)
```

tests/parser/types/test_function_returns.py
```diff
@@ -0,0 +1,13 @@
...
+    """
+
+    c = get_contract_with_gas_estimation(list_tester_code)
+    assert isinstance(c.foo()[1], list)
+    assert len(c.foo()[1]) == 2
+    assert c.foo()[0] == 0
+    assert c.foo()[1].sort() == [0, 1].sort()
```
jacqueswww: a single assert of == [0, [0, 1]] is more than enough ;)

tests/parser/types/test_function_returns.py
ehanoc: I've added the mixed base types and it works fine it seems.

Although when including bytes[10], i get something as follows : 

``` web3.exceptions.BadFunctionCallOutput: Could not decode contract function call zoo return data b'\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x01\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x01\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x02\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\nhellohello\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00' for output_types ['bytes', 'uint256', 'uint256[2]']```

I don't understand what's going on here. Shouldn't the output_types match with ['bytes[10],  'uint256', 'uint256[2]']] ?

tests/parser/types/test_function_returns.py
ehanoc: Actually, that reveals that inner lists returned are reversed. :open_mouth:  Any ideas of what might cause that?

jacqueswww: I think this is probably a misalignment issue. Try with clearer values like `[5, 6]`.

tests/parser/types/test_function_returns.py
jacqueswww: This probably has to do with dynamic offset not being incremented (and might be related to the bottom issue)?
If you don't figure it out push your latest commit and I can take a look at things.

ehanoc: Hey man, havent had the time yet. New job... Feel free to pull my branch. But ill try to take a look at it in the next week though

jacqueswww: Sure no problem, will pick it up ;)

PR #962, opened 2018-07-24
Better constant folding

### LINKED ISSUE #844: Improve literal min/max checking.
### What's your issue about?

From gitter:

We do fold the constants in python. That's why 2**256 - 1 throws out of range error
basically the AST parses 2 ** 256 to POW(2, 256), then we pick it up becomes a new literal. And then that new literal fails - because that literal is out of bounds as per our definition.
The problem I had was, I couldn't figure out how to know - this is the final stage literal.
But I assume now that I think about it some more, I could just move those limit checks to the assignment/comparison operator cases.
basically literals can be huge, but the moment they get used we do the checks.

### How can it be fixed?

jacqueswww:
### - What I did

Fixes #844.


tests/conftest.py
```diff
@@ -42,7 +42,7 @@ def __prepared_function(self, *args, **kwargs):
         if not kwargs:
             modifier, modifier_dict = 'call', {}
             fn_abi = [x for x in self._function.contract_abi if x['name'] == self._function.function_identifier].pop()
-            modifier_dict.update({'gas': fn_abi['gas'] * 1000})  # To make tests faster just supply some high gas value.
```
fubuloubu: How does this make tests faster? Genuinely curious

jacqueswww: If you supply a gas value, web3 doesn't run an estimateGas. estimateGas being very expensive as it does the the same call over and over to to estimate the gas.

fubuloubu: Oh, that makes complete sense! I wonder if we could just set it to the block gas limit since it automines and there's no reason for a transaction to cost more than 6 or 7 million gas (or whatever it is now)

fubuloubu: And then check for the actual execution amount after it runs

Issue #965, opened 2018-07-27
Optimize multiplication operation on Vyper


hskang9:
## Simple Summary
Optimize multiplication check as recent openzeppelin safemath [implementation](https://github.com/OpenZeppelin/openzeppelin-solidity/pull/522).

## Abstract
One of the things I like Vyper is that it supports safemath from the language.
Its logic is on [expr.py](https://github.com/ethereum/vyper/blob/master/vyper/parser/expr.py).
However, I found a typo in the comment:
from
``` # Checks that: a == 0 || a / b == b``` 
to
``` # Checks that: a == 0 || c(which is a*b) / a == b```

 and a space for gas optimization which was approved in Openzeppelin's SafeMath [implementation](https://github.com/OpenZeppelin/openzeppelin-solidity/blob/master/contracts/math/SafeMath.sol).

## Motivation
If this change is not approved, comment typo is not fixed and vyper uses more gas than solidity with safemath contract.

## Specification
Here is the difference of Opcode to be pushed to LLL parse tree:
### Before
assert(a==0 || (a*b / a) == b)
```
['assert', ['or', ['iszero', left],
['eq', ['div', ['mul', left, right], left], right]]],
['mul', left, right]
```

### After

if (a == 0) { return 0; } // Without verifying a\*b / a == b
assert(a*b/a==b);
```
[  'if', 
   ['eq', left, 0],
   [0], 
  ['seq',
     ['assert', ['eq', ['div', ['mul', left, right], left], right]],
     ['mul', left, right]
  ]
]

```

Reasons and verification is described on the link [here](https://github.com/OpenZeppelin/openzeppelin-solidity/pull/522)

## Backwards Compatibility
This version is compatible with the others but it uses less gas for an arithmetic operation.

## Copyright
Copyright and related rights waived via [CC0](https://creativecommons.org/publicdomain/zero/1.0/)

jacqueswww:
Wow good catch picking that up from SafeMath :smile:

fubuloubu:
I'm missing something about this comment in context of the code. How does checking if and only if `a == 0` in order to proceed with any multiplication help?

hskang9:
Oops. I passed out 3 hours ago and checked this. I will implement it right.

jacqueswww:
@fubuloubu so basically you save gas because you immediatly return 0 for `0/x` instead of doing all the checks and the math and return 0 anyways.

fubuloubu:
Okay, that makes way more sense now lol

PR #966, opened 2018-07-27
Optimize multiplication operation on Vyper

hskang9:
Applies VIP #965

### DIFF
--- a/vyper/parser/expr.py
+++ b/vyper/parser/expr.py
@@ -330,9 +330,10 @@ class Expr(object):
             new_unit = combine_units(left.typ.unit, right.typ.unit)
             if ltyp == rtyp == 'uint256':
                 o = LLLnode.from_list(['seq',
-                                        # Checks that: a == 0 || a / b == b
-                                        ['assert', ['or', ['iszero', left],
-                                        ['eq', ['div', ['mul', left, right], left], right]]],
+                                        # Checks that: a == 0
+                                        # Gas optimization: this is cheaper than asserting 'a' not being zero, but the
+                                        # benefit is lost if 'b' is also tested.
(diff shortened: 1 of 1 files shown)


vyper/parser/expr.py
```diff
@@ -330,9 +330,10 @@ def arithmetic(self):
...
             if ltyp == rtyp == 'uint256':
                 o = LLLnode.from_list(['seq',
-                                        # Checks that: a == 0 || a / b == b
-                                        ['assert', ['or', ['iszero', left],
-                                        ['eq', ['div', ['mul', left, right], left], right]]],
+                                        # Checks that: a == 0
+                                        # Gas optimization: this is cheaper than asserting 'a' not being zero, but the
+                                        # benefit is lost if 'b' is also tested.
+                                        ['assert', ['iszero', left]],
                                         ['mul', left, right]], typ=BaseType('uint256', new_unit), pos=getpos(self.expr))
```
jacqueswww: Looking at the solidity contract:
```
    if (a == 0) {
      return 0;
    }

    c = a * b;
    assert(c / a == b);
    return c
```
We are only left with `'assert', ['iszero', left]],` .
But the second assert with c / a * b still needs to occur?

jacqueswww: That's also why the CI keeps failing.

jacqueswww: If you want to emulate the the return 0 it would be, something like: 
```
['if', 
   ['eq', left, 0], 
   [0],
   ['seq'
       ['assert', ['eq', ['div', ['mul', left, right], left], right] ]],
       ['mul', left, right]]
]
```

hskang9: Thank you @jacqueswww I will test it and implement it right away.

PR #968, opened 2018-07-27
Fix multiple args.

jacqueswww:
### - What I did
When implementing multiple args, we didn't realise that the old use case of `vyper -f abi inputfile.vy` wasn't working anymore.

This changes multiple args to us comma separate list instead. Because it seems like argparse as absolutely no way of handling `-f abi -f json`.


fubuloubu:
Nice. Comma separated list makes the most sense

PR #974, opened 2018-07-31
test vyper with web3 v4.5.0

voith:
### - What I did
bumped web3 version to v4.5.0

### - How I did it
updated `setup.py` with the new web3 version

### - How to verify it
check if CI turns green.


jacqueswww:
Thanks! I want to try a bit with getting `eth-tester==0.1.0-beta.29` working now too. Will get back to you on that here :smile:

PR #977, opened 2018-08-01
Add vyper-lll.

### LINKED ISSUE #790: Expose LLL IR
The current compiler `v0.0.4` doesn't expose the LLL IR as it's own target. An implication of this is that you can't author LLL and then generate bytecode in an easy manner.

It would be ideal to eventually expose the LLL as a separate module that the vyper frontend (and any other frontend) could consume.

jacqueswww:
### - What I did
Fixes #790. 
Very basic for now, see test_lll_from_s_expression for an example.

Couldn't resist, a lisp joke.

### DIFF
--- a/bin/vyper-lll
+++ b/bin/vyper-lll
@@ -1,0 +1,39 @@
+#!/usr/bin/env python3
+
+import argparse
+import vyper
+
+from vyper import (
+    optimizer
+)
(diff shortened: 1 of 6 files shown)


jacqueswww:
@ralexstokes have look please :smile_cat:

ralexstokes:
looks good to me :)

PR #978, opened 2018-08-02
LLL: represent ints as hex

jacqueswww:
### - What I did
Using hex is more readable and useful than large int values:

```python
[seq,
  [return,
    0x0,
    [lll,
      [seq,
        [mstore, 0x1c, [calldataload, 0x0]],
        [mstore, 0x20, 0x10000000000000000000000000000000000000000],
        [mstore, 0x40, 0x7fffffffffffffffffffffffffffffff],
        [mstore, 0x60, -0x80000000000000000000000000000000],
        [mstore, 0x80, 0x12a05f1fffffffffffffffffffffffffdabf41c00],
        [mstore, 0xa0, -0x12a05f20000000000000000000000000000000000],
        # Line 2
        [if,
          [eq, [mload, 0x0], '0x455e01f8' <puliccc>],
          [seq, [assert, [iszero, callvalue]], stop]],
        /* Default function */ [revert, 0x0, 0x0]],
      0x0]]]
```


fubuloubu:
:+1: Hex, much more readable output

PR #979, opened 2018-08-04
added version option in command line

MohamedLEGH:
### - What I did
The command line vyper now can print version number with  
vyper --version
### - How I did it
https://docs.python.org/3/library/argparse.html

`parser.add_argument('--version', action='version', version='%(prog)s 2.0')`

### - How to verify it
vyper --version

I did "make test" and all tests were passed (96% coverage)
 
command line print version number
### - Cute Animal Picture
image

### DIFF
--- a/bin/vyper
+++ b/bin/vyper
@@ -11,7 +11,8 @@ from vyper import compile_lll
 
 sys.tracebacklimit = 0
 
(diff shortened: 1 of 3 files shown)


MohamedLEGH:
Added vyper-lll --version and vyper-serve --version

PR #987, opened 2018-08-09
Default parameters support

### LINKED ISSUE #903: Allow default parameters
This is a reduced scope feature as compared to adding Function Overloading #884.

# Introduction

ERC-721 includes two functions with the same name but one includes an extra parameter. The ERC specification states that the less parameter one will function equivalently as if the more parameter one were called with a specific value.

# Specification

Allow default parameters at function declaration.

Here is a very relevant example. Includes NatSpec documentation as per #898.

*Sorry, not sure if my bytes declaration is correct.*

```py
## 
# @notice Transfers the ownership of an NFT from one address to another address
# @dev Throws unless `msg.sender` is the current owner, an authorized
#  operator, or the approved address for this NFT. Throws if `_from` is
#  not the current owner. Throws if `_to` is the zero address. Throws if
#  `_tokenId` is not a valid NFT. When transfer is complete, this function
#  checks if `_to` is a smart contract (code size > 0). If so, it calls
#  `onERC721Received` on `_to` and throws if the return value is not
#  `bytes4(keccak256("onERC721Received(address,uint256,bytes)"))`.
# @param _from The current owner of the NFT
# @param _to The new owner
# @param _tokenId The NFT to transfer
# @param data Additional data with no specified format, sent in call to `_to`
@public
@payable
def safeTransferFrom(from: address, to: address, token_id: uint256, data: bytes = ""):
    ...
```

The Vyper compiler will process this as two separate function selectors. Actually (1 + number of default parameters) function selectors. This follows the Python convention of allowing parameters to be omitted if a default parameter is specified. This differs from Python in that Python function arguments can be specified out of order.

The Vyper compiler can optimize byte code more efficiently than Solidity currently does because Vyper will have more knowledge about the relationship between `safeTransferFrom(a,b,c,d)` and `safeTransferFrom(a,b,c)`.

# Literature review

Vyper has [expressed opinions](https://github.com/ethereum/vyper/blob/master/README.md) on function overloading in general:

* This can cause lots of confusion on which function is called at any given time.
* It makes the code much harder to search through as you have to keep track on which call refers to which function.
* Thus it's easier to write missleading code (foo("hello") logs "hello" but foo("hello", "world") steals your funds).

Strictly speaking, a default parameter is a function overload. From the EWM there are two function descriptors, but from Vyper, there is one function. I believe the opinions above apply to allowing arbitrary function overloads, whereas this issue is only a very specific kind of overload.

With this proposed feature, we centralize implementation for both code paths in one function. This solves the issue of tracking which call refers to which function, they are the same function. Also this implementation is much safer than Solidity. In Solidity, as this can cause confusion. The foo() logs but foo(param) steals funds example above does not apply because BOTH are the same function.

# Discussion

Honestly I don't feel like this is a compromise. I believe this actively promotes Vyper's goal to improve human-readability of code and prevent misleading code.

Pinging people from function overloading discussion: @maurelian @fubuloubu @jacqueswww @ben-kaufman

jacqueswww:
### - What I did
Fixes #903.

### - How I did it
Lots of wrestling, and changing of structs.

### - How to verify it
Check the tests.

Add support for default parameters.

### DIFF
--- a/bin/vyper
+++ b/bin/vyper
@@ -9,7 +9,7 @@ from vyper.parser.parser import parse_to_lll
 from vyper.parser import parser_utils
 from vyper import compile_lll
 
-sys.tracebacklimit = 0
+sys.tracebacklimit = 10
 
 parser = argparse.ArgumentParser(description='Vyper {0} programming language for Ethereum'.format(vyper.__version__))
 parser.add_argument('input_file', help='Vyper sourcecode to compile')
--- a/bin/vyper
+++ b/bin/vyper
@@ -1,6 +1,7 @@
 #!/usr/bin/env python3
 import argparse
 import json
+import os
 import sys
 import vyper
 
@@ -9,7 +10,7 @@ from vyper.parser.parser import parse_to_lll
 from vyper.parser import parser_utils
 from vyper import compile_lll
 
-sys.tracebacklimit = 10
+sys.tracebacklimit = os.environ.get('VYPER_TRACEBACK_LIMIT', 0)
 
 parser = argparse.ArgumentParser(description='Vyper {0} programming language for Ethereum'.format(vyper.__version__))
 parser.add_argument('input_file', help='Vyper sourcecode to compile')
--- a/tests/parser/functions/test_default_function.py
+++ b/tests/parser/functions/test_default_function.py
@@ -31,6 +31,27 @@ def __default__():
     assert w3.eth.getBalance(c.address) == w3.toWei(0.1, 'ether')
 
 
+def test_basic_default_default_param_function(w3, get_logs, get_contract_with_gas_estimation):
+    code = """
+Sent: event({sender: indexed(address)})
+@public
+@payable
+def fooBar(a: int128 = 12345) -> int128:
+    log.Sent(ZERO_ADDRESS)
+    return a
+
+@public
+@payable
+def __default__():
+    log.Sent(msg.sender)
+    """
+    c = get_contract_with_gas_estimation(code)
+
+    logs = get_logs(w3.eth.sendTransaction({'to': c.address, 'value': 10**17}), c, 'Sent')
+    assert w3.eth.accounts[0] == logs[0].args.sender
+    assert w3.eth.getBalance(c.address) == w3.toWei(0.1, 'ether')
+
+
 def test_basic_default_not_payable(w3, assert_tx_failed, get_contract_with_gas_estimation):
     code = """
 Sent: event({sender: indexed(address)})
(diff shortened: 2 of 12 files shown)


bin/vyper
```diff
@@ -9,7 +9,7 @@ from vyper.parser.parser import parse_to_lll
 from vyper.parser import parser_utils
 from vyper import compile_lll
 
-sys.tracebacklimit = 0
+sys.tracebacklimit = 10
```
fubuloubu: Why this change?

tests/parser/functions/test_default_function.py
```diff
@@ -31,6 +31,27 @@ def __default__():
...
+@public
+@payable
+def __default__():
+    log.Sent(msg.sender)
+    """
+    c = get_contract_with_gas_estimation(code)
+
+    logs = get_logs(w3.eth.sendTransaction({'to': c.address, 'value': 10**17}), c, 'Sent')
+    assert w3.eth.accounts[0] == logs[0].args.sender
+    assert w3.eth.getBalance(c.address) == w3.toWei(0.1, 'ether')
```
fubuloubu: It might be a little clearer here to rewrite:
```python
amount = w3.toWei(0.1, 'ether')
logs = get_logs(w3.eth.sendTransaction({'to': c.address, 'value': amount}), c, 'Sent')
assert w3.eth.accounts[0] == logs[0].args.sender
assert w3.eth.getBalance(c.address) == amount
```

tests/parser/functions/test_default_parameters.py
```diff
@@ -0,0 +1,158 @@
...
+@public
+def fooBar(a: bytes[100], b: int128, c: bytes[100] = "testing", d: uint256 = 999) -> (bytes[100], int128, bytes[100], uint256):
+    return a, b, c, d
+    """
+    c = get_contract(code)
+    c_default = b"testing"
+    d_default = 999
+
+    # c set, 7d default value
+    assert c.fooBar(b"booo", 12321, b'woo') == [b"booo", 12321, b'woo', d_default]
```
fubuloubu: `web3` returns a list instead of a tuple?

jacqueswww: Yup it does

tests/parser/functions/test_default_function.py
jacqueswww: sure

bin/vyper
jacqueswww: slipped through, I think I will make this so you can set it with an env variable.

tests/parser/functions/test_default_parameters.py
fubuloubu: This seems like a bug with web3 then...

PR #991, opened 2018-08-10
added default function test with send/call

### LINKED ISSUE #899: Default function test with send
### What's your issue about?

- send (CALL opcode with 0 gas value) has a  3200 gas limit.
- Write a test case for using send to a default function, showing more gas can be sent if using `call`.

#### Cute Animal Picture
^.^

evgeniuz:
### - What I did
Implemented test that calls default function with `send` and `raw_call` and tests behavior of gas usage.

Fixes #899.

### - How I did it
`send` won't revert transaction in case it fails, so I've tested that value transfer did not happen and default function wasn't executed (variable was not changed).

### - How to verify it
Run tests.


evgeniuz:
Not sure if I can do something about CI failing.

jacqueswww:
Yes seems so, very strange dependency error related to https://github.com/erikrose/parsimonious/issues/137, somehow.

jacqueswww:
@evgeniuz  Looks good, please just fix the one pep8 error:

tests/parser/functions/test_send.py:56:35: E711 comparison to None should be 'if cond is None:'

evgeniuz:
Fixed PEP8 error.

ceresstation:
Thanks @evgeniuz, paying this out now!

jacqueswww:
Awesome, thanks :)

PR #1012, opened 2018-09-06
Less gas for private functions (WIP)

### LINKED ISSUE #901: VIP: reduce the cost and difficulty of code reuse in common situations
## Preamble

    VIP: <to be assigned>
    Title:
    Author: <list of authors' names and optionally, email addresses>
    Type: <Standard Track | Informational | Meta>
    Status: Draft
    Created: <date created on, in ISO 8601 (yyyy-mm-dd) format>

## Simple Summary

1. I shouldn't need to copy and paste the same code snippets into 5 functions to valid authorization. 
2. I shouldn't feel like copying and pasting code is more efficient or cheaper. 

## Abstract

In order to call a within the same contract function, Vyper uses the `CALL` opcode, which send a new message to call functions within the same contract. This has some nice safety benefits. 

Unfortunately it will result in developers using anti-patterns to save on gas, or access `msg.sender` in a function call.

## Motivation

The benefit of using `CALL` to access code in the same contract is to create a new execution context, with no risk of side effects from memory access. I admit that I find this to be quite an elegant use of the EVM for safety.

### Problems with using `CALL`

Unfortunately, there are two major issues with the use of the `CALL` opcode for calling functions within the contract:

#### 1. Environment values change unexpectedly

At least two important environment opcodes will return different values: `CALLER` (ie. `msg.sender`) and `CALLVALUE` (ie. `msg.value`). This complicates the use of functions for permission checking. The following is a naive vyper translation of the common "ownable" pattern for a simple storage solidity contract. 

```python
owner: address
value: public(int128)

@public
def __init__():
    self.owner = msg.sender

@private
def checkOwner():
  assert msg.sender == self.owner # this will REVERT every time

@public
def writeValue(_value: int128):
    self.checkOwner()
    self.value = _value
```

#### 2. Gas costs disincentivizes code reuse

The `CALL` opcode costs 700 gas (+ input data costs + function dispatching again). By contrast Solidity uses `JUMP` (2 gas) to call a function.

This imposes a large penalty on code reuse. Regardless Vyper's design philosophy favoring safety over gas efficiency, developers will respond to this incentive by simply copying and pasting boilerplate code. This is dangerous, and difficult to audit. 

## Specification: 

I'm sorry, I don't have one specification. Below are some options I see for mitigating these issues.

From the *Motivation* section above, we have 3 parameters along which to analyze approaches to calling a function in the EVM. 

1. Gas cost 
2. Preservation of env variables
3. Safety benefits

### 1. Use `DELEGATECALL` instead of `CALL`

1. Gas cost: NEUTRAL (same as using CALL)
2. Env variables: GOOD (gives us back `CALLER` and `CALLVALUE`)
3. Memory Safety: BAD (executes the code in the same memory context, nullifying reasons for using `CALL` in the first place)

I proposed this earlier, but I don't think this is a good solution. 

### 2. Just use `JUMP` like Solidity

1. Gas cost: GOOD (Much lower gas cost)
2. Env variables: GOOD (Preserves `CALLER` and `CALLVALUE`)
3. Memory Safety: BAD (Greater risk of compiler bugs)

This is worth considering. 

### 3. Implement a safer analog to Solidity's modifiers

These could have significant restrictions on them, like no access to `MSTORE` or `SSTORE`, and not accepting arguments. I assume this would be done by with `JUMP`, or just inlining the same code each time it's needed. 

1. Gas cost: GOOD (Lower execution cost, but might increase size of contract if inlining is used)
2. Env variables: GOOD (Preserves `CALLER` and `CALLVALUE`)
3. Memory Safety: Depends on implementation

### 4. Create special variables which persist across message calls

Built-in vars like `caller` and `callvalue` can be created, which are magically passed to a function in a message call. 

My code above would thus be:

```python
@private
def checkOwner():
  assert _sender == self.owner # <- replaced msg.sender with a magical global variable `sender`

@public
def writeValue(_value: int128):
    self.checkOwner)
    self.value = _value
```

This feels a bit funky to me. Probably would be hard to implement safely, and obscures the true functioning of the EVM.

### 5. Status Quo

My problem in the code sample above can be addressed by passing the values I need as arguments: 

```python
@private
def checkOwner(_sender: address):
  assert _sender == self.owner # <- replaced msg.sender with _sender

@public
def writeValue(_value: int128):
    self.checkOwner(msg.sender) # <- pass msg.sender as an argument 
    self.value = _value
```
Maybe that's OK? But it requires a deeper understanding of how both the EVM and Vyper work. 

## Backwards Compatibility

Dependent on approach. 

## Copyright

jacqueswww:
### - What I did
Fixes #901.

### - How I did it
No comment.

### - How to verify it
Well... tests I guess.

Use JUMP instead of CALL for private functions.

### - Cute Animal Picture

whatsapp image 2018-09-06 at 12 13 21

### DIFF
--- a/tests/parser/features/decorators/test_private.py
+++ b/tests/parser/features/decorators/test_private.py
@@ -13,20 +13,50 @@ def returnten() -> int128:
     assert c.returnten() == 10
 
 
-def test_private_with_more_vars(get_contract_with_gas_estimation):
+def test_private_with_more_vars(get_contract):
     private_test_code = """
 @private
 def afunc() -> int128:
+    a: int128 = 4
+    b: int128 = 40
+    c: int128 = 400
+    return a + b + c
+
+
+@public
+def return_it() -> int128:
+    a: int128 = 111
+    b: int128 = 222
+    c: int128 = self.afunc()
+    return a + b + c
+    """
+
+    c = get_contract(private_test_code)
+    assert c.return_it() == 777
+
+
+def test_private_with_more_vars_nested(get_contract_with_gas_estimation):
+    private_test_code = """
+@private
+def more() -> int128:
+    a: int128 = 11
+    b: int128 = 100
+    c: int128 = 1
+    return a + b + c
+
+@private
+def afunc() -> int128:
     a: int128 = 1
-    return 444
+    a += self.more()
+    return 444 + self.more()
 
 @public
 def return_it() -> int128:
     a: int128 = 222
     b: int128 = 111
-    c: int128 = self.afunc()
+    c: int128 = self.more()
     return a + b + c
     """
 
     c = get_contract_with_gas_estimation(private_test_code)
-    assert c.return_it() == 777
+    assert c.return_it() == 444
--- a/tests/parser/features/decorators/test_private.py
+++ b/tests/parser/features/decorators/test_private.py
@@ -211,17 +211,60 @@ def _test(a: int128) -> (int128, int128):
 
 @public
 def test(a: int128) -> (int128, int128):
-    b: int128
-    c: int128
-    b, c = self._test(a)
-    return b, c
+    return self._test(a)
     """
 
     c = get_contract_with_gas_estimation(code)
 
     assert c.test(11) == [13, 2]
 
+
+def test_private_return_bytes(get_contract_with_gas_estimation):
+    code = """
+@private
+def _test() -> (bytes[100]):
+    b: bytes[50] = "hello                   1           2"
+    return b
+
+
+@public
+def test() -> (bytes[100]):
+    d: bytes[100]
+    d = self._test()
+    return d
+
+@public
+def test2() -> (bytes[100]):
+    d: bytes[100]
+    return self._test()
+    """
+    c = get_contract_with_gas_estimation(code)
+
+    assert c.test() == b"hello                   1           2"
+    # assert c.test2() == b"hello                   1           2"
+
+
+def test_private_return_tuple_bytes(get_contract_with_gas_estimation):
+    code = """
+@private
+# @public
+def _test(a: int128, b: bytes[50]) -> (int128, bytes[100]):
+    return a + 2, concat("badabing:", b)
+
+@public
+def test(a: int128, b: bytes[50]) -> (int128, bytes[100], bytes[50]):
+    c: int128
+    d: bytes[100]
+    c, d = self._test(a, b)
+    return c, d, b
+    """
+
+    c = get_contract_with_gas_estimation(code)
+
+    assert c.test(11, b"test") == [13, b"badabing:test", b"test"]
+
 # Return types to test:
 # 1.) ListType
 # 2.) BytesArray
 # 3.) Straight tuple return `return self.priv_call() -> (int128, bytes[10]`
+# 4.) Add test to copy bytearray from storage and return
--- a/tests/parser/features/decorators/test_private.py
+++ b/tests/parser/features/decorators/test_private.py
@@ -12,4 +12,21 @@ def returnten() -> int128:
     c = get_contract_with_gas_estimation(private_test_code)
     assert c.returnten() == 10
 
-    print("Passed private function test")
+
+def test_private_with_more_vars(get_contract_with_gas_estimation):
+    private_test_code = """
+@private
+def afunc() -> int128:
+    a: int128 = 1
+    return 444
+
+@public
+def return_it() -> int128:
+    a: int128 = 222
+    b: int128 = 111
+    c: int128 = self.afunc()
+    return a + b + c
+    """
+
+    c = get_contract_with_gas_estimation(private_test_code)
+    assert c.return_it() == 777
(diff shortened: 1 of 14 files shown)


jacqueswww:
@haydenadams please run your uniswap code on this branch as see if there are any errors, test that fail.

I want to run as many tests as possible before merging. With EIP-1380 also out there, vyper has 2 two plans ready now ;)

fubuloubu:
This update is a big one, hopefully I can review it next week. Perhaps @pipermerriam can get some eyes on it so we can make sure this doesn't cause major issues?

tests/parser/features/decorators/test_private.py
```diff
@@ -12,4 +12,387 @@ def returnten() -> int128:
...
+def __init__():
+    self.greeting = "Hello "
+
+@private
+def set_greeting(_greeting: bytes[20]):
+    a: uint256 = 333
+    b: uint256 = 334
+    c: uint256 = 335
+    d: uint256 = 336
+    if a + b + c + d == 1338:
```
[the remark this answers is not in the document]
fubuloubu: ;) lol

tests/parser/features/decorators/test_private.py
```diff
@@ -12,4 +12,387 @@ def returnten() -> int128:
...
+def addition(a: uint256, b: uint256 = 1) -> uint256:
+    return a + b
+
+
+@public
+def add_one(a: uint256) -> uint256:
+    return self.addition(a)
+
+@public
+def added(a: uint256, b: uint256) -> uint256:
```
fubuloubu: Should also assert that `a`, `b` were not modified in a test somewhere

tests/parser/features/decorators/test_private.py
```diff
@@ -12,4 +12,387 @@ def returnten() -> int128:
...
+    c = get_contract_with_gas_estimation(private_test_code)
+
+    assert c.add_one(20) == 21
+    assert c.added(10, 20) == 30
+
+
+def test_private_return_tuple(get_contract_with_gas_estimation):
+    code = """
+@private
+# @public
```
fubuloubu: Remove this comment? Is this comment intentional?

If intentional, it should be added to syntax tests not here.

tests/parser/features/decorators/test_private.py
```diff
@@ -12,4 +12,387 @@ def returnten() -> int128:
...
+@private
+def set_greeting(_greeting: bytes[20]):
+    a: uint256 = 333
+    b: uint256 = 334
+    c: uint256 = 335
+    d: uint256 = 336
+    if a + b + c + d == 1338:
+        self.greeting = _greeting
+
+@public
```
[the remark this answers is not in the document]
fubuloubu: Further look: Mix these in. It shoudn't make a difference, but it would be good to mix it up.

tests/parser/features/decorators/test_private.py
```diff
@@ -12,4 +12,387 @@ def returnten() -> int128:
...
+def test4() -> (bytes[100]):
+    d: bytes[100] = 'xyzxyzxyzxyz'
+    return self.get_msg()
+    """
+
+    c = get_contract_with_gas_estimation(code)
+
+    assert c.test() == b"hello                   1           2"
```
fubuloubu: Maybe put this string as a var up above so spacing is only set once.

tests/parser/features/decorators/test_private.py
```diff
@@ -12,4 +12,387 @@ def returnten() -> int128:
...
+    assert c.test2() == b"hello                   1           2"
+    assert c.test3(b"alice") == b"alice"
+    c.set(b"hello daar", transact={})
+    assert c.test4() == b"hello daar"
+
+
+def test_private_bytes_as_args(get_contract_with_gas_estimation):
+    code = """
+@private
+def _test(a: bytes[40]) -> (bytes[100]):
```
fubuloubu: I think the style guide should *suggest* that `_name` be the convention for private functions.

tests/parser/types/test_string_literal.py
```diff
@@ -67,6 +67,7 @@ def baz(s: int128, L: int128) -> bytes[100]:
         if x * y == 999:
             return self.moo
         """ % (("c" * i), ("c" * i), ("c" * i))
+        print(kode)
```
fubuloubu: Remove?

jacqueswww:
Thanks, will have a look/fix them on Monday.

setup.py
```diff
@@ -25,15 +25,12 @@
         'pytest-runner'
     ],
     tests_require=[
+        'pytest',
```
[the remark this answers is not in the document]
jacqueswww: yip

haydenadams:
> @haydenadams please run your uniswap code on this branch as see if there are any errors or tests that fail.
> 
> I want to run as many tests as possible before merging. With EIP-1380 also out there, vyper has 2 two plans ready now ;)

@jacqueswww  Been delayed on this since I've been traveling. Looking at it now.

PR #1017, opened 2018-09-12
Define equality checking of types more cleanly

davesque:
### - What I did

Noticed some possible improvements to be made to the API for equality checking in `NodeType` subclasses.  Did a tiny bit of cleanup there.

### - How I did it

* Added `__eq__` implementation on `NodeType` base class which always checks that two instances are of the same class.  This should be more future proof since it doesn't require the class to be explicitly stated.
* Converted `NodeType` to abstract base class
* Added abstract method `eq` which checks for equality particular to each subclass

### - How to verify it

Run the test suite.

Internal change.  No need for description.

### - Cute Animal Picture

Cute animal picture

### DIFF
--- a/vyper/types/types.py
+++ b/vyper/types/types.py
@@ -1,3 +1,4 @@
+import abc
 import ast
 import copy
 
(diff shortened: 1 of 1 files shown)


fubuloubu:
Nice, thanks Dave!

Coverage change is inconsequential, `eq()` is not checked for `NodeType`, but it is not intended to be used in this way.

davesque:
@fubuloubu Actually, maybe I should put in a "no coverage" pragma on the abstract method stub.

fubuloubu:
Yes! Need more of those TBH. A lot of the exceptions aren't actually reachable, and are there to account for change in the codebase

Issue #990, opened 2018-08-10
VIP: Constant Variables


nrryuya:
## Preamble
```
VIP: 990
Title: Constant Variables
Author: Ryuya Nakamura (@nrryuya)
Type: Standard Track
Status: Draft
Created: 10-8-2018
```

## Simple Summary
Adding constant variables (like [Solidity's one](https://solidity.readthedocs.io/en/v0.4.24/contracts.html#constant-state-variables)) for readability and usability.

## Abstract
Adding constant variables which is assigned from an constant expression and can be used in whole the contract. Every occurrence is replaced by the constant expression at compile time.

## Motivation
Constant variables reduces hard coding of same value repeatedly and improve readability. It also helps code auditing.

## Specification
Spec on declaring constant variables.

```
# Block interval of Plasma chain
CHILD_BLOCK_INTERVAL: constant(uint256) = 1000
```

## Backwards Compatibility
No problem

## Copyright
Copyright and related rights waived via [CC0](https://creativecommons.org/publicdomain/zero/1.0/)

nrryuya:
I wonder 
- []specification on declaring constants in Python's style
- []types allowed as constants(uint256, int128, address, decimal, bytes?)

jacqueswww:
I think we should leave types of the equation, and just follow the principal that each constant is a literal.
```python
CHILD_BLOCK_INTERVAL: constant = 100
```
Also I think there is merit in limiting constants to only be defined at the top of a program (also outside functions).

ed-ilyin:
I think constants should have types.
Currently I have to do like this within function.
```
    secondsPerMonth:uint256(sec/mon) = 2592000 # 30×24×60×60
    seconds:timedelta = _durationInMonths * secondsPerMonth
```
typed constants will make a code cleaner

jacqueswww:
@nrryuya yes, above the all the functions just after globals perhaps?

jacqueswww:
Starting on this.

PR #1018, opened 2018-09-13
Refactor global context.

jacqueswww:
### - What I did

Whilst working on #990, the following refactoring became necessary to simply usage of commmonly shared attributes in parser.py.


jacqueswww:
parser.py is systematically getting smaller ! :smile_cat:

Issue #642, opened 2018-01-23
VIP: Standardize compiler output artifacts


fubuloubu:
## Preamble

VIP: <to be assigned>
Title: Standardize compiler output artifacts
Author: @fubuloubu
Type: Standard
Status: Draft
Created: 2018-01-22

## Simple Summary
Change default compiler output to match solc for compatibility with existing/future tools

## Abstract
The solidity compiler outputs a JSON-formatted string containing the compilation artifacts for all compiled contracts. The format is:
```python
{
  "contracts" : {
    "path/to/contractA.ext" : {
      "abi" : "[\" ... \"]", # Not sure why this is in quotes actually...
      "bin" : "0x...",
      "bin-runtime" : "0x...",
      ... # Additional stuff
    },
    "path/to/contractB.ext" : { ... },
    ...
  },
  "version" : "some-compiler-version-string"
}
```
It would be great if Vyper matched this output, at least as a default, so that it would enable easier integration with existing (and future) tools

## Motivation
My personal motivation is that I want to write a set of tools for Ethereum's Python ecosystem that make it easy to test, deploy, and interact with contracts. Some of the existing tools already leverage Solidity's output for this purpose (or otherwise hardcode themselves to solidity's style). Vyper, being the new kid on the block, should integrate with these tools, and I believe the best way of doing that is by using the same output (`--combined-json` in solidity) so that these tools can be useful.

My particular vision for a project written in this ecosystem would have the following folder structure:
```bash
project/
    contracts/
        contracts.json
        # arbitrarily nested contract files: .sol, .vy, etc.
    tests/  # pytest-ethereum: py.test plugin for contract testing
        config.yml  # References ../contracts.json, auto-generates pytest fixtures with given args
        # arbitrarily nested tests: test_*.py
    deployments/  # eth-deployer: enables easy deployment management
        config.yml  # References ../contracts.json, contains deployment args
        addresses.yml  # Deployment address tracking file
        # eth-deployer would have a console for interactions
        # Simulation and deployment files, etc.
    ...
    # Formal requirement verification, simulations, coverage testing, etc...
```
It doesn't have to look like this in the long run, my point is that it all builds off this common artifact output, and JSON is a good format to store these artifacts in. Later down the road, we can validate the compiler version present in the artifact output file and check some checksum hashes of the input files (stored in the compiler artifacts) and enable fast rebuilds when only certain files were changed in the `contracts/` directory (updating alongside similar output from solc). But I digress...

[...]

PR #1033, opened 2018-09-25
Standardize compiler output #642

fubuloubu:
Solves #642
Modified compiler module to have accessible API function for direct use
Leveraged API for script to shorten untested logic

### - What I did
Got tired of compiling 3 separate times to obtain what I wanted through the `vyper.compiler` API,
so I made my own API

Also added the ability to compile multiple files at once

### - How I did it
Refactored code out of `bin/vyper`

### - How to verify it
Tests all still work, also try `vyper -f abi,bytecode examples/*.vy`

Refactored compiler API; script can process multiple files

### - Cute Animal Picture

annoyed dolphin

### DIFF
--- a/bin/vyper
+++ b/bin/vyper
@@ -18,9 +18,6 @@ if __name__ == '__main__':
 
     args = parser.parse_args()
 
-    if args.show_gas_estimates:
-        parser_utils.LLLnode.repr_show_gas = True
-
     def uniq(seq):
         exists = set()
         if isinstance(seq, list):
@@ -34,11 +31,12 @@ if __name__ == '__main__':
     output['contracts'] = {}
     output['version'] = vyper.__version__
 
+    print_output = False
     for fname in args.file:
         with open(fname, 'r') as fh:
             code = fh.read()
         try:
-            file_output = api(code)
+            file_output = api(code, show_gas_estimates=args.show_gas_estimates)
         except Exception as e:
             print("File", fname, "failed to compile:")
             print()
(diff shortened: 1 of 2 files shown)


jacqueswww:
Cool will check this out still properly, but we might have to think if we only want the output to only be machine readable, and perhaps add something like `-f combined_json` instead, as that would be backwards incompatible. Also - I use `-f ir` extensively (with colors). And wrapping it all into json makes this unfeasible.

fubuloubu:
Sure! I mean, we can massage this around a bit more for script usage but my concept was to make the compiler API more straightforward so it can be leveraged in other Python tools more directly (`interface = vyper.compile(code)` instead of having to do 3 things to get `interface`)

The script can then handle file inputs (even multiple files) and how the output ends up being displayed for each item. Moving the assembly pretty-printer to where it is might not be the best option for that piece, but it shouldn't live in the script if it is more broadly useful

fubuloubu:
Well, we should ask the question "how do we think this will be used?"

We could do what solidity did and have a "json mode" that accepts input and produces output only with json. We could have a "combined json" flag that switches the output to json from newline prints.

Having the API available in the compiler is the biggest thing for me, it allows tools like Populus and what not to leverage vyper directly. Imagine `contract = w3.eth.contract(**vyper.compiler.api(code))` :)

jacqueswww:
Weird... I thought I closed this already. Must've been during the time github was resyncing it's databases... anyways closing because #1089 is where this work moved on.

PR #1034, opened 2018-09-25
Change ERC20 example to newer syntax

fubuloubu:
### - What I did
Noticed this file didn't compile when I tried it, so I fixed it

### - How I did it
Magic!

### - How to verify it
`vyper examples/tokens/ERC20.vy`

No change

### - Cute Animal Picture
donkey

### DIFF
--- a/examples/tokens/ERC20.vy
+++ b/examples/tokens/ERC20.vy
@@ -14,9 +14,9 @@ Approval: event({_owner: indexed(address), _spender: indexed(address), _value: u
 name: public(bytes32)
 symbol: public(bytes32)
 totalSupply: public(uint256)
-decimals: public(int128)
-balances: int128[address]
-allowed: int128[address][address]
+decimals: public(uint256)
+balances: uint256[address]
+allowed: uint256[address][address]
 
 
 @public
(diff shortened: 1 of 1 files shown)


fubuloubu:
No idea, I didn't write it lol. Just re-wrote to fix some bad practices and the compilation error.

Prior to my re-write, it did the `if` flow (bad!), which I changed to `assert`s (good!)

Differences are...

Vypercoin added: https://github.com/ethereum/vyper/blob/f5d5446c9cab212dc698ec94c077b34deb8695d8/examples/tokens/vypercoin.vy#L72
Rewrite removed `convert`: https://github.com/ethereum/vyper/blob/f5d5446c9cab212dc698ec94c077b34deb8695d8/examples/tokens/vypercoin.vy#L25

---

More comes down to a naming convention thing? Do name it ERC20 (alongside ERC721 example) or Vypercoin?

jacqueswww:
Super strange! Yes one probably originated to be a completely `uint256` version (vs the original vypercoin int128 one), and now both examples have converged to be the same. ERC20.vy doesn't have any tests, so I would say we keep vypercoin.vy and remove ERC20.vy. And then we can rename vypercoin.vy to ERC20.vy if it's clearer for new comers?

fubuloubu:
`deposit()` function? where?

jacqueswww:
Oh wrong ERC20! deposit is in the solidity compatible file: https://github.com/ethereum/vyper/blob/0e2fbb89b798be1b07d7dc32213b9b37d4f8df58/examples/tokens/ERC20_solidity_compatible/ERC20.vy#L21

fubuloubu:
We have too many ERC20 tokens lol

fubuloubu:
Perhaps we should refine to an "ERC Implementations" folder?

haydenadams:
Technically one version of ERC20.vy was formally verified, which is interesting.

https://github.com/runtimeverification/verified-smart-contracts/blob/master/erc20/vyper/README.md

fubuloubu:
Closing per #1133

PR #1036, opened 2018-09-26
Add Access-Control-Allow-Origin: * to all vyper-server requests.

jacqueswww:

### DIFF
--- a/bin/vyper-serve
+++ b/bin/vyper-serve
@@ -46,7 +46,7 @@ class VyperRequestHandler(BaseHTTPRequestHandler):
             self.send_response(200)
             self.send_cors_all()
             self.end_headers()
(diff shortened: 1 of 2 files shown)


nrryuya:
IMO, this modification is needed? 
```
    def do_OPTIONS(self):
        self.send_response(200)
        self.send_cors_all()
        self.send_header("Access-Control-Allow-Headers", "X-Requested-With, Content-type")
        self.end_headers()
```

I just tried only with google chrome, though.

jacqueswww:
OK cool, I will add to `send_cors_all` ;)

nrryuya:
Yes, when I did a experiment without response code with Chrome, `ERR_INVALID_HTTP_RESPONSE` occurs. 
(Sorry for not explaining enough)

PR #1042, opened 2018-10-02
Change convert to use type instead of string of 'type'

### LINKED ISSUE #1026: Convert function takes string instead of type for conversion to
### What's your issue about?
If you forget to add quotes to the type you want to convert to, Vyper will give an unhelpful error:
```python
convert(some_int128, uint256)
# raises:
#    AttributeError: 'Name' object has no attribute 's'
```

### How can it be fixed?
So, catching this error might work out to fix this issue, but I think the underlying issue is that the second argument (`convertTo`) is a string instead of a typename. This actually makes it a little more unintuitive to write conversions as you could misspell the name and not get visual feedback from your IDE (assuming you have syntax highlighting up)

I would suggest turning this into a VIP to modify the syntax of convert such that a valid typename is supplied as the second argument

#### Cute Animal Picture
pony

jacqueswww:
### - What I did
Fixes #1026
Change the convert style to use types instead of string. (`VIP1026 <https://github.com/ethereum/vyper/issues/1026>`_)

### DIFF
--- a/vyper/types/convert.py
+++ b/vyper/types/convert.py
@@ -1,4 +1,5 @@
 import ast
+import warnings
 
 from vyper.functions.signature import (
     signature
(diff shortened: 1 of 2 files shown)


vyper/types/convert.py
```diff
@@ -96,11 +99,20 @@ def to_bytes32(expr, args, kwargs, context):
 
 
 def convert(expr, context):
-    output_type = expr.args[1].s
+
+    if isinstance(expr.args[1], ast.Name):
+        output_type = expr.args[1].id
+    else:
+        raise ParserException(
```
fubuloubu: What do you think about a deprecation warning instead?

jacqueswww: yeah sure, why not ;)

fubuloubu:
I meant instead of an exception lol. Right now you get both a warning and an exception

jacqueswww:
@fubuloubu Yes but we need both, to only have depreciation warning would only cover strings. But other stuff could be passed in that obviously isn't compatible as parameter 2.

fubuloubu:
Ah, makes sense!

Issue #905, opened 2018-06-17
Style guide


fulldecent:
Add a coding style guide to Vyper and use it consistently.

Currently examples include both camel rase (https://github.com/ethereum/vyper/blob/master/docs/structure-of-a-contract.rst `storedData `) and snake case (https://github.com/ethereum/vyper/blob/235e1126c292985b98988043bad79b71bb08bb7d/examples/stock/company.v.py `total_shares `) for the same type of thing.

I do not know which one is canonical and this is a distraction from writing code.

# Work plan

- [ ] Decide the canonical style
- [ ] Document it at https://vyper.readthedocs.io/en/latest/
- [ ] Update code examples to use this style
- [ ] Notify external parties like https://vyper.online and request they also adopt it

# Discussion

Function names, parameter names, events and public variable names are user-visible. Maybe in the future all contract calls will be hidden from the end user -- but for the foreseeable future, people will be confirming transactions on Etherscan and using MetaMask to confirm details before they click "accept".

In order to be consistent with other deployed contracts, I recommend we adopt from [Solidity Style Guide](https://solidity.readthedocs.io/en/v0.4.24/style-guide.html) anything that affects the end-user. This includes function names, parameter names, etc.

fubuloubu:
Solidity uses camel case and has defined several standards. These would be different if using snake case, so we may have to accept that begrudgingly

maurelian:
Solidity's style guide is based on PEP8, so converting the guide from Solidity to Vyper should be doable. 

I do think Vyper (and future languages) are obligated to adopt camelCase for external functions, but it seems reasonable to me that private functions, and all other names can follow Python conventions.

jacqueswww:
We can also consider to add some of the style rules to the compiler (e.g. all public function should be camelcase).
This is a feature that I believe will make it much easier for people to adopt the style :tongue:

jacqueswww:
As discussed on https://github.com/ethereum/vyper/pull/925, we will probably have to go with camelcasing all methods like solidity. The thought process goes like:

1.) Public method require it thanks to ERC's
2.) To keep method casing consistent we have to match it accross internal & public functions.

I still think we should only apply the hard camelcase check on public methods though.

fulldecent:
For reference, here are the parts of the Solidity style guide that are visible outside just Solidity.

SOURCE: https://solidity.readthedocs.io/en/v0.4.24/style-guide.html

---

# Style Guide [EXCERPTS]

## Naming Conventions

[...]

PR #1052, opened 2018-10-20
Text Update: camelCase to snake_case

Dan-Nolan:
### - What I did

Changed a variable name. Noticed all the other variables are snake_cased while this one was camelCased.

### - How I did it

With my keyboard :)

### - How to verify it

Take a look!

No Change

### - Cute Animal Picture

image

### DIFF
--- a/docs/structure-of-a-contract.rst
+++ b/docs/structure-of-a-contract.rst
@@ -18,7 +18,7 @@ State variables are values which are permanently stored in contract storage.
 
 ::
 
-  storedData: int128
+  stored_data: int128
 
 See the :ref:`types` section for valid state variable types and
 :ref:`visibility-and-getters` for possible choices for


Dan-Nolan:
Actually, I noticed a couple other places in the Vyper by Example section where camelCase is used. Should we update these as well, or is there a reason why camelCase might be used for state variables?

fubuloubu:
So, mixedCase (CamelCase uppercases the first letter always I think) is used because we need to ensure compatibility with standards that Solidity has already set, being the first language out of the gate and thus ahead on standards.

The way solidity works under the hood (and thus **all** Ethereum smart contracts by convention) is that the first 4 bytes of the call data, when you make a call to an address, are loaded with a slice of the keccak hash of the function signature. So, a function that has signature `myFun(a: address, b: bytes32) -> uint256: ...` has a compressed signature (removing spaces and arg names) of `myFun(address,bytes32)` (the return type is not included in this signature), which gives a hash of `0x160f22da...` that is used to identify which function you are trying to call when you call a contract at address `X`.

So, what this means is that if we want to be compliant with things like ERC-20, our function names (such as `transferFrom`) really matter what case they are in, since snake case would give you a totally different hash and thus not be compliant with the spec. Same goes with state variables, since we sometimes use the `public` keyword to generate a getter, so functions like `totalySupply` must also have matching case in order to generate the same getter.

Now, that does not mean all variables have to have the same casing, but if we've come this far we might as well adopt it all to make it consistent. It is therefore a recommendation of the style guide (and a requirement for ERC compatibility) to use mixedCase for **everything** even though it's not the same as Python. The context switching between the two is difficult, but helps our users disambiguate between the two syntaxes a little quicker hopefully.

---

See [here](https://solidity.readthedocs.io/en/develop/abi-spec.html) for more information.

NOTE: Unfortunately, the ABI spec is still maintained alongside the Solidity documentation, which means it basically maintains the ABI spec. We should do something about that! Compilers Ring FTW!

fubuloubu:
All that being said, I'm going to reject this PR. Would appreciate it if you wanted to PR this information into our docs in the appropriate place if it is not clear. We have internalized this information, but it is always great to have outside perspectives point holes in our documentation so more people know what is going on! Thanks!

Dan-Nolan:
Good to know, thanks for the background information! 

Just now seeing the discussion in #905 and #925, which has been illuminating as well.

PR #1056, opened 2018-10-22
Add formats to help for command line

fubuloubu:


bin/vyper
```diff
@@ -78,6 +64,13 @@ def get_source_map_asm_list(code):
 
 
 if __name__ == '__main__':
+    
+    warnings.simplefilter('always')
+    sys.tracebacklimit = 0
+    tb_limit = os.environ.get('VYPER_TRACEBACK_LIMIT')
+    if tb_limit:
+        sys.tracebacklimit = int(tb_limit)
```
jacqueswww: Considering these are used in a global scope, I think having them outside main might make sense?

jacqueswww: Correct, these are all related to CLI usage ;)

jacqueswww:
Closing this as discussed, busy with the updated one of this ;)

PR #1060, opened 2018-10-30
fix vypercoin erc20 compliance, optimize gas costs

PavelCore:
### - What I did
vypercoin.vy violated SHOULD requirement of ERC20 standard. Moreover, the code contained some redundant checks, that consumes gas.
### - How I did it
By fixing vypercoin.vy file. Let's go over changes:

1. The following part of the code makes a simple check for the overflow in the `totalSupply` variable. I moved it to the beginning of the constructor so that less gas will be spent in the case of failure (since nothing will be written to the storage).
```
    assert _initialSupply * 10 ** _decimals >= _initialSupply
    self.totalSupply =_initialSupply * 10 ** _decimals
    self.balances[msg.sender] = self.totalSupply
```

2. This event SHOULD be present according to the ERC20 standard, since new tokens are created in the constructor.
```
log.Transfer(ZERO_ADDRESS, msg.sender, self.totalSupply)
```
Citation from [standard description](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-20.md):

> A token contract which creates new tokens SHOULD trigger a Transfer event with the _from address set to 0x0 when tokens are created.

3. The following check is redundant. The sum of all user balances is equal to totalSupply, thus this assert never fails.
```
    assert self.balances[_to] + _amount >= self.balances[_to]
```

4. Transactions in ethereum are atomic. That means, that EVM state would not change between these two lines. Thus, the first line is redundant.
```
    self.allowed[msg.sender][_spender] = 0
    self.allowed[msg.sender][_spender] = _amount
```

### - How to verify it
Run tests for vypercoin.vy file
None
### - Cute Animal Picture
image

### DIFF
--- a/examples/tokens/vypercoin.vy
+++ b/examples/tokens/vypercoin.vy
@@ -19,11 +19,13 @@ allowed: uint256[address][address]
 @public
 def __init__(_name: bytes32, _symbol: bytes32, _decimals: uint256, _initialSupply: uint256):
 
+    assert _initialSupply * 10 ** _decimals >= _initialSupply
+    self.totalSupply =_initialSupply * 10 ** _decimals
+    self.balances[msg.sender] = self.totalSupply
     self.name = _name
     self.symbol = _symbol
     self.decimals = _decimals
-    self.totalSupply =_initialSupply * convert(10, uint256) ** _decimals
-    self.balances[msg.sender] = self.totalSupply
+    log.Transfer(ZERO_ADDRESS, msg.sender, self.totalSupply)
 
 
 # What is the balance of a particular account?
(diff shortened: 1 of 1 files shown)


fubuloubu:
LGTM

3. Is also interesting because we should never have an overflow to begin with, not sure who added this check.

fubuloubu:
P.S. test case was executed for the update right? It is in the test suite?

PavelCore:
If I understand your question correctly, then yes: there are tests for vypercoin.vy.

travis CI output:
```
tests/examples/tokens/test_vypercoin.py ......                           [  7%]
```

fubuloubu:
Awesome, just double checking the test was run. Sometimes the examples don't seem to make it into our test suite.

PR #1062, opened 2018-11-01
Change doc due to the changes on #1042

hskang9:
#1042 has changed the convert function to include type as itself.

### - What I did
I changed the doc's content

### - How I did it
I removed '' from the argument for destination of ``convert()``

### - How to verify it

Check #1042


fubuloubu:
Awesome! Thanks!

PR #1063, opened 2018-11-02
Blind auction example, adapted from "Solidity by Example" 

jakerockland:
### - What I did

Wrote an example Vyper contract modeled off of the "Blind Auction" example in "Solidity by Example".

### - How I did it

Adapted the corresponding Solidity contract to Vyper as closely as possible. Added appropriate tests for the contract.

### - How to verify it

```
make
vyper examples/auctions/blind_auction.vy

make test
```

Added a "Blind Auction" example contract modeled off of the "Blind Auction" example in "Solidity by Example".

### DIFF
--- a/examples/auctions/blind_auction.vy
+++ b/examples/auctions/blind_auction.vy
@@ -1,5 +1,4 @@
-# Blind Auction
-# Adapted to Vyper from [Solidity by Example](https://github.com/ethereum/solidity/blob/develop/docs/solidity-by-example.rst#blind-auction-1)
+# Blind Auction # Adapted to Vyper from [Solidity by Example](https://github.com/ethereum/solidity/blob/develop/docs/solidity-by-example.rst#blind-auction-1)
 
 # Note: because Vyper does not allow for dynamic arrays, we have limited the
 # number of bids that can be placed by one address to 128 in this example
@@ -20,12 +19,12 @@ ended: public(bool)
 highestBid: public(wei_value)
 highestBidder: public(address)
 
-# State of bids
-bids: ({blindedBid: bytes32, deposit: wei_value}[128])[address]
-bidCounts: int128[address]
+# State of
+bids: map(address, {blindedBid: bytes32, deposit: wei_value}[128])
+bidCounts: map(address, int128)
 
 # Allowed withdrawals of previous bids
-pendingReturns: wei_value[address]
+pendingReturns: map(address, wei_value)
 
 
 # Create a blinded auction with `_biddingTime` seconds bidding time and
--- a/examples/auctions/blind_auction.vy
+++ b/examples/auctions/blind_auction.vy
@@ -1,0 +1,183 @@
+# Blind Auction
+# Adapted to Vyper from [Solidity by Example](https://github.com/ethereum/solidity/blob/develop/docs/solidity-by-example.rst#blind-auction-1)
+
+# Event for logging that auction has ended
+AuctionEnded: event({_highestBidder: address, _highestBid: wei_value})
+
+# Auction parameters
+beneficiary: public(address)
+biddingEnd: public(timestamp)
+revealEnd: public(timestamp)
+
+# Set to true at the end of auction, disallowing any new bids
+ended: public(bool)
+
+# Final auction state
+highestBid: public(wei_value)
+highestBidder: public(address)
+
+# State of bids
+# Note: because Vyper does not allow for dynamic arrays, we have limited the
+# number of bids that can be placed by one address to 128 in this example
+bids: ({blindedBid: bytes32, deposit: wei_value}[128])[address]
+bidCounts: int128[address]
+
+# Allowed withdrawals of previous bids
+pendingReturns: wei_value[address]
+
+
+# Create a blinded auction with `_biddingTime` seconds bidding time and
+# `_revealTime` seconds reveal time on behalf of the beneficiary address
+# `_beneficiary`.
+@public
+def __init__(_beneficiary: address, _biddingTime: timedelta, _revealTime: timedelta):
+    self.beneficiary = _beneficiary
+    self.biddingEnd = block.timestamp + _biddingTime
+    self.revealEnd = self.biddingEnd + _revealTime
+
+
+# Place a blinded bid with:
+# if fake:
+#     fakeInt = 1
+# else:
+#     fakeInt = 0
+# _blindedBid = sha3(concat(
+#       convert(value, bytes32),
+#       convert(fakeInt, bytes32),
+#       secret)
+# )
+#
+# The sent ether is only refunded if the bid is correctly revealed in the
+# revealing phase. The bid is valid if the ether sent together with the bid is
+# at least "value" and "fake" is not true. Setting "fake" to true and sending
+# not the exact amount are ways to hide the real bid but still make the
+# required deposit. The same address can place multiple bids.
+@public
+@payable
+def bid(_blindedBid: bytes32):
+    # Check if bidding period is still open
+    assert block.timestamp < self.biddingEnd
+
+    # Check that payer hasn't already placed maximum number of bids
+    # Note: because Vyper does not allow for dynamic arrays, we have limited the
+    # number of bids that can be placed by one address to 128 in this example
+    numBids: int128 = self.bidCounts[msg.sender]
+    assert numBids < 128
+
+    # Add bid to mapping of all bids
+    (self.bids[msg.sender])[numBids] = {blindedBid: _blindedBid, deposit: msg.value}
+    self.bidCounts[msg.sender] += 1
+
+
+# Returns a boolean value, `True` if bid placed successfully, `False` otherwise.
+@private
+def placeBid(bidder: address, value: wei_value) -> bool:
+    # If bid is less than highest bid, bid fails
+    if (value <= self.highestBid):
+        return False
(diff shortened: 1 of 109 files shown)


examples/auctions/blind_auction.vy
```diff
@@ -0,0 +1,63 @@
+# Blind Auction
+# Adapted to Vyper from [Solidity by Example](https://github.com/ethereum/solidity/blob/develop/docs/solidity-by-example.rst#blind-auction-1)
+
+# Custom struct for bids
+bid: {
+    blindedBid: bytes32,
+    deposit: wei_value
+}
```
fubuloubu: This creates a global named `bid` that is a struct of the layout provided. We don't have support for named structs *yet*, but we're working on it! (#300 )

examples/auctions/blind_auction.vy
```diff
@@ -0,0 +1,63 @@
...
+# Auction parameters
+beneficiary: public(address)
+biddingEnd: public(timestamp)
+revealEnd: public(timestamp)
+
+# Set to true at the end of auction, disallowing any new bids
+ended: public(bool)
+
+# State of bids
+bids: public([address])
```
fubuloubu: It looks like you are missing a "to" value for the mapping (we really need to improve this syntax), which I am guessing you mean to write `bids: public(bid[address])`

As noted above `bid` is not a named struct, so this will not work. The following will work:
```python
bids: public({
    blindedBid: bytes32,
    deposit: wei_value
}[address])
```

examples/auctions/blind_auction.vy
```diff
@@ -0,0 +1,63 @@
...
+
+# Final auction state
+highestBidder: public(bid[address])
```
fubuloubu: As noted above `bid` is not a named struct, so this will not work.

I would also sort of expect this to just be `highestBidder: address`, which can be used to look up the owner of the current highest bid (which probably doesn't mean much after revealing, you could probably delete it). It seems like you are doing this after a commit-reveal period, I would add some comments to that effect to differentiate globals from stage 1 (commit) and stage 2 (reveal).

examples/auctions/blind_auction.vy
```diff
@@ -0,0 +1,63 @@
...
+    # /// place multiple bids.
+    # function bid(bytes32 _blindedBid)
+    #     public
+    #     payable
+    #     onlyBefore(biddingEnd)
+    # {
+    #     bids[msg.sender].push(Bid({
+    #         blindedBid: _blindedBid,
+    #         deposit: msg.value
+    #     }));
```
fubuloubu: It looks like the Solidity example uses a dynamic array to store the bidders, which we do not support. It is often better to track this kind of state with a mapping, and then emit an event tracking the bids as they occur. The bid logs are accessible external to the smart contract for whatever dapp/front end you have, so you can collect and iterate over those if desired (listening to the event stream)

Some guidance on this is that for tracking state that doesn't require ordering, a mapping is often better (and cheaper!) than an array.

fubuloubu:
@jakerockland it's no problem! You used the WIP tag, so it is okay to push incomplete work for safe keeping and getting reviews like this.

@jacqueswww can help you with the LLL, definitely seems like an unhandled exception.

jakerockland:
@jacqueswww any chance you could help nudge me in the right direction for debugging the unhandled exception?

jakerockland:
Ah no worries, will loop back with you next week @jacqueswww and in the meantime will work on a test case for the new example.

jakerockland:
@jacqueswww @fubuloubu 

Just pushed some commits and have bit of an update.

Identified that the source of the uncaught exception was this block in my contract, within the `placeBid(bidder: address, value: wei_value) -> bool` function:

```
if (value <= self.highestBid):
    return False
```

As a fix for now to allow me to keep working, I've re-factored the `placeBid` function as follows:

```
# Returns a boolean value, `True` if bid placed successfully, `False` otherwise.
@private
def placeBid(bidder: address, value: wei_value) -> bool:
    bidSuccess: bool

    # If bid is less than highest bid, bid fails
    if (value <= self.highestBid):
        bidSuccess = False
    # Otherwise bid succeeds
    else:
        bidSuccess = True

        # Refund the previously highest bidder
        if (self.highestBidder != ZERO_ADDRESS):
            self.pendingReturns[self.highestBidder] += self.highestBid

        # Place bid successfully and update auction state
        self.highestBid = value
        self.highestBidder = bidder

    return bidSuccess
```

Not sure if having refactoring like the above is actually how contracts should be written in Vyper? Or if one should be allowed to do things like:

```
if (conditionIsTrue):
    return True # Return early from within if statement
```

Did some other minor refactoring and started on a test case, but is currently only testing initialization, not contract functionality.

jacqueswww:
Great! Those type of return statements are allowed - will have a look.

jakerockland:
@jacqueswww awesome, thanks so much.

fubuloubu:
p.s. Really cool example!

jakerockland:
@jacqueswww Ah rad will fix, actually still working on more comprehensive tests as well. Will remove the WIP in the title and ping you guys when I think it is ready for final review.

jacqueswww:
Goos to hear! Will check then :)

jakerockland:
@jacqueswww I pinged you about this on Gitter but may be actually better/easier to ask here where it is searchable at a later date:

Quick question—whenever I try to run a test individually that uses the `assert_tx_failed` fixture I get some variant this error:

```
w3 = <web3.main.Web3 object at 0x10e29a630>
tester = <eth_tester.main.EthereumTester object at 0x10e29a240>
auction_contract = <tests.conftest.VyperContract object at 0x10ec69048>
assert_tx_failed = <function assert_tx_failed.<locals>.assert_tx_failed at 0x10e904c80>

    def test_early_reveal(w3, tester, auction_contract, assert_tx_failed):
        k1, k2, k3 = w3.eth.accounts[:3]
    
        # Move time slightly forward (still before reveal time)
        tester.time_travel(
            tester.get_block_by_number('latest')['timestamp'] + 10
        )
    
        # Try to reveal early
        _values = [0] * 128
        _fakes = [False] * 128
        _secrets = [b'\x00' * 32] * 128
        _values[0] = 175
        _fakes[0] = True
        _secrets[0] = (9876543).to_bytes(32, 'big')
        _values[1] = 275
        _fakes[1] = True
        _secrets[1] = (9876543).to_bytes(32, 'big')
>       assert_tx_failed(lambda: auction_contract.reveal(
            2,
            _values,
            _fakes,
            _secrets,
            transact={"value": 0, "from": k3})
        )

tests/examples/auctions/test_blind_auction.py:106: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
tests/conftest.py:287: in assert_tx_failed
    tester.revert_to_snapshot(snapshot_id)
venv/lib/python3.7/site-packages/eth_tester/main.py:499: in revert_to_snapshot
    self.backend.revert_to_snapshot(snapshot)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <eth_tester.backends.pyevm.main.PyEVMBackend object at 0x10e2c9940>
snapshot = b'\x85\xc6\tm\xc8,\x9fTZ\xe1\x07\xdbuee\x90t\x87o\xf7\x0c\x05\xd6\xd2,\x9d"\xd4\x85\xf9\x06\x0e'

    def revert_to_snapshot(self, snapshot):
        block = self.chain.get_block_by_hash(snapshot)
        chaindb = self.chain.chaindb
        if block.number > 0:
>           chaindb._set_as_canonical_chain_head(chaindb.db, block.header.hash)
E           TypeError: _set_as_canonical_chain_head() takes 2 positional arguments but 3 were given

```

Any idea what I'm doing wrong in running the test individually (just using the `pytest -k filepath` command), it works when running the entire test suite?

jacqueswww:
Very strange, will take a proper look tomorrow. But seems like something changed in eth_tester. Maybe try a different (older) version?

jakerockland:
Hrm, sounds good, I'll fiddle with it 😅

fubuloubu:
@njgheorgita can help?

examples/auctions/blind_auction.vy
```diff
@@ -0,0 +1,183 @@
...
+        if (i >= _numBids):
+            break
+
+        # Get bid to check
+        bidToCheck: {blindedBid: bytes32, deposit: wei_value} = (self.bids[msg.sender])[i]
+
+        # Check against encoded packet
+        value: wei_value = _values[i]
+        fake: bool = _fakes[i]
+        # TODO: the block below can eventually be removed and boolean `fake`
```
jakerockland: @fubuloubu @jacqueswww finished adding pretty comprehensive tests for this example, only outstanding thing is addressing this TODO, which I will do once #1077 is merged, but otherwise I think the example is ready for review 😄

jacqueswww:
@jakerockland Good stuff! I am quickly mergin `map()` PR, after that update I will merge this PR.

jakerockland:
Sounds great @jacqueswww. Can you ping me when you merge that and I'll make sure to update the example this PR with the latest syntax?

PR #1071, opened 2018-11-09
implemenation of units in ABI

### LINKED ISSUE #983: Add units to ABI output
## Simple Summary
We have units. It would be nice to add them to the ABI for use externally

## Abstract
Add `"units":"Full unit name"` to input listings, output listings, and event listings. As this is a optional readability enhancement for programmers using the smart contract through it's ABI, the full name is better to use than it's shorthand.

## Motivation
Units are a great addition to Vyper for internal use -- it reduces occurrences of incompatible units being mixed with each other in Vyper programs -- so why not extend that to external users of the programs by having the units available alongside type information. This way, front end programmers and CLI interactions can have a bit more context without having to read the program e.g. "Is this input wei or ether?"

Please note this would be an OPTIONAL part of the the ABI specification (not that one exists), similar to `gas`

## Specification
Vyper program:
```python
units: {
    m: "Meter",
    s: "Second"
}

@public
def velocity(d: uint256(m), t: uint256(s)) -> uint256(m/s):
    return d / t
```
returns ABI:
```json
[
    {
        "constant": false,
        "gas": 442,
        "inputs": [
            {
                "name": "d",
                "type": "uint256",
                "unit": "Meter"
            },
            {
                "name": "t",
                "type": "uint256",
                "unit": "Second"
            }
        ],
        "name": "velocity",
        "outputs": [
            {
                "name": "out",
                "type": "uint256",
                "unit": "Meter per Second"
            }
        ],
        "payable": false,
        "type": "function"
    }
]
```
We can change some of the conjugate units using a lexical conversion of their operators:
```python
N * m  # "Newton-Meter"
m / s  # "Meter per Second"
m / s**2  # "Meter per Second squared"
N * m / s**2  # "Newton-Meter per Second squared"
```

## Backwards Compatibility
No backwards incompatibilities

## Copyright

evgeniuz:
### - What I did
Added support for units in ABI output (function inputs/outputs and events). Fixes #983.
### - How I did it
Used `print_unit` on `type.unit` where relevant in places where ABI output is assembled.
### - How to verify it
Create contract with units in it, compile it with `-f abi`
### - Cute Animal Picture
snek

### DIFF
--- a/tests/parser/features/test_units.py
+++ b/tests/parser/features/test_units.py
@@ -1,8 +1,3 @@
-from decimal import Decimal
-
-from vyper.exceptions import TypeMismatchException, EventDeclarationException
-
-
 def test_function_with_units(get_contract_with_gas_estimation):
     code = """
 units: {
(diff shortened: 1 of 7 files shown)


evgeniuz:
Added basic support for units in functions/events. Lexical conversion is a bit more complicated, trying to figure out how to pass detailed descriptions, as `custom_units` only has unit names in it.

evgeniuz:
Passed `custom_unit_descriptions` from context through `to_abi_dict`. This is mostly complete, going to add removal of `unit` from ABI if it's empty (https://github.com/ethereum/vyper/issues/983#issuecomment-437937224) and unit tests.

evgeniuz:
Now removing unit from ABI if empty. All current tests are passing except single one that used `sec` unit, changed it to accommodate new behaviour.

Changes are ready for review, only thing left to do is to add testing.

evgeniuz:
Added tests.

tests/parser/features/test_units.py
```diff
@@ -0,0 +1,168 @@
...
+    return f * d / (t * t)
+
+@public
+def bar(a: uint256(m), b: uint256(m), c: uint256(m)) -> uint256(m**3):
+    return (a * b * c)
+    """
+
+    c = get_contract_with_gas_estimation(code)
+
+    assert c._classic_contract.abi == [{
```
fubuloubu: Checking the entire ABI like this is probably fragile. I would suggest checking subsets of the ABI by breaking it out. In particular, gas values might change. No reason to assert that.

vyper/signatures/function_signature.py
```diff
@@ -136,28 +139,41 @@ def from_definition(cls, code, sigs=None, custom_units=None, contract_def=False,
...
+
+        for abi_output in abi_outputs:
+            delete_unit_if_empty(abi_output)
 
-        return [{"type": x, "name": "out"} for x in res]
+        return abi_outputs
 
-    def to_abi_dict(self):
-        return {
+    def to_abi_dict(self, custom_units_descriptions={}):
```
jacqueswww: Please use custom_units_descriptions=None and then `if custom_units_descriptions is None ... ` instead. https://docs.quantifiedcode.com/python-anti-patterns/correctness/mutable_default_value_as_argument.html

Mutable default parameters can cause a lot of havoc ;)

Issue #1064, opened 2018-11-02
VIP: Allow `bool` as an input type to built-in `convert` function


jakerockland:
## Simple Summary
Adding `bool` as an allowed input type to the built-in `convert` function so that one can do for example:

```python
flagSet: bool = True
flagSetInt: int128 = convert(flagSet, "int128")
```

## Abstract
Currently the built in `convert` function is specified as:

def convert(a, b) -> c:
  """
  :param a: value to convert
  :type a: either decimal, int128, uint256 or bytes32
  :param b: the destination type to convert to
  :type b: type of either decimal, int128, uint256 or bytes32

  :output c: either decimal, int128, uint256 or bytes32
  """
 
I'm proposing we re-specify it as:

def convert(a, b) -> c:
  """
  :param a: value to convert
  :type a: either bool, decimal, int128, uint256 or bytes32
  :param b: the destination type to convert to
  :type b: type of either decimal, int128, uint256 or bytes32

  :output c: either decimal, int128, uint256 or bytes32
  """

Where True <-> 1 and False <-> 0.

## Motivation
There are situations in which it would be convenient to be able to translate a boolean value into an integer or byte value to be able to pack it into an array of bytes with other data. Currently this seems to only be achievable using conditional logic, whereas the built in `convert` function supports similar behavior for other input types.

## Specification
The technical specification for this is a re-iteration of what is in "Abstract" above.  

## Backwards Compatibility
This VIP is backwards compatible.

## Copyright
Copyright and related rights waived via [CC0](https://creativecommons.org/publicdomain/zero/1.0/)

fubuloubu:
I think this one is probably okay to move forward with, another related one is #1074

It should basically just be possible to convert any basic (32 byte) type to any other basic type.

jakerockland:
Awesome, I’ll start on tackling both. 😄👌

jakerockland:
@fubuloubu quick question, when working on functionality like this, what is the best way to run a single test case that I've written, as opposed to running the entire test suite?

fubuloubu:
`pytest -k path/to/test_case.py`

PR #1077, opened 2018-11-13
Allow bool as an input type to built-in convert function

jakerockland:
### - What I did

Proposed solution to: #1064

### - How I did it

Updated `vyper/types/convert.py` to support boolean inputs.

### - How to verify it

I've added corresponding tests for the output types that can be convereted to from boolean values, so they can be run with:

`make test`

Updated built-in convert function to allow bool as an input type

### - Cute Animal Picture

image


tests/parser/types/test_bytes.py
```diff
@@ -4,6 +4,19 @@
...
+    exp_code = """
+@public
+def testConvertBytes32(flag: bool) -> bytes32:
+    flagBytes: bytes32 = convert(flag, bytes32)
+    return flagBytes
+    """
+
+    c = get_contract_with_gas_estimation(exp_code)
+    assert c.testConvertBytes32(False) == b"\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00"
+    assert c.testConvertBytes32(True) == b"\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x01"
```
jacqueswww: I think just slicing `c.testConvertBytes32(True)[-1] == b'x01'` here would be more readable. If we want make sure about the 32 byte padding perhaps just an assert len() == 32 as well.

jakerockland:
@jacqueswww Yeah that's a great point. I included the entire 32 bytes to make sure the padding was as expected, but I think making that change for readability and adding the second assert is the way to go. Just pushed a commit.

jakerockland:
Fixes #1064

jakerockland:
Added one last commit to fix style guideline issue that was throwing Travis CI off.

fubuloubu:
@jakerockland, you're a beast!

jakerockland:
Realized that there are already test files for conversions in the `tests/parser/functions` directory. I think some refactoring is still due per #1084, but think it would be better to refactor as I did in recent commits as opposed to all into one file.

PR #1081, opened 2018-11-14
Set valency to 0 on return statements.

jacqueswww:
### - What I did
Set valency to 0 on return statements.

### DIFF
--- a/tests/parser/features/decorators/test_private.py
+++ b/tests/parser/features/decorators/test_private.py
@@ -476,3 +476,30 @@ def a() -> bool:
     c2 = get_contract(code2)
     assert c1.a() is True
     assert c2.a() is True
+
+
+def test_private_nested_if_return(get_contract):
+    code = """
+
+@private
+def _test(z: int128) -> bool:
+    y: int128 = 1
+
+    if (z <= 0):
+        return True
+    else:
+        y = 2
+
+    return False
+
+
+@public
+def test(z: int128) -> bool:
+    return self._test(z)
+    """
+
+    c = get_contract(code)
+
+    assert c.test(-1) is True
+    assert c.test(0) is True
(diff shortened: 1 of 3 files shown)


fubuloubu:
What the heck is valency? And what problem was actually solved with this PR? I noticed something with private returns that have branches

jakerockland:
Am also unclear on what exactly valency refers to here so would love an explainer 😅 

To make it more clear what this solves though @fubuloubu previously this was the situation:

```python
# Some sort of valency problem (this doesn't compile)
@private
def foo(input: int128) -> bool:
    if (input > 100):
        return False

    return True


# Workaround (this compiles)
@private
def bar(input: int128) -> bool:
    result: bool

    if (input > 100):
        result = False
    else:
        result = True

    return result
```

jakerockland:
Hrm yeah if there is good reason to prevent returning from within a branch, it may be worth opening another issue to better document that somewhere/to try to catch that at compile time and throw an exception.

fubuloubu:
It's a little late for me to jump down the rabbit hole on this, but #590 mentions it

jacqueswww:
@jakerockland @fubuloubu valency indicates that a value is pushed on the stack. Before the jump-to-private change a return statement obviously had a valency of 0. But because of the jumping, we have to override valency and make sure the stack is in a sane state. This is where seq_unchecked (seq without an additional POP) and the likes come in.

With the above bug, the private return statement had a valency of 1 which miss-matched. With the above fix I basically override it to 0, which is the same as the traditional RETURN op code.

fubuloubu:
Can we call it "stack depth" or something? Valency makes me thing of chemistry _shudders_ 😂🤣

jakerockland:
Thanks for the explainer @jacqueswww, that makes more sense to me now.

PR #1082, opened 2018-11-15
Allow converting sliced bytes to uint256

jakerockland:
### - What I did

Fixes #1072 (this may currently be only a partial fix, pending discussion on issue)

### - How I did it

Updated `convert.py` to allow converting to `uint256` using existing internal `byte_array_to_num` method that supports sliced bytes. Added relevant test to `test_uint256.py`.

### - How to verify it

`make test`

### - Description for the changelog

Added support for converting sliced `bytes` to `uint256`.

### - Cute Animal Picture

image

PR #1085, opened 2018-11-15
Do not allow conversion from bytes array > 32 to int128 at compile time

jakerockland:
### - What I did

Related somewhat to conversation on #1082. Catch at compile-time if trying to convert `bytes` array of max length greater than 32 to `int128` type.

### - How I did it

Updated `convert.py` to do a check on max length when converting from bytes type (my understanding as to why `SizeLimits.in_bounds()` doesn't work for `bytes` type is that it is being passed by reference not by value?).

Added relevant tests to `test_int128.py`

### - How to verify it

`make test`

Patch to prevent conversion from bytes array greater than 32 bytes to int128.

### - Cute Animal Picture

image

### DIFF
--- a/vyper/types/convert.py
+++ b/vyper/types/convert.py
@@ -38,13 +38,13 @@ def to_int128(expr, args, kwargs, context):
             ['clamp', ['mload', MemoryPositions.MINNUM], in_node,
                         ['mload', MemoryPositions.MAXNUM]], typ=BaseType('int128', in_node.typ.unit), pos=getpos(expr)
         )
-    
+
     elif input_type is 'bytes':
         if in_node.typ.maxlen > 32:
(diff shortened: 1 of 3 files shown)


vyper/types/convert.py
```diff
@@ -30,23 +30,28 @@
 @signature(('uint256', 'bytes32', 'bytes', 'bool'), '*')
 def to_int128(expr, args, kwargs, context):
     in_node = args[0]
-    typ, len = get_type(in_node)
-    if typ in ('uint256', 'bytes32'):
+    input_type, len = get_type(in_node)
```
jacqueswww: Just change len to underscore and we can merge ;)

jakerockland: @jacqueswww done! 🙌 😅

Issue #1072, opened 2018-11-09
Converting sliced bytes to non-bytes32 doesn't work


fubuloubu:
### Version Information
* vyper Version: 0.1.0b4

### What's your issue about?
Vyper doesn't like converting a slice of `bytes` (slice len <=32) to a type other than `bytes32`:

```python
# Compiles!
@public
@constant
def foo(_arg: bytes[100]) -> bytes32:
    return convert(slice(_arg, start=20, len=32), bytes32)

# Does not compile....
@public
@constant
def bar(_arg: bytes[100]) -> uint256:
    return convert(slice(_arg, start=20, len=32), uint256)

# Compiles! (Workaround)
@public
@constant
def baz(_arg: bytes[100]) -> uint256:
    return convert(convert(slice(_arg, start=20, len=32), bytes32), uint256)

# Note: above does not work for addresses!
```

### How can it be fixed?
Make it so!

#### Cute Animal Picture

![fluffy pony](https://i.ytimg.com/vi/sNv1s5lAZts/maxresdefault.jpg)

jakerockland:
@fubuloubu I'm starting to look into this one. One note for now, this guy also compiles:

```
# Compiles!
@public
@constant
def who(_arg: bytes[100]) -> int128:
    return convert(slice(_arg, start=20, len=32), int128)
```

Seems like we may want to re-name this issue to be more specific to converting from `bytes` to `int128`, though perhaps we do want to also add converting from `bytes` to `address`? Though perhaps not converting from `bytes` to `decimal` as converting from `bytes32` to `decimal` is also not currently supported. Thoughts?

jakerockland:
Yeah it definitely is a bit of a rabbit hole that has been unearthed haha 😂. Happy to work on that table though yeah, that sounds like a good plan of action to me 👍

PR #1087, opened 2018-11-17
Fix bug with convert(2**256 - 1, int128)

jakerockland:
### - What I did

In the process of starting work on a better spec-ed table of expected conversion outcomes per @fubuloubu and I's conversation on #1072, I noticed some pretty strange behavior that allowed for converting from _some_ uint256 that were outside of the maximum allowed range for `int128`.

More specifically I found that when running this test:

```python
def test_convert_from_uint256(get_contract_with_gas_estimation):
    code = """
@public
def test(foo: uint256) -> int128:
    return convert(foo, int128)
    """

    assert c.test(0) == 0
    assert c.test(2**127 - 1) == 2**127 - 1

    assert c.test(2**256 - 1) == -1 # FIXME: This line I expected to throw a runtime exception but didn't.
    assert c.test(2**127) # FIXME: Whereas this line through the expected runtime exception.
```

Can open an issue if necessary, though I think this PR may just serve as both the issue and the solution to the issue 😅 

### - How I did it

Changed psuedo-op code being used on convert to do a clamp between 0 and maximum unsigned integer.

### - How to verify it

Review the tests added to `test_convert_int128.py` and run `make test`

Fixed bug that allowed `convert(2**256 -1, int128)`

### - Cute Animal Picture

image

### DIFF
--- a/tests/parser/functions/test_convert_int128.py
+++ b/tests/parser/functions/test_convert_int128.py
@@ -63,6 +63,54 @@ def from_bool(flag: bool) -> int128:
     assert c.from_bool(True) == 1
 
 
+def test_convert_from_uint256(get_contract_with_gas_estimation):
+    code = """
+@public
+def test(foo: uint256) -> int128:
+    return convert(foo, int128)
+    """
+
+    c = get_contract_with_gas_estimation(code)
+    assert c.test(0) == 0
+    assert c.test(2**127 - 1) == 2**127 - 1
+
+
+def test_out_of_range_from_uint256(assert_tx_failed, get_contract_with_gas_estimation):
+    code = """
+@public
+def test(foo: uint256) -> int128:
+    return convert(foo, int128)
+    """
+
+    c = get_contract_with_gas_estimation(code)
+    assert_tx_failed(lambda: c.test(2**127))
+    assert_tx_failed(lambda: c.test(2**256 - 1))
+
+
+def test_out_of_range_from_uint256_at_compile(assert_compile_failed, get_contract_with_gas_estimation):
+    code = """
+@public
+def test() -> int128:
+    return convert(2**127, int128)
+    """
+
+    assert_compile_failed(
+        lambda: get_contract_with_gas_estimation(code),
+        InvalidLiteralException
+    )
+
+    code = """
+@public
+def test() -> int128:
+    return convert(2**256 - 1, int128)
+    """
+
+    assert_compile_failed(
+        lambda: get_contract_with_gas_estimation(code),
+        InvalidLiteralException
+    )
+
+
 def test_convert_bytes32_to_num_overflow(assert_tx_failed, get_contract_with_gas_estimation):
     code = """
 @public
--- a/vyper/types/convert.py
+++ b/vyper/types/convert.py
@@ -31,7 +31,7 @@ from vyper.utils import (
 def to_int128(expr, args, kwargs, context):
     in_node = args[0]
     input_type, _ = get_type(in_node)
-    if input_type in ('uint256', 'bytes32'):
+    if input_type is 'bytes32':
         if in_node.typ.is_literal and not SizeLimits.in_bounds('int128', in_node.value):
             raise InvalidLiteralException("Number out of range: {}".format(in_node.value), expr)
         return LLLnode.from_list(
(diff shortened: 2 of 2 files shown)


vyper/types/convert.py
```diff
@@ -31,13 +31,21 @@
 def to_int128(expr, args, kwargs, context):
     in_node = args[0]
     typ, len = get_type(in_node)
-    if typ in ('uint256', 'bytes32'):
+    if typ is 'bytes32':
         if in_node.typ.is_literal and not SizeLimits.in_bounds('int128', in_node.value):
             raise InvalidLiteralException("Number out of range: {}".format(in_node.value), expr)
         return LLLnode.from_list(
             ['clamp', ['mload', MemoryPositions.MINNUM], in_node,
                         ['mload', MemoryPositions.MAXNUM]], typ=BaseType('int128', in_node.typ.unit), pos=getpos(expr)
```
jacqueswww: @jakerockland very strange that this clamp did not kick in? The limit below doesn't take the min into account.

jakerockland: @jacqueswww 

Yeah it is quite strange, I spent a good while tinkering with it trying to understand _why_ that is the case that `['clamp', ['mload', MemoryPositions.MINNUM], in_node, ['mload', MemoryPositions.MAXNUM]]` catches the fail case of (2**127) but not (2**256 - 1) and was getting pretty mind boggled. 

I think I may need some direction in being pointed to where the pseudo-op codes are implemented and assembled to their composure of op-codes? Or just some direction more generally of where to look deeper into this?

I do (probably biasedly haha) think that my proposed solution here of `['uclample', in_node, SizeLimits.MAXNUM]` is a bit more elegant/readable given that we know the input is unsigned. 

Do you know the performance differences of `uclample` vs `clamp` or where I could find some reference to that/dig deeper?

jacqueswww: The code for clamp can be found in `https://github.com/ethereum/vyper/blob/master/vyper/compile_lll.py#L254`; and yes I agree your approach is probably better  & is less gas.

jakerockland: Aha that is exactly what I was hoping to dig into in more detail. Thanks so much @jacqueswww 😄

jacqueswww: @jakerockland any luck figuring out why that clamp didn't work before?

jakerockland: @jacqueswww Just took another look at this with fresh eyes and actually I believe so but am not 100%.

So I think the real problem with what the behavior without this PR is was not that `clamp` is broken somehow, but rather that the bounds of the `clamp` weren't correct. 

To make this clear, you can replace `['uclample', in_node, SizeLimits.MAXNUM]` on my branch, with either `['clamp', 0, in_node, ['mload', MemoryPositions.MAXNUM]]` or `['clamp', 0, in_node, SizeLimits.MAXNUM]` and the tests I added to catch the error behavior I found pass as expected.

I think clamping between the min and max of the `int128` type isn't safe when converting from `uint256` because when calling a function like this below the expression `2**256 - 1` is evaluated to be -1 and then the -1 is clamped?

```python
    code = """
@public
def test(foo: uint256) -> int128:
    return convert(foo, int128)
    """

    assert c.test(2**256 - 1) == -1
```

Does that seem like what is going on to you? 

Also one thing that I'm a bit unclear about is under what circumstances (lets say limited to conversions for now) one would want to do this `['clamp', ['mload', MemoryPositions.MINNUM], in_node, ['mload', MemoryPositions.MAXNUM]]` instead of this `['clamp', SizeLimits.MINNUM, in_node, SizeLimits.MAXNUM]`?

Given all this, for the interest of this PR it seems we _could_ switch things to use `clamp` instead of `uclample`, with the lower bound fixed, but if the latter is better for other reasons (gas usage) it seems both more clear and more efficient to use that.

jakerockland: @jacqueswww Just pushed a merge conflict resolution, looking at this again with the rest of the upstream changes merged in, there are definitely other areas where the clamping going on as-is is a bit wonky/unnecessary (ex. with clamping a bool when converting from bool) which is somewhat my fault due to previous commits.

First thing I will do as part of work towards #1093 is cleaning up the clamping currently going on in type conversions. This can either be its own PR or part of WIP towards #1093.

tests/parser/functions/test_convert_int128.py
```diff
@@ -1,8 +1,61 @@
 from vyper.exceptions import (
+<<<<<<< HEAD
```
fubuloubu: Checked in a merge conflict!

jakerockland: @fubuloubu I believe that's on an old commit 😅I messed up and committed `8bfebe1` before I should have but fixed with `caaae76` which is latest. I can squash though all the commits here if you'd like and update the history?

fubuloubu: Yep, squash it plz!

vyper/types/convert.py
davesque: Maybe a side comment, but shouldn't comparisons of string values use `==` and not `is`?  I know Python can intern strings but I don't think that applies in this case.

tests/parser/functions/test_convert_int128.py
jakerockland: Haha already did! 😅

vyper/types/convert.py
jacqueswww: @jakerockland as for the initial problem, looks like we seriously have to look into fixing constant folding.

PR #1089, opened 2018-11-19
642 standardise compile call

### LINKED ISSUE #642: VIP: Standardize compiler output artifacts

jacqueswww:
### - What I did

Fixes #642, friendlier compile function.

### DIFF
--- a/bin/vyper
+++ b/bin/vyper
@@ -32,37 +32,6 @@ def uniq(seq):
     return [x for x in seq if not (x in exists or exists.add(x))]
 
 
-def get_asm(asm_list):
-    output_string = ''
-    skip_newlines = 0
-    for node in asm_list:
-        if isinstance(node, list):
-            output_string += get_asm(node)
-            continue
-
-        is_push = isinstance(node, str) and node.startswith('PUSH')
-
-        output_string += str(node) + ' '
-        if skip_newlines:
-            skip_newlines -= 1
-        elif is_push:
-            skip_newlines = int(node[4:]) - 1
-        else:
-            output_string += '\n'
-    return output_string
-
-
-def get_source_map(code):
-    asm_list = compile_lll.compile_to_assembly(optimizer.optimize(parse_to_lll(code, runtime_only=True)))
-    c, line_number_map = compile_lll.assembly_to_evm(asm_list)
-    # Sort line_number_map
-    out = OrderedDict()
-    keylist = line_number_map.keys()
-    for k in sorted(keylist):
-        out[k] = line_number_map[k]
-    return json.dumps(out)
-
-
 def get_source_map_asm_list(code):
     line_no_width = 6
     asm_list = compile_lll.compile_to_assembly(optimizer.optimize(parse_to_lll(code, runtime_only=True)))
--- a/bin/vyper
+++ b/bin/vyper
@@ -6,11 +6,8 @@ import sys
 import vyper
 import warnings
 
-from collections import OrderedDict
-from vyper import compiler, optimizer
-from vyper.parser.parser import parse_to_lll
-from vyper.parser import parser_utils
-from vyper import compile_lll
+from vyper import compiler
+
 
 warnings.simplefilter('always')
 sys.tracebacklimit = 0
@@ -20,7 +17,7 @@ if tb_limit:
 
 parser = argparse.ArgumentParser(description='Vyper programming language for Ethereum')
 parser.add_argument('--version', action='version', version='{0}'.format(vyper.__version__))
-parser.add_argument('input_file', help='Vyper sourcecode to compile')
+parser.add_argument('input_files', help='Vyper sourcecode to compile', nargs='+')
 parser.add_argument('-f', help='Format to print', default='bytecode', dest='format')
 parser.add_argument('--show-gas-estimates', help='Show gas estimates in ir output mode.', action="store_true")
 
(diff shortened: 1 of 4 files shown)


fubuloubu:
When compiling multiple files, no context is given for an error.

Suggestion: always show file name when displaying an error.

```bash
$ vyper examples/tokens/*.vy
vyper.exceptions.TypeMismatchException: line 26:4 Typecasting from base type uint256 to int128 unavailable
```

fubuloubu:
Also, can we add the format options to the help output? Maybe even restrict the options in argparse?

fubuloubu:
Lastly, add the compiler version to the `combined_json` output e.g. `{"contracts":{ current output },"version":"0.2.0-beta.4"}`

bin/vyper
```diff
@@ -41,7 +41,8 @@ if __name__ == '__main__':
 
     # Combined json output
     if args.format == 'combined_json':
-        out = compiler.compile(codes, ['bytecode', 'bytecode_runtime', 'abi', 'source_map'], output_type='dict')
+        out = vyper.compile(codes, ['bytecode', 'bytecode_runtime', 'abi', 'source_map'], output_type='dict')
```
fubuloubu: I was picturing that the `vyper.compile` interface would only take a single source program and compile it, not multiple. Can multiple files be handled here in this script instead?

e.g.
```python
out = {}
for filename, code in codes.items():
    out[filename] = vyper.compile(code, returns=['bytecode', 'bytecode_runtime', 'abi', 'source_map'])
```

bin/vyper
```diff
@@ -53,7 +54,7 @@ if __name__ == '__main__':
         for f in uniq(args.format.split(',')):
             formats.append(translate_map.get(f, f))
 
-        out_list = compiler.compile(codes, formats, output_type='list')
+        out_list = vyper.compile(codes, formats, output_type='list')
```
fubuloubu: Can `output_type` logic be handled in this script instead?. e.g. return it as a dict and handle list conversion here.

vyper/exceptions.py
```diff
@@ -78,3 +78,15 @@ class FunctionDeclarationException(ParserException):
 
 class EventDeclarationException(ParserException):
     pass
+
+
+class ContractException(Exception):
```
fubuloubu: RE: handling multiple files with `vyper.compile` - I don't think this is the clearest way to solve this. Can handle exceptions in the script and log them there. "issue while compiling X:"

bin/vyper
jacqueswww: Yes but then we loose capability of handling multiple inputs on the vyper module. That's why I chose to make the vyper binary skinny instead.

bin/vyper
jacqueswww: As mentioned above, I prefer having this as part of the vyper module, otherwise nobody else can do multiple input / multiple output (vyper-debug for example). 
As it stands compile does support a single string input.

vyper/exceptions.py
jacqueswww: I think actually making the `compile` interface return only a data structure on compilation faults might be even cleaner for integration of the compiler module?

jacqueswww: See below commit, I think this solves both scenarios ;)

fubuloubu:
I'll try to review this today

jacqueswww:
`combined_json`

bin/vyper
```diff
@@ -20,7 +17,7 @@ if tb_limit:
 
 parser = argparse.ArgumentParser(description='Vyper programming language for Ethereum')
 parser.add_argument('--version', action='version', version='{0}'.format(vyper.__version__))
-parser.add_argument('input_file', help='Vyper sourcecode to compile')
+parser.add_argument('input_files', help='Vyper sourcecode to compile', nargs='+')
 parser.add_argument('-f', help='Format to print', default='bytecode', dest='format')
```
fubuloubu: can you add a list of options here so the help displays what possible choices are?

vyper/exceptions.py
fubuloubu: Yeah, I like that much better

bin/vyper
jacqueswww: @fubuloubu  I think we are ready to merge now? :man_dancing:

PR #1094, opened 2018-11-22
Support BoolOps for > 2 arguments

### LINKED ISSUE #1075: BoolOps cannot accept >2 arguments
### Version Information

* vyper Version: 0.2.0b4

### What's your issue about?
```python
# Fails: Expected two arguments for a bool op
def foo():
    a: bool = True and True and False

# Fails: Expected two arguments for a bool op
def foo():
    a: bool = True or True or False

# This passes for some reason lol
def foo():
    a: bool = True or True and False

# This also passes lol
def foo():
    a: bool = True and (True and False)
```

### How can it be fixed?
Handle chained BoolOps similar to how regular Ops work (addition etc)

#### Cute Animal Picture
insane rhino

jakerockland:
### - What I did

Implemented support for BoolOps with greater than two arguments. Closes #1075 

### - How I did it

Made sure all expressions in [BoolOps `values`](https://docs.python.org/3.7/library/ast.html#abstract-grammar) were iterated through and handled.

### - How to verify it

Review test file `test_bool_ops.py` and run `make test`.

Add support for BoolOps with greater than two arguments

### - Cute Animal Picture

image

### DIFF
--- a/tests/parser/syntax/test_bool_ops.py
+++ b/tests/parser/syntax/test_bool_ops.py
@@ -18,6 +18,6 @@ def foobar() -> bool:
     """
 
     c = get_contract_with_gas_estimation(code)
-    assert c.foo() == False
-    assert c.bar() == True
-    assert c.foobar() == False
+    assert c.foo() is False
+    assert c.bar() is True
+    assert c.foobar() is False
--- a/tests/parser/syntax/test_bool_ops.py
+++ b/tests/parser/syntax/test_bool_ops.py
@@ -1,0 +1,23 @@
+
+def test_convert_from_bool(get_contract_with_gas_estimation):
+    code = """
+@public
+def foo() -> bool:
+    val: bool = True and True and False
+    return val
+
+@public
+def bar() -> bool:
+    val: bool = True or True or False
+    return val
+
+@public
+def foobar() -> bool:
+    val: bool = False and True or False
+    return val
+    """
+
+    c = get_contract_with_gas_estimation(code)
+    assert c.foo() == False
+    assert c.bar() == True
+    assert c.foobar() == False
--- a/tests/parser/syntax/test_bool_ops.py
+++ b/tests/parser/syntax/test_bool_ops.py
@@ -15,9 +15,27 @@ def bar() -> bool:
 def foobar() -> bool:
     val: bool = False and True or False
     return val
+
+@public
+def oof() -> bool:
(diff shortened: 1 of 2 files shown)


vyper/parser/expr.py
```diff
@@ -586,22 +586,47 @@ def load_bytearray(side):
...
+            # Check for calls at assignment
+            if self.context.in_assignment and isinstance(value, ast.Call):
+                raise StructureException("Boolean operations with calls may not be performed on assignment", self.expr)
+
+            # Check for boolean operations with non-boolean inputs
+            _expr = Expr.parse_value_expr(value, self.context)
+            if not is_base_type(_expr.typ, 'bool'):
+                raise TypeMismatchException("Boolean operations can only be between booleans!", self.expr)
+
+            # TODO: Handle special case of literals and simplify at compile time
```
jakerockland: In the interest of keeping changes modular, not sure if this is something that should be implemented as part of this PR, or as a separate PR after this one is merged.

jacqueswww: I think we should leave this out for now. The plan is to make constant folding it's own step that happens before it even gets to this translation step :)

tests/parser/syntax/test_bool_ops.py
```diff
@@ -0,0 +1,23 @@
...
+    return val
+
+@public
+def bar() -> bool:
+    val: bool = True or True or False
+    return val
+
+@public
+def foobar() -> bool:
+    val: bool = False and True or False
```
jacqueswww: Please add a test case for like 5 ops or something both even & odd number of ops ;)

vyper/parser/expr.py
jakerockland: Sounds good to me! 😊 Should Ieave the comment there for now for posterity or want me to pull it out?

tests/parser/syntax/test_bool_ops.py
jakerockland: Sounds good will do soon! 😊👌

jacqueswww: Leave it, might motivate us at some point? :P

vyper/parser/expr.py
jacqueswww: Leave it, might motivate us to do it sooner.

jakerockland:
@jacqueswww added a few more tests! 😄

PR #1096, opened 2018-11-26
Update ERC721 token

nrryuya:
### - What I did
Update example implementation of [ERC721](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-721.md) token. Specifically, 
- supported ERC165 (it's required in the EIP)
- added mint/burn function
- made it clear that assertions required in the EIP is implemented. 
- commented about other assertions not included in the EIP but needed.
- added tests with more test cases

### - How I did it
I referred to the previous implementation and [OpenZeppelin's Solidity implementation](https://github.com/OpenZeppelin/openzeppelin-solidity/blob/master/contracts/token/ERC721/ERC721.sol).

### - How to verify it
Test. Formal verification is [WIP](https://github.com/LayerXcom/verified-vyper-contracts/tree/master/erc721). 

Update example ERC721 token

### - Cute Animal Picture
image

### DIFF
--- a/examples/tokens/ERC721.vy
+++ b/examples/tokens/ERC721.vy
@@ -291,14 +291,14 @@ def setApprovalForAll(_operator: address, _approved: bool):
 ### MINT & BURN FUNCTIONS ###
 
 # @dev Function to mint tokens
-#      Throws if `msg.sender`` is not the minter.
+#      Throws if `msg.sender` is not the minter.
 #      Throws if `_to` is zero address.
 # @param to The address that will receive the minted tokens.
 # @param tokenId The token id to mint.
 # @return A boolean that indicates if the operation was successful.
 @public
 def mint(_to: address, _tokenId: uint256) -> bool:
-    # Throws if `msg.sender`` is not the minter
+    # Throws if `msg.sender` is not the minter
     assert msg.sender == self.minter
     # Throws if `_to` is zero address
     assert _to != ZERO_ADDRESS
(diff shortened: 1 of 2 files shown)


examples/tokens/ERC721.vy
```diff
@@ -62,18 +62,39 @@ ownerToNFTokenCount: uint256[address]
 # @dev Mapping from owner address to mapping of operator addresses.
 ownerToOperators: (bool[address])[address]
 
+# @dev Address of minter, who can mint a token
+minter: address
 
-# @dev Contract constructor. Per the Transfer event spec; during contract creation, any number of
-#      NFTs may be created and assigned without emitting Transfer.
+# @dev Mapping of interface id to bool about whether or not it's supported
+supportedInterfaces: bool[bytes32]
```
fubuloubu: Does `public` generate the correct getter `supportsInterface(_interfaceID: bytes32) -> bool`?

fubuloubu: hmm, yeah I guess "explicit is better than implicit", just seems like a waste of `public`'s utility

nrryuya:
Added a few minor fixes

examples/tokens/ERC721.vy
nrryuya: >seems like a waste of public's utility

Yeah that's also true.

fulldecent:
Please compare also to the reference implementation at https://github.com/0xcert/ethereum-erc721

PR #1117, opened 2018-12-04
Add security phases to interaction function

sgryphon:
### - What I did

Update example for safe remote purchase with security phases in the interaction function.

### - How I did it

I think the example is correct, although I haven't tested it (I haven't quite got viper running yet on my machine).

### - How to verify it

Create a contract address to participate in a sample old contract that triggers the exploit (recursive call into the received() function). Check the exploit works.

Try again with the updated example and see if it prevents the exploit.

Updated examples to follow good practice security phases for functions that have external interaction.

### - Cute Animal Picture

What is it with the animals?

### DIFF
--- a/examples/safe_remote_purchase/safe_remote_purchase.vy
+++ b/examples/safe_remote_purchase/safe_remote_purchase.vy
@@ -17,6 +17,7 @@ value: public(wei_value) #Value of the item
 seller: public(address)
 buyer: public(address)
 unlocked: public(bool)
+ended: public(bool)
 #@constant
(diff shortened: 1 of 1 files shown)


fubuloubu:
I like the explicit addition of this check in the code, vs the more implicit back end code I mentioned prior. Perhaps a compiler error could be added, but this is a more complex static analysis check that a tool like Slither could detect

jacqueswww:
As mentioned on the ticket, the contract was not vulnerable because of the `assert self.buyer == msg.sender` statement.

However I think it can't hurt to keep this flow consistent.

PR #1125, opened 2018-12-07
Make built-in constant op code generation more efficient

### LINKED ISSUE #1121: MAX[MIN]_INT128 makes error in range check
### Version Information

* vyper Version: [the latest](https://github.com/ethereum/vyper/commit/97ed5fec89947d41fd91f5190c4478ffcadbb9fb) and 0.1.0b4

### What's your issue about?

I got `TypeError` in compiling these sentences.
```
a: int128 = MAX_INT128
```
```
b: int128 = MIN_INT128
```
```
  File "/Users/ryuya.nakamura/.pyenv/versions/vyper/lib/python3.6/site-packages/vyper-0.1.0b4-py3.6.egg/vyper/utils.py", line 109, in in_bounds
    return cls.MINNUM <= value <= cls.MAXNUM
TypeError: '<=' not supported between instances of 'int' and 'str'
```

#### Cute Animal Picture

image

jakerockland:
### - What I did

Added better tests, refactored, and made op-code generation for built-in constants more efficient.

This also fixes #1121

### - How I did it

Built-in constants should use the values from `SizeLimits` in `utils.py`

### - How to verify it

`make test`

Make built-in constant op code generation more efficient.

### - Cute Animal Picture

image

### DIFF
--- a/tests/parser/types/numbers/test_constants.py
+++ b/tests/parser/types/numbers/test_constants.py
@@ -54,6 +54,59 @@ def test_arithmetic(a: int128) -> int128:
     assert c.test_arithmetic(5000) == 2**127 - 1 - 5000
 
 
+def test_builtin_constants_assignment(get_contract_with_gas_estimation):
+    code = """
+@public
+def foo() -> int128:
+    bar: int128 = MAX_INT128
+    return bar
+
+@public
+def goo() -> int128:
+    bar: int128 = MIN_INT128
+    return bar
+
+@public
+def hoo() -> bytes32:
+    bar: bytes32 = EMPTY_BYTES32
+    return bar
+
+@public
+def joo() -> address:
+    bar: address = ZERO_ADDRESS
+    return bar
+
+@public
+def koo() -> decimal:
+    bar: decimal = MAX_DECIMAL
+    return bar
+
+@public
+def loo() -> decimal:
+    bar: decimal = MIN_DECIMAL
+    return bar
+
+@public
+def zoo() -> uint256:
+    bar: uint256 = MAX_UINT256
+    return bar
+    """
+
+    c = get_contract_with_gas_estimation(code)
+
+    assert c.foo() == 2**127 - 1
+    assert c.goo() == -2**127
+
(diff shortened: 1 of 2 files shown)


jacqueswww:
@jakerockland this is more efficient for runtime execution, but create execution wil grow. Just something to keep in mind.

fubuloubu:
Our contract sizes are already pretty bloated haha. Is there a corresponding optimization that could reduce the impact here?

fubuloubu:
Wouldn't the code size be smaller if you store it in the bytecode actually?

jacqueswww:
this is roughly what the ASM looks like:
```
![image](https://user-images.githubusercontent.com/6917456/49807806-ad8ba580-fd63-11e8-8cab-687f35c028a2.png)

```

So an MLOAD (1byte opcode) + 1byte (memory address) .

jakerockland:
Hrm yeah that makes sense, I guess it is partially a question of how frequently one expects the same built-in constant to be used in a given contract? To me the runtime improvement seemed desirable from shaving off the extra `mload`s but maybe there is more to unpack here. #1134 is also relevant to this discussion.

jacqueswww:
Each byte in the CREATE costs 200 gas IIRC.

jacqueswww:
I completely agree, I presonally would prefer the runtime increase.

fubuloubu:
Sounds like there's a balance there. The "optimization" would be:
1. If the runtime size is very near to the EIP170 Limit, don't put it in the runtime X times
2. For all X usages of constant in Y functions, if Y gas costs are "reasonable" then the `mload` is preferred
3. If there there is heavy usage in a given function, making the gas costs "unreasonable" for that function, put it in the runtime instead

fubuloubu:
Hmm, I think the long term is to put this code into an "optimize" layer where this decision is made based on other metrics (like the ones I mentioned). Short term is to pick one or the other (and add a comment about these tradeoffs for later)

jakerockland:
@fubuloubu Yeah that makes sense to me! In the short term my vote is for runtime efficiency hehe 😉

PR #1126, opened 2018-12-07
Implemented missing decimal conversions

jakerockland:
### - What I did

Implemented decimal conversions not yet implemented based on #1093. This is building off #1111. 

### - How I did it

Added cases to support additional input types in `to_decimal` method in `convert.py`. Added appropriate tests to test new behavior.

### - Description for the changelog

Support decimal conversions for all basic type inputs.

### - Cute Animal Picture

image

PR #1134, opened 2018-12-11
Optimized conversion clamping

jakerockland:
### - What I did

Optimized clamping for conversions. Work towards #1093 

### - How I did it

Removed unnecessary `mload` instructions.

### - How to verify it

Tests haven't changed, so:

`make test`

Optimized conversion clamping.

### - Cute Animal Picture

image


jacqueswww:
@jakerockland @fubuloubu see now it gets difficult, I think for clamps we should probably use `mload` always. These could seriously effect your bytesize.

fubuloubu:
Great point, exactly why it's a balance that can be optimized lol. The clamps themselves can be optimized if proven unnecessary for a given statement

jakerockland:
Yeah that makes sense @fubuloubu @jacqueswww. Will make mental note of that to update #1126 and to make a new PR makes sure `mload` rather than literals consistency across conversion clamping. 👍

PR #1138, opened 2018-12-12
Conversion clamp optimizations

jakerockland:
### - What I did

This is built off of #1126 and related to conversation on #1137 and #1134. This optimization makes it such that we never add erroneous clamps for literals that we already know at compile-time are within bounds.

### - How I did it

After checking if literal, if not throwing an out of bounds error, can do conversion without clamping.

### - How to verify it

`make test`

Optimize to never add clamps with literals

### - Cute Animal Picture

image


fubuloubu:
Would a better optimization of converting a literal be to actually do the conversion at compile time and have it be a literal of that different type?

jakerockland:
@fubuloubu I'm a bit confused, isn't that what this PR does?

fubuloubu:
Just read it at a high level so somewhat my bad, but it doesn't appear to cover the case I was actually thinking of:

This is what I was thinking:
```python
FOO: constant(decimal) = 100.0

def bar() -> uint256:
    return convert(FOO, uint256)
```

In this case, `bar()` should just return a constant of 100 that is handled as a `uint256` in the code since we can do the conversion during compilation and know that the literal does not exceed the bounds.

jakerockland:
@fubuloubu haha so funny enough, that code throws an error:

```
vyper.exceptions.TypeMismatchException: line 1: Invalid value for constant type, expected decimal
```

It looks like there is an unrelated bug with custom decimal constants. Opening up a separate issue.

However, given the code:

```python
FOO: constant(int128) = 100

@public
def bar() -> uint256:
    return convert(FOO, uint256)
```

We get the compiled LLL of:

```
[seq,
  [return,
    0,
    [lll,
      [seq,
        # Line 3
        [if,
          [eq, [mload, 0], '4273672062' <bar()>],
          [seq,
            # Line 5
            [mstore, 0, [clampge, 100, 0]],
            [return, 0, 32],
            # Line 3
            stop]],
      0]]]
```

Whereas with the code:


```python
@public
def bar() -> uint256:
    return convert(100, uint256)
```

We get the compiled LLL of:

```
[seq,
  [return,
    0,
    [lll,
      [seq,
        # Line 1
        [if,
          [eq, [mload, 0], '4273672062' <bar()>],
          [seq,
            # Line 3
            [mstore, 0, 100],
            [return, 0, 32],
            # Line 1
            stop]],
      0]]]
```

If I understand what you're suggesting, is a valid re-scoping of the problem just that constants should be treated like literals?

If so, that seems like a different optimization all-together and likely should be its own PR?

fubuloubu:
Ah, that is a fair point (constants should be treated like literals). I would think this might make a better VIP suggestion to make sure it doesn't steamroll anything else
---
A better way to scope my corner case is:
```python
@public
def bar() -> uint256:
    return convert(100.0, uint256)
```
Does the decimal literal have to be clamped? Or can `convert(100.0, uint256)` be interpreted as `100`?

Does that make sense?

jakerockland:
Yeah that makes sense to me re: moving that to its own VIP 👍 

Regarding that corner case, avoiding clamping of literals is what this PR is designed to do 😅 

It won't currently work with the exact code example you posted, just because conversion from `decimal` itself is not fully supported up to the spec in VIP #1093.

However, in attempting to clarify with another example, I'm realizing that this optimization may have already been being done somewhere else in the code? I'm a bit confused here as to where exactly though.

For the code:

```python
@public
def bar() -> uint256:
    return convert(100.0, uint256)
```

The current LLL from `master` is, though I would expect it to include clamping given the state of `convert.py` on `master`:

```
[seq,
  [return,
    0,
    [lll,
      [seq,
        # Line 1
        [if,
          [eq, [mload, 0], '4273672062' <bar()>],
          [seq,
            # Line 3
            [mstore, 0, 170141183460469231731687303715884105727],
            [return, 0, 32],
            # Line 1
            stop]],
      0]]]
```

With this PR it is:

```
[seq,
  [return,
    0,
    [lll,
      [seq,
        # Line 1
        [if,
          [eq, [mload, 0], '4273672062' <bar()>],
          [seq,
            # Line 3
            [mstore, 0, 170141183460469231731687303715884105727],
            [return, 0, 32],
            # Line 1
            stop]],
      0]]]
```

As another note, this branch may need to be updated before merging with better literal tests (noticing some weird stuff is broken for converting to `int128`) and to include this optimization in the `to_decimal` conversions—though this could be handled in a separate PR as well.

fubuloubu:
Side note: we really need an AST lol. It takes so much effort to parse LLL for me

PR #1147, opened 2018-12-17
Document default value of optional args

nrryuya:
### - What I did
Documented default value of optional args in create_with_code_of

Document default value of optional args in create_with_code_of


nrryuya:
Fixed in #1148 to avoid a conflict.

PR #1156, opened 2018-12-18
Use the withdraw pattern to avoid the 'King of the Ether' security vulnerability

smarx:
### - What I did

I switched the simple auction code example to use the withdraw pattern. This avoids a security vulnerability where a malicious contract places a bid and then refuses all refunds (preventing others from outbidding them).

### - How I did it

I removed the `send` from `bid()` and instead added a `pendingReturns` state variable like in the blind auction example. I added a `withdraw()` function for bidders to actually pull those funds out.

### - How to verify it

I made sure the code still compiles, but I didn't do any other testing to verify functionality.

### - Cute Animal Picture

image

### DIFF
--- a/examples/auctions/simple_open_auction.vy
+++ b/examples/auctions/simple_open_auction.vy
@@ -13,6 +13,9 @@ highestBid: public(wei_value)
 # Set to true at the end, disallows any change
 ended: public(bool)
 
+# Keep track of refunded bids so we can follow the withdraw pattern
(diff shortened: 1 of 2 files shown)


smarx:
Sorry, just discovered via Travis that there are tests in place for this example, and those tests need to be updated. I will do that.

jacqueswww:
@smart-t Looks good! Always better to have securest as possible examples.

PR #1162, opened 2018-12-28
Add is_contract to the docs

nrryuya:
### - What I did
Added is_contract to the docs

Add is_contract to the docs


nrryuya:
#1165 is correct.

PR #1165, opened 2018-12-31
Add a member of address `is_contract` to the docs

nrryuya:
### - What I did
Added a member of address `is_contract` to the docs

Add a member of address `is_contract` to the docs


jacqueswww:
Good catch!

PR #1209, opened 2019-01-21
Fix the types of name and symbol in ERC20

nrryuya:
### - What I did
Fixed the types of `name` and `symbol` in ERC20.
(Thanks @fubuloubu !)

### - How I did it
`bytes` is more appropriate as a substitute for `string` in Solidity so I changed the type of `name` and `symbol` in ERC20 example.

### - How to verify it
Tests.

Fixed the types of `name` and `symbol` in ERC20.

### - Cute Animal Picture

image

### DIFF
--- a/vyper/premade_contracts.py
+++ b/vyper/premade_contracts.py
@@ -2,8 +2,8 @@ import ast
 
 erc20 = """
 class ERC20():
-    def name() -> bytes[40]: constant
-    def symbol() -> bytes[5]: constant
+    def name() -> bytes[64]: constant
+    def symbol() -> bytes[32]: constant
     def decimals() -> uint256: constant
     def balanceOf(_owner: address) -> uint256: constant
     def totalSupply() -> uint256: constant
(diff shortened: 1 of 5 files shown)


vyper/premade_contracts.py
```diff
@@ -2,8 +2,8 @@
 
 erc20 = """
 class ERC20():
-    def name() -> bytes32: constant
-    def symbol() -> bytes32: constant
+    def name() -> bytes[40]: constant
+    def symbol() -> bytes[5]: constant
```
jacqueswww: Because of the way bytearray works we could just make them `bytes[64]` and `bytes[32]` it occupies the same space, but is more flexible?

nrryuya: Sounds nice 👍

nrryuya: cc @fubuloubu

fubuloubu: WFM

PR #1212, opened 2019-01-22
Remove the description of `None` from docs

nrryuya:
### - What I did
`None` is not supported (#1106) so I removed the description from docs.

Removed the description of `None` from docs


jakerockland:
@nrryuya @jacqueswww  that section of the docs may be a place that it is helpful to add another line noting the usage of the `clear()` function, something like: 

> In Vyper, there is no ``null`` option like most programing languages have. Thus, every variable type has a default value. In order to check if a variable is empty, you will need to compare it to its type's default value.
> If you would like to reset a variable to its type's default value, use the built-in `clear()` function.

PR #1214, opened 2019-01-22
Add Contract Interfaces

jacqueswww:
### - What I did
Fixes #855.


fubuloubu:
This feature needs a corresponding larger example to show how it works

PR #1215, opened 2019-01-23
Ensure call to function make_struct_type has correct number of arguments

ipwright:
The call to <make_struct_type> has a missing argument, which may cause a run-time failure. This commit fixes the call to include the correct number of arguments.

### - What I did

I added the missing <constants> parameter to the call to <make_struct_type>

### - Friendly disclaimer

I work for Semmle, and noticed the bug with our LGTM code analyzer:
https://lgtm.com/projects/g/ethereum/vyper/snapshot/dist-1506710546241-1548102310071/files/vyper/types/types.py?sort=name&dir=ASC&mode=heatmap#x62d2d89c83c310e4:1

I helped develop LGTM's code quality algorithms, and I'm interested in knowing how much fixing and clean-up effort is required to move a project up a grade. E.g., Vyper is doing great for JavaScript but has a few issues in Python.
[Language grade: JavaScript](https://lgtm.com/projects/g/ethereum/vyper/context:javascript)
[Language grade: Python](https://lgtm.com/projects/g/ethereum/vyper/context:python)
https://lgtm.com/projects/g/ethereum/vyper/

If you enable LGTM automated code review on Vyper you can prevent these issues prior to merging.
https://lgtm.com/projects/g/ethereum/vyper/ci/

Best wishes!
Ian.

### DIFF
--- a/vyper/types/types.py
+++ b/vyper/types/types.py
@@ -327,7 +327,7 @@ def parse_type(item, location, sigs=None, custom_units=None, custom_structs=None
                 return ContractType(item.args[0].id)
         # Struct types
         if (custom_structs is not None) and (item.func.id in custom_structs):
-            return make_struct_type(item.id, location, custom_structs[item.id], custom_units, custom_structs)
+            return make_struct_type(item.id, location, custom_structs[item.id], custom_units, custom_structs, constants)
         if not isinstance(item.func, ast.Name):
             raise InvalidTypeException("Malformed unit type:", item)
         base_type = item.func.id


fubuloubu:
@ipwright where does the JS check come from? IIRC we don't have any JS code.

ipwright:
@fubuloubu I think you've got one JavaSript file in:
https://lgtm.com/projects/g/ethereum/vyper/latest/files/examples/wallet/sign.js?sort=name&dir=ASC&mode=heatmap
So the A+ in this case just tells us that there's no problems with this single file.

fubuloubu:
Gotcha! I forgot about that lol

jacqueswww:
@ipwright pretty cool will have a proper look at all the output sometime :)

Why is one not supposed to add additional attributes?
image

ipwright:
Hi @jacqueswww 
This is an LGTM `warning` (probably, but not necessarily, indicates a potential bug, but not as severe as an `error`). If you click on the `?` symbol you get an explanation of each alert:
https://lgtm.com/rules/9990086/
So adding additional attributes, without overriding the equality operator, could lead to undesired behavior in the subclass. Hope this helps.

I do recommend turning on LGTM automated code review
https://lgtm.com/projects/g/ethereum/vyper/ci/
for your project, since you get automatically informed of new alerts prior to merging (the service is free BTW).

I'm also considering opening PRs for some of the other potential sources of error (as I mentioned I worked on our code quality algorithms, and I want to get a feel for the amount of effort required to move up a grade). Would that be welcome? (No worries if not).

Best wishes.
Ian.

jacqueswww:
From the output, these rules are useful:
- Unused local variable
- First argument of a method is not named 'self'
- Unreachable code
At first glance they are also easy to fix ;)

Issue #1237, opened 2019-02-08
ERC20 Contract example - no checks on transferFrom()


ssteiger:
### Version Information
--
### What's your issue about?

The example contract at:
https://github.com/ethereum/vyper/blob/master/examples/tokens/ERC20.vy
In line 74.

No checks are performed in the `transferFrom()` function.
I am not sure but couldn't anyone just pass any `from/to` address combination and effectively steal their funds?
Doesn't there need to be some check if the `_from` account has authorized the sender?

### How can it be fixed?

Option 1: Add something like this: 
`assert _from == msg.sender`

Option 2: Make the function `@private`

Option 3: Add something like this:
`assert self.allowances[_to][msg.sender] >= _value`
`self.allowances[_to][msg.sender] -= _value`

Option 4: Both option 1 and option 2

#### Cute Animal Picture

![image](https://user-images.githubusercontent.com/10925195/52504900-1c616f80-2bea-11e9-8551-b68369d7615d.png)

fubuloubu:
So, this is a subtle thing. Option 3 basically already occurs.

In this line:
https://github.com/ethereum/vyper/blob/master/examples/tokens/ERC20.vy#L85

We are decrementing the sender's allowance as allocated by `_from`. If that allowance is lower than the amount we want to transfer, the function will revert (because we bake in SafeMath by default).

Should add a comment there to that effect (and also to `transfer`)

jacqueswww:
Closing as this was merged.

PR #1238, opened 2019-02-08
ERC20 contract example: Added comment for clarifying why not using assert

ssteiger:
ref: #1237

### - Cute Animal Picture

image

### DIFF
--- a/examples/tokens/ERC20.vy
+++ b/examples/tokens/ERC20.vy
@@ -64,6 +64,8 @@ def transfer(_to : address, _value : uint256) -> bool:
     @param _to The address to transfer to.
     @param _value The amount to be transferred.
     """
+    # NOTE: vyper does not allow unterflows
+    #       so checks for sufficient funds/allowances are done implicitly
     self.balances[msg.sender] -= _value
     self.balances[_to] += _value
     log.Transfer(msg.sender, _to, _value)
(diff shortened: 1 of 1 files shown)


examples/tokens/ERC20.vy
```diff
@@ -64,6 +64,8 @@ def transfer(_to : address, _value : uint256) -> bool:
     @param _to The address to transfer to.
     @param _value The amount to be transferred.
     """
+    # NOTE: vyper does not allow unterflows
+    #       so checks for sufficient funds/allowances are done implicitly
```
fubuloubu:
```suggestion
    #       so the following subtraction would revert on insufficient balance
```

Also, `underflows` above it

examples/tokens/ERC20.vy
```diff
@@ -80,6 +82,8 @@ def transferFrom(_from : address, _to : address, _value : uint256) -> bool:
      @param _to address The address which you want to transfer to
      @param _value uint256 the amount of tokens to be transferred
     """
+    # NOTE: vyper does not allow unterflows
+    #       so checks for sufficient funds/allowances are done implicitly
```
fubuloubu:
```suggestion
    #       so the following subtraction would revert on insufficient balance
```
Also, `underflows` above it

examples/tokens/ERC20.vy
```diff
@@ -80,6 +82,8 @@ def transferFrom(_from : address, _to : address, _value : uint256) -> bool:
...
     self.balances[_from] -= _value
     self.balances[_to] += _value
     self.allowances[_from][msg.sender] -= _value
```
fubuloubu:
```suggestion
    # NOTE: vyper does not allow underflows
    #      so the following subtraction would revert on insufficient allowance
    self.allowances[_from][msg.sender] -= _value
```

fubuloubu:
Awesome, thanks for taking a deeper look!

PR #1253, opened 2019-02-18
Convert from `decimal` to `uint256` and to `int128`

jakerockland:
### What I did

Work towards #1093, added conversion from `decimal` type to `uint256` and `int128`.

### How I did it

Updated `convert.py`, updated the tests.

### How to verify it

Run the tests: `make test`!

Added conversions from `decimal` to `uint256` and to `int128`.

### Cute Animal Picture

image


fubuloubu:
Typically conversions like these do `floor`, I'm okay with it it it's in the docs.

jakerockland:
Actually minor detail that the result is truncated rather than floored (only relevant difference for signed values). This is because the EVM opcode “SDIV” uses truncation. But eah I think that given the behavior is clear in the docs, it’s fine to do so implicitly.

On that note, I will make a docs PR soon!!

PR #1257, opened 2019-02-19
Remove mutable data structure as default function argument

pipermerriam:
### What was wrong

The functions `vyper.compiler.compile_code` and `vyper.compiler.compile_codes` use a list as a default for a function argument.  This is a common error.  While the current usage was benign, were this value to be mutated during the course of the function, those mutations would persist through the next function call.

### How I fixed it.

The function now uses a default argument of `None` and has new logic to put the appropriate default in place if no value has been explicitly provided.

- Remove usage of mutable default for function argument in `vyper.compiler.compile_code` and `cyper.compiler.compile_codes`

### Cute Animal Picture

baby-fox_624

### DIFF
--- a/vyper/compiler.py
+++ b/vyper/compiler.py
@@ -113,7 +113,7 @@ output_formats_map = {
 
 def compile_codes(codes, output_formats=None, output_type='list', exc_handler=None, interface_codes=None):
     if output_formats is None:
-        ouptup_formats = ('bytecode',)
+        output_formats = ('bytecode',)
 
     out = OrderedDict()
     for contract_name, code in codes.items():
--- a/vyper/compiler.py
+++ b/vyper/compiler.py
@@ -111,7 +111,9 @@ output_formats_map = {
 }
 
 
-def compile_codes(codes, output_formats=['bytecode'], output_type='list', exc_handler=None, interface_codes=None):
+def compile_codes(codes, output_formats=None, output_type='list', exc_handler=None, interface_codes=None):
+    if output_formats is None:
+        ouptup_formats = ('bytecode',)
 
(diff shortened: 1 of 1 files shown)


vyper/compiler.py
```diff
@@ -139,6 +141,6 @@ def compile_codes(codes, output_formats=['bytecode'], output_type='list', exc_ha
         raise Exception('Unknown output_type')
 
 
-def compile_code(code, output_formats=['bytecode']):
+def compile_code(code, output_formats=None):
```
cburgdorf: @pipermerriam does it feel wrong to you to just use `output_formats=('bytecode',)` here, in other words use an immutable data structure to mitigate the risk?

vyper/compiler.py
```diff
@@ -111,7 +111,9 @@ def get_source_map(code, contract_name, interface_codes=None):
 }
 
 
-def compile_codes(codes, output_formats=['bytecode'], output_type='list', exc_handler=None, interface_codes=None):
+def compile_codes(codes, output_formats=None, output_type='list', exc_handler=None, interface_codes=None):
+    if output_formats is None:
+        ouptup_formats = ('bytecode',)
```
fubuloubu: Typo

vyper/compiler.py
pipermerriam: Yes, I chose this because of the other function also specifying this as a default and thus, by using `None` I could avoid duplicating the default parameter across the two functions.

I could get around this by encoding it as a constant, but this felt better to me.

pipermerriam:
@fubuloubu thanks for catching that.  I committed your suggested change.

PR #1269, opened 2019-02-21
Blockspace context managers

jacqueswww:
### What I did

Part of #1259.

- Uses context manager patter for assignment, assert, for loop and block scoping.

### Cute Animal Picture

### DIFF
--- a/vyper/parser/context.py
+++ b/vyper/parser/context.py
@@ -72,20 +72,30 @@ class Context():
         # store global context
         self.global_ctx = global_ctx
 
-    def set_in_assignment(self, state: bool):
-        self.in_assignment = state
-
-    def set_in_for_loop(self, name_of_list):
-        self.in_for_loop.add(name_of_list)
-
-    def remove_in_for_loop(self, name_of_list):
-        self.in_for_loop.remove(name_of_list)
-
     def is_constant(self):
         return self.constancy == Constancy.Constant or self.in_assertion
 
-    def set_in_assertion(self, val):
-        self.in_assertion = val
+    #
+    # Context Managers
+    # - Context managers are used to ensure proper wrapping of scopes and context states.
+
+    @contextmanager
+    def set_in_for_loop_scope(self, name_of_list):
+        self.in_for_loop.add(name_of_list)
+        yield
+        self.in_for_loop.remove(name_of_list)
+
+    @contextmanager
+    def set_in_assignment(self):
+        self.in_assignment = True
+        yield
+        self.in_assignment = False
+
+    @contextmanager
+    def set_in_assertion(self):
+        self.in_assertion = True
+        yield
+        self.in_assertion = False
 
     @contextmanager
     def make_blockscope(self, blockscope_id):
--- a/vyper/parser/context.py
+++ b/vyper/parser/context.py
@@ -1,4 +1,5 @@
-from enum import Enum
+import contextlib
+import enum
 
 from vyper.utils import (
     MemoryPositions,
@@ -15,10 +16,9 @@ from vyper.exceptions import (
 from vyper.signatures.function_signature import (
     VariableRecord,
 )
-from contextlib import contextmanager
 
 
-class Constancy(Enum):
+class Constancy(enum.Enum):
     Mutable = 0
     Constant = 1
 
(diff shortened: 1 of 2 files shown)


vyper/parser/context.py
```diff
@@ -71,25 +72,35 @@ def __init__(self, vars, global_ctx, sigs=None, forvars=None, return_type=None,
         # store global context
         self.global_ctx = global_ctx
 
-    def set_in_assignment(self, state: bool):
-        self.in_assignment = state
+    def is_constant(self):
+        return self.constancy == Constancy.Constant or self.in_assertion
```
pipermerriam: For enums it is typically better to do `self.constancy is Constancy.Constant` since enums use singletons.

vyper/parser/context.py
```diff
@@ -15,6 +15,7 @@
 from vyper.signatures.function_signature import (
     VariableRecord,
 )
+from contextlib import contextmanager
```
pipermerriam: *convention suggestion*

https://github.com/ethereum/snake-charmers-tactical-manual/blob/master/style-guide.md#avoid-from-x-import-y-for-most-standard-library-imports

We have a convention of avoiding `from X import Y` for **most** standard library imports.  There are a few exceptions to this, but we use this as a general rule.  Detailed explanation in the link above, but the short explanation is that when I see `contextmanager` in the code of a module, I need to scroll to the top and find the import to be confident it is what I think it is (or maybe I'm not familiar with contextlib and so I won't even have that intuition).  But when I see `contextlib.contextmanager` it's much easier to reason about without scrolling up to the import.

vyper/parser/context.py
```diff
@@ -71,25 +72,35 @@ def __init__(self, vars, global_ctx, sigs=None, forvars=None, return_type=None,
...
+        yield
         self.in_for_loop.remove(name_of_list)
 
-    def is_constant(self):
-        return self.constancy == Constancy.Constant or self.in_assertion
+    @contextmanager
+    def set_in_assignment(self):
+        self.in_assignment = True
+        yield
+        self.in_assignment = False
```
pipermerriam: I *think* the current pattern is right, but I want to point out that in many cases, you will want to structure this as:

```
try:
    yield
finally:
    self.in_assignment = False
```

Without this, the final cleanup step won't happen if an exception occurs within the context block.  I'm not sure which would be your desired behavior in this situation but it appears that the current code mirrors the previous functionality.

vyper/parser/context.py
```diff
@@ -71,25 +72,35 @@ def __init__(self, vars, global_ctx, sigs=None, forvars=None, return_type=None,
...
 
-    def set_in_assertion(self, val):
-        self.in_assertion = val
+    @contextmanager
+    def set_in_assertion(self):
```
pipermerriam: Maybe these would be more semantically meaningful names if the methods were `assertion_scope`.

```python
with self.context.assertion_scope():
    ...
```

I thought of `assertion_context` as well but the duplicate use of the term context bugged me a bit.

vyper/parser/stmt.py
```diff
@@ -323,12 +324,13 @@ def call(self):
             return external_call.make_external_call(self.stmt, self.context)
 
     def parse_assert(self):
-        tmp = self.context.in_assertion  # backup value
+
+        tmp_assertion_value = self.context.in_assertion  # backup value
```
pipermerriam: I think you can fix this by adjusting the `set_in_assertion` method to the following.

```python
@contextlib.contextmanager
def set_in_assertion():
    prev_value = self.in_assertion
    self.in_assertion = True
    yield
    self.in_assertion = prev_value
```

This will allow nested assertion contexts that correctly revert to the original value after each block (only reverting to `False` after the final block exits.  This all assumes this is desired behavior in all cases.

jacqueswww:
@pipermerriam thanks! Will check it out tomorrow, end of day here now. Much appreciated :smiley_cat:

vyper/parser/context.py
jacqueswww: Noted: Yes, I think for these cases we don't have sockets/files/dbs etc. to close so it's _soft_ with statement ;)

vyper/parser/context.py
jacqueswww: Makes perfect sense, thanks :+1:

vyper/parser/context.py
jacqueswww: I agree let's call them _context scopes_ or _scopes_ for short.

vyper/parser/stmt.py
jacqueswww: Brilliant! Didn't think of that at all.

ndoudaffy:
Application for loan

PR #1276, opened 2019-02-27
Add Circle CI config and lint whitelist

davesque:
### What I did

Cherry picked and modified Piper's Circle CI config from PR #1262 .  Added all files with linting errors to an explicit whitelist.  It seems like a good strategy here is to merge this PR first, then create a series of subsequent PRs that address linting errors in individual files or groups of files.  Those PRs would fix the errors and also remove the files from the whitelist.  This should also allow linting PRs to be independent of each other.

### Cute Animal Picture

Put a link to a cute animal picture inside the parenthesis-->

PR #1262, opened 2019-02-19
Convert to CircleCI config

pipermerriam:
### What I did

Converted the project to use CircleCI for automated testing.

### Why?

- We've found it to be at least as reliable as travis and potentially more reliable.
- We've found it to be faster than travis.
- The EF has a paid plan so there's more build capacity available.

### Cute Animal Picture

baby-fox-in-den

PR #1283, opened 2019-02-27
vyper/optimizer.py lint

davesque:

### DIFF
--- a/vyper/optimizer.py
+++ b/vyper/optimizer.py
@@ -32,12 +32,27 @@ def search_for_set(node, var):
 
 def replace_with_value(node, var, value):
     if node.value == "with" and node.args[0].value == var:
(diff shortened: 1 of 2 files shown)


vyper/optimizer.py
```diff
@@ -87,17 +171,47 @@ def optimize(node):
...
+            node.pos,
+            node.annotation,
+            add_gas_estimate=node.add_gas_estimate,
+            valency=node.valency,
+        )
+    elif _is_with_without_set(node, argz):
+        # TODO: This block is currently unreachable due to
+        # `_is_with_without_set` unconditionally returning `False` this appears
+        # to be because this "is actually not such a good optimization after
+        # all" accordiing to previous comment.
```
fubuloubu: @jacqueswww, should evalulate this

jacqueswww: I created an issue https://github.com/ethereum/vyper/issues/1306, I suspect we can just drop the optimisation.

PR #1286, opened 2019-02-27
vyper/parser/expr.py lint

davesque:

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -94,7 +94,6 @@ exclude=
     vyper/optimizer.py
     vyper/parser/constants.py
     vyper/parser/context.py
-    vyper/parser/expr.py
     vyper/parser/external_call.py
     vyper/parser/global_context.py
     vyper/parser/lll_node.py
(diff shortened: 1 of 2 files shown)


jacqueswww:
@davesque @fubuloubu reading this change I can see why 80 character should become optional from pep8, we are in era of 4k displays now. And this greatly limits a programmer (80 characters being a magic constant chosen from tty's).

davesque:
@jacqueswww We're not using an 80 char limit.  We use 100 as the limit: https://github.com/ethereum/vyper/blob/master/tox.ini#L23

Also, I think limiting the width is reasonable since people often view code in split terminal windows or editor panes.  And we're already using the 100 char limit in almost every other ethereum python package.

davesque:
The one problem I have with some of these formatting edits is that they modify nested list definitions which had the appearance of lisp s-expression style to have a style more typically encountered in other languages.  I don't love this because I understand that the lists are essentially meant to represent lisp code.

On the other hand, python just isn't lisp and I don't think we should pretend that it is.  I'd eventually be in favor of finding a way to get rid of all the list literal juggling in vyper altogether.

jacqueswww:
Yeah, I like the lispness because LLL is a lisp so why change it. List serves this purpose. I mentioned before that we should not let linting dictate our IR style.

LLL has served us well as an IR, for vyper64 we use LLVM which has it's own Library.

fubuloubu:
I like Piper's suggestions of an easier, Pythonic API for creating LLL "shortcuts" internally. I quite like the idea/look of it, and it formats better in our code. s-expr syntax is better for outside the compiler

PR #1291, opened 2019-02-27
vyper/parser/parser.py lint

davesque:

### DIFF
--- a/vyper/parser/parser.py
+++ b/vyper/parser/parser.py
@@ -1,6 +1,6 @@
 import ast
 import copy
-
+import functools
 
 from vyper.exceptions import (
     ParserException,
@@ -133,12 +133,16 @@ def generate_default_arg_sigs(code, contracts, global_ctx):
     return sig_fun_defs
 
 
+def _default_sig_formatter(sig, custom_units_descriptions):
+    return sig.to_abi_dict(custom_units_descriptions)
+
+
 # Get ABI signature
 def mk_full_signature(code, sig_formatter=None, interface_codes=None):
 
     if sig_formatter is None:
         # Use default JSON style output.
-        sig_formatter = lambda sig, custom_units_descriptions: sig.to_abi_dict(custom_units_descriptions)
+        sig_formatter = _default_sig_formatter
 
     o = []
     global_ctx = GlobalContext.get_global_context(code, interface_codes=interface_codes)
@@ -150,7 +154,8 @@ def mk_full_signature(code, sig_formatter=None, interface_codes=None):
 
     # Produce function signatures.
     for code in global_ctx._defs:
-        sig = FunctionSignature.from_definition(code,
+        sig = FunctionSignature.from_definition(
+            code,
             sigs=global_ctx._contracts,
             custom_units=global_ctx._custom_units,
             custom_structs=global_ctx._structs,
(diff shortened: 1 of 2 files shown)


vyper/parser/parser.py
```diff
@@ -569,32 +707,52 @@ def parse_func(code, sigs, origcode, global_ctx, _vars=None):
...
                 sig_chain.append([
                     'if', sig_compare,
                     ['seq',
                         private_label,
-                        LLLnode.from_list(['mstore', context.callback_ptr, 'pass'], annotation='pop callback pointer', pos=getpos(code)) if sig.private else ['pass'],
+                        ['pass'] if not sig.private else LLLnode.from_list([
+                            'mstore',
+                            context.callback_ptr,
+                            'pass',
+                        ], annotation='pop callback pointer', pos=getpos(code)),
```
fubuloubu: Explain this change please

davesque: My guess is that it made the linter happier *or* it seems a bit clearer to indicate that the expression is ternary by putting the syntax for that on the first line.

fubuloubu: Read it again with a clearer head, just switched the order of the if..then..else statement

vyper/parser/parser.py
```diff
@@ -287,48 +334,87 @@ def parse_tree_to_lll(code, origcode, runtime_only=False, interface_codes=None):
...
             raise StructureException(error_message)
 
     return LLLnode.from_list(o, typ=None)
 
 
+def _mk_calldatacopy_copier(pos, sz, mempos):
+    return ['calldatacopy', mempos, ['add', 4, pos], sz]
+
+
+def _mk_codecopy_coppier(pos, sz, mempos):
```
jacqueswww: @davesque typo ;)

vyper/parser/parser.py
```diff
@@ -775,17 +971,33 @@ def check_list_type_match(provided):  # Check list types match.
             pos = context.vars[arg.id].pos
             check_list_type_match(context.vars[arg.id].typ.subtype)
             mem_offset = 0
-            for i in range(0, size):
+            for _ in range(0, size):
                 arg2 = LLLnode.from_list(pos + mem_offset, typ=typ, location='memory')
                 # p_holder = context.new_placeholder(BaseType(32)) if i > 0 else placeholder
```
jacqueswww: remove this comment please ;)

vyper/parser/parser.py
```diff
@@ -800,7 +1012,7 @@ def pack_logging_data(expected_data, args, context, pos):
 
     # Unroll any function calls, to temp variables.
     prealloacted = {}
-    for idx, (arg, expected_arg) in enumerate(zip(args, expected_data)):
+    for idx, (arg, _expected_arg) in enumerate(zip(args, expected_data)):
```
jacqueswww: `_expected_arg` doesn't seem to be used use `_` instead.

davesque: I had thought to just use the underscore prefix in case someone wonders what is being ignored. Using the prefix tells flake8 to ignore it.

jacqueswww: Ah, OK that works too :+1:

davesque:
@jacqueswww Addressed comments and rebased.

PR #1293, opened 2019-02-27
vyper/parser/self_call.py lint

davesque:

### DIFF
--- a/vyper/parser/self_call.py
+++ b/vyper/parser/self_call.py
@@ -1,3 +1,5 @@
+import itertools
+
 from vyper.exceptions import (
     ConstancyViolationException
 )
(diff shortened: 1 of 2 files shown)


vyper/parser/self_call.py
```diff
@@ -163,30 +185,42 @@ def call_self_private(stmt_expr, context, sig):
...
-        ['seq_unchecked'] +
-        pre_init +
-        push_local_vars +
-        push_args +
-        jump_to_func +
-        pop_return_values +
-        pop_local_vars +
-        [returner]
-    )
+    call_body = list(itertools.chain(
```
fubuloubu: Can you explain this change?

davesque: Flake8 doesn't like binary operators hanging off the beginning or end of lines.  That replacement code is functionally equivalent to using `+` to concatenate lists.

PR #1301, opened 2019-02-27
vyper/compiler.py lint

davesque:


vyper/compiler.py
```diff
@@ -153,13 +203,14 @@ def compile_codes(codes, output_formats=None, output_type='list', exc_handler=No
                 raise Exception('Unsupported format type %s.' % output_format)
 
             try:
-                out.setdefault(contract_name, {})[output_format] = output_formats_map[output_format](
+                out.setdefault(contract_name, {})
+                out[contract_name][output_format] = output_formats_map[output_format](
```
fubuloubu: This change seems a little odd

davesque: Actually, I happened to see this earlier and I think it's appropriate.  The previous statement was confusing since it may not be clear that `setdefault` returns either the value that was set as the default or the value that was already present.  In the replacement code, it's very clear that a key is being set in a nested mapping of some sort.

PR #1324, opened 2019-03-05
Create forwarder to

### LINKED ISSUE #1177: `create_with_code_of()` name is somewhat misleading
### What's your issue about?

I was not familiar with the inner workings of the Vyper compiler before diving deep into the Uniswap codebase.
While inspecting their contracts I always assumed that `create_with_code_of()` would `CODECOPY` the whole code of the target contract but, once I dug into Vyper and much to my big surprise, it creates a forwarder (which makes a lot of sense reusability-wise).

### How can it be fixed?

Like I was fooled I believe that others might be and I would propose a more expressive name for this (given its name is already long), something along the lines of: `create_forwarder_to` or `create_delegated_copy_of`

jacqueswww:
### What I did
Fixes #1177.
Rename create_with_code_of to create_forwarder_to.

### Cute Animal Picture

### DIFF
--- a/docs/built-in-functions.rst
+++ b/docs/built-in-functions.rst
@@ -376,11 +376,11 @@ Note: To give it a more Python like syntax, the assert function can be called wi
 
 Emits a log without specifying the abi type, with the arguments entered as the first input.
 
-**create_with_code_of**
+**create_forwarder_to**
 -----------------------
 ::
 
-  def create_with_code_of(a, value=b):
+  def create_forwarder_to(a, value=b):
     """
     :param a: the address of the contract to duplicate.
     :type a: address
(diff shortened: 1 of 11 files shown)


docs/installing-vyper.rst
```diff
@@ -225,10 +225,16 @@ and try compiling a contract:
...
 
+::
+
+
+To install the latest beta version use:
+
+::
 
-(Note that installing the above snap is the latest master)
+    sudo snap install vyper --beta --devmode
```
fubuloubu: Can this install instructions be another PR?

jacqueswww: done

PR #1325, opened 2019-03-05
assure function

### LINKED ISSUE #711: VIP: Enable Dynamic Analysis/Symbolic Execution Checks
## Preamble

    VIP: <to be assigned>
    Title: assertions that should never-ever fail
    Author: Yoichi Hirai <yoichi@ethereum.org>
    Type: Standard Track
    Status: Draft
    Created: 2018-03-21

## Simple Summary

Add a special type of assertion that should never ever fail.  When a static analyzer can fire it, the program surely has a bug.

## Abstract

Static analyzers can detect bugs, but only when the desired properties are specified.  The easiest way is to insert a "never-to-fail" assertions in the program.

## Motivation

Mythril has [a feature to detect when 0xfe is somehow reachable](https://github.com/b-mueller/mythril/blob/master/security_checks.md).  KEVM can soon do that.

`assert` in vyper is expected to fail sometimes, and it is compiled to `REVERT`.  `REVERT` might indicate a mistake of the caller, a mistake of the programmer, or a mistake of the compiler.  So, static analyzers cannot shout "this is a bug".

Instead, if Vyper has a never-to-fail assertion that translates to `INVALID (0xfe)`, static analyzers can confidently shout "this is a bug. See, this execution reaches `INVALID`".

Alternatives are specially formatted comments like [ACSL](https://frama-c.com/acsl.html) or [JML](https://en.wikipedia.org/wiki/Java_Modeling_Language), but they are hard to learn.

## Specification

Add a additional custom constant to the `assert` statement:

```python
    assert amount != 0, UNREACHABLE
```
as well as
```python
    raise UNREACHABLE
```
## Backwards Compatibility

A program containing a name `assure` will cause compillation errors.

## Copyright

jacqueswww:
### What I did
Fixes #711.

### How to verify it
Play with the assure function.

Added assure function.

### Cute Animal Picture

### DIFF
--- a/docs/built-in-functions.rst
+++ b/docs/built-in-functions.rst
@@ -351,16 +351,33 @@ CAUTION! This method will delete the contract from the Ethereum blockchain. All 
 ----------
 ::
 
-  def assert(a):
+  def assert(a, reason):
     """
     :param a: the boolean condition to assert
     :type a: bool
+    :type reason: string_literal
     """
 
 Asserts the specified condition, if the condition is equals to true the code will continue to run.
 Otherwise, the OPCODE ``REVERT`` (0xfd) will be triggered, the code will stop it's operation, the contract's state will be reverted to the state before the transaction took place and the remaining gas will be returned to the transaction's sender.
 
+An optional reason string literal can be supplied to the assert statement to help a programmer indicate why an assertion failed, this is done using `Error(string)` method as specified in `EIP838 <https://github.com/ethereum/EIPs/issues/838>`_.
+
 Note: To give it a more Python like syntax, the assert function can be called without parenthesis, the syntax would be ``assert your_bool_condition``. Even though both options will compile, it's recommended to use the Pythonic version without parenthesis.
+
+
+**assure**
+----------
+::
+
+  def assure(a):
+    """
+    :param a: the boolean condition to assure
+    :type a: bool
+    """
+
+Assure is the same as an `assert` function, but uses the OPCODE ``INVALID`` (0xfe) instead, when triggerd the code will stop it's operation, the contract's state will be reverted to the state before the transaction took place and the remaining gas will NOT be returned. The purpose of this function is for use with static analyzers and similar tools, and in general is not recommended for uses outside of this use case.
+
 
 **raw_log**
 -----------
--- a/tests/parser/features/test_assert.py
+++ b/tests/parser/features/test_assert.py
@@ -17,15 +17,14 @@ def foo():
     """
     c = get_contract_with_gas_estimation(code)
     a0 = w3.eth.accounts[0]
-    pre_balance = w3.eth.getBalance(a0)
-    tx_hash = c.foo(transact={'from': a0, 'gas': 10**6, 'gasPrice': 10})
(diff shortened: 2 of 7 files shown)


fubuloubu:
Would like a few more examples showing the use of `assure` alongside `assert` (namely, `assert` should be used for pre-conditions, and `assure` for post-condition invariants that _should_ never be violated)

It may even be useful to have a few `assert` post-conditions (non-invariant checks) 5o show that `assure` is sort of a "last resort" mechanism used in only specific use cases.

jacqueswww:
@fubuloubu to be honest, I am not keen that we have 2 different asserts, and I still don't think having two types is a good idea (even more so because we have reason string support). So I would rather not advertise the use case, people that will want to use it will know how in any case. We don't currently have a examples for assert, so I don't think we should have examples on assure either then - people just pick it up and go a long.

jacqueswww:
Having thought about it some more. I am actually OK with having the function, especially because of the tool support (invalid being the de facto).
I also think having support for what these tools/analyzers offer are great.
@fubuloubou - I am still keen on you writing the examples though, as I think you will write it better than myself.

fubuloubu:
Okay, I assigned myself to add documentation on this feature and how to use it effectively. Let's keep #711 open until that is complete.

jacqueswww:
Cool, thanks :) it's been a while since I have thought about the analyzers, do we still have that model checker issue somewhere?

fubuloubu:
It would be fantastic to get someone to review this new feature.

cc @maurelian @wuestholz @japesinator @montyly could you aid us?

wuestholz:
@fubuloubu @jacqueswww Great to see this being added! In general, any solution that allows to distinguish between parameter validation (`require` in Solidity or REVERT in EVM) and properties that should (in principle) always hold (`assert` in Solidity and INVALID in EVM) will be very useful to developers and tool builders.

The ideal solution would probably be to change the semantics of the existing `assert` keyword to compile to the INVALID opcode and introduce a new `require` (or `assume`) keyword that is used for parameter validation (using REVERT). This would be in line with Solidity and less confusing to developers. I was actually involved in a similar discussion there (https://github.com/ethereum/solidity/pull/1702) way back that led to the introduction of `require`.

On the other hand, I understand that you might not want to change the semantics and then your solution will work as long as it is well-documented.

jacqueswww:
@fubuloubu @wuestholz the main reason we didn't want to change `assert` is that Vyper is rooted in python. Where `assert` gets used for input validation, as well as test assertions, this is why the terminology was chosen as such in #711 - `assert` with revert is also the natural choice someone writing a contract should use by default (because of the gas refund). I think it makes sense to leave the naming as is (I remember discussing this at length regarding the positives and negatives of either - and then settling on above.).

jacqueswww:
Meeting notes: closing in favour a proposal to be done where `assure` gets removed and `assert <condition>, UNREACHABLE` is used instead.

fubuloubu:
@maurelian @wuestholz @japesinator @montyly we've made a decision to modify the syntax to what @jacqueswww has proposed here: https://github.com/ethereum/vyper/issues/711#issuecomment-471652086

Please give us feedback!

PR #1337, opened 2019-03-11
[WIP] Add tests for script invocation

davesque:
### What I did

Added a test suite to ensure proper functioning of bin scripts.

### How I did it

Added `pytest-console-scripts` plugin to testing requirements and bumped required `pytest` version to take advantage of built-in temp file fixtures.  Added test suite for scripts.

### How to verify it

Run tests.

### Cute Animal Picture


tests/bin/test_vyper.py
```diff
@@ -0,0 +1,62 @@
...
+    ret = script_runner.run('vyper', '--version')
+
+    assert ret.stdout == __version__ + '\n'
+
+
+@pytest.mark.script_launch_mode('subprocess')
+def test_compile_one_file(script_runner, tmp_vy_file_1):
+    ret = script_runner.run('vyper', '-f', 'bytecode', tmp_vy_file_1)
+
+    assert ret.stdout == '0x6100c256600035601c52740100000000000000000000000000000000000000006020526f7fffffffffffffffffffffffffffffff6040527fffffffffffffffffffffffffffffffff8000000000000000000000000000000060605274012a05f1fffffffffffffffffffffffffdabf41c006080527ffffffffffffffffffffffffed5fa0e000000000000000000000000000000000060a052630dbe671f60005114156100b85734156100ac57600080fd5b600160005260206000f3005b60006000fd5b6100046100c2036100046000396100046100c2036000f3\n'  # noqa: E501
```
fubuloubu: @jacqueswww how fragile would it be to test against raw bytecode for these simple programs? It shouldn't have to be updated too often due to language semantics changes, right?

tests/bin/test_vyper.py
```diff
@@ -0,0 +1,62 @@
+import pytest
+
+from vyper import (
+    __version__,
+)
+
+
+@pytest.fixture
+def make_tmp_vy_file(tmp_path):
+    name_counter = 0
```
pipermerriam: Using an `itertools.count` for this type of thing can be a bit cleaner in cases like this where you just need blind increments.  I *think* it allows you to remove the need for `nonlocal` and in theory you can inline the incrementing into the string formating via `f'{tmp{next(name_counter)}.vy'` if you're feeling brave.

davesque:
Couldn't having a more flexible test be just as brittle but in a way that's harder to detect?  For example, imagine we're searching for the preamble and push op.  Then, someone breaks the compiler by causing the push op to appear in the constructor instead of the method.  We wouldn't know anything was wrong.  On the other hand, if you are making a change that has that effect, you expect to see tests break.  Looking for a literal value gives the highest likelihood that the tests will fail in an expected way.

pipermerriam:
> Couldn't having a more flexible test be just as brittle

Yes, though I'd argue that these are *integration* tests or something like that, and like you said, we focus on catching *functionality* errors at a lower level, and design this suite to ensure the CLI compilation produces the *expected* output.   I recognize the impreciseness of that definition and the somewhat circular definition.  Only trying to brainstorm ways to get similar guarantees from the tests while reducing likelihood that they end up being brittle.

jacqueswww:
Just bombing in to say this is PR leet congrats :-)

davesque:
@jacqueswww @pipermerriam Just as a reminder, I see this PR as being very far from complete! It was really just the first couple of steps. :)

davesque:
@jacqueswww Okay, haha.  I just looked at the PR number and now I get what you meant :).

fubuloubu:
@davesque any updates on this one?

davesque:
@fubuloubu Not currently.  I can potentially get back on this after doing the eth-abi strategies and juggle it with the AST class work.

fubuloubu:
Awesome! Would be helpful I think, we've had a few strange issues since the script is largely untested.

davesque:
@pipermerriam I'm looking over this discussion some time later now and, in retrospect, I completely agree with the points you were making earlier.  We could probably find ways to make things a bit more robust.

fubuloubu:
@davesque The original premise of this PR is now no longer necessary due to #1576, although I am sure some of the work can be reused to the new configuration as the compiler endpoints in `vyper/cli` are undertested (but should be easier to test now!)

PR #1346, opened 2019-03-11
Update/fix coverage reporting config

davesque:
### What I did

Updated config for coverage reporting to coveralls.

### How I did it

Added appropriate requirements to test extras.  Updated `tox.ini` to run coverage upload wrapper script.  Added appropriate env var to circle CI project environment (this isn't reflected in PR changes).

### How to verify it

Trigger a circle CI build.

### Cute Animal Picture

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -54,6 +54,7 @@ jobs:
       - image: circleci/python:3.7
         environment:
           TOXENV: py37-core
+
 workflows:
   version: 2
   test:
--- a/.coveralls.yml
+++ b/.coveralls.yml
@@ -1,0 +1,2 @@
+service_name: circle-ci
+parallel: true
--- a/setup.py
+++ b/setup.py
@@ -6,6 +6,8 @@ from setuptools import setup, find_packages
 test_deps = [
     'pytest>=3.6',
     'pytest-cov==2.4.0',
+    'coverage==4.0.3',
+    'python-coveralls==2.9.1',
     'pytest-xdist==1.18.1',
     'py-evm==0.2.0a39',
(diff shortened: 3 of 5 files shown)


fubuloubu:
This doesn't turn negative coverage into a blocker for merging, right?

In the past, we've only explored coverage misses at major checkpoints.

davesque:
@fubuloubu I guess I'm not sure.  That would happen if the `coveralls` binary returned non-zero status for coverage under a certain level.  I'll check if that happens.

davesque:
@fubuloubu I don't think that happens.  However, I've now got to figure out why Circle CI is ignoring my project env var setting.

davesque:
Seems like the env var issue has to do with my lacking permissions for our circle CI account.  I'm going to try and work with our devops guys who manage the account to confirm whether or not this is the case.

Once I get that worked out, I believe this config should get us back up and running with coverage.  The one other issue that might pop up is that the coveralls utility might not merge coverage data from parallel test workers properly.  It claims to handle that but I thought I saw it misreporting it earlier when I was doing some manual testing.

davesque:
I'm in touch with Circle CI support to try and get this worked out.  This issue shouldn't be happening and it feels like it's a problem on their end.

davesque:
As it turns out, there is a setting on circle CI that controls whether or not build environment variables are set for builds from PR forks.  I didn't notice that but it totally makes sense why it's there.  Otherwise, by simply creating a PR, someone could retrieve those values by including malicious code in a PR, whether or not it gets merged.

I think this PR should be good to merge.  It just means that we'll only get coverage reporting on the master branch which is what we want anyway.

PR #1348, opened 2019-03-12
Remove redundant 'seq' in initializer code

davesque:
### What I did

Call me crazy, but I think Vyper has been inserting an unnecessary `seq` construct into all initializer code.

**Update**: @jacqueswww has pointed out that this extra `seq` is optimized away.

### How I did it

Took it out.  Fixed one broken test.

### How to verify it

Run tests.

### Cute Animal Picture

### DIFF
--- a/tests/parser/features/test_gas.py
+++ b/tests/parser/features/test_gas.py
@@ -31,5 +31,5 @@ def __init__():
     """
     parser_utils.LLLnode.repr_show_gas = True
     out = parse_to_lll(code)
-    assert '35303' in str(out)[:28]
+    assert '35273' in str(out)[:28]
     parser_utils.LLLnode.repr_show_gas = False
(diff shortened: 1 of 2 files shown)


davesque:
@jacqueswww Am I confused here or is this actually a thing?

jacqueswww:
It's possible, is the bytecode shorter?

jacqueswww:
@davesque last time I checked the optimizer will remove the extra `seq` so I don't think this will actually have an effect on the final bytecode.

davesque:
@jacqueswww Yeah, now I see that happening in the optimizer.  It improves the gas estimate but probably doesn't change the bytecode length.

davesque:
Confirmed that nothing changes:
```
$ cat test.vy
x: int128

@public
def __init__():
    self.x = 1
$ g co initializer
Switched to branch 'initializer'
Your branch is up to date with 'origin/initializer'.
$ vyper test.vy > test1.out
$ g co master
Switched to branch 'master'
Your branch is up to date with 'origin/master'.
$ vyper test.vy > test2.out
$ diff test1.out test2.out
$
```
Guess we can merge this anyway to make the gas estimates a bit better.

jacqueswww:
Yes 'course :)

PR #1350, opened 2019-03-13
Some lint fixes in vyper.optimizer

davesque:
### What I did

Made some small lint fixes in `vyper.optimizer`.

### How to verify it

Run the tests.

### Cute Animal Picture

### DIFF
--- a/vyper/optimizer.py
+++ b/vyper/optimizer.py
@@ -1,3 +1,5 @@
+import operator
+
 from vyper.parser.parser_utils import (
     LLLnode,
 )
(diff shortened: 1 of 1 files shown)


vyper/optimizer.py
```diff
@@ -60,11 +62,11 @@ def replace_with_value(node, var, value):
...
 arith = {
-    "add": (lambda x, y: x + y, '+'),
-    "sub": (lambda x, y: x - y, '-'),
-    "mul": (lambda x, y: x * y, '*'),
-    "div": (lambda x, y: x // y, '/'),
-    "mod": (lambda x, y: x % y, '%'),
+    "add": (operator.add, '+'),
+    "sub": (operator.sub, '-'),
+    "mul": (operator.mul, '*'),
+    "div": (operator.floordiv, '/'),
```
pipermerriam: Interesting, this is opposite of python right since python uses `//` for floored division.  Might be another place where we're going to surprise python developers.

jacqueswww: @pipermerriam hmm not really, this is optimizer on the LLL which emulates what the EVM would've done - not how vyper handles division.
We don't support floor division in vyper because you have to be explicit about truncating values opting for `floor(a / 2)` instead.

PR #1354, opened 2019-03-14
Some cleanup/type annotations in parser submodules

davesque:
### What I did

This is a bit random but I was poking around reading the code and decided to bite this off and do some refactors and type annotations in the `parser` submodules.  This will hopefully give an example of some of the kinds of changes I'd like to see in other parts of the code base as well.

### How to verify it

Run the tests.

### Cute Animal Picture

### DIFF
--- a/vyper/exceptions.py
+++ b/vyper/exceptions.py
@@ -78,3 +78,7 @@ class FunctionDeclarationException(ParserException):
 
 class EventDeclarationException(ParserException):
     pass
+
+
+class VersionException(ParserException):
+    pass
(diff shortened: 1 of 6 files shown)


davesque:
@fubuloubu FYI, I just split this up into a lot of separate commits which will hopefully make it easier to follow everything I did step by step.  I know you already approved but I figure I'd point that out in case it is helpful.  It should make things easier for @jacqueswww when he looks this over.

davesque:
@jacqueswww I'd like to do some work on top of this.  Looks good to merge?

PR #1357, opened 2019-03-16
Add env var passthrough for coveralls

davesque:
### What I did

Add an env var passthrough in the tox config for coveralls.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -27,6 +27,7 @@ use_parentheses = True
 [testenv]
 usedevelop = True
 passenv =
+    COVERALLS_REPO_TOKEN
     CIRCLECI
     CIRCLE_*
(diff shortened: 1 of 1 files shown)


jacqueswww:
@davesque I see it's updating on https://coveralls.io/github/ethereum/vyper, any luck getting it to show on the Pull requests?

davesque:
There's supposedly a way to get coveralls to infer that an API call is from circle CI. If we could figure that out, we could have coverage reporting on PR branches. Otherwise, the only way we can upload coverage reports is with an API token in an env var. We're currently not setting circle project env vars on PR jobs and I think it should remain that way. Otherwise, the API token will be accessible by contributors and that's probably not good even though the only actual risk is that they could fuck with our coverage reporting :) .

PR #1358, opened 2019-03-18
More parser submodule cleanup

davesque:
### What I did

More parser submodule refactors/cleanup.

* Added various docstrings
* Created more robust post-parsing AST node visitors
* Add tests for new behavior handled by refactored node visitors

* Fixed a bug that prevented certain nested unary subtractions from being recognized as negative literals.

### Cute Animal Picture

### DIFF
--- a/tests/parser/functions/test_convert_to_uint256.py
+++ b/tests/parser/functions/test_convert_to_uint256.py
@@ -222,6 +222,17 @@ def foo() -> uint256:
     )
 
     code = """
+@public
+def foo() -> uint256:
+    return convert(-(-(-27.2319)), uint256)
+    """
+
+    assert_compile_failed(
+        lambda: get_contract_with_gas_estimation(code),
+        InvalidLiteralException
+    )
+
+    code = """
 bar: decimal
 
 @public
--- a/vyper/parser/parser_utils.py
+++ b/vyper/parser/parser_utils.py
@@ -742,6 +742,29 @@ class OptimizingVisitor(ast.NodeTransformer):
             return node
 
 
+def annotate_and_optimize_ast(
+    parsed_ast: ast.Module,
+    source_code: str,
+    class_types: Optional[ClassTypes] = None,
+) -> None:
+    """
+    Performs annotation and optimization on a parsed python AST by doing the
+    following:
+
+    * Annotating all AST nodes with the originating source code of the AST
+    * Annotating class definition nodes with their original class type
+      ("contract" or "struct")
+    * Substituting negative values for unary subtractions
+
+    :param parsed_ast: The AST to be annotated and optimized.
+    :param source_code: The originating source code of the AST.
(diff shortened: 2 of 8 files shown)


davesque:
I'm realizing now there may be an issue with the `parser_utils.get_number_as_fraction` function and the new optimizing node visitor converting expressions like `-(-(-2))` into `-2`.  Need to take a closer look at this before we merge this PR.

davesque:
@jacqueswww Off the top of your head, can you think of any other issues that the new `-(-(-n)) == -n` pre-parse optimization would cause?

davesque:
Actually, my sense is that `get_number_as_fraction` will probably still work.  The `OptimizingVisitor`, borrowing code from what it replaces, substitutes each expression that it optimizes with the unary subtraction's operand.  So what ends up in the AST should be an expression with a correct reference to its unsigned representation in the original source code.  I'll verify that the tests run over this possibility.

vyper/parser/parser_utils.py
```diff
@@ -698,37 +704,65 @@ def make_setter(left, right, location, pos, in_function_call=False):
...
+        self.generic_visit(node)
 
-    MyVisitor().visit(_ast)
+        # Decorate class definitions with their respective class types
+        node.class_type = self._class_types.get(node.name)
 
+        return node
+
+
+class OptimizingVisitor(ast.NodeTransformer):
```
jacqueswww: I don't think this naming describe what it actually happens OptimizingVisitor is ambiguous, RewriteUnaryOp is clearer name for me.

tests/parser/functions/test_convert_to_uint256.py
```diff
@@ -221,6 +221,17 @@ def foo() -> uint256:
         InvalidLiteralException
     )
 
+    code = """
+@public
+def foo() -> uint256:
+    return convert(-(-(-27.2319)), uint256)
```
jacqueswww: Please add tests that check the ast output of the annotate_and_optimize_ast function.

vyper/parser/parser_utils.py
davesque: I chose that name because it's a more general term and I thought more such actions could be added to the same transformer in the future. But I'll make the change you're requesting.

davesque:
Yep the behavior is the same except for the `-(-(-n)) -> -n` behavior which is new. There were two independent functions before which were always being called together and in the same way. So it made sense to normalize things and convert them into one.

vyper/parser/parser_utils.py
jacqueswww: Yes I think we'll be adding another visitor for constant and type annotations here hopefully soon, but let's evaluate what to name them once we have coded all of those up.

jacqueswww:
Works for me :+1:

davesque:
I'll go ahead and merge this since I think @pipermerriam 's tied up with other things ATM.

PR #1361, opened 2019-03-19
Flexible imports relative to paths given at the CLI

mikeshultz:
### What I did

Adjusted the `vyper` CLI script to be more flexible when importing interfaces. The intention being that the current working directory is less important than the location of the given paths of the files making the import.

### How I did it

CLI script searches all relevant dirs given instead of just different extensions.

This uses all directories provided at the CLI as search paths and not just the specific file that does the importing.  I did this mostly to keep this PR simple, but can refactor to be specific to the importing file.  Thoughts?

### How to verify it

Run the `vyper` CLI script away from the directory containing the contract(s).  I didn't see any CLI unit tests so I did not add any for this.  But I can if requested.

Interface imports are now relative to file paths instead of current directory

### Cute Animal Picture

### DIFF
--- a/bin/vyper
+++ b/bin/vyper
@@ -82,6 +82,18 @@ def exc_handler(contract_name, exception):
     raise exception
 
 
+def import_file_paths(dirs, importpath):
+    """ Get the possible import paths for an interface import """
+    extensions = ('.vy', '.json')
+    valid_paths = list()
+    for d in dirs:
+        for ext in extensions:
+            resolved_path = os.path.join(d, os.path.normpath(importpath.replace('.', '/'))) + ext
+            if os.path.isfile(resolved_path):
+                valid_paths.append(resolved_path)
+    return valid_paths
+
+
 if __name__ == '__main__':
     if args.show_gas_estimates:
         parser_utils.LLLnode.repr_show_gas = True
@@ -114,18 +126,17 @@ if __name__ == '__main__':
             formats.append(translate_map.get(f, f))
 
         interface_codes = {}
+        import_paths = set()
 
-        for code in codes.values():
+        for code_path, code in codes.items():
             interface_codes.update(extract_file_interface_imports(code))
+            dir_path = os.path.dirname(code_path)
+            if dir_path:
+                import_paths.add(dir_path)
 
         if interface_codes:
             for interface_name, interface_path in interface_codes.items():
-                file_path = os.path.join(os.path.normpath(interface_path.replace('.', '/')))
-                extensions = ('.vy', '.json')
-                valid_paths = [
-                    file_path + ex
-                    for ex in extensions if os.path.exists(file_path + ex)
-                ]
+                valid_paths = import_file_paths(import_paths, interface_path)
                 if not valid_paths:
                     raise Exception('Imported interface "{}{.vy,.json}" does not exist.'.format(
                         interface_path,


fubuloubu:
LGTM, but I think @jacqueswww or @davesque had "opinions" about how the relative path thing should work.

jacqueswww:
I am not 100% sure, but python style import using `..` and `.` are in generally not preferred (pretty sure there is a pep8 rule for it). This is why I used the approach of not allow `..` and `.` for the time being when I made the interfaces change.

Would it help if we instead allow the base directory to be set via an environment variable or parameter to the compiler?
```
vyper --base-dir <root_dir>
```
OR
```
BASE_DIR="" vyper ...
```

mikeshultz:
To be clear, this PR doesn't alter import syntax or anything, just the file search paths of the `vyper` CLI command.

I agree there's a possibility of a conflict, though that's probably an edge case.  This can be refactored to be strict, so that it only checks relative to the file doing the imports and relative to the pwd.  I chose this method to keep the PR simple but I can dig into it further tomorrow if that's preferred.

`--base-dir` would be useful for my purposes, however I still think it would be counter-intuitive to the contract developer.  Imports of interfaces(not of packages) should be relative to the source file, IMO.  Especially since vyper doesn't have standard library path like `site-packages` unless I'm missing something.

jacqueswww:
Ah OK, @mikeshultz  still brain storming... 
This is the broader vision that I had when I made the choices, if I structured my project like so:
```
<project-dir>
 -> contracts
     -> token.vy
 -> interfaces
     -> minime_token.vy
```

token.vy would use:
```python
import interfaces.minime_token as MiniMeTokenInterface
implements: MiniMeTokenInterface
```

or alternatively 
```
import interfaces.minime_token as MiniMeTokenInterface
```
On would just run vyper like so:
```
vyper contracts/token.vy
```
This emulates something like a django project, where the root directory is the location of manage.py.
(hmmm that make me think, maybe we should do a manage.py :tongue: ).

If you want everything in one project dir, it would also work:

```
from x import XInterface
```
```
<root_dir>
x.vy
main.vy
```
But then the compiler would have to be run in the same directory.
Basically it's what python does but without `..` and `.`, or did I miss something ?

mikeshultz:
Here's the use case I'm working with:

Project directory structure

    |- contracts/
        |
        |- MyContract.vy
        | 
        |- interface
           |
           | - IMyContract.vy

MyContract.vy:

    import interface.IMyContract as IMyContract
    implements: IMyContract
    [...]

Invocation:

    vyper contracts/MyContract.vy

jacqueswww:
So the import statment should be:
```python
import contracts.interface.IMyContract as IMyContract
```
Right ?

fubuloubu:
`contracts/` is typically the root directory (with the files directly inside 90% of the time, or subdir'd underneath), so I would use that assumption.

Majority of the time `contracts/` is a sub directory of the dapp project, alongside `tests/` and `src/` (for front end code), plus `docs/` and the like.

mikeshultz:
That would work, yes.  Would the contract dev expect that?  That requires the dev to take the working directory of the compiler into account when implementing that statement in the source.  They may have no idea how/where the contracts will be compiled.

IMO, the comiler/script should take the hint from the source instead of the source trying to guess how the compiler will be invoked.

davesque:
> To be clear, this PR doesn't alter import syntax or anything, just the file search paths of the vyper CLI command.

Yep, I understand.  But it changes the behavior of known syntax from python from what it usually would be.  Non-relative imports (ones lacking a `.` or `..` prefix in the import path) already have a specific function in the language we're trying to mimic.  I really think that, if we want relative imports, they need to look like they do in python.  Otherwise, python developers will be confused by the behavior.

As far as command-line option vs. env var goes, for now I think I prefer the command-line option idea since it's more explicit.  Env vars can be inherited from a parent process which might not be desirable.

On another note, it seems like the existing code without this PR also allows for the possibility of strange behavior when "as" names are shared across different code files but may refer to different interface paths.  Here's the line where that happens: https://github.com/ethereum/vyper/blob/master/bin/vyper#L119 .  This seems like a bug to me.

jacqueswww:
@davesque yes looks like `interface_codes` will have to be have an additional on a per file basis. 

I checked pep8 and this is all it said about relative imports, I would not say it's too strict so we could add support for `.` and `..` ?
image

jacqueswww:
I added the issue to the meeting on Monday's agenda, but I think I have been persuaded from my initial vision to actually add support for relative imports. About 93.4% convinced on `..` though, but hey let's make it more like python and support it - don't think incorrect interfaces can cause security risks, as the main contract are the ones that count.

mikeshultz:
I'll close this, then.

If python compat is paramount, perhaps the concept of `VYPERPATH` (ala `PYTHONPATH`) should be used, maybe with an accompanying CLI option.  Could be useful in the future with implementing packages, like with like EthPM.

jacqueswww:
@mikeshultz yes that is something to keep in mind as well :+1:

fubuloubu:
Lastly, if the scripts in `vyper/bin` have too much logic in them that they need to be tested (moreso than a simple integration test) I think it may be necessary to do more of that logic within the compiler.

I still like the division of labor that the compiler doesn't know anything about file locations, but perhaps there is a median suggestion there.

We should probably have a few test cases for these "integration" tests though, especially with the addition of this feature.

jacqueswww:
@davesque actually has WIP PR for testing vyper-bin already ;) so should be sorted soon.

davesque:
@mikeshultz Cool. Thanks for your work on this even though we're going to take a different route.

Issue #1364, opened 2019-03-22
msg missing from reserved_words


mikeshultz:
### Version Information

* vyper Version (output of `vyper --version`): master
* OS: n/a
* Python Version (output of `python --version`): n/a
* Environment (output of `pip freeze`): n/a

### What's your issue about?

`msg` appears to be missing from `vyper.utils.reserved_words`

### How can it be fixed?

Add `msg` to `vyper.utils.reserved_words`

mikeshultz:
Since `contract` is also in `reserved_words`, is `struct` missing as well?

davesque:
@mikeshultz Yep, I think those words belong in those collections.  Making a PR now to address this and other related things.

jacqueswww:
One of the faster issues we can close ;)

PR #1365, opened 2019-03-22
Fixes/updates in vyper.utils

davesque:
### What I did

Fixed problem mentioned in issue #1364 .  Updated reserved words list and did some other miscellaneous cleanup.

### How to verify it

Run the tests.

* The keywords `struct` and `msg` were incorrectly being allowed as variable names.  Those keywords are now properly treated as reserved words.

### Cute Animal Picture

### DIFF
--- a/tests/parser/exceptions/test_variable_declaration_exception.py
+++ b/tests/parser/exceptions/test_variable_declaration_exception.py
@@ -75,6 +75,16 @@ x: int128
 def foo():
     x = 5
     """,
+    """
+@public
+def foo():
+    msg: bool = True
+    """,
+    """
+@public
+def foo():
+    struct: bool = True
+    """,
 ]
 
 
--- a/tests/parser/syntax/test_ann_assign.py
+++ b/tests/parser/syntax/test_ann_assign.py
@@ -65,8 +65,8 @@ struct S:
     b: decimal
 @public
 def foo() -> int128:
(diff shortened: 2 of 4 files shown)


davesque:
@jacqueswww I wonder if it really matters that words like this can be used as variable names.  I wonder what some real world examples are of how this would lead to potential security problems or whatever.

fubuloubu:
```python
msg_sender: address = msg.sender  # Weird, but okay I guess

... # Bunch of stuff

transfer(msg_sender, self.balance)  # Could be easy to miss
```

davesque:
@fubuloubu I'm not following. Can you elaborate?

davesque:
@fubuloubu Ahh, I get it now.  That's a good example I guess.

jacqueswww:
@davesque It's basically to avoid ambiguity. You could make a struct with an address called `msg` and the actual address is someting else.

davesque:
@jacqueswww Yeah, I guess in the context of smart contracts, the stakes are higher.  I was thinking that Python allows name shadowing in a lot of cases, but it's a different language with different goals.

jacqueswww:
Merging, all good - and got 2 thumbsup :)

jacqueswww:
@davesque yes exactly. Good example is we also have `blockscopes` in vyper which prohibits shadowing:

```python
for i in range(1, 100):
     ...
# <i is inaccessible here>
```

PR #1378, opened 2019-04-01
Add missing persistent address test cases.

### LINKED ISSUE #1376: Cannot return ERC20 token address from contract
Hi There,

I'm trying to compile the following contract running `master`:

**Input contract**

```python
from vyper.interfaces import ERC20

token: ERC20

@public
@constant
def tokenAddress() -> address:
    return self.token
```

**Output**

```
vyper.exceptions.TypeMismatchException: line 8: Trying to return base type address(ERC20), output expecting address
    return self.token
-----------^
```

I can fix it temporarily by changing the `address` to an `address(ERC20)`, this compiles to runtime bytecode, but doesn't compile when extracting the ABI:

**Input contract 2 `(-f abi)`**

```python
from vyper.interfaces import ERC20

token: ERC20

@public
@constant
def tokenAddress() -> address(ERC20):
    return self.token
```

**Output**

```
vyper.exceptions.InvalidTypeException: line 7: Invalid base unit
def tokenAddress() -> address(ERC20):
------------------------------^
```

jacqueswww:
### What I did
Fixes #1376 

### Cute Animal Picture

### DIFF
--- a/tests/parser/syntax/test_interfaces.py
+++ b/tests/parser/syntax/test_interfaces.py
@@ -1,0 +1,133 @@
+import pytest
+
+from vyper import (
+    compiler,
+)
+from vyper.exceptions import (
+    InvalidTypeException,
+    StructureException,
+    TypeMismatchException,
+    VariableDeclarationException,
+)
+
+fail_list = [
+    ("""
+from vyper.interfaces import ERC20
+a: public(ERC20)
+@public
+def test():
+    b: uint256 = self.a
+    """, TypeMismatchException),
+    ("""
+from vyper.interfaces import ERC20
+aba: public(ERC20)
+@public
+def test():
+    self.aba = ERC20
+    """, VariableDeclarationException),
+    ("""
+from vyper.interfaces import ERC20
+
+a: address(ERC20) # invalid syntax now.
+    """, StructureException),
+    ("""
+from vyper.interfaces import ERC20
+
+@public
+def test():
+    a: address(ERC20) = ZERO_ADDRESS
+    """, InvalidTypeException),
+    ("""
+a: address
+
+@public
+def test():  # may not call normal address
+    assert self.a.random()
+    """, StructureException),
+    ("""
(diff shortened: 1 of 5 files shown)


jacqueswww:
@Pet3ris could you try uniswap with this branch, also I remembered: Hayden has a more up to date vyper branch at https://github.com/Uniswap/contracts-vyper/tree/vyper_updates if that is a good base to start off of, I believe it was changed for beta7 or beta8 last.

Pet3ris:
@jacqueswww - my uniswap migration works nicely now using `-> address`. Will also check out the other branches - thanks very much for the fix and pointer!

tests/parser/syntax/test_interfaces.py
```diff
@@ -0,0 +1,133 @@
+import pytest
+
+from vyper import (
+    compiler,
+)
+from vyper.exceptions import (
+    InvalidTypeException,
```
jacqueswww: @fubuloubu I think I have all the valid/failed interface syntax pegged in this test case now. Could you just give it a once over please :)

tests/parser/syntax/test_interfaces.py
```diff
@@ -0,0 +1,133 @@
...
+b: ERC20
+@public
+def test(input: address):
+    assert self.b.totalSupply() == ERC20(input).totalSupply()
+    """,
+    """
+from vyper.interfaces import ERC20
+
+contract Factory:
+   def getExchange(token_addr: address) -> address: constant
```
fubuloubu: Can you do a `getToken(exchange_addr: address) -> address`  case that does the opposite of `test()`?

jacqueswww: @fubuloubu Opposite in what way ?

fubuloubu: Test that you can get exchange address for the token (`ERC20` is an internal interface, `Factory` is defined in code), then test that you can get the token for a given exchange.

PR #1434, opened 2019-05-12
Fix typo in function sqrt

rushikeshacharya:
### What I did
Fix typo in docs/built-in-function.rst/ : sqrt
### Cute Animal Picture

Put a link to a cute animal picture inside the parenthesis-->

PR #1437, opened 2019-05-15
Fix typo

rushikeshacharya:
### What I did
I have made the changes in the sqrt function as suggested in PR #1434.
### Cute Animal Picture

### DIFF
--- a/docs/built-in-functions.rst
+++ b/docs/built-in-functions.rst
@@ -181,15 +181,15 @@ Returns ``sha256`` (SHA2 256bit output) hash of input.
 --------
 ::
 
-  def sqrt(a: decimal) -> decimal:
+  def sqrt(a: decimal) -> b:
     """
     :param a:
     :type a: decimal, larger than 0.0
 
-    :output sqrt: decimal
+    :output b: decimal
     """
 
-Returns the suare of the provided decimal number, using the Babylonian square root algorithm.
+Returns the square of the provided decimal number, using the Babylonian square root algorithm.
 
 
 **method_id**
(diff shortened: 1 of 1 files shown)


docs/built-in-functions.rst
```diff
@@ -181,15 +181,15 @@ Returns ``sha256`` (SHA2 256bit output) hash of input.
 --------
 ::
 
-  def sqrt(a: decimal) -> decimal:
+  def sqrt(a: decimal) -> b:
```
fubuloubu: This one is actually okay as is. Here `decimal` is the type that `sqrt` returns, according to it's function signature.

fubuloubu:
```suggestion
  def sqrt(a: decimal) -> decimal:
```

docs/built-in-functions.rst
```diff
@@ -181,15 +181,15 @@ Returns ``sha256`` (SHA2 256bit output) hash of input.
...
     """
     :param a:
     :type a: decimal, larger than 0.0
 
-    :output sqrt: decimal
+    :output b: decimal
```
fubuloubu:
```suggestion
    :output sqrt: decimal
```

docs/built-in-functions.rst
fubuloubu: reverting this change, was fine as is

docs/built-in-functions.rst
fubuloubu: reverting this change, was fine as is

davesque:
@fubuloubu I'm looking through this documentation article and the format is a bit confusing.  The article uses a kind of ad hoc docstring style that mimics the sphinx docstring format but doesn't really adhere to it.  It also specifies each function as an inline code block.  Seems like we should be using the sphinx reStructureText API as seen [here](http://www.sphinx-doc.org/en/stable/usage/quickstart.html#documenting-objects) to just get nicely formatted, linkable entries in the documentation.

fubuloubu:
I think that's a good point. Can you add an issue to track this docs restructuring? (no pun intended)

fubuloubu:
This PR no longer applies

PR #1453, opened 2019-05-25
basic storage example

benjyz:
### What I did

add basic storage example originally from truffle

### DIFF
--- a/examples/storage.vy
+++ b/examples/storage.vy
@@ -1,0 +1,10 @@
+stored_data: uint256
+
+@public
+def set(new_value : uint256):
+    self.stored_data = new_value
+
+@public
+@constant
+def get() -> uint256:
+    return self.stored_data


examples/storage.vy
```diff
@@ -0,0 +1,10 @@
+stored_data: uint256
+
+@public
+def set(new_value : uint256):
+    self.stored_data = new_value
+
+@public
+@constant
+def get() -> uint256:
+    return self.stored_data
```
Mc01: @benjyz In Python you want to ensure that your script ends with newline, I guess in Vyper also 😉

jacqueswww:
@benjyz we already have a example contract that is very similar (and has tests) https://github.com/ethereum/vyper/blob/master/examples/name_registry/name_registry.vy

So I am closing this PR, but feel free to add any other examples (with tests) that you think could be helpful :)

jacqueswww:
I see name registry uses map(). So we can add this example, but then we need tests see https://github.com/ethereum/vyper/blob/master/tests/examples/name_registry/test_name_registry.py 

As an example.

examples/storage.vy
```diff
@@ -0,0 +1,10 @@
+stored_data: uint256
```
jacqueswww: add comment that one could also use:
```
stored_date: public(uint256)
```
To create a getter.

jacqueswww:
Closing, no action for a long time - feel free to create a PR with tests.

PR #1495, opened 2019-06-25
Import built-in interfaces in token examples

sbrichards:
### What I did
https://vyper.readthedocs.io/en/v0.1.0-beta.10/structure-of-a-contract.html#built-in-interfaces

### Cute Animal Picture

image


fubuloubu:
These examples should probably be failing without this import... That's a bug!

Issue #1481, opened 2019-06-13
Documentation on testing a contract is wrong


choosy:
### Version Information

* vyper Version (output of `vyper --version`): 0.1.0b10
* OS: linux
* Python Version 3.7.3
* Environment (output of `pip freeze`):

### What's your issue about?
I'm reading through docs here: https://vyper.readthedocs.io/en/v0.1.0-beta.10/testing-deploying-contracts.html
in regards to testing a contract. 

It talks about two options, one using pyethereum>2.0.0 one using vyper-run.

Problem is, pyethereum is marked as deprecated in ethereum account. And vyper-run command does not exist. How should this be?

fubuloubu:
Thanks for the issue! That whole page needs to be revamped to use Web3py for testing and deployment. The Vyper API also changed.

iamdefinitelyahuman:
I think this issue can be closed.

fubuloubu:
Yes! Closed as of #1497

Issue #1233, opened 2019-02-07
Add doc tests


fubuloubu:
### What's your issue about?
Some of the docs are out of date. We should add doctests so that doesn't happen.

### How can it be fixed?
Check out how [Web3py](https://github.com/ethereum/web3.py) does it

#### Cute Animal Picture
![Giraffe Cuddling](https://c1.staticflickr.com/6/5509/29479940863_dd3901cd8a_b.jpg)

matnad:
I am starting on this.

theref:
I know this issue is a bit old, but docs are always a great first contribution to a project! Do you have any examples of where you *know* the docs are outdated and would benefit from doc tests?

PR #1497, opened 2019-06-25
Reworked test and deploy documentation

matnad:
# What I did

Work on issue #1481

- Split documentation for testing and deploying a contract
- Completely revamped documentation for testing to use pytest and eth_tester; complete with examples.
- Updated documentation for deploying contracts
- Added two new examples of very simple storage contracts that are used in the testing documentation
- Added two new test files for said storage contracts
- Updated the assert_tx_failed fixture in conftest with another optional argument to check for text in exceptions (so we can check for a specific ValidationError for example)

UPDATE:
- Split conftest.py into:
     * base_conftest.py: includes the base code and fixtures to start testing vyper contracts
     * conftest.py: includes the rest of the functions and fixtures to test our code. Uses base_conftest as a plugin.
- Adapted Docs to reflect this and included the base_conftest as literalinclude so the docs stays up to date
- No more non-included code in the doc

### How I did it
Modifying the .rst doc files and creating new example contracts with tests

### How to verify it
Proof read the documentation and check the contracts (they are simple dummy contracts to showcase testing)

Updated documentation for testing and deploying contracts.

### Cute Animal Picture


docs/testing-contracts.rst
```diff
@@ -0,0 +1,241 @@
...
+This documentation recommends the use of the `pytest <https://docs.pytest.org/en/latest/contents.html>`_ framework with the `ethereum-tester <https://github.com/ethereum/ethereum-tester>`_ package.
+Prior to testing, the vyper specific contract conversion and the blockchain related fixtures need to be set up. These fixtures will be used in every test file and should therefore be defined in `conftest.py <https://docs.pytest.org/en/latest/fixture.html#conftest-py-sharing-fixture-functions>`_.
+
+=================================
+Vyper Contract and Basic Fixtures
+=================================
+
+.. code-block:: python
+
+  import pytest
```
jacqueswww: Perhaps we should consider splitting our existing conftest.py into base_conftest.py and conftest.py which we can include here and it will stay up to date. In our existing conftest.py we just import the base functionality ;)

jacqueswww:
Very cool!

fubuloubu:
Now that we include concrete examples, I definitely think doctest should be used. This PR need not depend on that though, but that should be a next step.

docs/testing-contracts.rst
matnad: Good idea. I'll make a proposition for this tomorrow and add it to the PR

matnad:
> 
> 
> Now that we include concrete examples, I definitely think doctest should be used. This PR need not depend on that though, but that should be a next step.

Is someone already working on doctests?

fubuloubu:
No, but it's here: #1233

This issue doesn't depend on doctests, but we really should add them to prevent out of date syntax being used in the docs.

matnad:
Updated PR with request

Issue #1355, opened 2019-03-15
VIP - Allow equality checks between string types.


pipermerriam:
## Simple Summary

Ability to compare equality of two `string` variable is useful and somewhat trivial to support.

## Abstract

Currently, assuming two `string` type variable `a` and `b`, the compiler does not allow `a == b` comparisons.

Byte-wise comparison is expensive and violates the requirement that computation be "decidable", we can get around this by using `keccak`.

## Motivation

To allow equality checks between strings

## Specification

when the compiler encounters `a == b` when both `a` and `b` are strings, this should be converted to `keccak(a) == keccak(b)`.

Gas cost optimizations to be considered.

- Consider doing a *pre* check of their lengths
- Consider special casing `lengh-0` strings to always be equal

## Backwards Compatibility

None?

## Dependencies

Nope

## Copyright
Copyright and related rights waived via [CC0](https://creativecommons.org/publicdomain/zero/1.0/)

fubuloubu:
> Consider doing a pre-check of their lengths

In the happy case (lengths match) this will cost a small amount more, and `keccak(a) == keccak(b)` will still fail if `a` and `b` are not the same length.

> Consider special casing `length-0` strings to always be equal

Same criticism as above. `keccak(null_string) == keccak(null_string)`

jacqueswww:
@fubuloubu I will model the gas gains of doing the length precheck and then we can decide if it's worth it?

PR #1507, opened 2019-06-28
Allow equality checks between string types

jakerockland:
### What I did

Implemented VIP #1355, allowing equality checks between string types.

### How I did it

Check strings with length > 32 or unequal lengths by comparing `keccak256` hash of each.

### How to verify it

Run the tests: `make test`. More specifically: `pytest -k test_string` and `pytest -k test_bytes`

Allow equality checks between string types

### Cute Animal Picture

image

### DIFF
--- a/vyper/parser/keccak256_helper.py
+++ b/vyper/parser/keccak256_helper.py
@@ -1,0 +1,69 @@
+
+from vyper.parser.lll_node import (
+    LLLnode,
+)
+from vyper.parser.parser_utils import (
+    getpos,
+    make_byte_array_copier,
+)
+from vyper.types import (
+    BaseType,
+    is_base_type,
+)
+from vyper.utils import (
+    MemoryPositions,
+    bytes_to_int,
+    keccak256,
+)
+
+
+def keccak256_helper(expr, args, kwargs, context):
+    sub = args[0]
+    # Can hash literals
+    if isinstance(sub, bytes):
+        return LLLnode.from_list(
+            bytes_to_int(keccak256(sub)),
+            typ=BaseType('bytes32'),
+            pos=getpos(expr)
+        )
+    # Can hash bytes32 objects
+    if is_base_type(sub.typ, 'bytes32'):
+        return LLLnode.from_list(
+            [
+                'seq',
+                ['mstore', MemoryPositions.FREE_VAR_SPACE, sub],
+                ['sha3', MemoryPositions.FREE_VAR_SPACE, 32]
+            ],
+            typ=BaseType('bytes32'),
+            pos=getpos(expr),
+        )
+    # Copy the data to an in-memory array
+    if sub.location == "memory":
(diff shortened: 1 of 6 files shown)


jakerockland:
@jacqueswww updated 👍😄

vyper/parser/keccak256_helper.py
```diff
@@ -0,0 +1,69 @@
...
+        )
+    # Copy the data to an in-memory array
+    if sub.location == "memory":
+        # If we are hashing a value in memory, no need to copy it, just hash in-place
+        return LLLnode.from_list(
+            ['with', '_sub', sub, ['sha3', ['add', '_sub', 32], ['mload', '_sub']]],
+            typ=BaseType('bytes32'),
+            pos=getpos(expr),
+        )
+    elif sub.location == "storage":
```
jacqueswww: @jakerockland I just read this now, after having merged it. Will this not only hash the first 32bytes?

jakerockland: @jacqueswww hrm yeah this is an oversight on my part too 😓 

Given that this is the same code generation that is (and was) being used for the built-in `keccak256` function as well, I think the issue here is a bit broader in scope than the string comparison stuff, in that for any hashes of _any_ string literal/bytes array longer than 32 bytes, we are only hashing the first 32 bytes.

Given our definition of the built-in supporting string literals + byte arrays:

```python
def sha3(a) -> b:
  """
  :param a: value to hash
  :type a: either str_literal, bytes, bytes32

  :output b: bytes32
  """
```

I think we may want to open another ticket specifying that we need to fix the `keccak256` behavior all-around to hash _all_ bytes + need to add tests around this both in the context of string comparison but also just using the built-in.

I will have some time this evening that I can get started with writing tests around this to verify the issue + can open up a ticket from there.

Thoughts? 😅

jacqueswww: Hmm just check again, pretty sure keccak256/sha3 has tests for longer than 32 byte strings.

jakerockland: Hrm, yeah you're right we do have a couple of tests for that in `tests/parser/functions/test_keccak256.py`.

I need to do a bit more reading on the LLL here, but it does seem to me that if things currently work properly for hashing byte arrays longer than 32, that string comparison will work as well, as they use the same helper method? 

Likewise, if there was an issue with the string comparison for strings/byte arrays longer than 32, it would make me think we need to add some more tests with the `keccak256` built-in.

Does that make sense? @jacqueswww

jakerockland: Either way, it probably would be good to add more nuanced tests to both `functions/test_keccak256.py` and `types/test_string.py`, which I'm happy to do this evening @jacqueswww — from there it will be a lot more clear what exactly is wrong 😅

jacqueswww: Yeah, clearly morw tricky than one would initially think! :)

PR #1519, opened 2019-07-08
Add support for ListType in external contract call

brockelmore:
### What I did
Imported ListType from types and added case for get_external_contract_call_output

### How to verify it
I don't know the codebase well enough to make sure this makes sense. Just submitting this PR in case it really is this simple. Someone more knowledgeable should make any necessary security checks.

Added support for fixed-length arrays as output from external contract call

### Cute Animal Picture

### DIFF
--- a/vyper/parser/external_call.py
+++ b/vyper/parser/external_call.py
@@ -16,8 +16,8 @@ from vyper.parser.parser_utils import (
 from vyper.types import (
     BaseType,
     ByteArrayLike,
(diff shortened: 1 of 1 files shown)


jacqueswww:
Looks good, to add the test case see: https://github.com/ethereum/vyper/blob/master/tests/parser/features/external_contracts/test_external_contract_calls.py
Where you can just add a test case for getting a list from an external call.

PR #1524, opened 2019-07-10
Basic type conversions from `address` type

jakerockland:
### What I did

Implements missing conversions from `address` to other basic types, as specified in #1093 — note conversion from `address` to `bytes` type has yet to be implemented, but was holding off on this for this specific PR as  there are also a handful of other conversions _to_ `bytes` type that need to be implemented still and feel that could just be its own PR.

### How I did it

Updated `convert.py` to support new conversions.

### How to verify it

Run the tests `make test`

Add missing conversions from `address` type to other basic types

### Cute Animal Picture

image


vyper/types/convert.py
```diff
@@ -105,6 +106,17 @@ def to_int128(expr, args, kwargs, context):
                 pos=getpos(expr)
             )
 
+    elif input_type == 'address':
+        return LLLnode.from_list(
+            [
+                'signextend',
+                15,
+                ['and', in_arg, 0xFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF],
```
jacqueswww: I think using SizeLimits here would be cleaner?

jacqueswww: SizeLimits.ADDRSIZE specifically :)

jakerockland: @jacqueswww ah good call! 😅 I had overlooked that that constant existed — will make this change tonight.

jakerockland:
@jacqueswww pushed! 👍😄

Dylan-Kerler:
Is uint256 conversion implemented here? I don’t see it; only decimal, int128 and bool. Also, I don’t see these changes in the docs

iamdefinitelyahuman:
I believe it is implemented, yes.  Good catch, that should definitely be documented.

Issue #1526, opened 2019-07-15
PyPI vs Source versions


CarlBeek:
### Version Information

* vyper Version (output of `vyper --version`): Source: `0.1.0b10`, PyPI: `0.1.0b10`
* OS: osx
* Python Version (output of `python --version`): Python 3.6.8
* Environment (output of `pip freeze`): [Not super relevant to my issue, but here](https://gist.github.com/CarlBeek/91dcdf4994f2ee4b98e1b2591c369451)

### What's your issue about?

When building Vyper from source and using PyPI the compiler versions are different. Obviously this is the case, as changes are applied to master (and thus master is not really at version `0.1.0b10` despite what `version` says), but it caught myself and the team at RuntimeVerification off guard when formally verifying the [Eth2.0 deposit contract](https://github.com/ethereum/eth2.0-specs/tree/dev/deposit_contract).

There is nothing explicitly wrong with how things are currently handled, but I feel like it is bad dev-ex to run into this issue (especially considering the nature of code being compiled by Vyper). In addition, note that many projects are going to use PyPI for installation which does not provide a commit-hash (which is ridiculous IMO) so comparing versions is harder than it needs to be. (Also downloading source-code from PyPI doesn't give the `.git` folders, so checksums don't match either.)

### How can it be fixed?

- I think changes to master should be committed to a `dev` branch (or some similar branch) or `version` should point to `0.1.0b10-dev` or similar.
- A further suggestion: in your build instructions, I recon it would be nice if it recommended (optionally) checking out a tag first before running `make`.

jacqueswww:
For your purposes I would suggest build of a git hash in any case. We have tried introducing git hash into the pypi builds but experienced issues see https://github.com/ethereum/vyper/pull/1220.

CarlBeek:
While I don't necessarily agree with you that `master` should be ahead of the latest revision process, I do agree that it is done on the majority of projects and that my personal preferences on how to order a repository shouldn't weigh on that. :)

> Perhaps we could revisit introducing the commit hash into --version output ?

What if it is only present in versions that are not on PyPI? (So as not not to run into [PyPI's 'local version' rules](https://www.python.org/dev/peps/pep-0440/#local-version-identifiers)

Going to close this issue, because there is not something immediately actionable.

fubuloubu:
> What if it is only present in versions that are not on PyPI?

[...]

PR #1528, opened 2019-07-16
Added extended version info

fubuloubu:
### What I did
Changed how --version flag displays to include hash of all files in the vyper package

### How I did it
Manual directory traversal + hashing

### How to verify it
Need to verify it works under all deployed scenarios of vyper:
1. PyPI
2. Snapcraft
3. Docker (on Windows!)

Added version configuration info to `--version` flag

### Cute Animal Picture

fluffy hedgehog

fixes: #1526

### DIFF
--- a/vyper/__init__.py
+++ b/vyper/__init__.py
@@ -1,3 +1,8 @@
+import hashlib as _hashlib
+import os as _os
+from pathlib import (
+    Path as _Path,
+)
 import sys as _sys
 
 import pkg_resources as _pkg_resources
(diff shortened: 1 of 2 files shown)


vyper/__init__.py
```diff
@@ -12,7 +17,25 @@
...
 
 
+def _get_version_hash(root: _Path) -> str:
+    m = _hashlib.sha1()
+    for dirname, _, file_list in _os.walk(root):
+        for fname in file_list:
+            if fname[-3:] == '.py':
+                with open(_Path(dirname).joinpath(fname), 'rb') as f:
+                    m.update(f.read())
+    return m.digest().hex()
```
jacqueswww: @fubuloubu couldn't we just subprocess this to git instead? `git rev-parse --short HEAD` I am worried this causes more confusion.

vyper/__init__.py
```diff
@@ -12,7 +17,25 @@
...
 except _pkg_resources.DistributionNotFound:
     __version__ = '0.0.0development'
+
+
+# Append first 5 chars of sha1 hash of all files in package directory
+# NOTE: First 5 chars used so as not to confuse with git commit hash
+# NOTE: This should only be used a sanity check, please check git commits
+#       if unsure of version correctness
+extended_version = '{}+sha1.{}'.format(__version__,
+                                       _get_version_hash(_Path(__file__).parent)[:5])
```
jacqueswww: I am also worried if we do this, we make running vyper very slow?

vyper/__init__.py
fubuloubu: Does snapcraft, docker, and PyPI all have git?

vyper/__init__.py
fubuloubu: We could make this function lazy

jacqueswww: ye that might be a good idea.

fubuloubu:
Close in favor of #1535

PR #1530, opened 2019-07-17
Disallow conflicting method IDs

davesque:
### What was wrong?

It's possible to compile a vyper contract that contains conflicting methods.  That is, if a contract has two or more methods that produce the same internal routing ID (used for method selection during contract execution), it will still compile and create a contract with a security vulnerability.

### How did I fix it?

Added a check for duplicate method IDs in the contract interface.

### How to verify it?

Run the tests.

* Fixed a bug that allowed contracts with conflicting methods (methods that produce the same internal routing ID) to compile without error.

### Cute Animal Picture

### DIFF
--- a/tests/signatures/test_method_id_conflicts.py
+++ b/tests/signatures/test_method_id_conflicts.py
@@ -1,0 +1,60 @@
+import pytest
+
+from vyper import (
+    compiler,
+)
+from vyper.exceptions import (
+    StructureException,
+)
+
+FAILING_CONTRACTS = [
+    """
+@public
+@constant
+def gsf():
+    pass
+
+@public
+@constant
+def tgeo():
+    pass
+    """,
+    """
+@public
+@constant
+def withdraw(a: uint256):
+    pass
+
+@public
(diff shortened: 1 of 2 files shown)


davesque:
@jacqueswww @fubuloubu Not sure if this check is happening in the right place.  But I guess we can move it to wherever it needs to go.

davesque:
@fubuloubu Didn't realize that same code was used in that context but yes, it should throw in that case as well.  Although it should probably say that the problem is with an external interface.

jacqueswww:
Interfaces is fine I reckon, the other location could've been: `FunctionSignature.from_definition`.

davesque:
@fubuloubu @jacqueswww Actually, this will *not* throw for issues with external interfaces as far as I know.  For example, with this patch, this contract compiles just fine:
```python
contract TestContract:
    def gsf() -> uint256: constant
    def tgeo() -> uint256: constant

@public
@constant
def foo() -> uint256:
    return 1
```

PR #1542, opened 2019-07-23
Added snap build process and build files to gitignore

fubuloubu:
### What I did
Added snapcraft release process (for reference)

koala


fubuloubu:
Rules work!

PR #1543, opened 2019-07-23
Added badgers for PyPI and Snapcraft release pages

fubuloubu:
### What I did
Added some badges because who doesn't like badges?

Lovely Lynx


davesque:
What about badgers?
badger

fubuloubu:
Ugh, the _obvious_ choice for this PR's cute animal

fubuloubu:
Who doesn't like badgers?

PR #1545, opened 2019-07-24
Updates to snapcraft build config

fubuloubu:
### What I did
Noted we were using an older version of Python 3.6. Also noted we were missing a base snap leading to inconsistent build results.

### Cute Animal Picture
ugh


fubuloubu:
Confirm this fixes the segfault issue with the snap!

PR #1546, opened 2019-07-24
Ensure snapcraft version matches setup.py

fubuloubu:
### What I did
Have the bumpversion utility upgrade the version we upload to snapcraft


fubuloubu:
Works! Now creates snap with name `vyper_<version>_amd64.snap` instead of git nonsense with wrong beta

PR #1557, opened 2019-08-01
Add end offsets to AST and source map

iamdefinitelyahuman:
### What I did
Added end offsets for the Vyper AST nodes and the source map.

### How I did it
* Add tokens to the python AST nodes using the [`asttokens`](https://github.com/gristlabs/asttokens) library.
* Use the tokens to add `end_lineno` and `end_col_offset` attributes to the final Vyper AST. (Attributes with these names are included in the standard ast library [as of Python 3.8](https://github.com/python/cpython/pull/11605))
* include these new attributes when generating the ``LLLnode`` objects, and add their values to the source map.

### How to verify it
I've checked it manually but tests still need to be written, as well existing tests will need updating. I will do so if and when the concept is finalized / approved.

### Comments / Thoughts
This PR is meant as a proof of concept.  I assume it will need some discussion and work before it can be merged. The following questions come to mind:

* Is it acceptable to introduce the `asttokens` library as a requirement?  If not, I can instead take heavy inspiration from it and include the same functionality without requiring the import.
* Is the output format for the source map offsets the correct approach?  I've handled it as a 4 item tuple of `(lineno, col_offset, end_lineno, end_col_offset)`. This is effectively a breaking change for any program expecting exactly a two item tuple. Other approaches could be:
  * a 2x2 tuple as `[(lineno, col_offset), (end_lineno, end_col_offset)]`
  * switch to using exact positions as `(start, length)` - more in line with solc output but less-so with python ASTs. Also a larger deviation from the current output and so a larger breaking change.

### Cute Animal Picture

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -55,6 +55,7 @@ setup(
     python_requires='>=3.6',
     py_modules=['vyper'],
     install_requires=[
+        'asttokens>=1.1.13,<2',
         'pycryptodome>=3.5.1,<4',
     ],
     setup_requires=[
--- a/vyper/ast.py
+++ b/vyper/ast.py
@@ -14,7 +14,14 @@ from vyper.utils import (
     annotate_source_code,
 )
 
-BASE_NODE_ATTRIBUTES = ('node_id', 'source_code', 'col_offset', 'lineno')
+BASE_NODE_ATTRIBUTES = (
+    'node_id',
+    'source_code',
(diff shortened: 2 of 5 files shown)


iamdefinitelyahuman:
Based on the call I'm closing this. I will attempt to implement the same functionality without the `asttokens` import and open a new PR.

PR #1564, opened 2019-08-06
Move kwarg check to more meaningful location

davesque:
### What I did

Moved the recently added kwarg check for self calls to a more meaningful location.  Now, the check is done just before the line of code where it makes a difference.

### How to verify it

Run the tests.

### Cute Animal Picture


davesque:
This one's not critical.  Just seemed a bit more clear IMHO.

PR #1565, opened 2019-08-06
Add context information to some error messages

davesque:
### What I did

Added context information for a couple of compiler exceptions.

### Cute Animal Picture

### DIFF
--- a/vyper/parser/stmt.py
+++ b/vyper/parser/stmt.py
@@ -1020,11 +1020,13 @@ class Stmt(object):
                 "Cannot modify storage inside %s: %s" % (
                     self.context.pp_constancy(),
                     target.annotation,
-                )
+                ),
+                self.stmt,
             )
         if not target.mutable:
             raise ConstancyViolationException(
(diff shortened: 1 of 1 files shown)


vyper/parser/stmt.py
```diff
@@ -1020,11 +1020,13 @@ def get_target(self, target):
...
                     target.annotation,
-                )
+                ),
+                self.stmt,
             )
         if not target.mutable:
             raise ConstancyViolationException(
-                "Cannot modify function argument: %s" % target.annotation
+                "Cannot modify function argument: %s" % target.annotation,
+                self.stmt,
```
fubuloubu: Will it just concat the two strings together, or do you need a second `%s`?

davesque: `self.stmt` is the AST object that gets passed to the exception constructor. It's a separate expression from the formatted string above it.  Doing that allows to the exception to be formatted with a message that shows the location of the error in the Vyper source file.

Issue #1367, opened 2019-03-25
VIP: Add support for relative imports.


jacqueswww:
## Simple Summary
Allow using `.` and `..` to specify relative imports of interfaces.

## Abstract
Following a long discussion on https://github.com/ethereum/vyper/pull/1361. It came to light that frameworks and programmers need more felxibility in terms of importing interfaces. 

## Motivation
The motivation is critical for VIPs that add or change Vyper's functionality. It should clearly explain why the existing Vyper functionality is inadequate to address the problem that the VIP solves as well as how the VIP is in line with Vyper's goals and design philosophy.

## Specification

`..` will specify parent directory of the current `.vy`  being parsed.
`.` will specify current directory of the `.vy` being parsed.

e.g.

```python
from .token_interface import TokenInterface
```
Will create an interface names `TokenInterface` from a file currently located at the same directory as the `.vy` file that is currently being parsed. 

Environment variable:
Vyper shold be allowed to change it's lookup directory from CWD to one set in `VYPER_PATHS`.

## Backwards Compatibility
Is forward compatible.

## Dependencies
None.

## Copyright
Copyright and related rights waived via [CC0](https://creativecommons.org/publicdomain/zero/1.0/)

fubuloubu:
@mikeshultz We decided to take a different tactic to do what you were doing in #1361. We hope you might comment on it here :)

mikeshultz:
For what it's worth, the toolkit I'm working on now invokes the `vyper` module directly and does import resolution itself instead of relying on invoking the CLI script.

That said, I still think this is a good idea. And if this resolution is not implemented in the compiler(but the CLI script with the package), this is the kind of thing that should be documented for other tools utilizing Vyper.

Generally, maybe this is the kind of thing that would be good to get some documented rules/best practices for how to resolve absolute paths as well. *If* you all think the current working directory is the best place to resolve from, that should be documented.  Maybe absolute paths and the possibility of having a library `PATH` is a different discussion, however.

fubuloubu:
Maybe the compiler API can be written such that if the compiler encounters any import statements, it references a "lookup" mapping of those packages to the specific source files they reference, which must be provided by the CLI script (and any other tools). This keeps the compiler functionally pure, and mitigates introducing the concept of the filesystem into the compiler.

mikeshultz:
[...]

PR #1566, opened 2019-08-08
use pathlib instead of os.path

iamdefinitelyahuman:
### What I did
Replace `os.path` with `pathlib.Path` - it makes for much cleaner code and reduces headaches when dealing with Windows file systems.

### How to verify it
Run the tests.

### Cute Animal Picture
image

### DIFF
--- a/bin/vyper
+++ b/bin/vyper
@@ -4,7 +4,9 @@ from collections import (
     OrderedDict,
 )
 import json
-import os
+from pathlib import (
+    Path,
+)
 import sys
 from typing import (
     Any,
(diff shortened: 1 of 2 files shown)


iamdefinitelyahuman:
Did this as a lead-up to #1367 - implementing relative imports will mean a lot more interaction with the file system.

davesque:
@fubuloubu @iamdefinitelyahuman By the way, I did run the `vyper` bin manually a couple of times to make sure the affected code paths were still behaving as expected, but I think I or someone should revisit the script invocation pr #1337 .

PR #1570, opened 2019-08-08
add root path, search for imports in multiple folders

iamdefinitelyahuman:
### What I did
* Added a root path argument to `bin/vyper`
* Expanded functionality re: locating imports

### How I did it
The `root_path` is set via the `-p` command line argument.  If none is given, it defaults to the current working path.  When importing an interface, Vyper first looks in the same folder as the contract performing the import, then in the root path.  This is very similar to Python's normal import behavior.

For example, imagine the following file structure:

* `./contracts/foo.vy`
* `./contracts/interfaces/bar.vy`
* `./baz.vy`

If the the compile script is run as `vyper contracts/foo.vy`, any of the following import statements are valid within `foo.vy`:

* `import contracts.interfaces.bar as Bar`
* `import interfaces.bar as Bar`
* `import baz as Baz`

I have also added a check to prevent the bug described at https://github.com/ethereum/vyper/issues/1367#issuecomment-519484066. If two contracts assign the same name to different interfaces, an `Exception` is raised.

### How to verify it
I have performed manual testing. Currently there are no tests in place for invocation scripts. I intend to implement tests for `bin/vyper` over the course of completing #1367 and #1520.

### Motivation
I feel this is a necessary first step toward #1367, and will also be useful when handling source paths in the input JSON for #1520. I'm looking for feedback on my approach before working on these issues further. If this is :+1: I will next expand it to include `from` style imports.

### Cute Animal Picture
image


bin/vyper
```diff
@@ -114,28 +121,27 @@ def exc_handler(contract_name: ContractName, exception: Exception) -> None:
...
+    for file_path, code in codes.items():
+        parent_path = root_path.joinpath(file_path).parent
+        base_paths = (parent_path, root_path)
+        new_codes = extract_file_interface_imports(code)
+        for key, value in new_codes.items():
+            interface_path = get_interface_file_path(base_paths, value)
+            if key in interface_codes and interface_codes[key] != interface_path:
                 raise Exception(
-                    f'Imported interface "{interface_path}{{.vy,.json}}" does not exist.'
+                    f"Cannot assign same key to multiple interfaces - '{key}'"
```
davesque: I actually think the appropriate thing to do here is to have a separate import namespace for each code file.  This is how things work in python and I think we should try to duplicate that behavior.  For example, in python, if different files happened to do `from ... import ... as foo`, there'd be no problem because each file represents a different module with its own namespace.  There's a comment about this [here](https://github.com/ethereum/vyper/pull/1361#issuecomment-474635884) and some discussion.

iamdefinitelyahuman:
Further to this, there should be discussion around https://github.com/ethereum/vyper/pull/1566#issuecomment-519636827, how to best structure script invocation tests.

Currently all the files in `bin` are using an `if __name__ == '__main__':` pattern, and declared in `setup.py` via the `scripts` kwarg.  It might be worth switching to an `entry_points` setup calling a `main()` function, to simplify testing.  But I'm no expert with `setuptools`, I happily defer to anyone who is :)

bin/vyper
iamdefinitelyahuman: It wouldn't be hard to implement, it just means compiling contracts individually with a fresh call to `get_interface_codes` each time.  @fubuloubu?

fubuloubu: I like different namespaces better. Ultimately, compatibility at the contract layer relies upon the ABI typing and not having the right text. Being actually strongly typed vs. Python is something to consider.

iamdefinitelyahuman:
@davesque @fubuloubu see new commits

https://github.com/ethereum/vyper/blob/2c06ad6987921f026719f59efb15b1783e36f773/vyper/compiler.py#L233-L239

I'm not a huge fan of this :point_up:, but it was the best approach I could think of without introducing breaking changes to  `compile_codes`.  If the `interface_codes` kwarg is given as a single dict (the previous standard) it follows the original rules and thus we have the global namespace.  If given as a nested dict of `{'contract name': {'interface name': "interface path"}}` the namespaces are localized.

I kinda feel like breaking things might be better than the ambiguity of allowing both possibilities...

fubuloubu:
Yeah, didn't want to break things quite yet because tool integrators need to be made aware of the change.

iamdefinitelyahuman:
- [x] add root path
- [x] allow `import` from root path
- [x] relative and absolute `from` imports
- [x] derived interfaces
- [ ] allow aliasing of `from` imports ? (needs discussion)
- [ ] unit tests (awaiting #1574)

iamdefinitelyahuman:
Closing for now, will reopen a fresh PR when finished.

PR #1576, opened 2019-08-19
Move bin to vyper/cli, use entry_points

### LINKED ISSUE #1128: Python command script on Windows
### Version Information

* vyper Version: v0.1.0-beta.5
* OS: win
* Python Version (python --version): 3.7
pycryptodome==3.7.2
vyper==0.1.0b5

### What's your issue about?

After installation, the command `vyper` does not work on Windows.

The installation does put a shell script in the Python folder %AppData%\Local\Programs\Python\Python37\Scripts

However, although this is on the path, it does not run on Windows.

Other programs often put an exe or similar file there instead.

### How can it be fixed?

I put a simple vyper.cmd file in the same folder, with the contents:

@python C:\Dev\vyper\build\scripts-3.7\vyper %*

This is a hack, because it hard codes to my development path, but something similar would work if configured with the correct location installed.

#### Cute Animal Picture

### LINKED ISSUE #1574: Use entry_points instead of scripts
## Simple Summary
Vyper's compiler scripts should be installed via `entry_points` and made available as `vyper.cli`

## Abstract
Using `entry_points` instead of `scripts` removes the need for `if __name__ == "main":` within Vyper's command-line scripts.  Placing the functionality within a `main()` method simplifies writing unit tests. Migrating the scripts to `vyper/cli` allows them to be imported and accessed by other packages.

`entry_points` also provides [better cross-platform compatibility](https://setuptools.readthedocs.io/en/latest/setuptools.html#automatic-script-creation), and it's use is recommended over `scripts` within the [PyPA docs](https://python-packaging-user-guide.readthedocs.io/guides/distributing-packages-using-setuptools/#scripts).

## Motivation
* I need to write unit tests for #1570 
* 3rd party tooling written in python should be able to access the full range of the compiler's functionality without relying on a subprocess call to reach `bin/vyper`

## Specification
* modify `setup.py`
* replace `if __name__ == "main":` with `def main():` in all `bin` scripts
* move the `bin` folder to `vyper/cli`, add `.py` suffix to files

## Backwards Compatibility
Should not cause any issues as scripts will still be invoked via the same commands, and the `bin` folder was not previously available to import as a package.

## Copyright

iamdefinitelyahuman:
### What I did
Change the cli scripts into entry points per #1574.

### How I did it
* `setup.py` - remove `scripts` and add `entry_points`
* move `bin` folder to `vyper/cli`
* in all scripts, create a `_parse_cli_args` method as an entry point, and place the logic that was behind `if __name__ == "main"` into a 2nd method so it can be called by other packages
* remove lines from `tox.ini` that are no longer relevant

### How to verify it
The first commit moved the `bin` folder, so comparison from [this diff](https://github.com/ethereum/vyper/compare/432204eb57cb23b7021ee2cb9ceebd5111992307...iamdefinitelyahuman:entry_points?expand=1) is much easier for code review.

There were no tests previously. I haven't added any yet, but will be doing so in #1570 and then again for #1520. 

### Cute Animal Picture

image

Closes #1128 
Closes #1574


tox.ini
```diff
@@ -48,8 +48,5 @@ commands =
     flake8 {toxinidir}/vyper {toxinidir}/tests {toxinidir}/scripts
     flake8 {toxinidir}/bin/vyper {toxinidir}/bin/vyper-lll {toxinidir}/bin/vyper-serve
     mypy --follow-imports=silent --ignore-missing-imports --disallow-incomplete-defs -p vyper
-    mypy --follow-imports=silent --ignore-missing-imports --disallow-incomplete-defs bin/vyper
-    mypy --follow-imports=silent --ignore-missing-imports --disallow-incomplete-defs bin/vyper-lll
-    mypy --follow-imports=silent --ignore-missing-imports --disallow-incomplete-defs bin/vyper-serve
     mypy --follow-imports=silent --ignore-missing-imports --disallow-incomplete-defs scripts
-    isort --recursive --check-only --diff {toxinidir}/vyper {toxinidir}/tests {toxinidir}/bin {toxinidir}/scripts
+    isort --recursive --check-only --diff {toxinidir}/vyper {toxinidir}/tests {toxinidir}/scripts
```
fubuloubu: @jacqueswww is the stuff in `scripts/` still relevant?

vyper/cli/vyper_compile.py
```diff
@@ -0,0 +1,207 @@
...
+    bytecode (default) - Deployable bytecode
+    bytecode_runtime   - Bytecode at runtime
+    abi                - ABI in JSON format
+    abi_python         - ABI in python format
+    ast                - AST in JSON format
+    source_map         - Vyper source map
+    method_identifiers - Dictionary of method signature to method identifier.
+    combined_json      - All of the above format options combined as single JSON output.
+    interface          - Print Vyper interface of a contract
+    external_interface - Print Externa Contract of a contract, to be used as outside contract calls.
```
fubuloubu:
```suggestion
    external_interface - Print External Interface of a contract, to be used as outside contract calls.
```

Typo I noticed (from original)

tox.ini
jacqueswww: for time being yes.

tox.ini
```diff
@@ -48,8 +48,5 @@ commands =
     flake8 {toxinidir}/vyper {toxinidir}/tests {toxinidir}/scripts
     flake8 {toxinidir}/bin/vyper {toxinidir}/bin/vyper-lll {toxinidir}/bin/vyper-serve
```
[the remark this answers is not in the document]
iamdefinitelyahuman: ooops...  good catch

vyper/cli/vyper_compile.py
iamdefinitelyahuman: :+1:, i probably should have dedented that too

PR #1578, opened 2019-08-21
Expand import system

### LINKED ISSUE #1367: VIP: Add support for relative imports.

iamdefinitelyahuman:
### What I did
Expanded the import system to allow relative imports, and absolute imports from different folders.

Closes #1367

### How I did it
* Added a root path argument to `vyper` cli script
* Added support for absolute and relative imports using `from` statement
* Expanded searching functionality for absolute imports
* Separated interface code namespaces when compiling multiple contracts

See the updates to [`docs/structure-of-a-contract.rst`](https://github.com/iamdefinitelyahuman/vyper/blob/35a6e5a0729dbfafb63abee276885b7e0c0f3334/docs/structure-of-a-contract.rst#contract-interfaces) for implementation details.

Related discussion threads - #1367, #1570

### How to verify it
Run the tests.

* Added support for relative imports

### Cute Animal Picture
image


vyper/cli/vyper_compile.py
```diff
@@ -143,65 +154,89 @@ def exc_handler(contract_name: ContractName, exception: Exception) -> None:
...
 
-    for code in codes.values():
-        interface_codes.update(extract_file_interface_imports(code))
+    for file_path, code in codes.items():
+        interfaces[file_path] = {}
+        parent_path = root_path.joinpath(file_path).parent
 
-    if interface_codes:
+        interface_codes = extract_file_interface_imports(code)
         for interface_name, interface_path in interface_codes.items():
```
fubuloubu: This seems like a major change in control flow logic. Is this supposed to be indented under the previous `for` loop?

fubuloubu: To be clearer, in the prior setup the `if interface_codes` statement escaped the previous `for` loop. This change sets the `interface_codes` variable directly _however_ it does *not* escape the previous `for` loop anymore.

iamdefinitelyahuman: See https://github.com/ethereum/vyper/issues/1367#issuecomment-519484066 -  In the prior setup, the call to `extract_file_interface_imports` happened once per compile which resulted in a single shared interface namespace across all contracts being compiled. Making the call once per contract allows the namespaces to be isolated from each other.

docs/structure-of-a-contract.rst
```diff
@@ -165,120 +165,172 @@ Additional information about Ethereum Natural Specification (NatSpec) can be fou
...
-      FooBar(addy).one()
+    # External Contracts
+    contract Ballot:
+        def delegated(addr: address) -> bool: constant
+        def directlyVoted(addr: address) -> bool: constant
+        def giveRightToVote(voter: address): modifying
+        def forwardWeight(delegate_with_weight_to_forward: address): modifying
+        # ...
+
+The output can then easily be copy-pasted to be consumed.
```
jacqueswww: Very nice docs, thanks for the improvements!

vyper/cli/vyper_compile.py
```diff
@@ -143,65 +154,89 @@ def exc_handler(contract_name: ContractName, exception: Exception) -> None:
...
 
+def get_interface_file_path(base_paths: Sequence, import_path: str) -> Path:
+    relative_path = Path(import_path)
+    for path in base_paths:
+        file_path = path.joinpath(relative_path)
+        try:
+            suffix = next(i for i in ('.vy', '.json') if file_path.with_suffix(i).exists())
+            return file_path.with_suffix(suffix)
+        except StopIteration:
+            continue
```
jacqueswww: Not a necessary change - just something interesting I found: Apparently you can now pass None in to next to make the control flow nicer:
```python
suffix = next(i for i in ('.vy', '.json') if file_path.with_suffix(i).exists(), None)
if suffix:
   return file_path.with_suffix(suffix)

File: vyper/cli/vyper_compile.py [hunk 4]
```diff
@@ -143,65 +154,89 @@ def exc_handler(contract_name: ContractName, exception: Exception) -> None:
...
     codes: ContractCodes = OrderedDict()
     for file_name in input_files:
-        with open(file_name) as fh:
-            codes[file_name] = fh.read()
+        file_path = Path(file_name).resolve()
+        file_str = file_path.relative_to(root_path).as_posix()
+        with file_path.open() as fh:
+            codes[file_str] = fh.read()
 
+    version = False
```
jacqueswww: Rename the this variable to something more descriptive like `set_version`  / `show_version` ?

vyper/cli/vyper_compile.py
iamdefinitelyahuman: :+1:  it looks cleaner without the try/catch, will add

PR #1579, opened 2019-08-24
Compile-time check for missing args in private calls

iamdefinitelyahuman:
### What I did
Added a check for calls to private functions where no args are given but the call expects args. Prior to this PR, the following code would compile:

```python
@private
def bar(a: int128) -> int128:
    return 1

@public
def foo() -> int128:
    return self.bar()
```

I also tweaked the inputs to `pack_arguments`, allowing for code highlights in the exceptions raised there.

### How I did it
* `vyper.parser.self_call.call_self_private`
  * added a check when no args are given that raises if args were expected
* `vyper.parser.parser_utils.pack_arguments`
   * replaced input arg `pos` with `stmt_expr`, derive position using `getpos(stmt_expr)`
   * include `stmt_expr` when raising exceptions
   

### How to verify it
Run the tests.  I've added tests to check for args when none should be given, no args when args should be given, and the wrong number of args.

### Cute Animal Picture
image

### DIFF
--- a/tests/parser/features/external_contracts/test_external_contract_calls.py
+++ b/tests/parser/features/external_contracts/test_external_contract_calls.py
@@ -1,3 +1,5 @@
+import pytest
+
 from vyper.exceptions import (
     InvalidTypeException,
     StructureException,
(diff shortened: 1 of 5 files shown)


jacqueswww:
LGTM :racehorse:

PR #1580, opened 2019-08-24
AST end offsets and compressed source map

iamdefinitelyahuman:
### What I did
* Add end offsets to Vyper AST nodes and `pc_pos_map`
* Add jump map and compressed source map to compiler source map output

### How I did it
* Add tokens to the python AST nodes using the [`asttokens`](https://github.com/gristlabs/asttokens) library
* Use the tokens to add `end_lineno` and `end_col_offset` attributes to the final Vyper AST. (Attributes with these names are included in the standard ast library [as of Python 3.8](https://github.com/python/cpython/pull/11605))
* include these new attributes when generating the ``LLLnode`` objects, and add their values to `pc_pos_map` as returned by `vyper.compile_lll.assembly_to_evm`
* add `pc_jump_map` to `assembly_to_evm` output, in the format `{'program counter': [io-]}` - `i` indicates a jump instruction going into a private function, `o` coming out of a private function, `-` is an internal instruction such as a for loop or if / else statement.
* added `vyper.compiler.compress_source_map`: given source code, `pc_pos_map` and `pc_jump_map`, generates a compressed source map in the [same format used by `solc`](vyper.compiler)
* added `vyper.compiler.expand_source_map`: given a compressed source map string, expands it into a list of 4 item lists as `[start, length, contract_id, jump]`
* include `pc_jump_map` and `pc_pos_map_compressed` to compiler output given by `vyper.compiler.get_source_map`

### How to verify it
Run the tests.  I have added several new test cases around the jump map and compressed source map.

* Add end offsets to source map

### Cute Animal Picture
image


jacqueswww:
@iamdefinitelyahuman I think the CLI switch is missing?

iamdefinitelyahuman:
For now I've just included it as part of the output for `source_map` - my next target is #1520 where I plan to separate it all out a bit more.

If you'd like I can add another flag, `compressed_source_map` maybe?

jacqueswww:
I reckon that's fine for now, especially because of #1520 :+1:

PR #1586, opened 2019-08-27
Build process was missing git as a dependency for containers

fubuloubu:
Wasn't able to build snap or docker images for the library, noticed the `git` subprocess was failing because it wasn't installed


jacqueswww:
Ah off course! 🙈

PR #1593, opened 2019-08-30
Add check for only single @+@nonreentrant decorator.

jacqueswww:
### What I did
As the code currently stands, only a single nonreentrant decorator is supported.
### Cute Animal Picture

### DIFF
--- a/tests/parser/exceptions/test_structure_exception.py
+++ b/tests/parser/exceptions/test_structure_exception.py
@@ -185,6 +185,13 @@ struct S:
     x: int128
 s: S = S()
     """,
+    """
+@public
+@nonreentrant("B")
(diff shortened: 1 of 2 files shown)


jacqueswww:
As the code is currently written it only has support for one lock.

jacqueswww:
But could add support here for multiple, if you have a use case for me?

fubuloubu:
I was just thinking someone could have an advanced use case with methods `A`, `B`, and `C` where using `A` should disable `C` (but not `B`), etc.

Might be too advanced to be clear how it works, but was just thinking about that scenario. I have no specific use case though.

jacqueswww:
Yeah, @maurelian was saying the same thing. Will add it in later.

PR #1633, opened 2019-10-02
Remove for orelse for For

jacqueswww:
### What I did
We don't support For .. orelse. We should not let it blindly not generate code.

### Cute Animal Picture

```
^.^
```

### DIFF
--- a/vyper/ast.py
+++ b/vyper/ast.py
@@ -4,7 +4,7 @@ from itertools import (
 import typing
 
 from vyper.exceptions import (
-    CompilerPanic,
+    SyntaxException,
 )
 from vyper.settings import (
     VYPER_ERROR_CONTEXT_LINES,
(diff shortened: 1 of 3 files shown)


vyper/parser/stmt.py
```diff
@@ -534,7 +534,7 @@ def parse_for(self):
                 "..` or `for i in range(start, start + rounds): ..`"
             ), self.stmt.iter)
 
-        block_scope_id = id(self.stmt.orelse)
```
charles-cooper: i guess the original was a little weird

vyper/ast.py
```diff
@@ -43,9 +43,9 @@ def __init__(self, **kwargs):
             if field_name in self.get_slots():
                 setattr(self, field_name, value)
             elif value:
-                raise CompilerPanic(
-                    f'Unsupported non-empty value field_name: {field_name}, '
-                    f' class: {type(self)} value: {value}'
+                raise SyntaxException(
+                    f'Unsupported non-empty value (valid in Python, but invalid in Vyper) \n'
```
charles-cooper: I'm not convinced this is always the correct error message, but maybe it's not important.

vyper/parser/stmt.py
fubuloubu: Is there any concern here?

vyper/ast.py
fubuloubu: `SyntaxException` is probably a little clearer. There error message is fine.

fubuloubu:
@jacqueswww this is ready for merge, right?

jacqueswww:
Yes, it is.

vyper/parser/stmt.py
jacqueswww: @charles-cooper BTW with the new AST classes we have a node_id attribute now :) So we can replace all these hacks with using .node_id. Nice when old hacks are not necessary anymore.

PR #1678, opened 2019-11-04
Add support for Python3.8

iamdefinitelyahuman:
### What I did
Add support for Python 3.8

### How I did it
* Replace `Constant` python AST type with deprecated type names when generating Vyper AST nodes.
* Add python3.8 build to CI

### How to verify it
Run the tests

* support for Python3.8

### Cute Animal Picture
image

### DIFF
--- a/vyper/ast_utils.py
+++ b/vyper/ast_utils.py
@@ -75,34 +75,33 @@ def parse_python_ast(source_code: str,
                      source_id: int = 0) -> vyper_ast.VyperNode:
     if isinstance(node, list):
         return _build_vyper_ast_list(source_code, node, source_id)
-    elif isinstance(node, python_ast.AST):
-        class_name = node.__class__.__name__
-        if isinstance(node, python_ast.Constant):
-            if isinstance(node.value, (int, float)):
-                class_name = "Num"
-            elif isinstance(node.value, str):
-                class_name = "Str"
-            elif isinstance(node.value, bytes):
-                class_name = "Bytes"
-            elif node.value is None or isinstance(node.value, bool):
-                class_name = "NameConstant"
-        if not hasattr(vyper_ast, class_name):
-            raise SyntaxException(
-                f'Invalid syntax (unsupported "{class_name}" Python AST node).', node
-            )
-        vyper_class = getattr(vyper_ast, class_name)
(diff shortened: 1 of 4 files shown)


fubuloubu:
Huh, fuzzing actually caught something this time lol

iamdefinitelyahuman:
Failing test should be fixed by https://github.com/ethereum/vyper/pull/1679

fubuloubu:
@iamdefinitelyahuman rebase master

fubuloubu:
I'd like to merge #1657 first and rebase on that to ensure we're playing with all the latest toys

vyper/ast_utils.py
```diff
@@ -87,17 +76,33 @@ def parse_python_ast(source_code: str,
...
+        if isinstance(node, python_ast.Num):
+            class_name = "Num"
+        elif isinstance(node, python_ast.Str):
+            class_name = "Str"
+        elif isinstance(node, python_ast.Bytes):
+            class_name = "Bytes"
+        elif isinstance(node, python_ast.NameConstant):
+            class_name = "NameConstant"
         else:
+            class_name = node.__class__.__name__
```
jacqueswww: Not sure I understand the fix here, as far as I can tell all these types are replaced by Constant(value=) now, doesn't that mean we have to translate each constant type and port over the value attribute (e.g. `.s` for Str and `.n` for number) ?

jacqueswww: And would that mean something the replacement step would have to map something like `isinstance(node.value, str)` becomes 'vyper_ast.Str(s=node.value)`

iamdefinitelyahuman: See new commits based on our discussion :+1:

fubuloubu:
technically, `eth-tester` doesn't support 3.8 _yet_. That could be why this is an issue.

cc @carver

carver:
> technically, `eth-tester` (and `py-evm`) doesn't support 3.8 _yet_. That could be why this is an issue.
> 
> cc @carver

Ahh, it looks like it's the blake2b-py build that's not compatible somehow. cc @davesque

davesque:
Spending some time on this today.

davesque:
Just released blake2b-py v0.1.3 that includes support for python 3.8.

PR #1726, opened 2019-11-13
Fix fornat typo

### LINKED ISSUE #1725: AttributeError: 'str' object has no attribute 'fornat' (typo in vyper source)
Same version/environment issue as previous TSTL-discovered bugs.

```
cv0: constant(uint256) = block.difficulty
cv1: constant(uint256) = convert(cv0, uint256)
@public
def f0():
    assert cv0
```

gets:

```
(vyper-venv) cmp3149:vyper adg326$ vyper vfile.9938.vy 
Error compiling: vfile.9938.vy
AttributeError: 'str' object has no attribute 'fornat'
```

The fix is pretty trivial, see the line in `convert.py`:

`vyper/types/convert.py:            raise InvalidLiteralException("Unknown numeric literal type: {}".fornat(in_arg))`

I'll maybe do a tiny PR to fix now.

agroce:
### What I did

Fixed #1725 

### How I did it

tiny edit to change `fornat` to `format`

### How to verify it

run code from #1725

Fix typo in an error message for bad converts

### Cute Animal Picture

### DIFF
--- a/vyper/types/convert.py
+++ b/vyper/types/convert.py
@@ -192,7 +192,7 @@ def to_uint256(expr, args, kwargs, context):
                 pos=getpos(expr)
             )
         else:
-            raise InvalidLiteralException("Unknown numeric literal type: {}".fornat(in_arg))
+            raise InvalidLiteralException("Unknown numeric literal type: {}".format(in_arg))
 
     elif isinstance(in_arg, LLLnode) and input_type == 'int128':
         return LLLnode.from_list(


fubuloubu:
I noticed that you don't need the function definition for this exception condition to occur:
```python
C1: constant(uint256) = block.difficulty  # block.number also works... 
C2: constant(uint256) = convert(C1, uint256) 
```

fubuloubu:
There should definitely be a test for this condition if that exception was never hit. I can't figure out a good test case though that doesn't seem magical

fubuloubu:
Also, we should probably disallow environment variables from constant definitions entirely...

agroce:
True, the function is optional.  The TSTL reducer doesn't detect that, due to the setup right now not running just constant/var decl "contracts"

fubuloubu:
Not necessarily, since it wouldn't produce a non-trivial example (an end-user would never do that), but for testing purposes it makes it clearer where the syntax is failing.

agroce:
I'm doubtful it'll expose any bugs not detectable with some trivial function declaration, though arguably it means the tester can find them without producing a vallid-ish function, so I think I'll give it a shot

PR #1737, opened 2019-11-21
Fix coveralls

fubuloubu:
I think coveralls isn't pushing because of the `parallel` flag being set when the API is called.


iamdefinitelyahuman:
I've missed this so much.

PR #1744, opened 2019-11-27
Cleanup: Ethereum -> vyperlang (where appropiate)

fubuloubu:
Also changed some odds and ends around titles of things and the package
author


fubuloubu:
Wanted at least two reviews of this, make sure everyone agreed with the changes made

PR #1748, opened 2019-11-28
Added coverage reporting to Python 3.7/3.8

fubuloubu:
Codecov _should_ merge these?
(see https://docs.codecov.io/docs/merging-reports)

---

This is important for catching coverage data in case we make version-specific imports for a feature


fubuloubu:
Yay it worked!

PR #1749, opened 2019-11-29
Forgot f string

fubuloubu:
Found a location where f-string was missing

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -20,6 +20,7 @@ test_deps = [
 lint_deps = [
     'flake8>=3.7,<4',
     'flake8-bugbear>=19.8.0,<20',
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -20,6 +20,7 @@
 lint_deps = [
     'flake8>=3.7,<4',
     'flake8-bugbear>=19.8.0,<20',
+    'flake8-use-fstring>=1.0.0,<2.0.0',
```
fubuloubu: Package looks unmaintained... but simple and works well?

fubuloubu: Can always remove it if it gives us problems

fubuloubu:
@charles-cooper just sqash merge this one

PR #1752, opened 2019-11-29
Make MyPy it's own task

fubuloubu:


fubuloubu:
Not as big an improvement as I was hoping, but MyPy will certainly take longer once we've annotated all of our codebase

PR #1760, opened 2019-12-09
Upgrade to use as many CPUs as available locally

fubuloubu:
### What I did
`pytest-xdist` adds the ability to parallelize tests using the `-n [NUM]` flag. It also allows dynamically checking how many are available: https://github.com/pytest-dev/pytest-xdist#speed-up-test-runs-by-sending-tests-to-multiple-cpus

### Cute Animal Picture
roadrunner

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,5 +1,6 @@
 [tool:pytest]
-addopts = -n 4
+# use as many CPUs as available (via pytest-xdist)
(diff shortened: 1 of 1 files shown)


fubuloubu:
`auto` is a better default, because it doesn't over-create threads for test process

fubuloubu:
CI now has more consistent timing!

PR #1770, opened 2019-12-15
Removed unnecessary JS file

fubuloubu:
Turns out, the signing logic is tested with Python in `tests/examples/wallet/test_wallet.py`, therefore the JS file is unnecessary


fubuloubu:
Should remove JS analysis

PR #1774, opened 2019-12-19
Bugfix - Importing an interface that contains an 'implements' statement

iamdefinitelyahuman:
### The Issue
When importing a Vyper contract as an interface, a `StructureException` is raised if that file includes an `implements` statement.

### How I fixed it
In `vyper/signatures/interface.py`, I modified `extract_sigs` to exclude AST nodes related to `implements`.

### How to verify it
Run the tests. I added a new test case around this behavior.

### Cute Animal Picture
image


jacqueswww:
LGTM

PR #1751, opened 2019-11-29
AST Optimizer

fubuloubu:
### What I did
I've been wanting to do this for a long time. I think utilizing the AST more can make some of our type-checking and optimization code a lot clearer, nevermind the ability to better verify the code using mypy. This PR gets the work started by replacing `RewriteUnarySubVisitor`, which I noticed was buggy in #1721. It is based off the work of that PR, so start the review of this PR from 
0c3c5e8

### How I did it
Copypasta from Python's AST module

### How to verify it
https://greentreesnakes.readthedocs.io/en/latest/tofrom.html is a fantastic resource

### Description for the changelog
Added AST optimizer

### Cute Animal Picture
exhausted bee

PR #1775, opened 2019-12-20
Add Python AST compatibility to VyperNode classes

fubuloubu:
### What I did
Added property `_fields` and made `VyperNode` a subclass of Python's `ast.AST` class, giving us the ability to directly use features from Python's `ast` module on our own ast objects.

### How to verify it
```python
>>> from vyper.parser import parser
>>> src = open('examples/name_registry/name_registry.vy', 'r').read()
>>> ast = parser.parse_to_ast(src); [print(n.__class__.__name__) for n in ast]
AnnAssign
FunctionDef
FunctionDef
>>> from ast import iter_fields
>>> [print(n, "=", v.__class__.__name__) for n, v in iter_fields(ast[0])]
# AnnAssign fields:
target = Name
annotation = Call
value = NoneType
simple = int
# VyperNode fields:
source_code = str
node_id = int
src = str
lineno = int
col_offset = int
end_lineno = int
end_col_offset = int
```

### Cute Animal Picture
silly llama

### DIFF
--- a/vyper/ast.py
+++ b/vyper/ast.py
@@ -37,6 +37,12 @@ class VyperNode:
             for klass in cls.__class__.mro(cls)
         ))
 
+    @property
+    def _fields(self):
+        # NOTE: Returns all the fields of a node in a manner compliant
+        #       with the expectations of the Python AST module
+        return self.get_slots()
+
     def __init__(self, **kwargs):
 
         for field_name, value in kwargs.items():
--- a/vyper/ast.py
+++ b/vyper/ast.py
@@ -1,3 +1,4 @@
+from ast import AST
 from itertools import (
     chain,
 )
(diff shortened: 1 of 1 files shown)


fubuloubu:
This PR will let me refactor #1751 not to have to define my own copy of the python visitor functionality (see commit 821cf50)

jacqueswww:
@fubuloubu I thought the whole idea is that Vyper AST should not depend on python AST ?

fubuloubu:
I don't think it's necessary to fully differentiate between our node classes and the python's classes. Python's classes can help act as a tool to create and work with a grammar fully independent of Python's, if they are useful in that regard.

For example, if we can use it to completely remove some code from our codebase (such as around node display) then that's a net positive because we can rely on code with much higher assurance to it for the given task instead of writing our own. The eventual goal was to create a fully independent grammar customized to Vyper's needs, not just use different classes with the same configuration as Python's, making it harder to read and understand.

Regardless, it seems like there are problems with this approach according to the tests.

PR #1780, opened 2019-12-23
Fix inconsistent AST offsets in Python 3.8

iamdefinitelyahuman:
### The Issue

Python 3.8 has changed how line-based offsets are determined for some nodes. For example, with a `FunctionDef` that includes decorators, the decorators are included within the offsets in `<3.8`, they are not in `3.8`.

`asttokens` determines offsets using tokens - it does not have the same inconsistency between Python versions.

In `vyper/ast_utils.py` when generating the fields for nodes in `_build_vyper_ast_init_kwargs`, `asttokens` is used for both the line-based offsets and the absolute offset field `src`. The line-based offsets are then replaced by the Python AST offsets during the `get_slots()` iteration, creating an inconsistency between them and `src`.

### How I fixed it
Ignore all fields returned by `get_slots()` if they are in `vyper.ast.BASE_NODE_ATTRIBUTES`.

### How to verify it
Run the tests.

### Cute Animal Picture
image

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -59,7 +59,7 @@ setup(
     python_requires='>=3.6',
     py_modules=['vyper'],
     install_requires=[
-        'asttokens==1.1.13',
+        'asttokens==2.0.3',
         'pycryptodome>=3.5.1,<4',
     ],
     setup_requires=[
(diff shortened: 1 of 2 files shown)


iamdefinitelyahuman:
My first commit brought out another problem - the version of `asttokens` in use in Vyper had compatibility issues with Python 3.8 as well. These issues were fixed in https://github.com/gristlabs/asttokens/pull/36, which was included in version `2.0.0`. I've pushed another commit here to bump the dependency version to the latest release, `2.0.3`.

PR #1781, opened 2019-12-23
Fix Windows path issue when importing interfaces

iamdefinitelyahuman:
### The Issue
To prevent imports outside of the the root path, `vyper-json` prepends import paths with `/__vyper` and then ensures the final resolved path is still a subfolder of that path.  On Windows this check is failing with `ValueError: 'C:\\__vyper\\path.vy' does not start with '\\__vyper'`.  Oopsies.

You can see a detailed traceback for the issue [here](https://travis-ci.com/iamdefinitelyahuman/brownie/jobs/269839445).

### How I fixed it
Perform the `relative_to` check against a `Path` object, so that it works on any filesystem.

### How to verify it
I confirmed that it works in a Windows 10 VM. Not sure if anyone wants to add Windows to the CI...

### Cute Animal Picture
image

### DIFF
--- a/vyper/cli/vyper_json.py
+++ b/vyper/cli/vyper_json.py
@@ -152,9 +152,10 @@ def exc_handler_to_dict(file_path: Union[str, None],
 
 
 def _standardize_path(path_str: str) -> str:
-    path = Path("/__vyper/" + path_str.lstrip('/')).resolve()
+    root_path = Path('/__vyper').resolve()
+    path = root_path.joinpath(path_str.lstrip('/')).resolve()
     try:
-        path = path.relative_to("/__vyper")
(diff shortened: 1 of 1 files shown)


iamdefinitelyahuman:
Ignore my force pushes, was trying to get Windows CI to work but there's an upstream issue installing `pyethash` (https://github.com/ConsenSys/ethjsonrpc/issues/5) which is required by `py-evm` (https://github.com/ethereum/py-evm/issues/395). Unless we remove `py-evm` as a testing dependency, running tests on Windows is a no-go.

fubuloubu:
What if we don't run the full test-suite with Windows, but instead run just the CLI stuff?

PR #1783, opened 2019-12-26
Evm versions

### LINKED ISSUE #1230: VIP: EVM Ruleset Switch
## Simple Summary
People need to produce code for different versions of the EVM

## Abstract
Provide an interface to switch the VM rulesets used to compile, validate, and produce bytecode for a given source program.

## Motivation
It is important for developers to be able to produce bytecode for different versions of the EVM. This may include for future hardfork implementations (such as constantinople), or for chains that still use past rules (ETC or permissioned networks like Quorum). It might also be eventually useful to specify more granular or custom rulesets via `py-evm`, but that is out of scope for now.

## Specification
Solidity has `--evm-version VM_NAME` flag:
```bash
$ solc --help
...
  --evm-version version
                       Select desired EVM version. Either homestead, 
                       tangerineWhistle, spuriousDragon, byzantium (default) or
                       constantinople.
...
```
I suggest we use this.

The technical details should be as such:
1. The source program is compiled and/or statically analyzed for features at the AST or IR (LLL) layer
2. Features that are not possible in the chosen ruleset raise an exception
3. Features that have alternative implementations in the chosen ruleset generate that alternative, and note the modification (vs. the default)

## Backwards Compatibility
Fully backwards compatible

## Dependencies
No dependancies on other VIPs

## Copyright

### LINKED ISSUE #1654: Use SELFBALANCE opcode for balance queries
### What's your issue about?
Istanbul is changing the gas costs for balance queries via [EIP-1884](https://eips.ethereum.org/EIPS/eip-1884). Since it's significantly easier to query the local balance of the contract, and since that is a pretty wildly used feature that people use, that EIP also _reduces_ the costs for self balance checks via a new opcode.

### How can it be fixed?
An optimization should take place that replaces any self balance queries with the `SELFBALANCE` opcode. Either that, or querying it via `self.balance` should directly call this opcode.

Note: might be best to add this feature via #1230 ruleset switch so that it only uses this optimization under Istanbul rules

iamdefinitelyahuman:
### What I did
* Add the ability to target multiple EVM versions - closes #1230 
* Use the `SELFBALANCE` opcode when targeting istanbul - closes #1654 

### How I did it
* Expanded `vyper/opcodes.py`:
  * Added missing opcodes
  * Where gas cost changes between versions, a tuple is used instead of an integer. Values correspond to (byzantium, constantinople, istanbul) - `None` means the opcode is not available, and where the length is < 3 it means the price does not change for future versions.
  * Use getter functions for opcodes, instead of directly importing constants into other modules
* `vyper.compiler.compile_codes` now accepts `evm_version` as a kwarg. If not given, the default value is used (`vyper.opcodes.DEFAULT_EVM_VERSION`, currently byzantium)
* Added the option to select EVM version in `vyper.cli` modules 

### How to verify it
Run the tests.  I've added new test cases and parametrized a few existing ones.

### Cute Animal Picture
image

### DIFF
--- a/tests/cli/vyper_compile/test_compile_files.py
+++ b/tests/cli/vyper_compile/test_compile_files.py
@@ -20,3 +20,27 @@ def test_combined_json_keys(tmp_path):
 def test_invalid_root_path():
     with pytest.raises(FileNotFoundError):
         compile_files([], [], root_folder="path/that/does/not/exist")
+
+
+def test_evm_versions(tmp_path):
+    # should compile differently because of SELFBALANCE
+    code = """
+@public
+def foo() -> uint256(wei):
+    return self.balance
+"""
+
+    bar_path = tmp_path.joinpath('bar.vy')
+    with bar_path.open('w') as fp:
+        fp.write(code)
+
+    compile_data = compile_files(
+        [bar_path],
+        output_formats=['bytecode_runtime'],
+        evm_version="byzantium"
+    )
+    assert compile_data != compile_files(
+        [bar_path],
+        output_formats=['bytecode_runtime'],
+        evm_version="istanbul"
+    )
--- a/tests/cli/vyper_compile/test_compile_files.py
+++ b/tests/cli/vyper_compile/test_compile_files.py
@@ -34,13 +34,19 @@ def foo() -> uint256(wei):
     with bar_path.open('w') as fp:
         fp.write(code)
 
-    compile_data = compile_files(
+    byzantium_bytecode = compile_files(
         [bar_path],
-        output_formats=['bytecode_runtime'],
+        output_formats=['bytecode'],
         evm_version="byzantium"
-    )
-    assert compile_data != compile_files(
+    )[str(bar_path)]['bytecode']
+    istanbul_bytecode = compile_files(
         [bar_path],
-        output_formats=['bytecode_runtime'],
+        output_formats=['bytecode'],
         evm_version="istanbul"
-    )
+    )[str(bar_path)]['bytecode']
+
+    assert byzantium_bytecode != istanbul_bytecode
+
+    # SELFBALANCE opcode is 0x47
+    assert "47" not in byzantium_bytecode
+    assert "47" in istanbul_bytecode
--- a/vyper/opcodes.py
+++ b/vyper/opcodes.py
@@ -4,7 +4,7 @@ from typing import (
     Optional,
 )
 
-active_evm_version: int = None
+active_evm_version = None
 
 EVM_VERSIONS = {
     'byzantium': 0,
@@ -188,9 +188,18 @@ PSEUDO_OPCODES: Dict[str, List[Optional[int]]] = {
 COMB_OPCODES: Dict[str, List[Optional[int]]] = {**OPCODES, **PSEUDO_OPCODES}
 
 
-def set_evm_version(version_name):
-    global active_evm_version
-    active_evm_version = EVM_VERSIONS[version_name]
+def evm_wrapper(fn, *args, **kwargs):
+
+    def _wrapper(*args, **kwargs):
+        global active_evm_version
+        evm_version = kwargs.pop('evm_version', 'byzantium')
+        active_evm_version = EVM_VERSIONS[evm_version]
+        try:
+            return fn(*args, **kwargs)
+        finally:
+            active_evm_version = None
+
+    return _wrapper
 
 
 def _gas(opcode_name):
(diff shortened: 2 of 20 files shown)


iamdefinitelyahuman:
Things still needed:

- [ ] mypy / linting
- add test cases 
  - [ ] selecting different EVM versions in `vyper.compiler` and `vyper.cli`
  - [ ] gas prices with different rulesets
- [ ] update documentation

fubuloubu:
24f4071 adds missing check mentioned in #1652

vyper/parser/expr.py
```diff
@@ -382,6 +393,11 @@ def attribute(self):
             elif key == "tx.origin":
                 return LLLnode.from_list(['origin'], typ='address', pos=getpos(self.expr))
             elif key == "chain.id":
+                if get_active_evm_id() < 2:
+                    raise EvmVersionException(
+                        "chain.id is unavailable prior to istanbul ruleset",
+                        self.expr
```
fubuloubu: Like to have either `2` be replaced with `EVM_VERSIONS['istanbul']`, or the raw name replaced with a lookup (prefer the former)

Or even better, a helper that generates the exception based on a ruleset name

iamdefinitelyahuman: What about a method `is_min_evm_version(evm_version)` that returns a boolean?

iamdefinitelyahuman: It might also be cool to raise the exception here: https://github.com/vyperlang/vyper/blob/2448287461a4334b9224483f61eb3c6882dd1705/vyper/compile_lll.py#L455

It could serve as a catch-all. Then we don't need to worry about checking the validity of each specific opcode.  But I haven't checked if this might cause unintended issues elsewhere.

fubuloubu: I was picturing eventually it would live fully in the backend

fubuloubu:
I was picturing something more like `if version_check(begin='constaninople', end='istanbul'): ...` where it would check if the current version is within `begin...end` (inclusive), and either `begin` or `end` is optional (but not both)

jacqueswww:
Can we merge https://github.com/vyperlang/vyper/pull/1783/files before this one?

iamdefinitelyahuman:
@jacqueswww fine by me...  will rebase this after #1739 merges

tests/cli/vyper_compile/test_compile_files.py
```diff
@@ -20,3 +20,27 @@ def test_combined_json_keys(tmp_path):
...
+    compile_data = compile_files(
+        [bar_path],
+        output_formats=['bytecode_runtime'],
+        evm_version="byzantium"
+    )
+    assert compile_data != compile_files(
+        [bar_path],
+        output_formats=['bytecode_runtime'],
+        evm_version="istanbul"
+    )
```
fubuloubu: Should check that they do the same thing (and maybe that the gas cost is less for the optimization)

vyper/opcodes.py
```diff
@@ -1,173 +1,245 @@
...
+    'istanbul': 2,
 }
+DEFAULT_EVM_VERSION: str = "istanbul"
+
 
+# opcode as hex value
+# number of values removed from stack
+# number of values added to stack
+# gas cost (byzantium, constantinople, istanbul)
+OPCODES: Dict[str, List] = {
```
fubuloubu: Have to refactor this the way Jacques had it

fubuloubu:
G2G once those changes are addressed. I like how the version check works, very clean.

vyper/opcodes.py
iamdefinitelyahuman: Added mypy typing throughout `opcodes.py`

tests/cli/vyper_compile/test_compile_files.py
iamdefinitelyahuman: I expanded the test.  The cli functionality is merely a wrapper around `vyper.compile_codes`, and we have other testing of `SELFBALANCE` behavior accessed through that method.

vyper/opcodes.py
```diff
@@ -169,5 +197,67 @@
...
+    return _evm_opcodes[active_evm_version]
+
+
+def get_comb_opcodes() -> OpcodeRulesetMap:
+    return _evm_combined[active_evm_version]
+
 
-COMB_OPCODES: Dict[str, Tuple[Union[int, None], int, int, int]] = {**OPCODES, **PSEUDO_OPCODES}
+def version_check(begin: Optional[str] = None, end: Optional[str] = None) -> bool:
+    assert begin is not None or end is not None
```
fubuloubu: Should throw a usage exception of some sort

fubuloubu:
```suggestion
    if begin is None and end is None:
        raise CompilerPanic("Either beginning or end fork ruleset must be set.")
```

PR #1784, opened 2019-12-27
Allow relative imports beyond one parent level

### LINKED ISSUE #1703: Relative imports are broken beyond one parent level
The following code produces paths like `..../something.vy` instead of `../../../something.vy`:

https://github.com/ethereum/vyper/blob/c296b2d7532d913103aad494b749f8179a3acddc/vyper/signatures/interface.py#L225-L226

A path like `..../something.vy` is actually valid, so this can end up importing code from a surprising place. The expected behavior is [undocumented](https://vyper.readthedocs.io/en/latest/structure-of-a-contract.html#imports-via-import), so perhaps the intention was just to disallow more than two leading dots.

iamdefinitelyahuman:
### What I did
Allow relative imports beyond one parent level - fixes #1703 

### How I did it
* `vyper/signatures/interface.py` - properly handle `level > 2`
* `vyper/cli` - adjust checks for imports out of base folder

### How to verify it
Run the tests. I added / modified a few test cases to make sure it works.

### Cute Animal Picture
image


fubuloubu:
Just rebase and g2g

PR #1790, opened 2019-12-28
Minmax signing

### LINKED ISSUE #1690: Min and max can incorrectly use signed comparisons depending on argument order
```python
@public
@constant
def test1() -> uint256:
    return min(0, 2**255) # returns 2**255

@public
@constant
def test2() -> uint256:
    return min(2**255, 0) # returns 0
```

This is due to the following code:

https://github.com/ethereum/vyper/blob/c296b2d7532d913103aad494b749f8179a3acddc/vyper/functions/functions.py#L1174-L1177

iamdefinitelyahuman:
### What I did
Fix an error in `min` and `max` that caused incorrect evaluation with signed comparators - closes #1690 

### How I did it
Simplify / adjust logic in `vyper/functions/functions.py` - see code, it's fairly straightforward.

### How to verify it
Run the tests.  I added a new test case to check the example given in #1690.

### Cute Animal Picture
image

### DIFF
--- a/tests/parser/functions/test_minmax.py
+++ b/tests/parser/functions/test_minmax.py
@@ -185,3 +185,55 @@ def foo() -> uint256:
         lambda: get_contract_with_gas_estimation(code_2),
         TypeMismatchException
     )
+
+
+def test_unsigned(get_contract_with_gas_estimation):
+    code = """
+@public
+def foo1() -> uint256:
+    return min(0, 2**255)
+
+@public
+def foo2() -> uint256:
+    return min(2**255, 0)
+
+@public
+def foo3() -> uint256:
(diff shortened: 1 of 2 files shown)


tests/parser/functions/test_minmax.py
```diff
@@ -185,3 +185,29 @@ def foo() -> uint256:
...
+def foo2() -> uint256:
+    return min(2**255, 0)
+
+@public
+def foo3() -> uint256:
+    return max(0, 2**255)
+
+@public
+def foo4() -> uint256:
+    return max(2**255, 0)
```
fubuloubu: Can we do `2**127` and `-2**127` in both positions? Something that exercises the signed logic

PR #1778, opened 2019-12-21
Updates to vyper-json

iamdefinitelyahuman:
### What I did
* In the input json, allow `outputSelection` to be given as a subfield within `settings` - this is how Solidity formats it.
* Allow `istanbul` as a choice for the EVM version to target.  Though it doesn't change anything in how the compiler runs (yet), the code will still deploy and so the setting is valid.

### How I did it
Many hours of laborious coding.

### How to verify it
Run the tests.

### Cute Animal Picture
image

PR #1795, opened 2020-01-02
vyper-json - expect outputSelection within settings

iamdefinitelyahuman:
### What I did
* In the input json, allow `outputSelection` to be given as a subfield within `settings` - this is how Solidity formats it.

This is a cherry-pick from #1778 now that #1783 has merged.

### Cute Animal Picture
image

### DIFF
--- a/vyper/cli/vyper_json.py
+++ b/vyper/cli/vyper_json.py
@@ -223,7 +223,7 @@ def get_input_dict_interfaces(input_dict: Dict) -> Dict:
 
 def get_input_dict_output_formats(input_dict: Dict, contract_sources: ContractCodes) -> Dict:
     output_formats = {}
-    for path, outputs in input_dict['outputSelection'].items():
+    for path, outputs in input_dict['settings']['outputSelection'].items():
         if isinstance(outputs, dict):
             # if outputs are given in solc json format, collapse them into a single list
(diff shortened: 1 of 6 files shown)


vyper/cli/vyper_json.py
```diff
@@ -223,7 +223,11 @@ def get_input_dict_interfaces(input_dict: Dict) -> Dict:
 
 def get_input_dict_output_formats(input_dict: Dict, contract_sources: ContractCodes) -> Dict:
     output_formats = {}
-    for path, outputs in input_dict['outputSelection'].items():
+    try:
+        output_selection = input_dict['settings']['outputSelection']
+    except KeyError:
+        output_selection = input_dict['outputSelection']
```
fubuloubu: We're still in Beta, so I am fine with making this a breaking change instead of supporting older behavior.

In the future, would probably add a [`DeprecationWarning`](https://docs.python.org/3/library/exceptions.html#DeprecationWarning) instead.

fubuloubu: @jacqueswww thoughts?

jacqueswww: Should be fine to just make the change now.

jacqueswww: Sure; change w/o warning.

PR #1800, opened 2020-01-03
Include private methods when checking function selector collisions

### LINKED ISSUE #1687: Private functions can't have duplicate function selectors

iamdefinitelyahuman:
### What I did
Include private methods when checking for function selector collisions - closes #1687 

### How I did it
`vyper/signatures/interface.py` - do not exclude functions marked as private

### How to verify it
Run tests. I added test cases.

### Cute Animal Picture
image

### DIFF
--- a/vyper/signatures/interface.py
+++ b/vyper/signatures/interface.py
@@ -275,11 +275,10 @@ def find_signature_conflicts(sigs: Sequence[FunctionSignature]) -> Conflicts:
 
 
 def check_valid_contract_interface(global_ctx, contract_sigs):
-    public_func_sigs = [
-        sig for sig in contract_sigs.values()
-        if isinstance(sig, FunctionSignature) and not sig.private
-    ]
-    func_conflicts = find_signature_conflicts(public_func_sigs)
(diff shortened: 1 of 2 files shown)


vyper/signatures/interface.py
```diff
@@ -275,11 +275,8 @@ def find_signature_conflicts(sigs: Sequence[FunctionSignature]) -> Conflicts:
 
 
 def check_valid_contract_interface(global_ctx, contract_sigs):
-    public_func_sigs = [
-        sig for sig in contract_sigs.values()
-        if isinstance(sig, FunctionSignature) and not sig.private
-    ]
-    func_conflicts = find_signature_conflicts(public_func_sigs)
+    func_sigs = [sig for sig in contract_sigs.values() if isinstance(sig, FunctionSignature)]
```
fubuloubu: Add comments here as to why this collision check is done.

PR #1803, opened 2020-01-05
fix AttributeError when checking balance on non-address types

### LINKED ISSUE #1802: `AttributeError: 'Attribute' object has no attribute 'id'` on constants
### Version Information

* vyper Version (output of `vyper --version`): 0.1.0b14+commit.4ac5ef5
 OS: osx
* Python Version (output of `python --version`): 3.7.5
* Environment (output of `pip freeze`): same as previous TSTL issues

### What's your issue about?

Compiling:

```
cv0: constant(address) = block.coinbase.balance
```

produces:

```
Error compiling: vfile.67808.vy
```

iamdefinitelyahuman:
### What I did
Fix #1802 

### How I did it
Use `getattr` when checking for `self`

### How to verify it
Run the tests.

### Cute Animal Picture
image

### DIFF
--- a/vyper/parser/expr.py
+++ b/vyper/parser/expr.py
@@ -292,21 +292,22 @@ class Expr(object):
     def attribute(self):
         # x.balance: balance of address x
         if self.expr.attr == 'balance':
-            if getattr(self.expr.value, 'id', None) == "self" and version_check(begin="istanbul"):
-                return LLLnode.from_list(
-                    ['selfbalance'],
-                    typ=BaseType('uint256', {'wei': 1}),
-                    location=None,
-                    pos=getpos(self.expr),
-                )
             addr = Expr.parse_value_expr(self.expr.value, self.context)
             if not is_base_type(addr.typ, 'address'):
                 raise TypeMismatchException(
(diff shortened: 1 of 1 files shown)


vyper/parser/expr.py
```diff
@@ -292,21 +292,22 @@ def variables(self):
...
-                )
             addr = Expr.parse_value_expr(self.expr.value, self.context)
             if not is_base_type(addr.typ, 'address'):
                 raise TypeMismatchException(
                     "Type mismatch: balance keyword expects an address as input",
                     self.expr
                 )
+            if (
+                isinstance(self.expr.value, ast.Name) and
+                self.expr.value.id == "self" and
```
fubuloubu: Wouldn't `addr` be able to be used directly here now?

iamdefinitelyahuman: I don't know how to get from the parsed expr to `self`.  Not saying it's not possible, but I can't figure out how :grimacing:

fubuloubu: Eh, I'm just being anal

PR #1810, opened 2020-01-06
Factory example

fubuloubu:
### What I did
Added an example showing how to use `get_extcodehash` to create a registrar pattern where a child contract is deployed and registered separately from the parent, but the relationship is safely enforced.

### Cute Animal Picture
lazy tiger

### DIFF
--- a/examples/factory/Factory.vy
+++ b/examples/factory/Factory.vy
@@ -3,9 +3,10 @@ contract Exchange:
     def receive(_from: address, _amt: uint256): modifying
     def transfer(_to: address, _amt: uint256): modifying
 
+
 exchange_codehash: public(bytes32)
 # Maps token addresses to exchange addresses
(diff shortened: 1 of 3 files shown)


examples/factory/Factory.vy
```diff
@@ -0,0 +1,41 @@
...
+#       For example, allowing the deployer of this contract to change this
+#       value allows them to use a new contract if the old one has an issue.
+#       This would trigger a cascade effect across all exchanges that would
+#       need to be handled appropiately.
+
+
+@public
+def register():
+    # Verify code hash is the exchange's code hash
+    assert get_extcodehash(msg.sender) == self.exchange_codehash
```
fubuloubu:
```suggestion
    assert msg.sender.codehash == self.exchange_codehash
```

PR #1819, opened 2020-01-07
Add Snapcraft Release process

fubuloubu:
### What I did
Attempt at adding the Snapcraft release process to our CI

### How I did it
Copy the Makefile

### How to verify it
I have no idea


fubuloubu:
Note: The token added expires 1 year from today

PR #1827, opened 2020-01-09
Update grammar

fubuloubu:
### What I did
Simplified the grammar. Removed a lot of the unnecessary statements. Tried to make the resultant "parse tree" something that would be very usable inside Vyper (and directly convertible to a simplified version of our AST).

### Cute Animal Picture
penguin


fubuloubu:
This PR should be good as is, but more work could be done to simplify the grammar even more. Had not gotten much into the body statements yet.

fubuloubu:
Note: Some of the failures can be fixed by ignoring Lark's errors when running tests that are supposed to fail

jacqueswww:
Looks good to me :-) not sure why the tests are failing?

tests/grammar/test_dict_grammar.py
```diff
@@ -1,48 +0,0 @@
-import pytest
-from pytest import (
-    raises,
-)
-
-from lark.exceptions import (
```
jacqueswww: @fubuloubu  why remove these tests ?

fubuloubu:
@jacqueswww the failures have to do with some edge cases of units

tests/grammar/test_dict_grammar.py
fubuloubu: They no longer compile without more boilerplate because of how I changed the grammar. I could add the boilerplate to get them to compile, but I think some of them will be illegal syntax now. I stripped a lot out of the python setup

fubuloubu:
Rebasing on #1896 will settle failures

fubuloubu:
@iamdefinitelyahuman @charles-cooper ready for review

if you want to see it in action, import `LARK_GRAMMAR` from `tests/grammar/conftest.py` and then you can do `tree = LARK_GRAMMAR.parse(src_code)` to get the parse tree, and `print(tree.pretty())` to see it pretty printed.

For example:
```python
@public 
def foo(): 
    a: uint256 = 1 
```

Parse Tree (pretty-printed):
```
module
  function_def
    decorators
      decorator public
    function_sig        foo
    body
      declaration
        variable
          a
          type  uint256
        1
```

PR #1842, opened 2020-02-03
ABI interface fixes

### LINKED ISSUE #1832: Cannot generate AST from ABI containing string or bytes
### Version Information
* vyper Version: 0.1.0b16
* OS: linux
* Python Version: 3.7

### What's your issue about?
When importing a `json` file as an interface, if any inputs or outputs use type `string` or `bytes`, there is a `TypeError` raised:

```python
  File "vyper/signatures/interface.py", line 72, in abi_type_to_ast
    slice=ast.Index(256)
TypeError: __init__() takes 1 positional argument but 2 were given
```

It seems `ast.Index` is incorrect being initiated for these types:

https://github.com/vyperlang/vyper/blob/d4c53204ce4bb3a314443f6d6deadbb2b2111cdd/vyper/signatures/interface.py#L59-L73

### LINKED ISSUE #1841: ABI issue with bytes and string arrays inside tuples
### Version Information

* vyper Version (output of `vyper --version`): `0.1.0b16`
* OS: linux
* Python: `3.8`

### What's your issue about?

Returning a tuple that contains a bytes or string array that came from another contract call gives incorrect return data:

```python
def test_bytes_inside_tuple(get_contract):
    code = """
@public
@constant
def test_return() -> (bytes[6], uint256):
    return b"potato", 42
    """
    c1 = get_contract(code)

    code = """
contract jsonabi:
    def test_return() -> (bytes[6], uint256): constant

@public
def test_values(a: address) -> (bytes[6], uint256):
    """

    c2 = get_contract(code)
    assert c2.test_values(c1.address) == [b"potato", 42]
```

The above test fails with:
```
AssertionError: assert [b'\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00*\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x06', 50855962295469179162425751602892202823040865300719823209625149558373627199488] == [b'potato', 42]
```

Some observations:

1. The returned bytes array is always exactly 64 bytes long. I've tried a range of values between `bytes[6]` and `bytes[256]`.
2. The location of the bytes array within the tuple does not seem to have an effect.
3. The test passes if you assign the individual tuple values to memory variables and then return an array of those variables.
4. The same behavior exists with `string`.
5. The error is not present when using a `uint[]` array.
6. The error is not present when calling from a public function into a private function in the same contract.

iamdefinitelyahuman:
### What I did
Fix issues with `bytes`, `string` and `fixed168x10` when importing an ABI as an interface. Fixes #1832  

### How I did it
1. When determining if an interface has been implemented, ignore the length on `bytes` and `string`.
2. Substitute `fixed168x10` for `decimal` when converting ABI to AST.
3. For `bytes` and `string` arrays, use a length of `1024*1024` in inputs, `1` in outputs. The compiler does not implicitly truncate values so this does not create an issue when a limit is exceeded.

### How to verify it
Run the tests. I expanded the cases around ABIs as interfaces to check each type as an input and output.

### Cute Animal Picture
image

### DIFF
--- a/tests/parser/functions/test_interfaces.py
+++ b/tests/parser/functions/test_interfaces.py
@@ -1,4 +1,6 @@
-from decimal import Decimal
+from decimal import (
+    Decimal,
+)
 
 import pytest
 
@@ -453,13 +455,13 @@ def test_json_interface_calls(get_contract, type_str, value):
     abi = compile_code(code, ['abi'])['abi']
     c1 = get_contract(code)
 
-    code = """
+    code = f"""
 import jsonabi as jsonabi
 
 @public
 @constant
-def test_call(a: address, b: {0}) -> {0}:
+def test_call(a: address, b: {type_str}) -> {type_str}:
     return jsonabi(a).test_json(b)
-    """.format(type_str)
+    """
     c2 = get_contract(code, interface_codes={'jsonabi': {'type': 'json', 'code': abi}})
     assert c2.test_call(c1.address, value) == value
--- a/tests/parser/functions/test_interfaces.py
+++ b/tests/parser/functions/test_interfaces.py
@@ -1,3 +1,5 @@
+from decimal import Decimal
+
 import pytest
 
 from vyper.compiler import (
@@ -277,60 +279,6 @@ def test():
     assert erc20.balanceOf(sender) == 1000
 
 
-def test_json_interface(get_contract):
-    code = """
-import folding as Folding
-
-implements: Folding
-
-@public
-def test(a: uint256) -> uint256:
-    return 1 + a
-
-
-@public
-def test2(a: uint256):
-    pass
-    """
-
-    interface_codes = {
-        'Folding': {
-            'type': 'json',
-            'code': [
-                {
-                    "name": "test",
-                    "outputs": [{
-                        "type": "uint256",
-                        "name": "out"
-                    }],
-                    "inputs": [{
-                        "type": "uint256",
-                        "name": "s"
-                    }],
-                    "constant": False,
-                    "payable": False,
-                    "type": "function",
-                },
-                {
-                    "name": "test2",
-                    "outputs": [],
-                    "inputs": [{
-                        "type": "uint256",
-                        "name": "s"
-                    }],
-                    "constant": False,
-                    "payable": False,
-                    "type": "function",
-                }
-            ]
-        }
-    }
-
-    c = get_contract(code, interface_codes=interface_codes)
-
-    assert c.test(2) == 3
-
-
 def test_units_interface(w3, get_contract):
     code = """
 import balanceof as BalanceOf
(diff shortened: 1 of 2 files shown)


michwill:
Excellent, does it also fix the same issue with structs which contain `uint256` arrays?

michwill:
E.g. this:
```python
Z256: constant(uint256) = 0 

struct A:
    many: uint256[3]
    one: uint256

@private
@constant
def foo(_many: uint256[3], _one: uint256) -> A:
    a: A = A({many: [Z256, Z256, Z256], one: 0})
    a.many = _many
    a.one = _one
    return a

@public
def bar():
    m: uint256[3] = [Z256, Z256, Z256]
    m[0] = 42
    m[1] = 2020
    m[2] = 21000000
    out: A = foo(m, 33)
    assert out.many[0] == m[0]
    assert out.many[1] == m[1]
    assert out.many[2] == m[2]
    assert out.one == 33
```

iamdefinitelyahuman:
@michwill good call, I'll have a look at that as well

michwill:
As far as I can tell, this PR doesn't fix it for structs, and probably the solution would be suitable for this PR, too

vyper/signatures/interface.py
```diff
@@ -85,18 +85,18 @@ def mk_full_signature_from_json(abi):
         for a in func['inputs']:
             arg = ast.arg(
                 arg=a['name'],
-                annotation=abi_type_to_ast(a['type']),
+                annotation=abi_type_to_ast(a['type'], 1048576),
```
fubuloubu: Add a comment that this supports a maximum-sized `string` or `bytes` dynamic structure of 1 Mb (1024 * 1024 * 1 byte)

tests/parser/functions/test_interfaces.py
```diff
@@ -468,3 +418,50 @@ def bar(a: address) -> uint256:
...
     c = get_contract(code)
     assert_tx_failed(lambda: c.bar(c.address))
+
+
+type_str_params = [
+    ('int128', -33),
+    ('uint256', 42),
+    ('bool', True),
+    ('address', "0x1234567890123456789012345678901234567890"),
+    ('bytes32', b"bytes32bytes32bytes32bytes32poop"),
```
fubuloubu: Freudian slip

vyper/signatures/interface.py
```diff
@@ -56,20 +60,16 @@ def render_return(sig):
     return ""
 
 
-def abi_type_to_ast(atype):
+def abi_type_to_ast(atype, idx):
```
fubuloubu: Can `idx` be given a better name? Like maybe `expected_size`

vyper/signatures/interface.py
```diff
@@ -85,18 +85,18 @@ def mk_full_signature_from_json(abi):
         for a in func['inputs']:
             arg = ast.arg(
```
fubuloubu: Add a note describing the issue that you found with ABI inputs that use `bytes` or `strings` and how the addition of an expected solves it. I have a feeling this brittleness will become an issue in future iterations, and it will be good to have an explanation to this effect somewhere in the code.

vyper/signatures/interface.py
```diff
@@ -85,18 +85,18 @@ def mk_full_signature_from_json(abi):
...
                 arg=a['name'],
-                annotation=abi_type_to_ast(a['type']),
+                annotation=abi_type_to_ast(a['type'], 1048576),
                 lineno=0,
                 col_offset=0
             )
             args.append(arg)
 
         if len(func['outputs']) == 1:
```
fubuloubu: Add a note similar to the above about `inputs`

vyper/signatures/interface.py
```diff
@@ -294,16 +294,14 @@ def check_valid_contract_interface(global_ctx, contract_sigs):
 
         for sig, func_sig in contract_sigs.items():
             if isinstance(func_sig, FunctionSignature):
-                # Remove units, as inteface signatures should not enforce units.
+                if sig not in funcs_left or func_sig.private:
+                    continue
```
fubuloubu: Can these two be separate `if` statements, with comments explaining why we ignore them?

iamdefinitelyahuman:
note to self, need to open another issue re: structs in json interfaces

PR #1854, opened 2020-02-11
AST refactor step 1

iamdefinitelyahuman:
### What I did
Created `vyper/ast` subfolder and moved several existing modules into it. This is a first step toward #1806. Opening a PR now since this touches a lot of files but no logic was modified, and I want to minimize the size of the diff when I actually start _doing stuff_.

### How I did it
* `vyper/ast.py` -> `vyper/ast/nodes.py` - this change means we avoid modules where the stdlib `ast` is imported as `python_ast` and the vyper package as `vyper_ast`.
* `vyper/ast_utils.py` -> `vyper/ast/utils.py`
* `vyper/parser/pre_parser.py` -> `vyper/ast/pre_parser.py` - as far as I can tell this is the only file within `vyper/parser` that involves parsing prior to AST generation, the rest is parsing to LLL.

### How to verify it
Run the tests. I modified imports in the tests as well, everything is passing locally.

### Cute Animal Picture
image


vyper/ast/__init__.py
```diff
@@ -0,0 +1,66 @@
+from vyper.ast.nodes import (  # noqa: F401
```
fubuloubu: Is there any reason we can't just batch import, and move any extraneous functionality we don't want as part of the API for this stage to another module?

vyper/ast/utils.py
```diff
@@ -15,13 +17,14 @@
...
-from vyper.parser.pre_parser import (
-    pre_parse,
-)
 from vyper.utils import (
     iterable_cast,
 )
 
+from . import (
+    nodes as vy_ast,
+)
```
fubuloubu: Just `import .nodes as vy_ast` directly. Add a `noqa` if flake8 is giving you problems.

vyper/ast/__init__.py
iamdefinitelyahuman: I worry that a wildcard import opens the door to unexpected namespace issues.  If we use `from vyper.ast.nodes import *` we also import things like `itertools.chain` and `vyper.utils.annotate_source_code`.

iamdefinitelyahuman:
As I started work on the logic I ran into another chunk of code I want to move, so amended this PR.

`annotate_ast` was part of `parser.parser_utils`, but it's only ever used by `ast_utils.parse_to_ast`. I moved it, and related classes, to `ast.annotation`, and renamed the function to `annotate_python_ast` so it's more obvious what it's doing.

vyper/ast/__init__.py
fubuloubu: There's gotta be a fancy way to do this important to just grab all subclasses of `VyperNode`

iamdefinitelyahuman: Yes, if you'd prefer I can definitely do some voodoo.

iamdefinitelyahuman: ben casts `sys.modules[__name__]` + `cls.__mro__`...  it's super effective!

PR #1869, opened 2020-03-02
Type checking / semantic analysis pass

iamdefinitelyahuman:
### What I'm doing
WIP for a type checking / semantic analysis pass.

This is still a work in progress, but far enough along that I feel the general structure won't change much.  I would very much appreciate any feedback on it.

### How I'm doing it
All the work is in `vyper/context`:
1. `context/namespace.py` contains the `Namespace` object, a `dict` subclass which holds all type and definition objects used during the pass.
1. `context/types` contains classes that represent the core Vyper types. The range of functionality available for each type is determined according to the attributes and methods that it contains. Docstrings in `context/types/bases.py` outline a blueprint for how types are defined.
2. `context/definitions` contains classes that represent definitions - the assignment of a type (or types) to a name. Definitions hold additional information about the assignment, and provide methods for interaction with the underlying type.
3. `context/validation` contains methods for generating the builtin namespace, and validating the module and function scoped namespaces.

### Todo
Work to be done within the scope of this PR. This will evolve over time.

#### Functionality
- [x] json abi imports
- [ ] block the use of `msg.sender` and `msg.value` in private methods
- [x] `in` comparator logic
- [x] `convert` builtin function
- [ ] reserved keywords
- [ ] constant folding on binary ops and builtin functions prior to type checking

#### Exceptions
Currently I'm mostly raising `StructureException` everywhere. As part of this refactor I would like to outline exactly what every custom exception means, document it, and enforce proper use for this and all subsequent refactors. I think a good long-term policy is that the compiler should never raise built-in python exceptions - if one is raised, it means something went wrong.

### Next Steps
Once this is good to go I see several next steps:

1. Integration into the existing code base. This will basically consist of plugging it in, fixing/rewriting broken tests, and removing now-obsolete logic in later stages of the code.
2. Further refactoring. With obsoleted code removed and additional type information available, we should be able to develop a plan for refactoring further stages of the code base. At present I haven't added any logic around annotating the AST with type information - I think it makes sense to do it once we better know exactly what information is useful and how it can be best delivered.
3. Documentation updates. Because this pass touches _all_ the syntax, I'm in a good position to do a thorough overhaul of the docs and explicitly outline everything about the language that's currently under-explained (or not covered at all).

### Cute Animal Picture
TBD - I will find the cutest animal picture once this PR is no longer a draft.


iamdefinitelyahuman:
Closing this for now, it needs a massive rebase.  It will rise again stronger and prettier.

PR #1870, opened 2020-03-03
Expand and document custom exceptions

iamdefinitelyahuman:
### What I did
1. Add docstrings for all custom exceptions.
2. Expand some of the more commonly used exceptions into different types that are more specific to what is happening.
3. Rename `ParserException` to `VyperException` and explicitly declare it to be an inherited base exception.
4. Update documentation and tests.

### Motivation
Related to #1869 - I think it's good to have all custom exception types very clear prior to major refactoring. I also think we should have a policy that the compiler never intentionally raises a builtin exception. An end-user receiving a builtin exception should be a clear sign of a bug within the compiler.

I don't intend to change any raise statements throughout the existing code. But I __do__ want to enforce their use in any code we add going forward, specifically in the refactors on the near horizon.

### Cute Animal Picture
image

### DIFF
--- a/vyper/exceptions.py
+++ b/vyper/exceptions.py
@@ -62,15 +62,15 @@ class SyntaxException(ParserException):
 
 
 class StructureException(ParserException):
-    """TODO"""
+    """Invalid structure for parsable syntax."""
+
+
+class VersionException(ParserException):
+    """Version string is malformed or incompatible with this compiler version."""
 
 
 class InvalidLiteralException(ParserException):
     """Invalid literal value."""
-
-
-class InvalidTypeException(ParserException):
-    """Invalid type declaration."""
 
 
 class VariableDeclarationException(ParserException):
@@ -85,8 +85,20 @@ class EventDeclarationException(ParserException):
     """Invalid event declaration."""
 
 
+class UndeclaredDefinition(ParserException):
+    """Reference to a definition that has not been declared."""
+
+
+class NamespaceCollsion(ParserException):
+    """Assignment to a name that is already in use."""
+
+
+class InvalidTypeException(ParserException):
+    """Type is invalid for an action."""
+
+
 class TypeMismatchException(ParserException):
-    """TODO"""
+    """Attempt to perform an action with multiple, incompatible types."""
 
 
 class ConstancyViolationException(ParserException):
@@ -94,11 +106,7 @@ class ConstancyViolationException(ParserException):
 
 
 class NonPayableViolationException(ParserException):
-    """Used msg.value in a nonpayable function."""
-
-
-class VersionException(ParserException):
-    """Version string is malform or incompatible with this version of Vyper."""
+    """msg.value in a nonpayable function."""
 
 
 class ArrayIndexException(ParserException):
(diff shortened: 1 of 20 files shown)


vyper/exceptions.py
```diff
@@ -51,74 +51,111 @@ def __str__(self):
...
+    # type cannot be packed
 
 
-class SyntaxException(ParserException):
-    pass
+class ConstancyViolationException(ParserException):
+    """State-changing action inside a constant function."""
+
+
+class NonPayableViolationException(ParserException):
```
iamdefinitelyahuman: `NonPayableViolationException` is only raised in one place in the entire codebase.

vyper/exceptions.py
```diff
@@ -51,74 +51,117 @@ def __str__(self):
...
+    """Used msg.value in a nonpayable function."""
+
+
+class VersionException(ParserException):
+    """Version string is malform or incompatible with this version of Vyper."""
 
 
 class ArrayIndexException(ParserException):
```
iamdefinitelyahuman: `ArrayIndexException` is also only raised from one place.

vyper/exceptions.py
```diff
@@ -51,74 +51,117 @@ def __str__(self):
...
         item.col_offset = syntax_error.offset
-        item.source_code = source_code
+        item.full_source_code = source_code
         super().__init__(message=f'SyntaxError: {syntax_error.msg}', item=item)
 
 
-class VariableDeclarationException(ParserException):
-    pass
+class SyntaxException(ParserException):
+    """Valid Python syntax, but invalid Vyper syntax."""
```
iamdefinitelyahuman: Does it make sense to have two types of syntax error? We might consider merging `PythonSyntaxException` with `SyntaxException`.

vyper/exceptions.py
```diff
@@ -51,74 +51,117 @@ def __str__(self):
...
+    # decimal exponentiation
+    # no type given
 
 
-class TypeMismatchException(ParserException):
-    pass
+class VariableDeclarationException(ParserException):
+    """Invalid variable declaration."""
+    # invalid name
+    # already declared / shadows existing value
```
iamdefinitelyahuman: already declared / undeclared are common themes across several exception types and so good candidates for new exceptions.

for shadowing / already declared vars: `NamespaceCollsion`
for references to things that don't exist: `UndeclaredDefinition`

vyper/exceptions.py
```diff
@@ -51,74 +51,117 @@ def __str__(self):
...
+    # unsupported statement type
+    # altering list during iteration
+    # invalid use of range()
 
 
 class InvalidLiteralException(ParserException):
-    pass
+    """Invalid literal value."""
+    # number out of range
+    # too many decimals
```
iamdefinitelyahuman: For numerics that are out of bounds or decimals with too many places.. `OverflowException`?

vyper/exceptions.py
fubuloubu: I agree with this. You can modify the wording of a Python syntax exception so that it's just a normal syntax exception (using raise from)

vyper/exceptions.py
fubuloubu: I'm good with this

vyper/exceptions.py
fubuloubu: That is much clearer

vyper/exceptions.py
fubuloubu: Seems a little non-descript to me

vyper/exceptions.py
fubuloubu: This one seems fine, if you exceed the index of a static array it should throw a unique exception

iamdefinitelyahuman:
Likely there will be more specialization of the exceptions as we move forward with refactoring, but I feel this level is Good Enough to get started.

With the docstrings at least it's very clear where each exception is meant to be used.

PR #1872, opened 2020-03-05
Removed RLP List from functions, tests & docs

realjohnward:
### What I did
Removed RLP List from functions, tests & docs

### How I did it
Deleted all snippets of code referring to RLP Lists EXCEPT for the code inside vyper/vyper/parser/stmt.py

### Cute Animal Picture


fubuloubu:
@realjohnward great work! At this point, we're unsure of the impact this will have for our users, but we at least wanted to implement #1406 into the same release where this was removed so those users have a better alternative they can use. @charles-cooper is working on that one, so it may be a while before we can merge this.

jacqueswww:
@fubuloubu  I believe scripts/* referring to rlp_decoder can also be removed?

fubuloubu:
@realjohnward rebase and we'll merge it!

Issue #1863, opened 2020-02-20
Add Natspec in combined_json


montyly:
Hi,

Would it be possible to add the natspec documentation to the combined json output?

Typically on the following:
```python
@public
def test(a: uint256) -> bool:
    """
    @author montyly
    @notice Nothing
    @dev This is a test
    @param a just a param
    @return always true
    """
    return True
```
It would be great if `vyper test.vy -f combined_json` could return the natspec to the json. Solidity output it as two dicts: https://solidity.readthedocs.io/en/latest/natspec-format.html#documentation-output.

This will help third-parties tools integration (such as https://github.com/crytic/crytic-compile/pull/66).

Thanks

fubuloubu:
Yes! Should be possible. I don't think we actually parse the natspec currently, but it should be easy for someone to build a parser for the docstrings of a method and output the results alongside the other output formats.

montyly:
# Task Description
- Write a parser for natspec
- Parse docstrings comments inside methods as natspec
- Add natspec comments to output from compiler

#  Acceptance criteria
- Extensive unit tests for parser
- Unit tests for vyper-cli and vyper-json
- Demonstrate compatibility with Solidity

# Required skills
- Python 3.6-3.8
- Familiarity with natspec
- Familiarity with Solidity

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 200.0 DAI (200.0 USD @ $1.0/DAI)  attached to it.__

 * If you would like to work on this issue you can 'start work' [on the Gitcoin Issue Details page](https://gitcoin.co/issue/vyperlang/vyper/1863/4092).
* Want to chip in? Add your own contribution [here](https://gitcoin.co/issue/vyperlang/vyper/1863/4092).
* Questions? Checkout <a href='https://gitcoin.co/help'>Gitcoin Help</a> or the <a href='https://gitcoin.co/slack'>Gitcoin Slack</a>
* $79,579.38 more funded OSS Work available on the [Gitcoin Issue Explorer](https://gitcoin.co/explorer)

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__Workers have applied to start work__.


These users each claimed they can complete the work by 1 month from now.
Please review their action plans below:


[...]

PR #1874, opened 2020-03-06
Move docstring out of body in FunctionDef AST

iamdefinitelyahuman:
### What I did
Store function docstrings in the `doc_string` field of the related AST node.

### How I did it
1. `vyper/ast/nodes.py` - When `FunctionDef.body[0]` is an `Expr` containing a `Str`, remove the `Expr` from `body` and add `Str.s` to `FunctionDef.doc_string`.
2. `vyper/parser/expr.py` - remove logic for `Str` - the only use it had was to raise if a value _wasn't_ a docstring.

### How to verify it
Run the tests. I updated a test case that was failing.

### Cute Animal Picture
image

### DIFF
--- a/tests/parser/syntax/test_invalids.py
+++ b/tests/parser/syntax/test_invalids.py
@@ -271,8 +271,9 @@ def sum(a: int128, b: int128) -> int128:
 must_fail('''
 @public
 def a():
-    "Test"
-''', InvalidLiteralException)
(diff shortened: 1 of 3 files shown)


iamdefinitelyahuman:
This is a stepping stone for #1863 - the natspec parser should fit in nicely after the AST is generated, and can work by looking at `FunctionDef.doc_string`.

PR #1886, opened 2020-03-16
Ast modifications

iamdefinitelyahuman:
### What I did
Simplify the structure of `Compare` and `Assign` nodes.

### How I did it
1. Python comparisons allow more than 2 comparators:  `x < y < z`. Because of this, `Compare` has a rather complex structure. Since Vyper only allows `x < y`, I've changed the `Compare` structure to be identical to `BinOp`.  Members are now `left`, `op`, `right`.
2. Python allows multiple targets for assignment: `a = b = 42`. Vyper does not.  To standardize `AnnAssign` with `Assign`/`AugAssign`, I've converted the `AnnAssign.targets` list member into a single `AnnAssign.target`.

Both of these changes occur in the respective `__init__` methods for the nodes and raise `SyntaxException` if syntax is valid python but not valid vyper.  This is the only user-facing change - prior to this PR these situations would raise `StructureException`.

### How to verify it
Run tests. I've updated where needed.

### Cute Animal Picture
image

### DIFF
--- a/vyper/ast/nodes.py
+++ b/vyper/ast/nodes.py
@@ -554,7 +554,16 @@ class Or(VyperNode):
 
 
 class Compare(VyperNode):
-    __slots__ = ('comparators', 'ops', 'left', 'right')
+    __slots__ = ('op', 'left', 'right')
+
+    def __init__(self, *args, **kwargs):
+        if len(kwargs['ops']) > 1 or len(kwargs['comparators']) > 1:
+            err_args = (kwargs['full_source_code'], kwargs['lineno'], kwargs['col_offset'])
+            raise SyntaxException("Cannot have a comparison with more than two elements", err_args)
+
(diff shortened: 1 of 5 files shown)


vyper/ast/nodes.py
```diff
@@ -112,6 +113,12 @@ def sortkey(key):
     )
 
 
+def _raise_syntax_exc(msg, ast_struct):
+    raise SyntaxException(
```
fubuloubu: Should check if fields are in `ast_struct` first, raising just `msg` if it's missing, else the full message

iamdefinitelyahuman:
Seems the docs build is failing because vyper.online is down, otherwise gtg

iamdefinitelyahuman:
Once #1889 merges this should be gtg as well.

PR #1891, opened 2020-03-17
Modifications to raw_call arguments

### LINKED ISSUE #1890: VIP: make gas and outsize optional arguments in raw_call
## Simple Summary
The `raw_call` builtin function should handle both `gas` and `outsize` as optional arguments, using the following defaults:

1. `gas`: `gasLeft`
2. `outsize`: `0`

This is an extension of #1879

### `gas`
In the majority of cases, the desired effect in a `raw_call` is to forward all remaining gas. To me this feels intuitive, even expected. I don't think it should require an explicit declaration.

### `outsize`
`outsize` is necessary for type checking when the result of a `raw_call` is assigned to a variable, but in cases where the result isn't stored it feels unnecessary.  I think that we should use a default value of `0`. In this case we can internally return `None`, which raises if assigned to any variable.

## Specification
Change the syntax in the following manner:

#### Current
```python
# sending all gas, no assignment of return value
raw_call(receiver, b"foo", 0, msg.gas)

# declaring gas forwarded and outsize
bar: bytes[32] = raw_call(receiver, b"foo", 32, 50000)
```

#### Proposed
```python
# sending all gas, no assignment of return value
raw_call(receiver, b"foo")

# declaring gas forwarded and outsize
bar: bytes[32] = raw_call(receiver, b"foo", outsize=32, gas=50000)
```

The syntax for the default behavior looks much cleaner, and the advanced use is more explicit and so easier to understand.

## Backwards Compatibility
This VIP reverses the breaking change that would be caused by #1879.

## Copyright

iamdefinitelyahuman:
### What I did
1. Make `gas` and `outsize` into optional kwargs in `raw_call`.
2. When `outsize` is `0`, do not return a value.
3. When `gas` is not given, forward all available gas.

Closes #1890 

### How I did it
Modifications to `vyper/functions/functions.py` - I refactored / optimized the LLL a bit, if `outsize` is 0, the return value isn't copied to memory. I added comments, hopefully it's clear enough.

### How to verify it
Run tests. I added a few new cases, mostly around `gas`.

### Cute Animal Picture
image

### DIFF
--- a/vyper/functions/functions.py
+++ b/vyper/functions/functions.py
@@ -641,8 +641,8 @@ false_value = LLLnode.from_list(0, typ=BaseType('bool', is_literal=True))
 @signature(
     'address',
     'bytes',
-    outsize='num_literal',
-    gas='uint256',
+    outsize=Optional('num_literal', 0),
+    gas=Optional('uint256', 'gas'),
     value=Optional('uint256', zero_value),
     delegate_call=Optional('bool', false_value),
 )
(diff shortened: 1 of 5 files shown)


vyper/functions/functions.py
```diff
@@ -679,54 +679,30 @@ def raw_call(expr, args, kwargs, context):
...
+        # if there is no return value, the return offset can be 0
+        ['add', output_node, 32] if outsize else 0,
+        outsize
+    ]
+
+    # build sequence LLL
+    seq = ['seq', copier, ['assert', call_lll]]
+    if outsize:
+        # only copy the return value to memory if outsize > 0
+        seq += [['mstore', output_node, outsize], output_node]
```
fubuloubu: This whole section (686-699) is sort of ugly with all the mutations. Is there a way to build all the parts separately and combine them without mutation?

PR #1893, opened 2020-03-18
Adjust constant AST fields

iamdefinitelyahuman:
### What I did
Modified the members of `Constant` ast nodes.

1. Values are now always stored at `value`, instead of `n`/`s`. This is a change that Python introduced in version 3.8.
2. `ast.Decimal.value` is set as a `decimal.Decimal` immediately when the node is created.
3. Values for `ast.Hex`, `ast.Binary`, and `ast.Octal` are represented as strings instead of their `int` equivalent.

### How I did it
Mostly it involved modifying `ast/annotation.py` and deleting some now redundant code.

I've also added `n` and `s` `@property` members as `value` aliases, so these attributes are still available. It is the path of less resistance as opposed to changing every reference within the remaining existing code base.  As we refactor many of those references will be removed anyway.

### How to verify it
Run the tests.

### Cute Animal Picture
image

### DIFF
--- a/vyper/ast/nodes.py
+++ b/vyper/ast/nodes.py
@@ -1,5 +1,4 @@
 import ast as python_ast
-import decimal
 import sys
 from typing import (
     Optional,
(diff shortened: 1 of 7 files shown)


vyper/ast/nodes.py
```diff
@@ -405,72 +404,56 @@ class ClassDef(VyperNode):
...
 class Num(Constant):
     # inherited class for all numeric constant node types
-    __slots__ = ('n', )
-    _translated_fields = {'value': 'n'}
+    __slots__ = ()
     _python_ast_type = "Num"
+    _translated_fields = {'n': 'value'}
+
+    @property
+    def n(self):
```
fubuloubu: How annoying would it be to have these throw Warnings?

iamdefinitelyahuman: 50 places in the code or so annoying lol

fubuloubu: At least add a TODO

fubuloubu: Also for `s`

PR #1896, opened 2020-03-19
Remove units

### LINKED ISSUE #1881: VIP: Temporarily remove type units
## Simple Summary
The current implementation of type units feels inconsistent and confusing. I would like to remove them from the language in their current form.

## Motivation
While working on type checking I keep running into issues with units. All of this is made much more difficult because there is very minimal documentation on the subject.

1. `uint256` is considered "unitless", so assignments such as `uint256(wei) = uint256` are permitted. It took me a while to realize this was the case, I assumed `uint256` was equivalent to `uint256(None)` and so should not be compatible with `uint256(wei)`.
2. Addition and subtraction between values with different units is disallowed, but multiplication and division is allowed and produces new compound unit types. This is completely undocumented and does not feel intuitive to me.
3. `wei_value` and `uint256(wei)`, despite both being `uint256` with unit `wei`, are incompatible. 
4. Same for `timestamp`, `timedelta`, and `uint256(sec)`.  Per the documentation, `timestamp` additionally cannot be added or subtracted with another `timestamp`. I was unable to check with this firsthand, as attempting to assign to `timestamp` or `timedelta` gives me a `TypeMistmatch`. These feel broken.

While I'm not inherently opposed to the idea of units, the current implementation feels more confusing than beneficial. I propose we remove units altogether while refactoring. We can revisit the concept once refactoring is complete and documentation is up-to-date, and hopefully come up with another approach that's cleaner and more consistent.

## Backwards Compatibility
It's a breaking change.

## Copyright

iamdefinitelyahuman:
### What I did
Remove type units. Closes #1881 

### How I did it
1. Remove the ability for `BaseType` to accept a unit or positional argument.
2. Remove references to units throughout the code base.
3. Replace several points with sanity checks that will raise if a unit is given (in case I missed anything).
4. Modify interfaces, the name is stored as `name` instead of `unit`.
5. Update / remove test cases.
6. Update examples and docs.

### How to verify it
Run the tests.

### Cute Animal Picture
image


vyper/parser/external_call.py
```diff
@@ -180,8 +180,13 @@ def make_external_call(stmt_expr, context):
             gas=gas,
         )
 
-    elif isinstance(stmt_expr.func.value, vy_ast.Attribute) and stmt_expr.func.value.attr in context.globals:  # noqa: E501
-        contract_name = context.globals[stmt_expr.func.value.attr].typ.unit
+    elif (
+        isinstance(stmt_expr.func.value, vy_ast.Attribute) and
+        stmt_expr.func.value.attr in context.globals
+        and hasattr(context.globals[stmt_expr.func.value.attr].typ, 'name')
```
fubuloubu: What's this line for?

iamdefinitelyahuman: Currently, interfaces are being represented as an `address` with a unit, where the unit is the name of the interface.  This solves the same issue by comparing the contract names without needing to use units.

PR #1898, opened 2020-03-22
NatSpec parser

### LINKED ISSUE #1863: Add Natspec in combined_json

iamdefinitelyahuman:
### What I did
Add NatSpec parsing functionality.  Closes #1863 

### How I did it
1. Add logic for parsing natspec in docstrings under `ast/natspec.py`
2. Add [stub files](https://www.python.org/dev/peps/pep-0561/) for `ast/__init__.py` and `ast/nodes.py` to avoid typing issues
3. Add `userdoc` and `devdoc` as compiler output options and enable in `vyper-compile` and `vyper-json`.

The output format is identical to that of Solidity `>=0.6.0`:
1. Docstrings _without_ any natspec tags are handled as `@notice`.
2. Errors are raised on invalid or duplicate `@param` fields.
3. Multiple `@return` values are supported, each value is assigned a name `_[n]` with `n` incrementing from `0`.

### How to verify it
Run tests.

### TODO
- [x] Add test cases
- [x] Update documentation

### Cute Animal Picture
image


iamdefinitelyahuman:
```

PR #1900, opened 2020-03-24
only allow ascii literals in byte arrays

### LINKED ISSUE #1805: Vyper creates invalid ASCII bytes
### Version Information

* vyper version: 0.1.0b14

### What's your issue about?

The following code
```
@public
def foo() -> string[5]:
   return "ä"
```
returns a bytes array b'\xe4' which is invalid ASCII.

### How can it be fixed?

I think the problem is this section in signature.py:
```
bytez = b''
for c in arg.s:
  if ord(c) >= 256:
       raise InvalidLiteralException(
            f"Cannot insert special character {c} into byte array",
             arg,
             )
   bytez += bytes([ord(c)])
return bytez
```
which checks that a character is valid ASCII. Since ASCII only uses 7 bits the check should be `ord(c) >= 128` instead of 256.

iamdefinitelyahuman:
### What I did
Fix #1805 

### How I did it
Change the check for `>=256` to `>=128`

### How to verify it
Run tests.

### Cute Animal Picture
image

### DIFF
--- a/vyper/utils.py
+++ b/vyper/utils.py
@@ -32,8 +32,8 @@ def fourbytes_to_int(inp):
 def string_to_bytes(str):
     bytez = b''
     for c in str:
-        if ord(c) >= 256:
-            raise InvalidLiteral(f"Cannot insert special character {c} into byte array")
+        if ord(c) >= 128:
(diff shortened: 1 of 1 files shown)


iamdefinitelyahuman:
Hrmm, maybe not...

PR #1903, opened 2020-03-28
AST updates and documentation

iamdefinitelyahuman:
### What I did
* rename `get_all_children` to `get_descendants`
* remove `Octal` - it will be very easy to re-add if we decide to later
* add `compare_nodes` method and remove equality check between nodes in `vyper/parser/stmt.py`
* add/update docstrings throughout `vyper/ast`
* add a top-level `README.md` explaining the structure of the module

### How I did it
So much documentation.

### How to verify it
Run the tests. I had to modify a few to make them pass but it's fairly 1:1

### Cute Animal Picture
image


vyper/ast/README.md
```diff
@@ -0,0 +1,127 @@
...
+`vyper.ast.utils.parse_to_ast` is the main function used to generate a Vyper AST
+from a source string. The process is as follows:
+
+1. In [`pre_parser.py`](pre_parser.py), Vyper source is modified into parseable
+Python. This primarily involves substituting out `contract` and `struct` statements
+for `class`.
+2. A Python AST is generated from the modified source.
+3. In [`annotation.py`](annotation.py), additional information is added to the
+Python AST, and some nodes are modified to aid conversion to the Vyper AST.
+4. In [`nodes.py`](nodes.py), the Python AST nodes are converted to Vyper
```
fubuloubu:
```suggestion
4. In [`nodes.py`](nodes.py), the modified Python AST nodes are converted to Vyper
```

vyper/ast/README.md
```diff
@@ -0,0 +1,127 @@
...
+* Green Tree Snakes: [Meet the Nodes](https://greentreesnakes.readthedocs.io/en/latest/nodes.html)
+
+### Traversing the AST
+
+Each node contains several methods that are used to traverse the AST. These methods
+can optionally include one or more filters in order to quickly search for children
+or parents that match a desired pattern.
+
+* `get_children`: Returns direct childen of the given node.
+* `get_descendants`: Returns direct and indirect childen of the given node.
```
fubuloubu:
```suggestion
* `get_descendants`: Returns direct and indirect children (grand-children, great-grand-children, etc.) of the given node.
```

vyper/ast/README.md
```diff
@@ -0,0 +1,127 @@
...
+* `get_parent`: Returns a parent of the given node.
```
fubuloubu: "the" because there is only one
```suggestion
* `get_parent`: Returns the parent of the given node.
```

vyper/ast/README.md
```diff
@@ -0,0 +1,127 @@
...
+* `get`: Recursive getter function for node attributes.
+
+To learn more about these methods, read their docstrings in the `VyperNode` class
+in [`nodes.py](nodes.py).
```
fubuloubu:
```suggestion
in [`nodes.py`](nodes.py).
```

vyper/ast/README.md
```diff
@@ -0,0 +1,127 @@
...
+
+The type information in `node.pyi` is applied to `node.py`. Any types given in
+`node.py` are only there to aid code readability - they are ignored by the type
+checker.
+
+The following resources are useful for familiarizing yourself with stubs:
+
+* [MyPy: Stub Files](https://mypy.readthedocs.io/en/stable/stubs.html)
+* [PEP 484: Stub Files](https://www.python.org/dev/peps/pep-0484/#stub-files)
+
```
fubuloubu: Add note "You must modify both the source file and the stub when you make changes to a source file with a corresponding stub"

vyper/ast/README.md
```diff
@@ -0,0 +1,127 @@
...
+```python
+# Correct
+from vyper import ast as vy_ast
+from vyper.ast import VyperNode
+
+# Incorrect - could be mistaken for the builtin ast library
+from vyper import ast
+
+# Incorrect - only the functionality within vyper.ast is considered "public"
```
fubuloubu: We talked about whether underscoring made sense, and agreed that we wouldn't do it as long as this sort of distinction was understood, correct?

fubuloubu: Maybe add a note that it's not enforced, but highly suggested you not do this.

iamdefinitelyahuman: Ooooh good call.  I'll add an explanation of when to prepend with underscores in #1871.

vyper/ast/README.md
iamdefinitelyahuman: I fixed the description but maybe the name of the method is wrong. When a filter is given, it returns the first parent, grandparent, great-great-great-granddaddy, etc. that matches the filter.  Perhaps `get_ascendant` or `get_ancestor` would be clearer?

vyper/ast/README.md
iamdefinitelyahuman: https://github.com/vyperlang/vyper/pull/1871/files#diff-260f2647b0ee0405377642ee077b51f1R40

iamdefinitelyahuman:
@fubuloubu this most recent force push expanded the docstring for `VyperNode.from_node` and added test cases for `from_node` and `compare_nodes`

vyper/ast/nodes.py
```diff
@@ -196,13 +251,52 @@ def __init__(self, parent: Optional["VyperNode"] = None, **kwargs: dict):
...
+        Returns
+        -------
+        Vyper node instance
+        """
+        ast_struct = {i: getattr(node, i) for i in VyperNode.__slots__ if not i.startswith('_')}
+        ast_struct.update(ast_type=cls.__name__, **kwargs)
+        return cls(**ast_struct)
+
+    @classmethod
+    def get_slots(cls) -> set:
```
fubuloubu: "slots" seems like a more technical name than we need to expose in this API. Why not `get_fields` instead?

vyper/ast/README.md
fubuloubu: `ancestor`

vyper/ast/nodes.py
iamdefinitelyahuman: :+1: `get_fields` seems much clearer and less scary sounding.  Will change.

Issue #1938, opened 2020-04-26
Remove 'swap' from the list of reserved keywords


gonchs:
### Version Information

* vyper Version (output of `vyper --version`): 0.1.0b17+commit.0671b7b
* OS: osx
* Using docker

### What's your issue about?

I need to call a function called "swap" on an external contract from the Vyper code. When I try to define the interface, I get an error: "Function name invalid. swap is a reserved keyword (Vyper language)".

`SWAP` is an opcode. But does it need to be a reserved keyword? Is it [used for anything](https://github.com/vyperlang/vyper/search?q=swap&unscoped_q=swap)?

### How can it be fixed?

Remove "swap" from the list of reserved keywords.

fubuloubu:
We're definitely being a little overzealous with the naming

PR #1939, opened 2020-04-26
Remove some reserved keywords

fubuloubu:
fixes: #1938

### What I Did
There's no reason for certain LLL keywords to be reserved for use with Vyper syntax

I removed:
`push`, `dup`, `swap`, `call`, `stop`

### Cute Animal Picture
cute gecko

### DIFF
--- a/vyper/utils.py
+++ b/vyper/utils.py
@@ -155,7 +155,7 @@ RESERVED_KEYWORDS = {
     # control flow
     'if', 'for', 'while', 'until', 'pass', 'def',
     # EVM operations
-    'send', 'selfdestruct', 'assert', 'raise',
+    'send', 'selfdestruct', 'assert', 'raise', 'throw',
     # special functions (no name mangling)
     'init', '_init_', '___init___', '____init____',
     'default', '_default_', '___default___', '____default____',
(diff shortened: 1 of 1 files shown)


iamdefinitelyahuman:
most of these i agree on, but `throw` has a well-defined meaning from solidity. It's also a reserved keyword in that language - so no chance of an issue like the one reported in #1938.  I'm not sure it should be an allowable name.

vyper/utils.py
```diff
@@ -155,12 +155,10 @@ def in_bounds(cls, type_str, value):
     # control flow
     'if', 'for', 'while', 'until', 'pass', 'def',
     # EVM operations
-    'push', 'dup', 'swap', 'send', 'call',
-    'selfdestruct', 'assert', 'stop', 'throw',
-    'raise', 'init', 'default',
-    # special function (name mangling)
-    '_init_', '___init___', '____init____',
-    '_default_', '___default___', '____default____',
+    'send', 'selfdestruct', 'assert', 'raise',
```
fubuloubu, line 158:
```suggestion
    'send', 'selfdestruct', 'assert', 'raise', 'throw',
```

PR #1942, opened 2020-04-27
Folding builtins

iamdefinitelyahuman:
### What I did
Add folding logic for `floor` and `ceil` builtin functions.

### How I did it
* add `ast/validation.py` which contains the `validate_call_args` method. This method checks the number of arguments used in a function call, and the key names for keyword arguments.
* in `functions/functions.py`, add an `evaluate` method to the classes for `floor` and `ceil`. This method accepts a `Call` node and returns an `Int` node when the value can be folded.
* add property-based tests to verify that the folded output is equivalent to the unfolded output.

### How to verify it
Run tests.

### Cute Animal Picture
image

### DIFF
--- a/vyper/functions/functions.py
+++ b/vyper/functions/functions.py
@@ -2,13 +2,18 @@ from decimal import (
     Decimal,
 )
 import hashlib
+import math
 
 from vyper import (
     ast as vy_ast,
 )
+from vyper.ast.validation import (
+    validate_call_args,
+)
 from vyper.exceptions import (
     ConstancyViolation,
     InvalidLiteral,
+    InvalidType,
     StructureException,
     TypeMismatch,
 )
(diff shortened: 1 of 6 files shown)


vyper/functions/functions.py
```diff
@@ -85,6 +90,14 @@ class Floor:
     _inputs = [("value", "decimal")]
     _return_type = "int128"
 
+    def evaluate(self, node):
+        validate_call_args(node, 1)
+        if not isinstance(node.args[0], vy_ast.Decimal):
+            raise InvalidType("Call cannot be folded", node)
```
fubuloubu, line 96: I feel like this should raise "Invalid type for arg" earlier somehow than during folding

iamdefinitelyahuman:
Closing for now, will reopen with a larger amount of code.

PR #1949, opened 2020-05-06
Folding builtins

iamdefinitelyahuman:
### What I did
Add constant folding logic for builtin functions.

### How I did it
* folding logic is found in the `evaluate` method of each class in `vyper/functions/functions.py`
* the high level execution is handled via `replace_builtin_functions` in `vyper/ast/folding.py`
* test cases are in `tests/functions/folding/`

The only folding implementation I'm not testing is `method_id`. This function should _always_ fold, so verifying the result of folding vs not-folding seems redundant. Once we add folding into the compiler process we can update the existing test cases around this function.

### Next Steps
Things to do next that are outside the scope of this PR.

* add documentation
    * updates to "Types" and "Builtin functions" within RTD
    * add docstrings to classes in `vyper/functions/functions.py`
    * add `README.md` to `vyper/functions`
* plug folding logic into main compiler process
* refactor out now-redundant code

### Cute Animal Picture
image

### DIFF
--- a/tests/functions/folding/test_bitwise.py
+++ b/tests/functions/folding/test_bitwise.py
@@ -1,0 +1,71 @@
+from hypothesis import (
+    given,
+    settings,
+    strategies as st,
+)
+import pytest
+
+from vyper import (
+    ast as vy_ast,
+    functions as vy_fn,
+)
+
+st_uint256 = st.integers(min_value=0, max_value=2 ** 256 - 1)
+
+
+@pytest.mark.fuzzing
+@settings(max_examples=50, deadline=1000)
+@given(a=st_uint256, b=st_uint256)
+@pytest.mark.parametrize('fn_name', ['bitwise_and', 'bitwise_or', 'bitwise_xor'])
+def test_bitwise(get_contract, a, b, fn_name):
+
+    source = f"""
+@public
+def foo(a: uint256, b: uint256) -> uint256:
+    return {fn_name}(a, b)
+    """
+    contract = get_contract(source)
+
+    vyper_ast = vy_ast.parse_to_ast(f"{fn_name}({a}, {b})")
+    old_node = vyper_ast.body[0].value
+    new_node = vy_fn.DISPATCH_TABLE[fn_name].evaluate(old_node)
+
+    assert contract.foo(a, b) == new_node.value
+
+
+@pytest.mark.fuzzing
+@settings(max_examples=50, deadline=1000)
+@given(value=st_uint256)
+def test_bitwise_not(get_contract, value):
+
+    source = f"""
+@public
+def foo(a: uint256) -> uint256:
+    return bitwise_not(a)
+    """
+    contract = get_contract(source)
+
+    vyper_ast = vy_ast.parse_to_ast(f"bitwise_not({value})")
+    old_node = vyper_ast.body[0].value
+    new_node = vy_fn.BitwiseNot().evaluate(old_node)
+
+    assert contract.foo(value) == new_node.value
+
+
+@pytest.mark.fuzzing
+@settings(max_examples=50, deadline=1000)
+@given(value=st_uint256, steps=st.integers(min_value=-256, max_value=256))
+def test_shift(get_contract, value, steps):
+
+    source = f"""
+@public
+def foo(a: uint256, b: int128) -> uint256:
+    return shift(a, b)
+    """
+    contract = get_contract(source)
+
+    vyper_ast = vy_ast.parse_to_ast(f"shift({value}, {steps})")
+    old_node = vyper_ast.body[0].value
+    new_node = vy_fn.Shift().evaluate(old_node)
+
+    assert contract.foo(value, steps) == new_node.value
--- a/tests/functions/folding/test_fold_as_wei_value.py
+++ b/tests/functions/folding/test_fold_as_wei_value.py
@@ -1,0 +1,60 @@
+from hypothesis import (
+    given,
+    settings,
+    strategies as st,
+)
+import pytest
+
+from vyper import (
+    ast as vy_ast,
+    functions as vy_fn,
+)
+
+denoms = [x for k in vy_fn.AsWeiValue.wei_denoms.keys() for x in k]
+
+
+st_decimals = st.decimals(
+    min_value=0,
+    max_value=2 ** 32,
+    allow_nan=False,
+    allow_infinity=False,
+    places=10,
+)
+
+
+@pytest.mark.fuzzing
+@settings(max_examples=10, deadline=1000)
+@given(value=st_decimals)
+@pytest.mark.parametrize('denom', denoms)
+def test_decimal(get_contract, value, denom):
+    source = f"""
+@public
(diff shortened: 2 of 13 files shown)


fubuloubu:
Lol, looks like you found a Hypothesis bug

tests/functions/folding/test_bitwise.py
```diff
@@ -0,0 +1,71 @@
...
+    vyper_ast = vy_ast.parse_to_ast(f"bitwise_not({value})")
+    old_node = vyper_ast.body[0].value
+    new_node = vy_fn.BitwiseNot().evaluate(old_node)
+
+    assert contract.foo(value) == new_node.value
+
+
+@pytest.mark.fuzzing
+@settings(max_examples=50, deadline=1000)
+@given(value=st_uint256, steps=st.integers(min_value=-1000, max_value=1000))
```
fubuloubu, line 57: range should be +/- 256, shifting 256+ in either direction should just make it be zero

tests/functions/folding/test_fold_as_wei_value.py
```diff
@@ -0,0 +1,60 @@
...
+@pytest.mark.parametrize('denom', denoms)
+def test_decimal(get_contract, value, denom):
+    source = f"""
+@public
+def foo(a: decimal) -> uint256:
+    return as_wei_value(a, '{denom}')
+    """
+    contract = get_contract(source)
+
+    vyper_ast = vy_ast.parse_to_ast(f"as_wei_value({value:.10f}, '{denom}')")
```
fubuloubu, line 37: `18f`, because 18 decimal places is the largest valid length for Ether denominations

tests/functions/folding/test_fold_as_wei_value.py
```diff
@@ -0,0 +1,60 @@
...
+
+denoms = [x for k in vy_fn.AsWeiValue.wei_denoms.keys() for x in k]
+
+
+st_decimals = st.decimals(
+    min_value=0,
+    max_value=2 ** 32,
+    allow_nan=False,
+    allow_infinity=False,
+    places=10,
```
fubuloubu, line 21: 18 decimal places

tests/functions/folding/test_fold_as_wei_value.py
```diff
@@ -0,0 +1,60 @@
...
+    ast as vy_ast,
+    functions as vy_fn,
+)
+
+denoms = [x for k in vy_fn.AsWeiValue.wei_denoms.keys() for x in k]
+
+
+st_decimals = st.decimals(
+    min_value=0,
+    max_value=2 ** 32,
```
fubuloubu, line 18: Probably don't need to do anything more than like 10 or 100. We want some < 1.0 values

tests/functions/folding/test_keccak_sha.py
```diff
@@ -0,0 +1,97 @@
...
+
+    vyper_ast = vy_ast.parse_to_ast(f"{fn_name}('''{value}''')")
+    old_node = vyper_ast.body[0].value
+    new_node = vy_fn.DISPATCH_TABLE[fn_name].evaluate(old_node)
+
+    assert f"0x{contract.foo(value).hex()}" == new_node.value
+
+
+@pytest.mark.fuzzing
+@given(value=st.text(alphabet=alphabet, min_size=0, max_size=100))
```
fubuloubu, line 36: Why not use bytes and then convert to a string?

vyper/functions/functions.py
```diff
@@ -85,6 +93,14 @@ class Floor:
     _inputs = [("value", "decimal")]
     _return_type = "int128"
```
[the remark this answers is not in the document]
iamdefinitelyahuman: Yeah, because the input is always a decimal. We could change this in the future tho...  50% of cases, `uint256` will be fine, just needs an underflow check.

vyper/functions/functions.py
```diff
@@ -448,6 +523,31 @@ class MethodID:
     _id = "method_id"
     _inputs = [("method", "str_literal"), ("type", "name_literal")]
 
+    # TODO once this is plugged in, build_LLL can be removed as this method is always foldable
```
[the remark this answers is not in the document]
iamdefinitelyahuman: None of this code is actively doing anything yet.  Need to refactor `vyper/compiler.py`, then plug this in, then it can be removed.

tests/functions/folding/test_bitwise.py
iamdefinitelyahuman: Still good to have tests that verify that the result is 0 when shifting >256

tests/functions/folding/test_fold_as_wei_value.py
iamdefinitelyahuman: I chose 10 because we only allow a decimal precision of 10.  If we fold a decimal with a precision of 18 then we're allowing an invalid type to get past.

tests/functions/folding/test_keccak_sha.py
iamdefinitelyahuman: That..  is a great idea.

tests/functions/folding/test_fold_as_wei_value.py
fubuloubu: Note: Would fail because of https://github.com/vyperlang/vyper/issues/1637

tests/functions/folding/test_fold_as_wei_value.py
fubuloubu: Similar note for https://github.com/vyperlang/vyper/issues/1637

tests/functions/folding/test_fold_as_wei_value.py
fubuloubu: Ah gotcha

fubuloubu: Note: Change this when https://github.com/vyperlang/vyper/issues/1807 is implemented

tests/functions/folding/test_bitwise.py
fubuloubu: >=256. I mean, there's no difference in behavior past 255, we'll just get less examples validating the real behavior (statistically, about 12 examples out of 50 will be non-trivial)

vyper/functions/functions.py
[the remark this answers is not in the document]
fubuloubu: Well, I mean it should allow more than 1 type (basically, pass through input type)

But, I think this is talking about what we ultimately want, not what we have today.

tests/functions/folding/test_keccak_sha.py
iamdefinitelyahuman: Unfortunately it just led to a bunch of confusion. Some bytes can't decode to a string, and some decoded values such as `\` cause issues.  The exclusion logic ended up being far more confusing than just using `alphabet` and I'm not even sure if I handled all cases.  So sticking with `alphabet`.

iamdefinitelyahuman:
@fubuloubu unless you have other comments I think this is ready for merging after #1956 merges.

PR #1956, opened 2020-05-07
Binary to bytes

iamdefinitelyahuman:
### What I did
Swap binary to bytes prior to generating the vyper AST

### How I did it
* In `ast/annotation.py`, convert from binary to bytes and create `Bytes` node.
* remove `Binary` node and all references
* add tests

### How to verify it
Run tests.

### Cute Animal Picture
image


vyper/ast/annotation.py
```diff
@@ -139,15 +139,30 @@ def visit_Num(self, node):
         value = node.node_source_code
 
         # deduce non base-10 types based on prefix
-        literal_prefixes = {'0x': "Hex", '0b': "Binary", '0o': "Octal"}
+        literal_prefixes = {'0x': "Hex", '0o': "Octal"}
```
fubuloubu, line 142: Can't we do this with Octal too?

fubuloubu: Maybe even Hex. We can try to convert it to a check-summed address, and if it works then it's a valid address literal. If it isn't, it's bytes (even if it's length is 20). This way, improperly check-summed addresses will throw later on because they're not address types.

PR #1958, opened 2020-05-09
Update ERC721.py

gulfpearl:
Duplicate safeTransferFrom method signatures corrected.

### What I did
The method signatures were the same for both safeTransferFrom methods. I changed one to being with no data.


fubuloubu:
Good catch!

PR #1961, opened 2020-05-09
Refactor generation of GlobalContext

iamdefinitelyahuman:
_(This PR requires and is built on top of #1960)_
### What I did
Refactor methods for LLL, natspec and interface generation.

Each of these functions requires either the source code or AST as an input, in order to generate a `GlobalContext` object. I've modified them so that they now receive `GlobalContext` as an input.  This way we can make the core logic in `vyper/compiler.py` more linear.

### How I did it
* In several methods, modified the input arguments from `code` or `ast` to `global_ctx`.
* Adjusted `vyper/compiler.py` to generate `global_ctx` prior to calling each of these methods.

The only place that a `GlobalContext` is generated now, outside of `compiler.py`, is when handling interfaces.  I think it will be possible to refactor this out as well, but I didn't want to do too much in a single PR.

### How to verify it
Run the tests.

### Cute Animal Picture
image

### DIFF
--- a/vyper/parser/parser.py
+++ b/vyper/parser/parser.py
@@ -163,12 +163,10 @@ def parse_other_functions(o,
 
 # Main python parse tree => LLL method
 def parse_tree_to_lll(
-    vyper_ast_node: vy_ast.Module,
     source_code: str,
+    global_ctx: GlobalContext,
     runtime_only: bool = False,
-    interface_codes: Optional[InterfaceImports] = None
 ) -> LLLnode:
-    global_ctx = GlobalContext.get_global_context(vyper_ast_node, interface_codes=interface_codes)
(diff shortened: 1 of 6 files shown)


vyper/parser/parser.py
```diff
@@ -163,12 +163,10 @@ def parse_other_functions(o,
 
 # Main python parse tree => LLL method
 def parse_tree_to_lll(
```
fubuloubu, line 165: Is this function appropriately named?

fubuloubu: I'd expect this function to take an ast node as an argument

iamdefinitelyahuman: `GlobalContext` actually contains all of the information from the AST, so it isn't necessary to pass in the AST itself.

PR #1964, opened 2020-05-11
Apply isort linter rule

iamdefinitelyahuman:
### What I did
* Adjust `isort` linter rules to be compatible with `black`
* Apply `black` to `vyper/ast` and `vyper/compiler`
* Pin `flake8` dependency to `3.7.9` - `3.8.0` causes significant breaking changes


setup.py
```diff
@@ -19,7 +19,7 @@
     'hypothesis[lark]>=4.53.2,<5',
 ]
 lint_deps = [
-    'flake8>=3.7,<4',
+    'flake8==3.7.9',
```
fubuloubu, line 22:
```suggestion
    'flake8>=3.7.9,<3.8.0',
```

tests/parser/features/test_internal_call.py
```diff
@@ -1,16 +1,9 @@
...
 
-from vyper.compiler import (
-    compile_code,
-)
-from vyper.exceptions import (
-    StructureException,
-    TypeMismatch,
-)
+from vyper.compiler import compile_code
+from vyper.exceptions import StructureException, TypeMismatch
```
fubuloubu, line 6: This is what `isort` wants? I would think 2+ imports would be the other way

vyper/__init__.py
```diff
@@ -2,10 +2,7 @@
 
 import pkg_resources as _pkg_resources
 
-from vyper.compiler import (  # noqa
-    compile_code,
-    compile_codes,
-)
+from vyper.compiler import compile_code, compile_codes  # noqa
```
fubuloubu, line 5: is `noqa` still needed?

tests/parser/features/test_internal_call.py
iamdefinitelyahuman: This is what `black` wants, in order to reduce the size of diffs.

vyper/__init__.py
iamdefinitelyahuman: Yes, but it shouldn't be a blanket `# noqa` here it should be `# noqa: F401`.  Good catch.

iamdefinitelyahuman: Fixed

PR #1967, opened 2020-05-13
AST constant folding

iamdefinitelyahuman:
### What I did
Add constant folding logic to the compilation process.

### How I did it
* Add a new phase to `compiler/phases.py` that calls `ast.folding.fold` to generate the folded AST. Because folding is a mutating action, I first perform a `deepcopy` of the original node. This way it is still available for compiler outputs that require it.
* Remove now-dead code in `parser/expr.py` and `functions/functions.py`
* Various fixes to minor bugs that were revealed after applying the new logic

### How to verify it
Run the tests. I moved / adjusted a few cases, but nothing substantial changed within the test suite.

### Cute Animal Picture
image


vyper/functions/functions.py
```diff
@@ -370,8 +371,8 @@ def evaluate(self, node):
         else:
             raise UnfoldableNode
 
-        hash_ = keccak256(value)
-        return vy_ast.Bytes.from_node(node, value=hash_)
+        hash_ = f"0x{keccak256(value).hex()}"
+        return vy_ast.Hex.from_node(node, value=hash_)
```
iamdefinitelyahuman, line 375: I had to change this back to `Hex` - `parser/expr.py` interprets a 32 byte string as `bytes[32]`. Once type checking is implemented we can try to fold out `Hex` altogether.

vyper/compiler/README.md
```diff
@@ -25,13 +25,15 @@ The compilation process has five broad phases:
 
 1. In [`vyper.ast`](../ast), the source code is parsed and converted to an
 abstract syntax tree.
-2. The [`GlobalContext`](../parser/global_context.py) object is generated from the
+2. In [`vyper.ast.folding`](../ast/folding.py), literal Vyper AST nodes are
+evaluated and replaced with the resulting values.
+3. The [`GlobalContext`](../parser/global_context.py) object is generated from the
 Vyper AST, analyzing and organizing the nodes prior to LLL generation.
-3. In [`vyper.parser.parser`](../parser/parser.py), the contextualized nodes are
```
fubuloubu, line 30: You can just use `1.` for all items in a markdown numbered list and it turns out okay. This way you don't have to modify them every time.

vyper/compiler/phases.py
```diff
@@ -144,13 +154,35 @@ def generate_ast(source_code: str, source_id: int) -> vy_ast.Module:
...
+
+
 def generate_global_context(
     vyper_module: vy_ast.Module, interface_codes: Optional[InterfaceImports],
 ) -> GlobalContext:
     """
     Generate a contextualized AST from the Vyper AST.
 
-    This is the second phase of compilation.
+    This is the third phase of compilation.
```
fubuloubu, line 185: I really don't think we should have these numbers spelled out like this.

iamdefinitelyahuman: Yeah, agreed.  I'll add a single comment that the complier phases execute in the same order as the functions are given.

vyper/compiler/README.md
iamdefinitelyahuman: Voodoo!

PR #1982, opened 2020-05-18
Fix memory length generated by assert_reason

### LINKED ISSUE #1981: Error strings are zero-padded to 32 bytes
### What's your issue about?
Vyper zero-pads error strings to 32 bytes within the bytecode. This in turn causes the string to occupy an additional memory slot, and can lead to [decoding weirdness](https://etherscan.io/tx/0xa0f0aa5fc4f56771a0b49ff66c752bcad2a32844ec4b173337086d6610571d65).

Here is an example contract:

```python
@public
def foo():
    raise "foo"
```

Here is the bytecode that handles the error string:

```python
PUSH4 0x8C379A0   # error string signature
PUSH2 0x140
MSTORE
PUSH1 0x20
PUSH2 0x160
MSTORE
PUSH1 0x3
PUSH2 0x180
MSTORE
PUSH32 0x666F6F00000000000000000000000000000  # "foo" with many trailing zeros
PUSH2 0x1A0
MSTORE 
PUSH2 0x180
POP
PUSH1 0x0
PUSH2 0x15
JUMPI
PUSH1 0x84  # returns 4 memory slots
PUSH2 0x15C
REVERT
```

### How can it be fixed?
Remove the zero-padding and use `PUSH[n]` where `n` is the actual length of the encoded string. This way we can also reduce the size in memory that we are returning.

ZumZoom:
Fixes #1981

### Description of issue

Code generation for `assert_reason` generates incorrect length for revert string which leads to [issues with revert reason decoding](https://etherscan.io/tx/0xa0f0aa5fc4f56771a0b49ff66c752bcad2a32844ec4b173337086d6610571d65).

### How I fixed it

Fixed length of memory passed to revert opcode generated by assert_reason.

### How to verify it

Check that [assert tests](https://github.com/vyperlang/vyper/blob/master/tests/parser/features/test_assert.py) are still passing.

Removed 32 bytes of junk at the end of revert reasons.

### Cute Animal Picture

### DIFF
--- a/tests/parser/features/test_assert.py
+++ b/tests/parser/features/test_assert.py
@@ -1,5 +1,6 @@
 import pytest
 from eth_tester.exceptions import TransactionFailed
+from eth_abi import decode_single
 
 from vyper.exceptions import ConstancyViolation, StructureException
 
@@ -192,3 +193,20 @@ def test(x: uint256[3]) -> bool:
     assert_tx_failed(lambda: c.test([5, 1, 3]))
     assert_tx_failed(lambda: c.test([1, 5, 3]))
     assert_tx_failed(lambda: c.test([1, 3, 5]))
+
+
+def test_assest_reason_revert_length(w3, get_contract):
+    code = """
+@public
+def test() -> int128:
+    assert 1 == 2, "oops"
+    return 1
+"""
+    c = get_contract(code)
+    w3.manager.provider.ethereum_tester.backend.is_eip838_error = lambda err: False
+    with pytest.raises(TransactionFailed) as e_info:
+        c.test()
+    error_bytes = eval(e_info.value.args[0])
+    assert len(error_bytes) == 100
+    msg = decode_single('string', error_bytes[36:])
+    assert msg == 'oops'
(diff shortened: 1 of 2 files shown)


iamdefinitelyahuman:
Closes #1981

iamdefinitelyahuman:
This looks right to me.  A followup to my comments in #1981, I think we actually do need the `PUSH32` to ensure the memory is zero'd.

tests/parser/features/test_assert.py
```diff
@@ -192,3 +193,20 @@ def test(x: uint256[3]) -> bool:
     assert_tx_failed(lambda: c.test([5, 1, 3]))
     assert_tx_failed(lambda: c.test([1, 5, 3]))
     assert_tx_failed(lambda: c.test([1, 3, 5]))
+
+
+def test_assest_reason_revert_length(w3, get_contract):
+    code = """
+@public
+def test() -> int128:
+    assert 1 == 2, "oops"
```
fubuloubu, line 202: oopsie

PR #1988, opened 2020-05-27
Validate constants

iamdefinitelyahuman:
### What I did
Validate constant values prior to folding.

### How I did it
Add a `validate` method to some AST nodes that raises if the node's value cannot be typed. This is useful for ensuring descriptive exceptions for e.g. numeric bounds issues, without having to duplicate code.

The `validate` methods are called from `ast.validate.validate_literal_nodes`, which runs after AST generation and prior to constant folding.

This PR results in the some dead code in `functions/functions.py` and `parser/**.py`.  I intend to handle that as part of the larger coming soon:tm: type-checker PR.

### How to verify it
Run the tests.

### Cute Animal Picture
image

### DIFF
--- a/vyper/ast/nodes.py
+++ b/vyper/ast/nodes.py
@@ -6,6 +6,8 @@ from typing import Any, Optional, Union
 
 from vyper.exceptions import (
     CompilerPanic,
+    InvalidLiteral,
+    OverflowException,
     SyntaxException,
     TypeMismatch,
     UnfoldableNode,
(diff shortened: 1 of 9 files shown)


vyper/ast/nodes.py
```diff
@@ -640,6 +655,16 @@ def n(self):
         # TODO phase out use of Num.n and remove this
         return self.value
 
+    def validate(self):
+        if self.value < -(2 ** 127):
```
fubuloubu, line 659: Can we at least grab this via `MIN_INT128`?

vyper/ast/nodes.py
```diff
@@ -640,6 +655,16 @@ def n(self):
...
+            raise OverflowException(
+                "Value is below lower bound for all numeric types", self
+            )
+        if self.value > 2 ** 256 - 1:
```
fubuloubu, line 663: Can we at least grab this via `MAX_UINT256`?

vyper/ast/nodes.py
```diff
@@ -666,6 +691,11 @@ class Decimal(Num):
 
     __slots__ = ()
 
+    def validate(self):
+        if self.value.as_tuple().exponent < -10:
```
fubuloubu, line 695: Is the number of decimal places a constant somewhere in the codebase?

PR #1992, opened 2020-05-29
Add test case for validate_version_pragma.

arjunaskykok:
### What I did

There is a method named validate_version_pragma which has not been tested. So I add a test case.


tests/ast/test_pre_parser.py
```diff
@@ -0,0 +1,29 @@
...
+
+from vyper.ast.pre_parser import validate_version_pragma
+from vyper.exceptions import VersionException
+
+
+def test_validate_version_pragma_valid(monkeypatch):
+    _mock_version(monkeypatch)
+    validate_version_pragma(' @version 0.1.0b17', (1, 0))
+    validate_version_pragma(' @version 0.1.3', (1, 0))
+    assert True
```
fubuloubu, line 11: Why is this line here?

tests/ast/test_pre_parser.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+def test_validate_version_pragma_invalid(monkeypatch):
+    _mock_version(monkeypatch)
+    with pytest.raises(VersionException):
+        validate_version_pragma(' @version 0.1', (1, 0))
+
+
+def _mock_version(monkeypatch):
+    monkeypatch.setattr("vyper.__version__", "0.1.0b17")
```
fubuloubu, line 29: Can we move this to the top, and make it it's own fixture?
```suggestion
@pytest.fixture
def mock_version(monkeypatch):
    def set_version(version=vyper.__version__):
        monkeypatch.setattr("vyper.__version__", version)
    return set_version
```

You would import `mock_version` instead of `monkey_patch`, and it also allows setting different versions to check (or leaving it empty and checking against whatever the latest is)

tests/ast/test_pre_parser.py
```diff
@@ -0,0 +1,29 @@
+import pytest
+
+from vyper.ast.pre_parser import validate_version_pragma
+from vyper.exceptions import VersionException
+
+
+def test_validate_version_pragma_valid(monkeypatch):
+    _mock_version(monkeypatch)
+    validate_version_pragma(' @version 0.1.0b17', (1, 0))
+    validate_version_pragma(' @version 0.1.3', (1, 0))
```
fubuloubu, line 10: Turn this into a parameterized test case:
```suggestion
@pytest.mark.parametrize('version', [
    vyper.__version__,
    '0.1.3',  # Maybe find some cleverer way of setting these,
    '1.0.0',  # so it's more scalable as we push new versions
])
def test_valid_version_pragma(monkeypatch, version):  # Shorter name too
    _mock_version(monkeypatch)
    validate_version_pragma(f' @version {version}', (1, 0))
```

tests/ast/test_pre_parser.py
```diff
@@ -0,0 +1,29 @@
...
+    with pytest.raises(VersionException):
+        validate_version_pragma(' @version 0.2.0b17', (1, 0))
+    with pytest.raises(VersionException):
+        validate_version_pragma(' @version 3.1.0b17', (1, 0))
+
+
+def test_validate_version_pragma_invalid(monkeypatch):
+    _mock_version(monkeypatch)
+    with pytest.raises(VersionException):
+        validate_version_pragma(' @version 0.1', (1, 0))
```
fubuloubu, line 25: Parametrize here too:
```suggestion
@pytest.mark.parametrize('version', [
    '0.2.0b17',
    '3.1.0b17',
])
def test_invalid_version_pragma(monkeypatch, version):
    _mock_version(monkeypatch)
    with pytest.raises(VersionException):
        validate_version_pragma(' @version {version}', (1, 0))
```

tests/ast/test_pre_parser.py
fubuloubu: Also consider making `(1, 0)` into a readable constant like `SRC_LINE`

tests/ast/test_pre_parser.py
arjunaskykok: Sorry, I just feel the need to add at least one "assert" statement in one test case. I'll remove that.

tests/ast/test_pre_parser.py
arjunaskykok: Roger that!

tests/ast/test_pre_parser.py
arjunaskykok: Aye, Sir!

tests/ast/test_pre_parser.py
arjunaskykok: Yes, Sir!

arjunaskykok:
> Just a few comments, overall it is good.
> 
> It would be nice to have a better way of working with versions, perhaps the [`bumpversion.Version`](https://github.com/peritus/bumpversion/blob/cc3c8cfd77380ef50eeac740efe627509a248101/bumpversion/__init__.py#L372-L394) object can be used to make it better, so we can add/substract, set pre-release, etc. from the version number we do have.

I'll investigate the bumpversion tool and see how I can integrate the tool into Vyper.

fubuloubu:
We use bumpversion to manage our version number, so it's a natural fit to import the package and see if we can use the data structure to make for easier testing some how. Something to explore.

tests/ast/test_pre_parser.py
```diff
@@ -0,0 +1,41 @@
+import pytest
+
+from bumpversion.version_part import VersionConfig
+from vyper import __version__
+from vyper.ast.pre_parser import validate_version_pragma
+from vyper.exceptions import VersionException
+
+SRC_LINE = (1, 0)
```
fubuloubu, line 8: Just add a comment that this is a dummy source line:
```suggestion
SRC_LINE = (1, 0)  # Dummy source line
```

arjunaskykok: Aye!

PR #1994, opened 2020-05-30
Update Install Instructions

fubuloubu:
### What I did
Reordered and streamlined the installation docs to put Docker first (to reduce the number of issues people have setting up vyper)

### Cute Animal Picture
crouching monkey, hidden panda

### DIFF
--- a/docs/installing-vyper.rst
+++ b/docs/installing-vyper.rst
@@ -1,10 +1,6 @@
 Installing Vyper
 ################
 
-Don't panic if the installation fails. Vyper is still under development and
-undergoes constant changes. Installation will be much more simplified and
-optimized after a stable version release.
-
 Take a deep breath, follow the instructions, and please
 `create an issue <https://github.com/vyperlang/vyper/issues>`_ if you encounter
 any errors.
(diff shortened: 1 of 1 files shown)


rayrrr:
Might I recommend that we consider replacing the entire Python 3.6 installation section of these instructions with a suggestion to use https://github.com/pyenv/pyenv as a cross-platform means of making Python 3.6 available on devs' systems?

fubuloubu:
I use pyenv, and works really well for me. Will add that suggestion

rayrrr:
Cool @fubuloubu thanks. One other thought along these lines...since this itself is a compilable language perhaps we should actually consider https://asdf-vm.com instead of `pyenv`. That way if we eventually want to offer a Vyper plugin for it, we will already have some experience using it...

fubuloubu:
Hosting the docs changes here: https://vyper.readthedocs.io/en/fubuloubu-install-notes/installing-vyper.html

docs/installing-vyper.rst
```diff
@@ -178,75 +126,48 @@ If any unexpected errors or exceptions are encountered, please feel free to `ope
...
+It is **strongly recommended** to install Vyper in **a virtual Python
+environment**, so that new packages installed and dependencies built are
+strictly contained in your Vyper project and will not alter or affect your
+other development environment set-up.
 
-    docker run -v $(pwd):/code/ -it --entrypoint /bin/bash vyperlang/vyper
-    root@d35252d1fb1b:/code# vyper <contract_file.vy>
 
-The normal paramaters are also supported, for example:
+To create a new virtual environment for Vyper run the following commands:
```
rayrrr, line 157: Can we make the commands use the new `pyenv` approach as recommended above rather than manual creation of a virtualenv?

fubuloubu: poetry is really interesting, but I haven't used it yet. Is it very beginner friendly?

docs/installing-vyper.rst
```diff
@@ -178,75 +126,48 @@ If any unexpected errors or exceptions are encountered, please feel free to `ope
...
-    docker run -v $(pwd):/code vyperlang/vyper /code/<contract_file.vy>
+Creating a virtual environment
+==============================
 
-Alternatively you can log into the docker image and execute vyper on the prompt.
-::
+It is **strongly recommended** to install Vyper in **a virtual Python
+environment**, so that new packages installed and dependencies built are
+strictly contained in your Vyper project and will not alter or affect your
+other development environment set-up.
```
fubuloubu, line 154:
```suggestion
other development environment set-up. For easy virtualenv management, we recommend `Poetry <https://github.com/python-poetry/poetry>`_.
```

docs/installing-vyper.rst
rayrrr: @fubuloubu Very beginner friendly! Easy as Py. ;) I was using `pipenv` for a while but after many debates with a `poetry` evangelist peer I finally gave it a try and am now a convert. So, this is a vote for `poetry` for someone who went from manual `virtualenv` management to `mkvirtualenv` to `pipenv` to `poetry`. Give it a try and see what you think!

PR #2012, opened 2020-06-18
Streamline Optional Dependencies

fubuloubu:
### What I did
- Added `dev` and `docs` optionals and organized them better
- cleaned up makefile and gitignore a bit

### Cute Animal Picture
grooming

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,22 +11,46 @@ endif
 init:
 	${pip} install .
 
-dev-deps:
-	${pip} install .[test,lint]
+dev-init:
+	${pip} install .[dev]
 
 test:
-	python setup.py test
+	pytest
 
 lint:
 	tox -e lint
 
-clean: clean-build clean-pyc clean-test
+docs:
+	rm -f docs/vyper.rst
+	rm -f docs/modules.rst
+	sphinx-apidoc -o docs/ -d 2 vyper/
+	$(MAKE) -C docs clean
+	$(MAKE) -C docs html
+	# To display docs, run:
+	# open docs/_build/html/index.html (macOS)
+	# start docs/_build/html/index.html (Windows)
+	# xdg-open docs/_build/html/index.html (Linux)
+
+release: clean
+	python setup.py sdist bdist_wheel
+	twine check dist/*
+	#twine upload dist/*
+
+clean: clean-build clean-docs clean-pyc clean-test
 
 clean-build:
 	@echo Cleaning python build files...
 	@rm -fr build/
 	@rm -fr dist/
 	@rm -fr *.egg-info
+	@rm -f vyper/vyper_git_version.txt
+
+clean-docs:
+	@echo Cleaning doc build files...
+	@rm -rf docs/_build/
+	@rm -f docs/modules.rst
+	@rm -f docs/vyper.rst
+	@rm -f docs/vyper.*.rst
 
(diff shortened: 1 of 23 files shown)


setup.py
```diff
@@ -7,31 +7,43 @@
...
-
-
-extras = {
-    'test': test_deps,
-    'lint': lint_deps,
+extras_require = {
+    'test': [
+        'pytest>=5.2.0,<6',
+        'pytest-cov>=2.8.1,<3',
+        'coveralls[yaml]>=1.8.2,<2',
```
iamdefinitelyahuman, line 14: probably don't need this anymore since we use codecov

setup.py
```diff
@@ -7,31 +7,43 @@
...
+        'isort>=4.2.15,<5',
+        'mypy>=0.740,<1',
+    ],
+    'docs': [
+        'recommonmark',
+        'sphinx',
+        'sphinx_rtd_theme',
+    ],
+    'dev': [
+        'ipython',
```
iamdefinitelyahuman, line 35: it's a useful tool, but `ipython` as a dependency seems weird

setup.py
```diff
@@ -7,31 +7,43 @@
...
-    'test': test_deps,
-    'lint': lint_deps,
+extras_require = {
+    'test': [
+        'pytest>=5.2.0,<6',
+        'pytest-cov>=2.8.1,<3',
+        'coveralls[yaml]>=1.8.2,<2',
+        'pytest-xdist>=1.30.0,<2',
+        'eth-tester[py-evm]>=0.3.0b1,<0.4',
+        'web3>=5.2.0,<5.3.0',
```
iamdefinitelyahuman, line 17: web3 is up to `5.11.1` and we get a warning from something in it..  maybe time to bump this one as well

Makefile
```diff
@@ -11,22 +11,46 @@ endif
...
 	@rm -fr dist/
 	@rm -fr *.egg-info
+	@rm -f vyper/vyper_git_version.txt
+
+clean-docs:
+	@echo Cleaning doc build files...
+	@rm -rf docs/_build/
+	@rm -f docs/modules.rst
+	@rm -f docs/vyper.rst
+	@rm -f docs/vyper.*.rst
```
iamdefinitelyahuman, line 53: not sure these lines do anything

setup.py
fubuloubu: nothing under `dev` is a required dependency, just helpful

Makefile
fubuloubu: they do, these are intermediate build files for sphinx

fubuloubu:
@iamdefinitelyahuman I had to make a bunch of changes for updated linter rules in d2b1461 

You might have to rebase off of it

fubuloubu:
Hypothesis [`v5.8.6`](https://hypothesis.readthedocs.io/en/latest/changes.html#v5-8-6) changed a warning for function scoped pytest fixtures, however the warning message gives an unpredictable exception due to how it is being composed. This affects parameterized test cases only.

Slated to be fixed: https://github.com/HypothesisWorks/hypothesis/issues/2469

fubuloubu:
Hopefully HypothesisWorks/hypothesis#2469 gets released soon, but this PR has a fixed pin that solves the problem for now.

PR #2017, opened 2020-06-21
Type check tests and refactoring

iamdefinitelyahuman:
### What I did
* first round of unit tests for `vyper/context`
* refactor out `build_type_from_ann_assign`
* move function constancy checks to `FunctionNodeVisitor`
* properly handle ABI array types

Next up, refactoring (gutting) `parser`

### How to verify it
Run the tests...  there are 930-ish new ones.

### Cute Animal Picture
image

### DIFF
--- a/tests/functional/context/conftest.py
+++ b/tests/functional/context/conftest.py
@@ -1,0 +1,31 @@
+import pytest
+
+from vyper import ast as vy_ast
+from vyper.context import namespace as ns
+
+
+@pytest.fixture(scope="session")
+def build_node():
+    """
+    Yields a helper function for generating a single Vyper AST node.
+    """
+
+    def _build_node(source):
+        source = f"""'I am a docstring.'\n{source}"""
+        ast = vy_ast.parse_to_ast(source).body[0]
+        if isinstance(ast, vy_ast.Expr):
+            ast = ast.value
+        return ast
+
+    yield _build_node
+
+
+@pytest.fixture
+def namespace():
+    """
+    Yields a clean `Namespace` object.
+    """
+    obj = ns.get_namespace()
+    obj.clear()
+    yield obj
+    obj.clear()
--- a/tests/functional/context/conftest.py
+++ b/tests/functional/context/conftest.py
@@ -1,7 +1,7 @@
 import pytest
 
 from vyper import ast as vy_ast
-from vyper.context import namespace as ns
+from vyper.context.namespace import get_namespace
 
 
 @pytest.fixture(scope="session")
@@ -11,6 +11,7 @@ def build_node():
     """
 
     def _build_node(source):
+        # docstring ensures string nodes are properly generated, not turned into docstrings
         source = f"""'I am a docstring.'\n{source}"""
         ast = vy_ast.parse_to_ast(source).body[0]
         if isinstance(ast, vy_ast.Expr):
@@ -25,7 +26,7 @@ def namespace():
     """
     Yields a clean `Namespace` object.
     """
-    obj = ns.get_namespace()
+    obj = get_namespace()
     obj.clear()
     yield obj
     obj.clear()
--- a/tests/functional/context/types/test_pure_types.py
+++ b/tests/functional/context/types/test_pure_types.py
@@ -1,0 +1,131 @@
+import pytest
+
+from vyper.context.types.value.address import AddressPureType
+from vyper.context.types.value.array_value import (
+    BytesArrayPureType,
+    StringPureType,
+)
+from vyper.context.types.value.boolean import BoolPureType
+from vyper.context.types.value.bytes_fixed import Bytes32PureType
+from vyper.context.types.value.numeric import (
+    DecimalPureType,
+    Int128PureType,
+    Uint256PureType,
+)
+from vyper.exceptions import (
+    InvalidLiteral,
+    OverflowException,
+    StructureException,
+    UnexpectedNodeType,
+    UnexpectedValue,
+)
+
+BASE_SOURCE = """
+"Add a docstring so the target test node doesn't become one."
+"""
+
+PURE_TYPES = {
+    AddressPureType: "address",
+    BoolPureType: "bool",
+    Bytes32PureType: "bytes32",
+    BytesArrayPureType: "bytes[1]",
+    DecimalPureType: "decimal",
+    Int128PureType: "int128",
+    StringPureType: "string[1]",
+    Uint256PureType: "uint256",
+}
+
+
+VALID_LITERALS = {
+    AddressPureType: ["0x6B175474E89094C44Da98b954EedeAC495271d0F"],
+    BoolPureType: ["True", "False"],
+    Bytes32PureType: ["0x6B175474E89094C44Da98b954EedeAC495271d0F4Da98b954EedeAC495271d0F"],
+    BytesArrayPureType: [
+        "b''",
+        "b'this is thirty three bytes long!!'",
+        r"b'\xbe\xef'",
+        "0x6B175474E89094C44Da98b954EedeAC495271d0F",
+        "0x6B175474E89094C44Da98b954EedeAC495271d0F4Da98b954EedeAC495271d0F",
+    ],
+    DecimalPureType: ["-1.666", "3.31337", "8008135.0", "1.2345678901"],
+    Int128PureType: ["-1", "0", "12", "42"],
+    StringPureType: ["''", "'hello'", "'this is thirty three chars long!!'", "'-42'"],
+    Uint256PureType: ["0", "12", "42"],
+}
+
+INVALID_LITERALS = [
+    "None",
+    "115792089237316195423570985008687907853269984665640564039457584007913129639936",
+    "115792089237316195423570985008687907853269984665640564039457584007913129639935.1",
+]
+INVALID_NODES = [
+    "self.{}",
+    "[{}]",
+    "({},)",
+    "{}()",
+    "{}(foo)",
+    "{}.foo",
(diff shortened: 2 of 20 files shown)


fubuloubu:
I rebased `next` off of `master`, and now there's a bunch of conflicts

tests/functional/context/types/test_pure_types.py
```diff
@@ -0,0 +1,131 @@
...
+    "{}[1]",
+    "self.{}[1]",
+]
+
+
+@pytest.mark.parametrize("PureType", VALID_LITERALS.keys())
+@pytest.mark.parametrize(
+    "source", [x for v in VALID_LITERALS.values() for x in v] + INVALID_LITERALS
+)
+def test_from_literal(build_node, PureType, source):
```
fubuloubu, line 79: Lol, wtf is even happening here?

fubuloubu: All of these tests are like... way too clever lol

tests/functional/context/types/test_type_from_abi.py
```diff
@@ -0,0 +1,51 @@
...
+from vyper.context.types.indexable.sequence import ArrayDefinition
+from vyper.context.types.utils import get_type_from_abi
+from vyper.exceptions import UnknownType
+
+BASE_TYPES = ["int128", "uint256", "bool", "address", "bytes32"]
+
+
+@pytest.mark.parametrize("type_str", BASE_TYPES)
+def test_base_types(namespace, type_str):
+    pure_type = get_pure_types()[type_str]
```
fubuloubu, line 13: Man, I really don't like the term "pure types", it's very strange.

Basic? Base? Built-in?

vyper/context/types/utils.py
```diff
@@ -39,11 +35,18 @@ def get_type_from_abi(
...
+        if "[" in type_string:
+            value_type_string, length_str = type_string.rsplit("[", maxsplit=1)
+            length = int(length_str.rstrip("]"))
+            value_type = get_type_from_abi({"type": value_type_string}, is_constant=is_constant)
+            return ArrayDefinition(value_type, length, is_constant, is_public)
+
         return namespace[type_string]._type(is_constant=is_constant, is_public=is_public)
-    except KeyError:
+    except (KeyError, ValueError, InvalidType, UnknownType):
         raise UnknownType(f"ABI contains unknown type: {type_string}") from None
```
fubuloubu: Would be better if this was split it up better:
```python
    if "[" in type_string:
        value_type_string, length_str = type_string.rsplit("[", maxsplit=1)  # Shouldn't fail
        length_str = length_str.rstrip("]")
        if len(length_str) != 1:
            raise ...  # Invalid syntax
        try:
            length = int(length_str)  # Maybe also check if length has decimal places or something?
        except KeyError as e:
            raise ... from e  # Length is not an integer type
        if 0 > length or length < SOME_NUMBER:
           raise ...  # Length is not a valid range
        try:
            value_type = get_type_from_abi({"type": value_type_string}, is_constant=is_constant)
        except (KeyError, ValueError, InvalidType, UnknownType) as e:  # Not sure of the errors this throws
            raise UnknownType(f"ABI contains unknown type: {type_string}") from e
        return ArrayDefinition(value_type, length, is_constant, is_public)
    else:
        return namespace[type_string]._type(is_constant=is_constant, is_public=is_public)
```

All of this logic might live better in it's own function though

vyper/context/validation/module.py
```diff
@@ -58,34 +61,51 @@ def visit_AnnAssign(self, node):
...
+        if isinstance(annotation, vy_ast.Call):
+            # the annotation is a function call, e.g. `foo: constant(uint256)`
+            call_name = annotation.get("func.id")
+            if call_name in ("constant", "public"):
+                validate_call_args(annotation, 1)
+                if call_name == "constant":
+                    # declaring a constant
+                    is_constant = True
+
+                elif call_name == "public":
```
fubuloubu, line 80: followed by an `else` case that throws if it's not an expected `call_name`?

vyper/context/validation/module.py
```diff
@@ -58,34 +61,51 @@ def visit_AnnAssign(self, node):
...
+                    # declaring a public variable
+                    is_public = True
+                # remove the outer call node, to handle cases such as `public(map(..))`
```
fubuloubu, line 83: Really want to do `Map[typ, typ]` syntax instead...

vyper/context/validation/module.py
```diff
@@ -58,34 +61,51 @@ def visit_AnnAssign(self, node):
...
+            if not node.value:
+                raise VariableDeclarationException("Constant must be declared with a value", node)
+            if not check_literal(node.value):
+                raise ConstancyViolation("Value must be a literal", node.value)
+
+            validate_expected_type(node.value, type_definition)
+            try:
+                self.namespace[name] = type_definition
+            except VyperException as exc:
+                raise exc.with_annotation(node) from None
```
fubuloubu, line 97: why `raise ... from None` is being used everywhere?

vyper/context/validation/module.py
```diff
@@ -94,19 +114,19 @@ def visit_ClassDef(self, node):
...
             self.namespace["self"].add_member(func.name, func)
         except VyperException as exc:
             raise exc.with_annotation(node) from None
 
+    def visit_Import(self, node):
+        _add_import(node, self.interface_codes, self.namespace)
+
+    def visit_ImportFrom(self, node):
+        _add_import(node, self.interface_codes, self.namespace)
+
```
fubuloubu, line 129: Why is there no difference between these two?

vyper/exceptions.py
```diff
@@ -235,11 +235,22 @@ def __init__(self, message=""):
         self.message = message
 
     def __str__(self):
-        return f"{self.message} Please create an issue."
+        return (
+            f"{self.message}\n\nThis is an unhandled internal compiler error. "
+            "Please create an issue on Github to notify the developers."
```
fubuloubu, line 240: Maybe add a link to do this?

tests/functional/context/conftest.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+@pytest.fixture(scope="session")
+def build_node():
+    """
+    Yields a helper function for generating a single Vyper AST node.
+    """
+
+    def _build_node(source):
+        source = f"""'I am a docstring.'\n{source}"""
```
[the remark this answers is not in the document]
iamdefinitelyahuman: The docstring is required because if the first node is a string, it gets parsed as a docstring insead of a regular node in the body.  I agree it's not intuitive, I'll add a comment.

vyper/context/validation/module.py
iamdefinitelyahuman: Line 74, we can't get here unless `call_name` is `public` or `constant`.  Possible line 80 should just be an `else` but I felt this was more readable.

vyper/context/validation/module.py
iamdefinitelyahuman: Yeah - I didn't wanna add type checking and modify syntax at the same time though.  Already a big enough diff :smile:

vyper/context/validation/module.py
iamdefinitelyahuman: It ensures you don't see the previous traceback that we're currently handling.  In this case, `with_annotation` creates a new exception by appending the annotation onto the old one.  Without `from None` the error message would be two copies of the exact same traceback.

vyper/context/validation/module.py
iamdefinitelyahuman: They're different node types but the logic to process them is near-identical.  One is `import x`, other is `from .. import x`.

iamdefinitelyahuman: I could refactor a bit to handle the not-identical bits prior to calling the function, though.

tests/functional/context/types/test_pure_types.py
iamdefinitelyahuman: added a comment

vyper/context/validation/module.py
iamdefinitelyahuman: Done.

vyper/context/types/utils.py
iamdefinitelyahuman: * `if len(length_str) != 1:` - this would block arrays of a length of 10 or more
* the bounds checks on length are in `ArrayDefinition`

iamdefinitelyahuman: I made the edits here - https://github.com/vyperlang/vyper/pull/2017/commits/76df4088b9f4c69b0016a493c459bb86ffc72daa

I don't like it though, too much duplicated code.  The exceptions being raised are different depending on the line, the but reason is always near-identical. I think a catchall exception handler is a much cleaner approach.

fubuloubu: It gives more specific error messages, so why does it hurt?

vyper/context/validation/module.py
fubuloubu: This would make it easier to add other options later on

vyper/context/types/utils.py
iamdefinitelyahuman: Doesn't hurt.  Lets go with it.

tests/functional/context/types/test_type_from_abi.py
iamdefinitelyahuman: This will touch a lot of files.  Once this PR is good to merge, I'll make the name change in a new one.

fubuloubu: Sounds good!

PR #2019, opened 2020-06-21
Change actions trigger

fubuloubu:


iamdefinitelyahuman:
What is your reasoning for this change?  I rely on the CI to test my code as I work... this change makes my normal workflow not-possible and slows my productivity quite a bit.

fubuloubu:
It was running the CI twice for any branch, which was getting to be annoying and eating our CI time

PR #2022, opened 2020-06-22
Remove dead code, round 1

iamdefinitelyahuman:
### What I did
Remove dead code and replace `VyperException` subclasses with `VyperInternalException` subclasses in `vyper/parser/*.py`

### How I did it
Type-check related logic in `parser` has been removed, or replaced with internal exceptions that indicate an unexpected compiler error:

* when the check involves structure validation of a node, such that a later attempt to access the incorrect structure will result in an attribute or key error, I've removed the check altogether
* when I'm not certain the mistake would get caught in the above manner, I've instead replaced it with an internal error or be returning `None`

In the major functions, I'm using a pattern where a check is made on the return value, and if `None`, an internal error is raised.  This reduces the number of unreachable lines of code while still providing a layer of safety in case type checking is insufficient.

All of this is a stop-gap on the way to a full refactor of the `parser/` module - we just need to make sure that if the type check layer misses something, the same exception __isn't__ raised in `parser/` so that we can address the error.

Most of the deleted code was found by looking at this coverage report: https://codecov.io/gh/iamdefinitelyahuman/vyper/tree/685223bf473283ba8e305eb5e6b87bb5e589caf7/vyper

I'm not finished. I still have a short list of exceptions in `parser/` that the test cases hit, where I need to implement logic.  Another smaller but more logic-heavy PR will be coming soon to cover them.

### How to verify it
Observe that test are still passing.

### Cute Animal Picture
image

### DIFF
--- a/vyper/context/validation/utils.py
+++ b/vyper/context/validation/utils.py
@@ -21,6 +21,7 @@ from vyper.exceptions import (
     UndeclaredDefinition,
     UnknownAttribute,
     VyperException,
+    ZeroDivisionException,
 )
 
 
@@ -100,6 +101,14 @@ class _ExprTypeChecker:
     def types_from_BinOp(self, node):
         # binary operation: `x + y`
         types_list = get_common_types(node.left, node.right)
+
+        if (
+            isinstance(node.op, (vy_ast.Div, vy_ast.Mod))
+            and isinstance(node.right, vy_ast.Num)
+            and not node.right.value
+        ):
+            raise ZeroDivisionException(f"{node.op.description} by zero", node)
+
         return _validate_op(node, types_list, "validate_numeric_op")
 
     def types_from_BoolOp(self, node):
--- a/vyper/parser/expr.py
+++ b/vyper/parser/expr.py
@@ -241,8 +241,6 @@ class Expr(object):
                 [obj],
                 typ=BaseType(typ, is_literal=True),
                 pos=getpos(self.expr))
-        elif self.context.constants.ast_is_constant(self.expr):
-            return self.context.constants.get_constant(self.expr.id, self.context)
         else:
             raise VariableDeclarationException(f"Undeclared variable: {self.expr.id}", self.expr)
 
--- a/vyper/parser/expr.py
+++ b/vyper/parser/expr.py
@@ -1,14 +1,9 @@
-import warnings
-
 from vyper import ast as vy_ast
 from vyper.exceptions import (
     EvmVersionException,
-    InvalidLiteral,
-    NonPayableViolation,
     StructureException,
+    TypeCheckFailure,
     TypeMismatch,
-    VariableDeclarationException,
-    ZeroDivisionException,
 )
 from vyper.opcodes import version_check
 from vyper.parser import external_call, self_call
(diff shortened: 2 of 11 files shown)


vyper/context/validation/utils.py
```diff
@@ -100,6 +101,14 @@ def types_from_Attribute(self, node):
     def types_from_BinOp(self, node):
         # binary operation: `x + y`
         types_list = get_common_types(node.left, node.right)
+
+        if (
+            isinstance(node.op, (vy_ast.Div, vy_ast.Mod))
+            and isinstance(node.right, vy_ast.Num)
+            and not node.right.value
```
fubuloubu, line 108: Why do a zero check like this?

vyper/parser/expr.py
```diff
@@ -122,32 +103,21 @@ def integer(self):
...
-        # if not SizeLimits.MINDECIMAL * den <= num <= SizeLimits.MAXDECIMAL * den:
         if not (SizeLimits.MINNUM * den <= num <= SizeLimits.MAXNUM * den):
-            raise InvalidLiteral("Number out of range: " + numstring, self.expr)
+            return
         if DECIMAL_DIVISOR % den:
-            raise InvalidLiteral(
-                "Type 'decimal' has maximum 10 decimal places",
-                self.expr
-            )
+            return
```
fubuloubu, line 111: Why return `None`?

vyper/parser/expr.py
```diff
@@ -122,32 +103,21 @@ def integer(self):
...
+    def parse_Hex(self):
         orignum = self.expr.value
-        if len(orignum) == 42:
-            if checksum_encode(orignum) != orignum:
-                raise InvalidLiteral(
-                    "Address checksum mismatch. If you are sure this is the "
-                    f"right address, the correct checksummed form is: {checksum_encode(orignum)}",
-                    self.expr
-                )
+        if len(orignum) == 42 and checksum_encode(orignum) == orignum:
```
fubuloubu, line 120: Why no else case?

vyper/parser/expr.py
```diff
@@ -383,56 +311,35 @@ def attribute(self):
...
+    def parse_BinOp(self):
         left = Expr.parse_value_expr(self.expr.left, self.context)
         right = Expr.parse_value_expr(self.expr.right, self.context)
 
         if not is_numeric_type(left.typ) or not is_numeric_type(right.typ):
-            raise TypeMismatch(
-                f"Unsupported types for arithmetic op: {left.typ} {right.typ}",
-                self.expr,
-            )
+            return
```
fubuloubu, line 342: Why return `None`?

vyper/parser/expr.py
```diff
@@ -572,13 +461,11 @@ def arithmetic(self):
             elif ltyp == rtyp == 'int128':
                 arith = ['exp', 'l', 'r']
 
-            else:
-                raise TypeMismatch('Only whole number exponents are supported', self.expr)
-        else:
-            raise StructureException(f"Unsupported binary operator: {self.expr.op}", self.expr)
+        print(arith)
```
fubuloubu, line 464: lol

vyper/parser/parser_utils.py
```diff
@@ -37,6 +36,17 @@
...
+
+    def _wrapped(*args, **kwargs):
+        return_value = fn(*args, **kwargs)
+        if return_value is None:
+            raise TypeCheckFailure(f"{fn.__name__} did not return a value")
+        return return_value
+
+    return _wrapped
+
+
```
fubuloubu, line 49: I definitely don't like this, very magical

fubuloubu: If it's only temporary, it's okay I suppose

vyper/parser/expr.py
iamdefinitelyahuman: ooops

vyper/context/validation/utils.py
iamdefinitelyahuman: This handles cases where the left side isn't a literal value, so it wasn't caught during folding.

vyper/parser/parser_utils.py
iamdefinitelyahuman: All of it is temporary, I hope :)

_"All of this is a stop-gap on the way to a full refactor of the `parser/` module - we just need to make sure that if the type check layer misses something, the same exception isn't raised in `parser/` so that we can address the error."_

vyper/parser/expr.py
iamdefinitelyahuman: I'm using a pattern where a check is made on the return value, and if `None`, an internal error is raised. This reduces the number of unreachable lines of code while still providing a layer of safety in case type checking is insufficient.

PR #2024, opened 2020-06-23
Fix TypeError in do_GET

andrelfpinto:
Fix 'TypeError: can only concatenate str (not "bytes") to str'  in do_GET. Encoding should be after concatenation.


fubuloubu:
Thanks!

Issue #1966, opened 2020-05-13
No compilation error raised when Map in Struct


henry-hz:
### Version Information

* vyper Version (output of `vyper --version`): 0.1.0b17+commit.eaa6754
* OS: Debian GNU/Linux 10 (buster)
* Python Version - 3.6.2
* Environment (output of `pip freeze`):

asttokens==2.0.3
pkg-resources==0.0.0
pycryptodome==3.9.7
six==1.14.0
vyper==0.1.0b17

### What's your issue about?

Can't populate a single struct without adding an inside mapping when using a struct declaration variable inside a function

```
struct Teddy:
    id: uint256
    year: uint256
    component: map(uint256, uint256)

teddies: map(uint256, Teddy)

@public
def __init__():
    self.teddies[0].id = 0
    self.teddies[0].year = 2019

@public
def create_teddy(teddy_name: string[20]):
  teddy: Teddy = Teddy({
        id: 3,
        year: 1922
      })
```

output:
```
vyper.exceptions.InvalidType: line 4:15 No mappings allowed for in-memory types, only fixed-size arrays
     3     year: uint256
---> 4     component: map(uint256, uint256)
----------------------^
```

### How can it be fixed?

tip: strange, cleaning the function 'create_teddy' you just compile it... the context is affecting how the compiler 'understand' the struct declaration. I also tried to:

```
 self.teddies[0] = Teddy({
        id: 3,
        year: 1922
      })
```

but got a 
```
vyper.exceptions.TypeMismatch: line 15:2 Keys don't match for structs, missing component
```

fubuloubu:
I'm not sure I understand the issue.

It is true that we currently disallow a struct from containing an inner mapping, so the `vyper.exceptions.InvalidType` error makes sense. It is also true that you cannot initialize a struct without all of it's members present, so the `vyper.exceptions.TypeMismatch` also makes sense.

Is your bug about the relative ordering of these two issues, and how they change when you add a new function? Ordering of error reporting is not guaranteed if you change the contents of the source file, certain checks happen in different stages so one may occur earlier than another check.

henry-hz:
Thanks for the answer,
interesting, seems that it's not clear in the docs, and people think that inner mapping is okay:
https://ethereum.stackexchange.com/questions/74467/vyper-mapping-inside-a-struct

In fact I do compile without all members, the error only appear when more conditions are in the same time:
1. struct with inner map
2. populate the struct using MyStruct({bla: bla})

[...]

PR #2027, opened 2020-06-23
Location and constancy logic

iamdefinitelyahuman:
### What I did
* Add `location` attribute to type objects, represent locations via an `Enum`
* Prevent use of mappings within structs (related issue: #1966)
* Constancy checks in `for` loops
* Remove more dead code in `parser/`

### How to verify it
Run tests.

### Cute Animal Picture
image

### DIFF
--- a/vyper/parser/stmt.py
+++ b/vyper/parser/stmt.py
@@ -1,10 +1,6 @@
 from vyper import ast as vy_ast
 from vyper.codegen.return_ import gen_tuple_return, make_return_stmt
-from vyper.exceptions import (
-    ConstancyViolation,
-    StructureException,
-    TypeCheckFailure,
-)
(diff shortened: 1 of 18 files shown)


vyper/parser/stmt.py
```diff
@@ -606,24 +602,21 @@ def _get_target(self, target):
                 raise_exception = True
```
fubuloubu, line 602: `list_name = target.value.id` above is no longer in use

fubuloubu: https://lgtm.com/projects/g/vyperlang/vyper/rev/pr-d5ea08f0db8196c51f8ab23dbe142eb2f58a1db5

iamdefinitelyahuman: ooops

Issue #1707, opened 2019-11-09
Array being iterated over can be modified in another function


smarx:
Modifying the array directly in the loop body is disallowed, so this could be considered a bug:

```python
x: uint256[1]

@private
def store_in_x(index: uint256, value: uint256):
    self.x[index] = value

@public
def foo():
    for n in self.x:
        # self.x[0] = n * 2 # blocked with "Altering list 'self.x' which is being iterated!"
        self.store_in_x(0, n * 2) # BUG: allowed
```

iamdefinitelyahuman:
Fixed in #2028

PR #2028, opened 2020-06-24
Circular call logic and exception annotations

iamdefinitelyahuman:
Fixes: #1707 

### What I did
* allow exceptions to be annotated with multiple nodes
* add circular call check
* check for modificaiton of iterated array in for loop, including via function calls
* misc fixes to interfaces and events
 
Sooooo close....

### How to verify it
Run tests. Read code.

### Cute Animal Picture
image


vyper/exceptions.py
```diff
@@ -31,68 +32,79 @@ class VyperException(Exception):
...
         self.lineno = None
         self.col_offset = None
 
-        if isinstance(item, tuple):
-            self.lineno, self.col_offset = item[:2]
-        elif hasattr(item, 'lineno'):
-            self.lineno = item.lineno
-            self.col_offset = item.col_offset
-            self.source_code = item.full_source_code
+        if len(items) == 1 and isinstance(items[0], tuple):
```
fubuloubu, line 54: I think instead of allowing a tuple, you should have to apply it via `*tuple_arg` now, then we can remove this `if-else` case

vyper/signatures/event_signature.py
```diff
@@ -69,13 +69,12 @@ def from_declaration(cls, code, global_ctx):
                     raise EventDeclarationException("Indexed arguments are limited to 32 bytes")
                 if topics_count > 4:
                     raise EventDeclarationException(
-                        f"Maximum of 3 topics {topics_count - 1} given",
-                        arg,
+                        f"Maximum of 3 topics {topics_count - 1} given", arg_item,
```
fubuloubu, line 72: Isn't this being triggered in another place in this PR?

fubuloubu:
This solves #1707 right?

vyper/exceptions.py
```diff
@@ -31,68 +32,79 @@ class VyperException(Exception):
...
         Exception initializer.
 
         Arguments
         ---------
         message : str
             Error message to display with the exception.
-        item : VyperNode | tuple, optional
-            Vyper ast node or tuple of (lineno, col_offset) indicating where
-            the exception occured.
+        *items : VyperNode | tuple, optional
```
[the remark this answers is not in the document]
iamdefinitelyahuman: https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_numpy.html#example-numpy-style-python-docstrings

vyper/exceptions.py
iamdefinitelyahuman: Have to preserve this to keep the old API..  we can phase it out once we've refactored away more of `parser`

iamdefinitelyahuman:
> This solves #1707 right?

Yes :)))

vyper/signatures/event_signature.py
iamdefinitelyahuman: Ahh yeah.  I should be raising internal exceptions here.  will modify.

fubuloubu: I mean, you're doing the same check twice, why is that?

vyper/context/types/event.py
```diff
@@ -41,13 +41,17 @@ def from_annotation(
...
         validate_call_args(node, 1)
         if not isinstance(node.args[0], vy_ast.Dict):
             raise StructureException("Invalid event declaration syntax", node.args[0])
         for key, value in zip(node.args[0].keys, node.args[0].values):
             if isinstance(value, vy_ast.Call) and value.get("func.id") == "indexed":
                 validate_call_args(value, 1)
+                if indexed.count(True) == 3:
+                    raise EventDeclarationException(
+                        "Event cannot have more than three indexed arguments", value
+                    )
```
fubuloubu, line 54: Here is the other check

vyper/signatures/event_signature.py
iamdefinitelyahuman: Temporary sanity check while type checking pass is new and old code hasn't been refactored.

PR #2032, opened 2020-06-25
Adjust return type of len

### LINKED ISSUE #1979: VIP: len return type should be uint256
## Simple Summary
The builtin `len` function should return `uint256`.

## Motivation
It is impossible to have a negative length. As such, `uint256` is the type I would expect to receive from this function. 

## Specification

```python
len(b: bytes) → uint256
```

## Backwards Compatibility
It's a small but breaking change.

## Copyright

iamdefinitelyahuman:
### What I did
Modify builtin `len` to return `uint256` instead of `int128`

Closes #1979 

### How I did it
* Adjust the return value in `vyper/functions/functions.py`
* Update test cases
* Update documentation

In the documentation, the example format I used is one I'd like to apply to all the builtins before the next release ships.

### How to verify it
Run the tests, read the docs.

### Cute Animal Picture
image

### DIFF
--- a/docs/built-in-functions.rst
+++ b/docs/built-in-functions.rst
@@ -56,9 +56,21 @@ contracts.
     * ``start``: start position of the slice
     * ``length``: length of the slice
 
-.. py:function:: len(b: bytes) -> int128
+.. py:function:: len(b: Union[bytes, string]) -> uint256
 
-    Returns the length of a given ``bytes`` list.
+    Return the length of a given ``bytes`` or ``string``.
+
+    .. code-block:: python
+
+        @public
+        @constant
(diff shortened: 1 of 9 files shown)


docs/built-in-functions.rst
```diff
@@ -56,9 +56,21 @@ contracts.
...
 
-    Returns the length of a given ``bytes`` list.
+    Return the length of a given ``bytes`` or ``string``.
+
+    .. code-block:: python
+
+        @public
+        @constant
+        def foo(b: string[32]) -> uint256:
+            return len(b)
```
fubuloubu, line 68:
```suggestion
        def foo(s: string[32]) -> uint256:
            return len(s)
```

PR #2033, opened 2020-06-25
Use uint256 in slice

### LINKED ISSUE #1986: VIP: start and length args in `slice` should be uint256
## Simple Summary
The `start` and `length` args in the builtin function `slice` should be expected as `uint256`.

## Motivation
We do not use negative values for indexing, so a negative start index or length already won't work. There is no need for `int128` here.

## Specification
Change the function signature to:

```python
slice(b: bytes, start: uint256, length: uint256) → bytes
```

## Backwards Compatibility
This is a breaking change.

## Copyright

iamdefinitelyahuman:
### What I did
Adjust input arguments for builtin function `slice` to use `uint256` instead of `int128`.

Closes #1986 

### How I did it
* Adjust the return value in `vyper/functions/functions.py`
* Update test cases
* Update documentation

### How to verify it
Run tests, read docs.

### Cute Animal Picture
image

### DIFF
--- a/docs/built-in-functions.rst
+++ b/docs/built-in-functions.rst
@@ -48,13 +48,27 @@ contracts.
     * ``value``: Value for the ether unit
     * ``unit``: Ether unit name (e.g. ``"wei"``, ``"ether"``, ``"gwei"``, etc.) indicating the denomination of ``value``.
 
-.. py:function:: slice(b: bytes, start: int128, length: int128) -> bytes
+.. py:function:: slice(b: Union[bytes, bytes32, string], start: uint256, length: uint256) -> Union[bytes, string]
 
-    Copies a list of bytes and returns a specified slice.
+    Copy a list of bytes and return a specified slice.
 
-    * ``b``: ``bytes`` or ``bytes32`` to be sliced
+    * ``b``: value being sliced
     * ``start``: start position of the slice
     * ``length``: length of the slice
+
+    If the value being sliced is a ``bytes`` or ``bytes32``, the return type is ``bytes``.  If it is a ``string``, the return type is ``string``.
+
+    .. code-block:: python
+
+        @public
+        @constant
+        def foo(s: string[32]) -> string[5]:
+            return slice(s, 4, 5)
+
+    .. code-block:: python
+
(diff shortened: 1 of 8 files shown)


docs/built-in-functions.rst
```diff
@@ -48,14 +48,28 @@ contracts.
     * ``value``: Value for the ether unit
     * ``unit``: Ether unit name (e.g. ``"wei"``, ``"ether"``, ``"gwei"``, etc.) indicating the denomination of ``value``.
 
-.. py:function:: slice(b: bytes, start: int128, length: int128) -> bytes
+.. py:function:: slice(b: Union[bytes, bytes32, string], start: uint256, length: uint256) -> Union[bytes, string]
```
fubuloubu, line 51: Does it make sense to be able to do a slice of a `bytes32`?

tests/parser/functions/test_slice.py
```diff
@@ -31,7 +31,7 @@ def test_test_slice2(get_contract_with_gas_estimation):
 def slice_tower_test(inp1: bytes[50]) -> bytes[50]:
     inp: bytes[50] = inp1
     for i in range(1, 11):
-        inp = slice(inp, 1, 30 - i * 2)
+        inp = slice(inp, 1, convert(30 - i * 2, uint256))
```
fubuloubu, line 34: The convert shouldn't be necessary, what's going wrong in this statement?

iamdefinitelyahuman: It's because logic in `parser` still implicitly casts `i` as an `int128`.  In the future the `convert` won't be necessary but today it is :(

docs/built-in-functions.rst
iamdefinitelyahuman: Yeah, I've personally used it when rolling my own tight storage packing logic.

fubuloubu: ehhhhhh....

tests/parser/functions/test_slice.py
fubuloubu: Okay gotcha, leave a TODO for that?

PR #2044, opened 2020-06-25
Modify event declaration syntax

### LINKED ISSUE #1864: VIP: Modify event declaration syntax to resemble that of struct
## Simple Summary
Change this:

```python
Transfer: event({_from: indexed(address), _to: indexed(address), _value: uint256})
```

to this:

```python
event Transfer:
    _from: indexed(address)
    _to: indexed(address)
    _value: uint256
```

## Motivation
The current approach to declaring an event reads as though there is a variable being created, with it's type defined from the result of a call to an `event` function.  This is misleading because:

1. The created event is not a variable. You cannot assign to it during declaration, or reference it except as a callable member of `log`.
2. `event`, though appearing to be a function, is not a function. And type annotations are not typically declared as the result of a function call.
3. The reliance on the ordering of dictionary keys inside `event` is not a standard python behavior until version 3.7. Declaring the fields in this way does not intuitively read as though they are sequenced.

By instead using the struct-style declaration, we make it more clear that this not a simple variable, and that the members inside are sequenced.

## Backwards Compatibility
This will be a rather large breaking change, we should definitely discuss it at length. If it is approved, it probably belongs in the `1.0rc` release.

## Copyright

iamdefinitelyahuman:
### What I did
Modify event syntax.

* declaration uses `class` structure, similar to `struct` and `interface`.
* `log` is now a statement

Closes #1864 

### How I did it
* Modified `nodes/pre_parser.py` to allow statement substitution.  I chose `yield` for the substitution because I think it unlikely we will ever have a use for this statement in Vyper, and it accepts an expression following the statement.  As a side effect, this paves the way to introducing other custom statements if we wish.
* Added `event` as a possible substitution for `class`
* Modified `nodes/annotation.py` to convert the `Yield` node into the new Vyper `Log` node.
* Adjusted the type check and parser logic for events.  Declaration is very similar to how `struct` is handled, emitting an event is a function call.

### How to verify it
Run the tests. I had to make a lot of changes here.

### TODO
- [ ] fix failing tests
- [ ] update documentation

### Cute Animal Picture


vyper/context/types/meta/event.py
```diff
@@ -0,0 +1,84 @@
...
+    NamespaceCollision,
+    StructureException,
+)
+
+
+class Event:
+    """
+    Event type.
+
+    Event are special types that exist as members of the builtin `log` object.
```
fubuloubu, line 20: This isn't true, is it?

vyper/context/types/meta/event.py
```diff
@@ -0,0 +1,84 @@
...
+
+    Attributes
+    ----------
+    indexed : list
+        A list of booleans indicating if each argument within the event is
+        indexed.
+    """
```
fubuloubu, line 27: Perhaps this should be a dict instead? Makes for easier lookup

vyper/context/types/meta/event.py
```diff
@@ -0,0 +1,84 @@
...
+    ---------
+    node : ClassDef
+        Vyper ast node defining the event
+    Returns
+    -------
+    Event
+    """
+
+    members: OrderedDict = OrderedDict()
+    indexed: List = []
```
fubuloubu, line 56: `OrderedDict[name, is_indexed]` instead?

vyper/context/types/meta/event.py
```diff
@@ -0,0 +1,84 @@
...
+        member_name = node.target.id
+        if member_name in members:
+            raise NamespaceCollision(
+                f"Event member '{member_name}' has already been declared", node.target
+            )
+
+        annotation = node.annotation
+        if isinstance(annotation, vy_ast.Call) and annotation.get("func.id") == "indexed":
+            validate_call_args(annotation, 1)
+            if indexed.count(True) == 3:
```
fubuloubu, line 73: ah, I see why you went for a List here...

vyper/context/validation/module.py
```diff
@@ -170,9 +170,13 @@ def visit_AnnAssign(self, node):
             raise exc.with_annotation(node) from None
 
     def visit_ClassDef(self, node):
-        type_ = self.namespace[node.class_type].build_primitive_from_node(node)
+        obj = self.namespace[node.class_type].build_primitive_from_node(node)
         try:
-            self.namespace[node.name] = type_
+            if hasattr(obj, "_member_of"):
+                member_key = obj._member_of
+                self.namespace[member_key].add_member(node.name, obj)
```
fubuloubu, line 177: What's going on here?

tests/grammar/vyper.lark
```diff
@@ -133,7 +136,7 @@ break_stmt: _BREAK
 continue_stmt: _CONTINUE
 
 // TODO: Remove "." as unnecessary
```
fubuloubu, line 138: Delete this line too lol

tests/grammar/vyper.lark
```diff
@@ -59,10 +59,13 @@ returns: _RETURN_TYPE type
...
+// Events can be composed of 1+ types or "pass"
+_EVENT_DECL: "event"
+event_member: NAME ":" type
 indexed_event_arg: NAME ":" "indexed" "(" type ")"
-// Events can have zero args
-event_def: NAME ":" "event" "(" "{" (variable | indexed_event_arg)? ("," (variable | indexed_event_arg))* "}" ")"
-
+event_body: _NEWLINE _INDENT ((variable | indexed_event_arg) _NEWLINE)+ _DEDENT
+event_pass_body: "pass"
+event_def: _EVENT_DECL NAME ":" ( event_body | event_pass_body )
```
fubuloubu, line 68: `_PASS: "pass"` is already defined below, so change this too:
```suggestion
event_def: _EVENT_DECL NAME ":" ( event_body | _PASS )
```

tests/parser/exceptions/test_structure_exception.py
```diff
@@ -46,11 +46,6 @@ def foo() -> int128:
     """
 interface F:
     def foo(): constant
-struct S:
-    x: int128
-    """,
-    """
-g: int128
 struct S:
     x: int128
```
fubuloubu, line 50: why'd you remove this case?

tests/parser/features/external_contracts/test_external_contract_calls.py
```diff
@@ -341,40 +341,6 @@ def bar(arg1: address, arg2: int128) -> int128:
     assert_tx_failed(lambda: get_contract(contract), exception=UnknownType)
 
 
-def test_external_contracts_must_be_declared_first_1(assert_tx_failed, get_contract):
```
fubuloubu, line 344: So, this PR also modifies the assumption of ordering?

vyper/exceptions.py
```diff
@@ -116,6 +120,7 @@ def __init__(self, message, source_code, lineno, col_offset):
         item.lineno = lineno
         item.col_offset = col_offset
         item.full_source_code = source_code
+        print("got this far")
```
fubuloubu, line 123:
```suggestion
```

iamdefinitelyahuman: This is the real reason for code reviews.

tests/grammar/vyper.lark
```diff
@@ -59,10 +59,12 @@ returns: _RETURN_TYPE type
...
 
+// Events can be composed of 1+ types or "pass"
+_EVENT_DECL: "event"
+event_member: NAME ":" type
 indexed_event_arg: NAME ":" "indexed" "(" type ")"
-// Events can have zero args
-event_def: NAME ":" "event" "(" "{" (variable | indexed_event_arg)? ("," (variable | indexed_event_arg))* "}" ")"
-
+event_body: _NEWLINE _INDENT ((variable | indexed_event_arg) _NEWLINE)+ _DEDENT
+event_def: _EVENT_DECL NAME ":" ( event_body | _PASS )
```
fubuloubu, line 67: Add a comment about "Events with no args use pass statement instead" above `event_def`

docs/logging.rst
```diff
@@ -12,20 +12,27 @@ This example is taken from the `sample ERC20 contract <https://github.com/vyperl
...
        ... Logic here to do the real work ...
 
        # All done, log the event for listeners
-       log.Transfer(msg.sender, _to, _amount)
+       log Transfer(msg.sender, _to, _value)
 
 Let's look at what this is doing. First, we declare two event types to log. The two events are similar in that they contain
 two indexed address fields. Indexed fields do not make up part of the event data itself, but can be searched by clients that
-want to catch the event. Also, each event contains one single data field, in each case called ``_value``. Events can contain several arguments with any names desired.
+want to catch the event. Also, each event contains one single data field, in each case called ``value``. Events can contain several arguments with any names desired.
```
fubuloubu, line 35: Name doesn't match above

tests/grammar/vyper.lark
```diff
@@ -59,10 +59,13 @@ returns: _RETURN_TYPE type
 function_sig: _FUNC_DECL NAME "(" [parameters] ")" [returns]
 function_def: [decorators] function_sig ":" body
 
+// Events can be composed of 1+ types
```
fubuloubu, line 62: "Events can have 0 or more members"

docs/logging.rst
iamdefinitelyahuman: The name in the event is `value`, the local memory variable is `_value`.

Issue #1969, opened 2020-05-13
VIP: Change Mapping Syntax (again)


fubuloubu:
## Simple Summary
Change mapping syntax from `var: map(key_type, val_type)` to `var: HashMap[key_type, val_type]`

## Motivation
Mapping syntax was changed in #564. While it made for easier understanding of how mappings work in practice, it precludes the use of the Python standard function `map(fn, iter) -> iter` for anything in Vyper to use. We propose adopting a MyPy-like type structure, which will make working with mappings (and nested mappings) more apparent, and open up the keyword `map` to future use as an iterator operation.

## Specification
The syntax for mappings would change from `var: map(key_type, val_type)` to `var: HashMap[key_type, val_type]`. This looks like the following in practice:
```python
struct MyStruct: ...

my_mapping: HashMap[uint256, MyStruct]

my_nested_map: HashMap[bytes32, HashMap[uint256, address]]

@public
def foo():
    self.my_mapping[1] = MyStruct({...})  # Same access as before
    self.my_nested_map[blk.hash][1] = msg.sender  # Same access as before
```

## Backwards Compatibility
This VIP will break the current syntax for defining mappings

## Dependencies
Depends on #564

## References
No references

## Copyright
Copyright and related rights waived via [CC0](https://creativecommons.org/publicdomain/zero/1.0/)

howardpen9:
![image](https://user-images.githubusercontent.com/36764495/92538151-87fd3a00-f270-11ea-9ecb-6e98d3bb3302.png)

howardpen9:
ok, then I know the reason why the code on this Sep 2019's paper couldn't work...

Thanks for the updated.

PR #2045, opened 2020-06-25
Update Map Syntax

fubuloubu:
Closes: #1969

Changed map syntax from `map(...)` to `HashMap[...]`

### Cute Animal Picture
weeeeee

### DIFF
--- a/vyper/context/types/indexable/mapping.py
+++ b/vyper/context/types/indexable/mapping.py
@@ -8,7 +8,7 @@ from vyper.context.types.bases import (
 )
 from vyper.context.types.utils import get_type_from_annotation
 from vyper.context.validation.utils import validate_expected_type
-from vyper.exceptions import CompilerPanic, StructureException
+from vyper.exceptions import StructureException
 
 
 class MappingDefinition(IndexableTypeDefinition):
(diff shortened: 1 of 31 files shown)


vyper/context/types/indexable/mapping.py
```diff
@@ -28,29 +27,32 @@ def get_index_type(self, node):
...
         location: DataLocation = DataLocation.UNSET,
         is_constant: bool = False,
         is_public: bool = False,
     ) -> MappingDefinition:
-        if isinstance(node, vy_ast.Name):
-            raise CompilerPanic("Node must be a call")
+        if not isinstance(node, vy_ast.Subscript):
+            raise CompilerPanic("Node must be a subscript")
+        if not isinstance(node.slice.value, vy_ast.Tuple):
+            raise CompilerPanic("Mapping must be a subscript with a Tuple Index")
```
iamdefinitelyahuman, line 44: This and the above shouldn't be `CompilerPanic` - they're easily reachable by the user.  I think `StructureException` is the most fitting.

iamdefinitelyahuman: Also, reference to `Tuple` isn't clear imo.  The average user has no idea that the 2 numbers are represented as a tuple in the AST. It should just be something like "subscript with 2 values"

fubuloubu: See if you like how I solved this in a887410

iamdefinitelyahuman: It's good!

Issue #2040, opened 2020-06-25
VIP: Rename @constant decorator to @view


fubuloubu:
## Simple Summary
Change the `@constant` decorator to `@view` for methods

## Motivation
To be consistent with the ABI `stateMutability` field, we should use the term `view` instead of `constant` when decorating a method

## Specification
Contracts and Vyper interfaces should be specified using the following syntax to denote a method that does not modify state:
```python
@view
@public
def getSender() -> address:
    return msg.sender
```

## Backwards Compatibility
Breaks backwards compatibility because of a name change.

## Dependencies
No dependencies

## References
-  #1931
- https://solidity.readthedocs.io/en/latest/abi-spec.html#json

## Copyright
Copyright and related rights waived via [CC0](https://creativecommons.org/publicdomain/zero/1.0/)

iamdefinitelyahuman:
Violent agreement!

PR #2049, opened 2020-06-26
Rename @constant decorator to @view

fubuloubu:
Fixes: #2040

### What I did
- `@constant` decorator is now `@view`
- `constant` state mutability field on external interfaces is now `view`
- handle ABIv2 `stateMutability` field in JSON ABI (as well as ABIv1 `payable`/`constant`)

Rename constant state mutability descriptor and decorator to view

### Cute Animal Picture
yay


vyper/context/types/function.py
```diff
@@ -167,8 +180,10 @@ def from_FunctionDef(
...
                         )
                     kwargs["is_public"] = bool(decorator.id == "public")
-                elif decorator.id in ("constant", "payable"):
-                    kwargs[f"is_{decorator.id}"] = True
+                elif decorator.id == "payable":
+                    kwargs["is_payable"] = True
+                elif decorator.id == "view":
+                    kwargs["is_constant"] = True
                 else:
                     raise FunctionDeclarationException(
```
iamdefinitelyahuman, line 188: It might make sense to temporarily add a special exception here for when the decorator is `@constant`, to let people know about the change.

fubuloubu: Fixed in 0c5023b

Issue #2041, opened 2020-06-25
VIP: Support @pure decorator


fubuloubu:
## Simple Summary
Support the use of `@pure` as a decorator for methods

## Motivation
To be consistent with the ABI `stateMutability` field, we should add support for specifying a method as `@pure`.

## Specification
A method that is decorated as `@pure` must not read any blockchain state such as storage variables, environment variables, querying contract code or account balances, or making a call to non`pure` methods, or to other contracts. In the event this happens, this should raise an exception that uniquely details a violation of reading blockchain state. Such a function might be useful for complex math or other purposes.

A method decorated with `@pure` looks like:
```python
@pure
@public
def getMaff() -> uint256:
    return 1 + 2
```

## Backwards Compatibility
No backwards compatibility issues

## Dependencies
No dependencies. Related to #2040

## References
-  #1931
- https://solidity.readthedocs.io/en/latest/abi-spec.html#json

## Copyright
Copyright and related rights waived via [CC0](https://creativecommons.org/publicdomain/zero/1.0/)

fubuloubu:
Remove this line in `docs/index.rst`:
Anything marked constant is not allowed to change the state

PR #2052, opened 2020-06-26
Add Support for Pure function mutability

fubuloubu:
Closes: #2041

### What I did
- [x] Complete refactor of function visibility handling
- [x] Complete refactor of function state mutability handling
- [x] Changed `is_constant` keyword to `is_immutable` on `BaseTypeDefinition` and all subclasses
- [x] Added additional logical verification around multiple mutability decorators (possible rendering a later check rendundant?)
- [x] Add test cases
- [x] Update Documentation

### How I did it
grepping and enums

### How to verify it
Look at new test cases

Add Support for Pure function mutability

### Cute Animal Picture
happy dog


vyper/context/types/function.py
```diff
@@ -26,6 +27,34 @@
...
+    PAYABLE = StringEnum.auto()
+
+    @classmethod
+    def from_abi(cls, abi_dict) -> "StateMutability":
+        """
+        Extract stateMutability from an entry in a contract's ABI
+        """
+        if "stateMutability" in abi_dict:
+            return cls(abi_dict["stateMutability"])
+        elif "payable" in abi_dict and abi_dict["payable"]:
```
iamdefinitelyahuman, line 48:
```suggestion
        elif abi_dict.get("payable"):
```

could do this for constant as well..  bit less verbose.

vyper/context/types/function.py
```diff
@@ -97,44 +131,37 @@ def from_abi(cls, abi: Dict) -> "ContractFunctionType":
...
+            function_visibility=FunctionVisibility.PUBLIC,
+            state_mutability=StateMutability.from_abi(abi),
+        )
 
     @classmethod
     def from_FunctionDef(
         cls,
         node: vy_ast.FunctionDef,
-        is_constant: Optional[bool] = None,
+        is_immutable: Optional[bool] = None,
```
fubuloubu, line 164: May have to change this back?

vyper/context/types/function.py
```diff
@@ -157,15 +184,24 @@ def from_FunctionDef(
...
 
         # decorators
         for decorator in node.decorator_list:
 
             if isinstance(decorator, vy_ast.Call):
+                if "nonreentrant" in kwargs:
+                    raise StructureException(
+                        f"nonreentrant decorator is already set with key: {kwargs['nonreentrant']}",
+                        node,
+                    )
```
fubuloubu, line 204: I think this overrides a later check

vyper/context/types/function.py
```diff
@@ -175,16 +211,20 @@ def from_FunctionDef(
...
-                    kwargs["is_payable"] = True
-                elif decorator.id == "view":
-                    kwargs["is_constant"] = True
+                    kwargs["function_visibility"] = FunctionVisibility(decorator.id)
+
+                elif StateMutability.is_valid_value(decorator.id):
+                    if "state_mutability" in kwargs:
+                        raise FunctionDeclarationException(
+                            f"Mutability is already set to: {kwargs['state_mutability']}", node
+                        )
```
fubuloubu, line 225: I think this check didn't exist before?

vyper/context/types/function.py
```diff
@@ -199,8 +239,15 @@ def from_FunctionDef(
...
-            raise FunctionDeclarationException("Visibility must be public or private", node)
+        if "function_visibility" not in kwargs:
+            options = " or ".join([v.value for v in list(FunctionVisibility)])
+            raise FunctionDeclarationException(
+                f"Visibility must be set to one of {options}", node,
+            )
+
+        if "state_mutability" not in kwargs:
+            # Assume nonpayable if not set at all (cannot accept Ether, but can modify state)
+            kwargs["state_mutability"] = StateMutability("nonpayable")
```
fubuloubu, line 250: This is now explicitly set always

vyper/context/types/function.py
```diff
@@ -175,16 +211,20 @@ def from_FunctionDef(
                 kwargs["nonreentrant"] = decorator.args[0].value
 
             elif isinstance(decorator, vy_ast.Name):
-                if decorator.id in ("public", "private"):
-                    if "is_public" in kwargs:
+                if FunctionVisibility.is_valid_value(decorator.id):
```
iamdefinitelyahuman, line 214: I like this a lot

vyper/context/types/function.py
```diff
@@ -175,16 +211,20 @@ def from_FunctionDef(
...
+                            f"Visibility is already set to: {kwargs['function_visibility']}", node
                         )
-                    kwargs["is_public"] = bool(decorator.id == "public")
-                elif decorator.id == "payable":
-                    kwargs["is_payable"] = True
-                elif decorator.id == "view":
-                    kwargs["is_constant"] = True
+                    kwargs["function_visibility"] = FunctionVisibility(decorator.id)
+
+                elif StateMutability.is_valid_value(decorator.id):
```
iamdefinitelyahuman, line 221: and again..  soo much cleaner

vyper/context/types/function.py
```diff
@@ -280,13 +327,20 @@ def from_AnnAssign(cls, node: vy_ast.AnnAssign) -> "ContractFunctionType":
...
         for item in arguments:
             args_dict[f"arg{len(args_dict)}"] = item
-        return cls(node.target.id, args_dict, len(arguments), return_type, is_public=True)
+        return cls(
+            node.target.id,
+            args_dict,
+            len(arguments),
+            return_type,
+            function_visibility=FunctionVisibility.PUBLIC,
+            state_mutability=StateMutability.NONPAYABLE,
```
fubuloubu, line 336: Are these the right defaults?

vyper/context/validation/local.py
```diff
@@ -112,15 +116,24 @@ def __init__(
...
-        if not getattr(self.func, "is_payable", False):
+        if self.func.mutability is StateMutability.PURE:
+            node_list = fn_node.get_descendants(
+                vy_ast.Attribute, {"value.id": {"self", "msg", "tx", "block", "chain"}}
+            )
+            if node_list:
+                raise ConstancyViolation(
+                    "not allowed to query contract or environment variables in pure functions",
+                    node_list[0],
+                )
```
fubuloubu, line 135: This check is new for the `pure` decorator

vyper/context/types/function.py
```diff
@@ -199,8 +239,15 @@ def from_FunctionDef(
             else:
                 raise StructureException("Bad decorator syntax", decorator)
 
-        if "is_public" not in kwargs:
-            raise FunctionDeclarationException("Visibility must be public or private", node)
+        if "function_visibility" not in kwargs:
+            options = " or ".join([v.value for v in list(FunctionVisibility)])
```
iamdefinitelyahuman, line 243: join can accept an iterator, no need for the square brackets here

vyper/context/validation/local.py
```diff
@@ -159,7 +172,10 @@ def visit_Assign(self, node):
             raise StructureException("Right-hand side of assignment cannot be a tuple", node.value)
         target = get_exact_type_from_node(node.target)
         validate_expected_type(node.value, target)
-        if self.func.is_constant and target.location == DataLocation.STORAGE:
+        if (
+            self.func.mutability in (StateMutability.PURE, StateMutability.VIEW)
```
fubuloubu, line 176: It would be nicer to do `self.func.mutability < StateMutability.VIEW`

vyper/context/validation/local.py
```diff
@@ -168,7 +184,10 @@ def visit_AugAssign(self, node):
...
```
fubuloubu, line 188: It would be nicer to do `self.func.mutability < StateMutability.VIEW`

vyper/context/validation/local.py
```diff
@@ -350,9 +369,20 @@ def visit_Expr(self, node):
             raise StructureException("To call an event you must use the `log` statement", node)
 
         if isinstance(fn_type, ContractFunctionType):
-            if self.func.is_constant and not fn_type.is_constant:
+            if self.func.mutability is StateMutability.VIEW and fn_type.mutability not in (
+                StateMutability.VIEW,
+                StateMutability.PURE,
+            ):
```
fubuloubu, line 375: It would be nicer to do `self.func.mutability is StateMutability.VIEW and fn_type.mutability > StateMutability.VIEW`

or better yet: `self.func.mutability < fn_type.mutability`, and combine with below

vyper/context/validation/local.py
```diff
@@ -350,9 +369,20 @@ def visit_Expr(self, node):
...
+                raise ConstancyViolation(
+                    "Cannot call a mutating function from a view function", node
```
iamdefinitelyahuman, line 377: the word "mutating" feels a bit out of place here.  it's technically correct, but i think most users would better understand "state modifying" or something similar.

vyper/context/validation/local.py
```diff
@@ -350,9 +369,20 @@ def visit_Expr(self, node):
...
+                )
+
+            if (
+                self.func.mutability is StateMutability.PURE
+                and fn_type.mutability is not StateMutability.PURE
+            ):
                 raise ConstancyViolation(
-                    "Cannot call a non-constant function from a constant function", node
+                    "Cannot call a stateful function from a pure function", node
```
iamdefinitelyahuman, line 385: Ahh..  "stateful" :point_up:  for previous comment.

vyper/context/validation/local.py
fubuloubu: Actually, is it okay that a `nonpayable` function can call a `payable` one?

vyper/context/types/function.py
fubuloubu:
```suggestion
            options = " or ".join(v.value for v in list(FunctionVisibility))
```

vyper/context/validation/local.py
fubuloubu:
```suggestion
                    "Cannot call a state modifying function from a view function", node
```

vyper/context/validation/local.py
fubuloubu: stateful does not mean state modifying though

vyper/context/types/function.py
fubuloubu: Shouldn't it be?
```suggestion
        elif abi_dict.get("payable", False):
```

fubuloubu:
> This pull request **introduces 1 alert** when merging [a23fb48](https://github.com/vyperlang/vyper/commit/a23fb4871025a741b1c04ce6470637c6025c561e) into [1c3809b](https://github.com/vyperlang/vyper/commit/1c3809b3f6df8591e55a1e7ef12e4315329f3fb1) - [view on LGTM.com](https://lgtm.com/projects/g/vyperlang/vyper/rev/pr-182c14c0762329a03faaf0bc9cbcc53caea9b18b)
> 
> **new alerts:**
> 
>     * 1 for First parameter of a method is not named 'self'

Oh yeah, I don't know what the hell `enum.Enum` is doing here, but this is fineeeee

vyper/context/validation/local.py
iamdefinitelyahuman: Yess

vyper/context/types/function.py
iamdefinitelyahuman: :+1:

vyper/context/types/function.py
iamdefinitelyahuman: `get` returns `None` if the key doesn't exist

vyper/context/validation/local.py
iamdefinitelyahuman: Yeah, nonpayable should be able to call payable.  It doesn't mean "can't transfer ETH", just "can't receive ETH".

vyper/context/types/function.py
iamdefinitelyahuman: I don't think so..  previously it was either `@constant` or nothing, so modifiying would be implied..  thus you couldn't decorate it both ways.

vyper/context/types/function.py
iamdefinitelyahuman: I like.

vyper/context/types/function.py
iamdefinitelyahuman: Looks right to me!

vyper/context/validation/local.py
iamdefinitelyahuman: I like it.  I'm not sure if it's actually better, but maybe we could pull that list of objects from `contract/environment.py` so that if the environment variables change in the future, we don't have to update this line as well?

iamdefinitelyahuman: `self` won't be anywhere in `environment.py` because it's special and magic, but I think the others would be available somewhere there.

fubuloubu: Yeah, I was thinking that. Does one exist? or just get the keys of something?

vyper/context/types/function.py
```diff
@@ -199,8 +239,15 @@ def from_FunctionDef(
...
-            raise FunctionDeclarationException("Visibility must be public or private", node)
+        if "function_visibility" not in kwargs:
+            options = " or ".join(v.value for v in list(FunctionVisibility))
+            raise FunctionDeclarationException(
+                f"Visibility must be set to one of {options}", node,
+            )
+
+        if "state_mutability" not in kwargs:
+            # Assume nonpayable if not set at all (cannot accept Ether, but can modify state)
+            kwargs["state_mutability"] = StateMutability("nonpayable")
```
fubuloubu, line 249: Better to use the enum type
```suggestion
            kwargs["state_mutability"] = StateMutability.NONPAYABLE
```

vyper/context/validation/local.py
```diff
@@ -353,11 +377,17 @@ def visit_Expr(self, node):
...
+            if (
+                self.func.mutability is StateMutability.VIEW
+                and fn_type.mutability > StateMutability.VIEW
+            ):
+                raise StateAccessViolation(
+                    "Cannot call a mutating function from a view function", node
                 )
 
+            if self.func.mutability is StateMutability.PURE:
+                raise StateAccessViolation("Cannot call a function from a pure function", node)
```
fubuloubu, line 389: I changed this just to always fail if you try and make a call as `@pure`

vyper/context/types/function.py
fubuloubu: Yes, before if you specified `nonreentrant` multiple times, this would silently update. So, not really a big deal.

fubuloubu:
Need to find a way to ignore LGTM on this issue...

fubuloubu:
Changing `modifying` to `nonpayable`, and supporting `payable` in interface definitions will happen in the last PR

vyper/context/types/utils.py
```diff
@@ -13,13 +14,65 @@
...
+    def values(cls) -> List[str]:
+        return [v.value for v in cls.options()]
+
+    # Comparison operations
+    def __eq__(self, other: object) -> bool:
+        if not isinstance(other, self.__class__):
+            super().__eq__(other)
+        return self is other
+
+    def __ne__(self, other: object) -> bool:
```
iamdefinitelyahuman, line 53: This isn't needed, python will automatically call to `__eq__` and invert the result.

vyper/context/types/utils.py
```diff
@@ -13,13 +14,65 @@
...
+        return list(cls)
+
+    @classmethod
+    def values(cls) -> List[str]:
+        return [v.value for v in cls.options()]
+
+    # Comparison operations
+    def __eq__(self, other: object) -> bool:
+        if not isinstance(other, self.__class__):
+            super().__eq__(other)
```
iamdefinitelyahuman, line 50: should this be returning the call to `super()`  ?  If not, I'm confused what this does.

vyper/context/types/utils.py
```diff
@@ -13,13 +14,65 @@
...
+        return self is other
+
+    def __ne__(self, other: object) -> bool:
+        return not self.__eq__(other)
+
+    def __lt__(self, other: object) -> bool:
+        if not isinstance(other, self.__class__):
+            super().__eq__(other)
```
iamdefinitelyahuman, line 58: same question above re: super

vyper/context/validation/local.py
```diff
@@ -168,17 +192,17 @@ def visit_Assign(self, node):
             raise StructureException("Right-hand side of assignment cannot be a tuple", node.value)
         target = get_exact_type_from_node(node.target)
         validate_expected_type(node.value, target)
-        if self.func.is_constant and target.location == DataLocation.STORAGE:
-            raise ConstancyViolation("Cannot modify storage in a constant function", node)
+        if self.func.mutability <= StateMutability.VIEW and target.location == DataLocation.STORAGE:
+            raise StateAccessViolation("Cannot modify storage in a pure or view function", node)
```
iamdefinitelyahuman, line 196:
```suggestion
            raise StateAccessViolation(f"Cannot modify storage in a {self.func.mutability} function", node)
```

vyper/context/validation/local.py
```diff
@@ -168,17 +192,17 @@ def visit_Assign(self, node):
...
 
     def visit_AugAssign(self, node):
         if isinstance(node.value, vy_ast.Tuple):
             raise StructureException("Right-hand side of assignment cannot be a tuple", node.value)
         target = get_exact_type_from_node(node.target)
         validate_expected_type(node.value, target)
-        if self.func.is_constant and target.location == DataLocation.STORAGE:
-            raise ConstancyViolation("Cannot modify storage in a constant function", node)
+        if self.func.mutability <= StateMutability.VIEW and target.location == DataLocation.STORAGE:
+            raise StateAccessViolation("Cannot modify storage in a pure or view function", node)
```
iamdefinitelyahuman, line 205: same suggestion as above

docs/compiler-exceptions.rst
```diff
@@ -30,9 +30,17 @@ of the error within the code:
 
     Raises when an array index is out of bounds.
 
-.. py:exception:: ConstancyViolation
+.. py:exception:: IteratorException
 
-    Raises when attempting to perform a modifying action within a constant or view-only context. For example, writing to storage in a ``@view`` function or modifying a constant variable.
+   Raises when an iterator is constructed or used incorrectly.
+
+.. py:exception:: StateAccessViolation
```
iamdefinitelyahuman, line 37: Can you move these to maintain the alphabetical sorting? :pray:

vyper/context/types/utils.py
fubuloubu: Can raise an error instead

iamdefinitelyahuman: Is there ever a situation where checking equality raises?  If the check fails it should just always return `False`.

fubuloubu: If the types don't match? Means you used it wrong e.g. `CompilerPanic`

fubuloubu:
SHUT UP BOT

Issue #1702, opened 2019-11-09
Struct getter collisions


smarx:
Due to the way getter names are generated for struct fields, collisions can happen:

```python
struct Test:
    foo: uint256
    __foo: uint256
    
a: public(Test)
a__: public(Test)
```

The error message is `Duplicate function name: a____foo`. This is better than an ambiguity, but perhaps the collision should be avoided altogether by using a separator that cannot appear in identifiers.

The following code is responsible for this naming:

https://github.com/ethereum/vyper/blob/c296b2d7532d913103aad494b749f8179a3acddc/vyper/parser/global_context.py#L230-L240

fubuloubu:
I was thinking about this a while ago, and there's no reason the getter can't be `method_id("a.foo")` instead of `method_id("a__foo")`. That would avoid this scenario because `.` is not a valid identifier character

fubuloubu:
Removed "bug" because this actually would change semantics, although not enough to really be a problem for anyone I don't think since it's only used externally

jacqueswww:
Meeting Notes: Refactor public getters for structs to return the full struct.

fubuloubu:
Note: originally this was done this way because we didn't have ABI handling for structs yet.

PR #2064, opened 2020-06-28
Update Struct public getter to return whole Struct

fubuloubu:
Closes: #1702

### What I did
We had some funky handling prior for struct members because of a bug in web3py. This PR makes the correct change to support returning structs from public getters directly, making it compatible with how Solidity handles it.

### How I did it
Just returning the struct name as a type directly instead of iterating over it's members

### How to verify it
I updated the tests to use the tuple sequence returned from the public getter. Also note the xfail on the one test case that couldn't be handled by web3py's current configuration. We can remove this once web3py releases a fix, or we change to using Brownie

Public getters for structs now return the whole struct

### Cute Animal Picture
exhausted

### DIFF
--- a/tests/examples/voting/test_ballot.py
+++ b/tests/examples/voting/test_ballot.py
@@ -16,33 +16,33 @@ def test_initial_state(w3, c):
     # Check chairperson is msg.sender
     assert c.chairperson() == a0
     # Check propsal names are correct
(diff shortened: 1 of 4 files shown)


fubuloubu:
@iamdefinitelyahuman don't jynx me broh

PR #2069, opened 2020-06-29
Add short circuiting for boolean operations

### LINKED ISSUE #1817: VIP: Implement short circuiting
## Simple Summary
Add short circuiting logic to binary operations.

## Motivation
1. This is pretty standard behavior in most programming languages - I believe the majority of users will assume that it already happens and write their code accordingly.
2. By not doing it, we make transactions less efficient through unnecessary evaluation.
3. We introduce a risk for exploits due to unexpected behavior:

```python
@private
def a() -> bool:
    return True

@private
def b() -> bool:
    steal_the_money()
    return True

@public
def foo():
    if a() or b():
        # something mostly harmless
```

## Specification
Treat `and` and `or` as short-circuit operators: evaluate them left to right, and stop evaluation as soon as an outcome is determined.

## Backwards Compatibility
This will alter behavior in contracts, but should not introduce any breaking changes.

## Copyright

iamdefinitelyahuman:
### What I did
Implement short circuiting for boolean operations.  Closes #1817 

### How I did it
Modifications to `parser/expr.py`:

1. Place a label at the end of the `BoolOp` LLL
2. then evaluate each set of values one at time and jump to the end if the conditions are correct

### How to verify it
Run the tests. I added quite a few cases, thanks to parametrization.

Added short circuit logic for boolean operations.

### Cute Animal Picture
image

### DIFF
--- a/tests/parser/features/test_short_circuiting.py
+++ b/tests/parser/features/test_short_circuiting.py
@@ -1,0 +1,175 @@
+import itertools
+
+import pytest
+
+
+def test_short_circuit_and_left_is_false(w3, get_contract):
+    code = """
+
+called_left: public(bool)
+called_right: public(bool)
+
+@private
+def left() -> bool:
+    self.called_left = True
+    return False
+
+@private
+def right() -> bool:
+    self.called_right = True
+    return False
+
+@public
+def foo() -> bool:
+    return self.left() and self.right()
+"""
+    c = get_contract(code)
+    assert not c.foo()
+
+    c.foo(transact={})
+    assert c.called_left()
+    assert not c.called_right()
+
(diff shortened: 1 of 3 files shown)


tests/parser/features/test_short_circuiting.py
```diff
@@ -0,0 +1,167 @@
...
+def right() -> bool:
+    self.called_right = True
+    return False
+
+@public
+def foo() -> bool:
+    return self.left() and self.right()
+"""
+    c = get_contract(code)
+    c.foo(transact={})
```
fubuloubu, line 27: Maybe also assert on the return value?

PR #2071, opened 2020-06-29
array arg clamps fix

### LINKED ISSUE #2003: Infinite loop from multidimensional array in calldata
### Version Information

* vyper Version (output of `vyper --version`): latest master
* OS: linux
* Python Version: `3.8.2`

### What's your issue about?

Using a multidimensional array in the function inputs, where the length of the >=2nd dimension is 6 or more, causes an infinite loop when calling the function.

For example, each of the following methods will compile but attempting to call them fails with out of gas:

```python
@public
def foo(a: uint256[1][6]):
    pass

@public
def bar(a: uint256[1][1][6]):
    pass
```

For comparison, these methods execute as expected:

```python
@public
def foo(a: uint256[6][1]):
    pass

@public
def bar(a: uint256[100][5][5][5]):
    pass
```

#### Some observations:

* The length of the first array element has no effect.
* The data type has no effect.
* The location of the array within calldata, and total number of arguments, has no effect.
* The number of dimensions, or dimension that exceeds a length of 5, has no effect.

### How can it be fixed?
Unsure at this time.

iamdefinitelyahuman:
### What I did
Adjust logic for args clamper when dealing with arrays.

Fixes #2003 

### How I did it
* Instead of recursively looping through each array within a multidimensional array, we immediately determine the base type and then check each item within the array.
* If the base type does not require clamping, don't loop

### How to verify it
Run tests.  I wasn't able to find many tests around arg clamping so I wrote some to test the entire behavior.

### Cute Animal Picture
image

### DIFF
--- a/vyper/parser/arg_clamps.py
+++ b/vyper/parser/arg_clamps.py
@@ -68,31 +68,44 @@ def make_arg_clamper(datapos, mempos, typ, is_init=False):
     # Lists: recurse
     elif isinstance(typ, ListType):
         if typ.count > 5 or (type(datapos) is list and type(mempos) is list):
-            subtype_size = get_size_of_type(typ.subtype)
-            i_incr = subtype_size * 32
+            # find ultimate base type
+            subtype = typ.subtype
+            while hasattr(subtype, "subtype"):
+                subtype = subtype.subtype
 
-            mem_to = subtype_size * 32 * (typ.count - 1)
+            # make arg clamper for the base type
+            offset = 288
+            clamper = make_arg_clamper(
+                ["add", datapos, ["mload", offset]],
+                ["add", mempos, ["mload", offset]],
+                subtype,
+                is_init,
+            )
+            if clamper.value == "pass":
+                # no point looping if the base type doesn't require clamping
+                return clamper
+
+            type_size = get_size_of_type(typ)
+            i_incr = get_size_of_type(subtype) * 32
+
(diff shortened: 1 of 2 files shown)


vyper/parser/arg_clamps.py
```diff
@@ -68,31 +68,44 @@ def make_arg_clamper(datapos, mempos, typ, is_init=False):
     # Lists: recurse
     elif isinstance(typ, ListType):
         if typ.count > 5 or (type(datapos) is list and type(mempos) is list):
-            subtype_size = get_size_of_type(typ.subtype)
-            i_incr = subtype_size * 32
+            # find ultimate base type
+            subtype = typ.subtype
+            while hasattr(subtype, "subtype"):
+                subtype = subtype.subtype
```
fubuloubu, line 74: lol I hate parser

vyper/parser/arg_clamps.py
```diff
@@ -68,31 +68,44 @@ def make_arg_clamper(datapos, mempos, typ, is_init=False):
...
 
-            mem_to = subtype_size * 32 * (typ.count - 1)
+            # make arg clamper for the base type
+            offset = 288
```
fubuloubu, line 77: hello magical number

iamdefinitelyahuman: Hrmm, good point.  I had to look up what it was, I should probably also make it easier for future devs.

PR #2073, opened 2020-06-30
Rename function visibility decorators

### LINKED ISSUE #2065: VIP: Change `public` and `private` keywords to `external` and `internal`
## Simple Summary
Change `public` and `private` keywords to `external` and `internal`

## Motivation
`private` is a confusing keyword for newcomers, as it doesn't mean what it should in typical contexts. Also, our use of `public` is more consistent with Solidity's `external` modifier, as we do additional checks, so it makes sense to change that too

## Specification
Just change the decorators

## Backwards Compatibility
This would not be backwards compatible

## Dependencies
No dependencies

## References
No references

## Copyright

iamdefinitelyahuman:
### What I did
Replace `@public` and `@private` with `@external` and `@internal`.

Closes #2065 

### How I did it
A lot of grepping, trial and error.

### How to verify it
Run tests.

### Cute Animal Picture
image

### DIFF
--- a/vyper/context/types/function.py
+++ b/vyper/context/types/function.py
@@ -28,8 +28,8 @@ from vyper.exceptions import (
 
 
 class FunctionVisibility(StringEnum):
-    PUBLIC = StringEnum.auto()
-    PRIVATE = StringEnum.auto()
+    EXTERNAL = StringEnum.auto()
+    INTERNAL = StringEnum.auto()
 
 
 class StateMutability(StringEnum):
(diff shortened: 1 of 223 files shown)


vyper/context/types/function.py
```diff
@@ -171,9 +171,8 @@ def from_FunctionDef(
         ---------
         node : FunctionDef
             Vyper ast node to generate the function definition from.
-        is_public : bool, optional
-            Boolean indicating if the function is public or private. Should only be
-            given if the visibility not is indicated via a decorator.
+        is_interface: bool, optional
+            Boolean indicating if the function definition is part of an interface.
```
fubuloubu, line 175: Why the change to `is_interface`?

PR #2075, opened 2020-07-01
Bounds checks for `as_wei_value`

### LINKED ISSUE #1637: Multiple issues with as_wei_value

iamdefinitelyahuman:
### What I did
* Add overflow check for `as_wei_value` when the given value is `uint256`
* Use correct math for `int128`
* Revert on negative values

Fixes #1637 

### How I did it
Modificaions to  `vyper/functions/functions.py::AsWeiValue.build_LLL`

### How to verify it
Run tests.  I added cases to verify the outputs and bounds checks.  I also removed an old test that's no longer useful, because the behavior it was testing is now being folded.

### Cute Animal Picture
image

### DIFF
--- a/tests/parser/types/value/test_as_wei_value.py
+++ b/tests/parser/types/value/test_as_wei_value.py
@@ -1,30 +1,93 @@
-def test_wei_conversion(get_contract_with_gas_estimation):
-    test_wei = """
-@external
-def return_2_finney() -> uint256:
-    return as_wei_value(2, "finney")
+from decimal import Decimal
 
-@external
-def return_3_finney() -> uint256:
-    return as_wei_value(2 + 1, "finney")
+import pytest
 
-@external
-def return_2p5_ether() -> uint256:
-    return as_wei_value(2.5, "ether")
+wei_denoms = {
+    "femtoether": 3,
+    "kwei": 3,
+    "babbage": 3,
+    "picoether": 6,
+    "mwei": 6,
+    "lovelace": 6,
+    "nanoether": 9,
+    "gwei": 9,
+    "shannon": 9,
+    "microether": 12,
+    "szabo": 12,
+    "milliether": 15,
+    "finney": 15,
+    "ether": 18,
+    "kether": 21,
+    "grand": 21,
+}
 
-@external
-def return_3p5_ether() -> uint256:
-    return as_wei_value(2.5 + 1.0, "ether")
 
+@pytest.mark.parametrize("denom,multiplier", wei_denoms.items())
+def test_wei_uint256(get_contract, assert_tx_failed, denom, multiplier):
+    code = f"""
 @external
-def return_2pow64_wei() -> uint256:
-    return as_wei_value(18446744073.709551616, "gwei")
+def foo(a: uint256) -> uint256:
+    return as_wei_value(a, "{denom}")
     """
 
-    c = get_contract_with_gas_estimation(test_wei)
+    c = get_contract(code)
 
-    assert c.return_2_finney() == 2 * 10 ** 15
-    assert c.return_3_finney() == 3 * 10 ** 15, c.return_3_finney()
-    assert c.return_2p5_ether() == 2.5 * 10 ** 18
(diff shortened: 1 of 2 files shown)


tests/parser/types/value/test_as_wei_value.py
```diff
@@ -1,30 +1,84 @@
...
+
+wei_denoms = {
+    ("femtoether", "kwei", "babbage"): 3,
+    ("picoether", "mwei", "lovelace"): 6,
+    ("nanoether", "gwei", "shannon"): 9,
+    ("microether", "szabo",): 12,
+    ("milliether", "finney",): 15,
+    ("ether",): 18,
+    ("kether", "grand"): 21,
+}
```
fubuloubu, line 13: It would be clearer if this were in a flat heirarchy
```suggestion
wei_denoms = {
    "femtoether": 3,
    "kwei": 3,
    "babbage": 3,
    "picoether": 6,
    "mwei": 6,
    "lovelace": 6,
    "nanoether": 9,
    "gwei": 9,
    "shannon": 9,
    "microether": 12,
    "szabo": 12,
    "milliether": 15,
    "finney": 15,
    "ether": 18,
    "kether": 21,
    "grand": 21,
}
```

tests/parser/types/value/test_as_wei_value.py
```diff
@@ -1,30 +1,84 @@
...
 
+
+@pytest.mark.parametrize("denom", [x for v in wei_denoms.keys() for x in v])
```
fubuloubu, line 16:
```suggestion
@pytest.mark.parametrize("denom", wei_denoms.keys())
```

tests/parser/types/value/test_as_wei_value.py
```diff
@@ -1,30 +1,84 @@
...
+    code = f"""
 @external
-def return_3_finney() -> uint256:
-    return as_wei_value(2 + 1, "finney")
+def foo(a: uint256) -> uint256:
+    return as_wei_value(a, "{denom}")
+    """
+    c = get_contract(code)
 
+    multiplier = next(v for k, v in wei_denoms.items() if denom in k)
```
fubuloubu, line 25:
```suggestion
    multiplier = wei_denoms[denom]
```

PR #2076, opened 2020-07-01
Check length of return data from external calls

### LINKED ISSUE #1840: No length check on bytes and string arrays
### Version Information

* vyper Version (output of `vyper --version`): `0.1.0b16`
* OS: linux
* Python: `3.8`

### What's your issue about?

When a contract expects to receive a fixed length bytes or string array, there is no check on the length of the received data.  In the following example, calling `Bar.bar(Foo.address)` returns `b"foobar"`.

Contract `Foo`:

```python
@public
def foo() -> bytes[6]:
    return b"foobar"
```

Contract `Bar`:

```python
contract Foo:
    def foo() -> bytes[3]: constant

a: bytes[3]

@public
def bar(_addr: address) -> bytes[3]:
    self.a = Foo(_addr).foo()
    return self.a
```

This is an issue because Vyper does not allow dynamic arrays but Solidity does. If a Vyper contract is interacting with a Solidity contract that returns `bytes` or `string` it must give an upper bound on the expected bytes length.

### How can it be fixed?
Check the length of the returned data when receiving a bytes or string array. If the length exceeds the given upper bound, truncate it.

### LINKED ISSUE #1978: Vyper doesn't check RETURNDATASIZE on external calls
### What's your issue about?
From https://github.com/vyperlang/vyper/issues/1383#issuecomment-480471399, we don't do a runtime check that the return data size matches what is expected:

A.vy:
```
@public
@constant
def foo():
    pass
```

B.vy:
```
contract A:
    def foo() -> uint256: constant

@public
def bar():
    # A.foo() should raise because it expects a RETURNDATASIZE of 32, but is actually 0
    # NOTE: It will instead believe the result was uint256(0) because it doesn't check this
    assert A(msg.sender).foo() == 42
```

### How can it be fixed?
Add a check to LLL for this on all calls

iamdefinitelyahuman:
### What I did
* Check the length of return data when making an external call.  Closes #1978 
* Validate that returned `bytes` and `string` data does now have a length longer than the max allowed for the given type. Closes #1840 

### How I did it
* Add a check that `RETURNDATASIZE` > expected length -1 after making a call.
* In cases where we are expecting return data, do not include the check involving `EXTCODESIZE`.  An address that is not a contract will not return any data, so this check isn't required.
* For dynamic types (bytes and string), assert that the length of the returned data does not exceed the max allowable length for the contract type

It is still possible for a call to return _more_ data than expected by the interface.  This is consistent with how Solidity handles the check.

### How to verify it
Run the tests. I added a few test cases, and removed one that is no longer valid.

### Cute Animal Picture
image

### DIFF
--- a/tests/parser/features/external_contracts/test_external_contract_calls.py
+++ b/tests/parser/features/external_contracts/test_external_contract_calls.py
@@ -67,14 +67,15 @@ def bar(arg1: address) -> int128:
     print("Successfully executed a complicated external contract call")
 
 
-def test_external_contract_calls_with_bytes(get_contract, get_contract_with_gas_estimation):
-    contract_1 = """
+@pytest.mark.parametrize("length", [3, 32, 33, 64])
+def test_external_contract_calls_with_bytes(get_contract, length):
+    contract_1 = f"""
 @external
-def array() -> bytes[3]:
+def array() -> bytes[{length}]:
     return b'dog'
     """
 
-    c = get_contract_with_gas_estimation(contract_1)
+    c = get_contract(contract_1)
 
     contract_2 = """
 interface Foo:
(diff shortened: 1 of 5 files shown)


vyper/parser/external_call.py
```diff
@@ -33,10 +34,10 @@ def external_call(node, context, interface_name, contract_address, pos, value=No
...
     )
     output_placeholder, output_size, returner = get_external_call_output(sig, context)
-    sub = [
-        "seq",
-        ["assert", ["extcodesize", contract_address]],
-    ]
+    sub = ["seq"]
+    if not output_size:
+        # if we are not expecting return data, check that a contract exists at the target address
+        sub.append(["assert", ["extcodesize", contract_address]])
```
fubuloubu, line 40: Why does this optimization make sense? Write it in a comment

vyper/parser/external_call.py
```diff
@@ -76,9 +77,13 @@ def external_call(node, context, interface_name, contract_address, pos, value=No
                 ],
             ]
         )
+    if output_size:
+        # when return data is expected, revert if the length is insufficient
+        static_output_size = get_static_size_of_type(sig.output_type) * 32
+        sub.append(["assert", ["gt", "returndatasize", static_output_size - 1]])
```
fubuloubu, line 83: What if it returns bytes?

Probably needs separate cases if it's static or dynamic... Perhaps done in another PR

tests/parser/features/external_contracts/test_external_contract_calls.py
```diff
@@ -795,3 +740,87 @@ def get_array(arg1: address) -> int128[3]:
...
+def bar(a: int128) -> (int128, int128):
+    return a, 789
+"""
+    contract_2 = """
+interface Bar:
+    def bar(a: int128) -> int128: view
+
+@external
+def foo(_addr: address) -> int128:
+    return Bar(_addr).bar(456)
```
fubuloubu, line 795: Does this mean that the following _also_ works?

```python
@external
def baz(_addr: address) -> (int128, int128):
    return Bar(_addr).bar(456)
```

fubuloubu: Should probably also check this:

```python
@external
def baz(_addr: address) -> int128:
    a, _ = Bar(_addr).bar(456)
    return a
```

iamdefinitelyahuman: The first would work, yes.  And the 2nd won't work because type checking will flag it as expecting the wrong number of arguments.

fubuloubu: Oh yeah, the interface `Bar` doesn't match the return type of bar in the other contract. So shouldn't both fail to compile?

iamdefinitelyahuman: They are seperate contracts - they can't see each other when compiling.  The check here is that when B calls A, but A returns _more_ calldata than expected, we don't raise.  This is consistent with how Solidity handles it.

fubuloubu: Yeah but in the first example the `Bar` interface specifies it returns an `int128`, but then in `baz` you're saying you expect it to return a `(int128, int128)`, which directly conflicts with the interface. Doesn't have anything to do with the fact the actual contract returns `(int128, int128)`, this should fail a compile-time check.

PR #2078, opened 2020-07-01
Namespace collisions

### LINKED ISSUE #1959: Interfaces allow duplicate definitions
### Version Information
* vyper Version (output of `vyper --version`): 0.1.0-beta.17

### What's your issue about?
This is apparently allowed...

`bar.vy`:
```python
@public
def bar():
    pass

@public
def bar():
    pass
```

`foo.vy`:
```python
import bar as Bar

bar: Bar
```

### How can it be fixed?
Make it so duplicates are caught when importing interfaces

iamdefinitelyahuman:
### What I did
* Add checks for duplicates names when declaring interfaces - fixes #1959 
* Improve the error message for namespace collisions when assigning storage variables. Currently it is "`Member [x] already exists in address`" - not sure how i missed this!

### How I did it
Minor edits.

### How to verify it
Run tests.

### Cute Animal Picture
image

### DIFF
--- a/vyper/context/types/function.py
+++ b/vyper/context/types/function.py
@@ -159,10 +159,7 @@ class ContractFunction(BaseTypeDefinition):
 
     @classmethod
     def from_FunctionDef(
-        cls,
-        node: vy_ast.FunctionDef,
-        is_interface: Optional[bool] = False,
-        include_defaults: Optional[bool] = True,
+        cls, node: vy_ast.FunctionDef, is_interface: Optional[bool] = False,
     ) -> "ContractFunction":
         """
         Generate a `ContractFunction` object from a `FunctionDef` node.
@@ -173,9 +170,6 @@ class ContractFunction(BaseTypeDefinition):
             Vyper ast node to generate the function definition from.
         is_interface: bool, optional
             Boolean indicating if the function definition is part of an interface.
-        include_defaults: bool, optional
-            If False, default arguments are ignored when parsing generating the
-            object. Used for interfaces.
 
         Returns
         -------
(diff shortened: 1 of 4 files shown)


vyper/context/types/meta/interface.py
```diff
@@ -140,13 +140,30 @@ def build_primitive_from_node(
...
-    functions = OrderedDict()
+    functions: OrderedDict = OrderedDict()
     for node in base_node.get_children(vy_ast.FunctionDef):
         if "external" in [i.id for i in node.decorator_list]:
-            func = ContractFunction.from_FunctionDef(node, include_defaults=True)
+            func = ContractFunction.from_FunctionDef(node)
+            if node.name in functions:
+                existing_args = list(functions[node.name].arguments)
+                new_args = list(func.arguments)
+                for a, b in zip(existing_args, new_args):
```
fubuloubu, line 150: I wish this chunk had a comment or two

vyper/context/types/function.py
```diff
@@ -265,10 +259,7 @@ def from_FunctionDef(
             )
 
         arguments = OrderedDict()
-        if include_defaults:
-            defaults = [None] * (len(node.args.args) - len(node.args.defaults)) + node.args.defaults
-        else:
-            defaults = [None] * len(node.args.args)
+        defaults = [None] * (len(node.args.args) - len(node.args.defaults)) + node.args.defaults
```
fubuloubu, line 262: This is because you should always include the defaults?

iamdefinitelyahuman: Yeah, at some point that kwarg stopped being used.  So it's ded code.

Issue #1440, opened 2019-05-20
VIP: Dynamic Arrays w/ Sizing


fubuloubu:
## Motivation
For inter-operation with Solidity and other languages that do not take Vyper's conservative approach to arrays, it is important to have a syntax that retains our principle of protecting against gas exhaustion attacks (and gas estimation) while encoding as a dynamic array type via the ABI.

## Specification
This looks like defining a new "meta" type called `DynArray`:
```python
@public
def foo(a: DynArray[uint256, 1:5]):
    ...  # a must have between 1 and 5 elements

@public
def bar(a: DynArray[uint256, 4]):
    ...  # a must have between 0 and 4 elements
```

In general, using this syntax, one can describe the minimum and the maximum of what a method will accept for a dynamic array. Min and max size have to be specified via `min_size:max_size` after the type that a dynamic array takes. If only a number is specified, `min_size` is assumed to be 0. If both are specified, the restriction `min_size < max_size` is enforced. Both can be a named constant.

When the calldata is parsed by the method, it will do an array bounds check to ensure the array's `len` is within the maximum and/or minimum specified. It will then operate on that array as a dynamic array, for example using he `for item in list` syntax for iteration.

This type of list will have limited abilities versus a static array. For example, it might not be allowed to be passed as an argument to an inner private function or to be referenced by index directly (these are possibilities, whatever the compiler needs to do to ensure safe usage of these types is TBD). This type of list may also not be allowed to be a state variable declaration (out-of-gas lock bugs possible), and potentially be disallowed from memory declaration usage as well.

Also, the built-in function `len(...)` should be expanded to allow the ability to query the current length variable of an `Array` type (currently it only queries the size of `bytes` objects)

## Backwards Compatibility
The current syntax for `bytes[N]`/`string[N]` uses this same behavior, so to make both syntax consistent we are changing to `Bytes[N]`/`Bytes[M:N]` and `String[N]`/`String[M:N]`.

## Dependencies
No dependencies

## Copyright
Copyright and related rights waived via [CC0](https://creativecommons.org/publicdomain/zero/1.0/)

jacqueswww:
Note from meeting: `bytes[min:max]` min & max will have to be forced syntax, for readability / clarity.

fubuloubu:
Something like `Array(type, max_size=5)` might be more readable:

```python
@public
def foo(a: Array(uint256, max_size=5)) -> uint256:
    # reverts if len(a) > 5
    sum: uint256 = 0
[...]

PR #2080, opened 2020-07-02
Make bytes and string uppercase

iamdefinitelyahuman:
### What I did
Rename `bytes` and `string` to `Bytes` and `String`

Related to #1440 

### How I did it
A lot of grepping.

### How to verify it
Run tests.

`bytes` and `string` have been changed to `Bytes` and `String`

### Cute Animal Picture
image

### DIFF
--- a/vyper/context/types/utils.py
+++ b/vyper/context/types/utils.py
@@ -91,7 +91,8 @@ def get_type_from_abi(
     type_string = abi_type["type"]
     if type_string == "fixed168x10":
         type_string = "decimal"
-    # TODO string and bytes
+    if type_string in ("string", "bytes"):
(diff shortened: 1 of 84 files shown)


vyper/context/types/utils.py
```diff
@@ -91,7 +91,8 @@ def get_type_from_abi(
     type_string = abi_type["type"]
     if type_string == "fixed168x10":
         type_string = "decimal"
-    # TODO string and bytes
```
fubuloubu, line 94: Why was this note there?

iamdefinitelyahuman: in anticipation of this, I guess :smile:

PR #2081, opened 2020-07-02
Docs updates

### LINKED ISSUE #1867: VIP: Remove assert and raise as callables
## Simple Summary
`assert` and `raise` should only be possible via statements, not callable functions.

## Motivation
Allowing these two commands to happen as statements _and_ functions creates ambiguity in the syntax without any added benefit that I can see. I believe being opinionated here will make for more readable smart contracts, and means we can reduce some complexity in the vyper codebase.

## Specification

Currently all of the following syntax is valid.  This VIP proposes that the second set of examples should be _invalid_.

```python
# keep this
assert foo, "oopsie"
raise "oopsie"

# remove this
assert(foo, "oopsie")
raise("oopsie")
```

## Backwards Compatibility
This will be a breaking change, if approved we might want to delay it until a major release.

## Copyright

### LINKED ISSUE #1917: Documentation Updates Megaissue
While working on #1915 I've run into some areas where the documentation is lacking. This issue is a list of topics that I think need work.  It may change over time.

- [x] `public` and `constant` as methods applied to storage variables
- [x] `self`
- [x] assignment
- [x] statements, expressions, control structure
- [x] scoping rules
- [x] `for` loops
- [x] tuples
- [x] contract objects
- [ ] memory layout of data types
- [ ] pass-by-reference / pass-by-value
- [ ] abi format
- [x] arithmetic functions (should be moved from types to builtin functions)
- [x] allowable literals for each type
- [x] examples for each of the builtin functions
- [x] `__init__` method

iamdefinitelyahuman:
### What I did
Big update to the documentation.

Closes #1867 
Closes #1917 

This PR depends on #2080 - I worked on top of commits there to avoid conflicts.

### How I did it
Lots of writing.

### How to verify it
Render and read the docs.

### Cute Animal Picture
image


docs/types.rst
```diff
@@ -301,153 +273,141 @@ Where ``x`` is a byte array and ``_start`` as well as ``_len`` are integer value
 .. index:: !bytes
 
 Fixed-size Byte Arrays
-======================
+----------------------
 
 **Keyword:** ``Bytes``
 
 A byte array with a fixed size.
```
fubuloubu, line 280: `Bytes` is a dynamic type

docs/types.rst
```diff
@@ -301,153 +273,141 @@ Where ``x`` is a byte array and ``_start`` as well as ``_len`` are integer value
...
+    # Setting values
+    exampleList = [10, 11, 12]
+    exampleList[2] = 42
+
+    # Returning a value
+    return exampleList[0]
+
+.. _types-struct:
 
 Structs
```
fubuloubu, line 339: This section could be more filled out, or link to a better section

docs/types.rst
```diff
@@ -301,153 +273,141 @@ Where ``x`` is a byte array and ``_start`` as well as ``_len`` are integer value
...
 Mappings
-========
+--------
 
-Mappings in Vyper can be seen as `hash tables <https://en.wikipedia.org/wiki/Hash_table>`_ which are virtually initialized such that
-every possible key exists and is mapped to a value whose byte-representation is
-all zeros: a type's default value. The similarity ends here, though: The key data is not actually stored
-in a mapping, only its ``keccak256`` hash used to look up the value. Because of this, mappings
-do not have a length or a concept of a key or value being "set".
+Mappings are `hash tables <https://en.wikipedia.org/wiki/Hash_table>`_ that are virtually initialized such that every possible key exists and is mapped to a value whose byte-representation is all zeros: a type's :ref:`default value <types-initial>`.
```
fubuloubu, line 368: Maybe mention that the default value is the same value produced by `empty(type)`

docs/types.rst
```diff
@@ -301,153 +273,141 @@ Where ``x`` is a byte array and ``_start`` as well as ``_len`` are integer value
...
-------
+******
 
 Mapping types are declared as ``HashMap[_KeyType, _ValueType]``.
-Here ``_KeyType`` can be any base or bytes type. Mappings, interfaces or structs are not support as key types.
-``_ValueType`` can actually be any type, including mappings.
 
-**Example:**
-::
+* ``_KeyType`` can be any base or bytes type. Mappings, interfaces or structs are not support as key types.
```
fubuloubu, line 380: Interfaces should eventually be allowed

docs/types.rst
```diff
@@ -479,7 +439,7 @@ Here you can find a list of all types and default values:
 .. _type_conversions:
 
 Type Conversions
```
fubuloubu, line 441: This table uses older [constant names](https://vyper.readthedocs.io/en/release-0.2.0/constants-and-vars.html#built-in-constants), and also doesn't display all on one page

docs/statements.rst
```diff
@@ -0,0 +1,114 @@
...
+The event must have been previously declared
+
+See :ref:`Event Logging<event-logging>` for more information on events.
+
+Assertions and Exceptions
+=========================
+
+Vyper uses state-reverting exceptions to handle errors. Exceptions trigger the ``REVERT`` opcode (``0xFD``) with the provided reason given as the error message. When an exception is raised the code stops operation, the contract's state is reverted to the state before the transaction took place and the remaining gas is returned to the transaction's sender. When an exception happen in a sub-call, it “bubbles up” (i.e., exceptions are rethrown) automatically.
+
+If the reason string is set to ``UNREACHABLE``, an ``INVALID`` opcode (``0xFE``) is used instead of ``REVERT``. In this case, calls that revert do not receive a gas refund.
```
fubuloubu, line 85:
```suggestion
If the reason string is set to ``UNREACHABLE``, an ``INVALID`` opcode (``0xFE``) is used instead of ``REVERT``. In this case, calls that revert do not receive a gas refund. This is not a recommended practice for general usage, but is available for interoperability with various tools that use the ``INVALID`` opcode to perform dynamic analysis.
```

docs/toctree.rst
```diff
@@ -0,0 +1,48 @@
...
+
+.. toctree::
+    :caption: Language Description
+    :maxdepth: 2
+
+    structure-of-a-contract.rst
+    types.rst
+    constants-and-vars.rst
+    statements.rst
+    control-structures.rst
```
fubuloubu, line 25: Should these two be switched?

docs/control-structures.rst
```diff
@@ -0,0 +1,259 @@
...
+=============
+
+The ``for`` statement is a control flow construct used to iterate over a value:
+
+.. code-block:: python
+
+    for i in <ITERABLE>:
+        ...
+
+The iterated value can be an array, or generated from the builtin ``range`` function.
```
fubuloubu, line 208:
```suggestion
The iterated value can be a static array, or generated from the builtin ``range`` function.
```

I mean, we can probably iterate over a dynamic array, eventually.... because we know max size!

docs/control-structures.rst
```diff
@@ -0,0 +1,259 @@
...
+You can also iterate over a literal array, as long as a common type can be determined for each item in the array:
+
+.. code-block:: python
+
+    for i in [4, 23, 42]:
+        ...
+
+Some restrictions:
+
+* You cannot iterate over a multi-dimensional array.  ``i`` must always be a base type.
```
fubuloubu, line 232: We can probably eventually relax this to simply being "sized"

docs/scoping-and-declarations.rst
```diff
@@ -0,0 +1,185 @@
...
+    def foo() -> (int128: int128):
+        return 2, 3
+
+    @external
+    def bar():
+        a: int128 = 0
+        b: int128 = 0
+
+        # the return value of `foo` is assigned using a tuple
+        (a, b) = self.foo()
```
fubuloubu, line 53:
```suggestion
        (a, b) = self.foo()
        # Can also skip the parenthesis
        a, b = self.foo()
```

docs/built-in-functions.rst
```diff
@@ -3,176 +3,309 @@
...
+        512
 
-.. py:function:: len(b: Union[Bytes, String]) -> uint256
+Chain Interaction
+=================
 
-    Return the length of a given ``Bytes`` or ``String``.
+.. py:function:: create_forwarder_to(target: address, value: uint256 = 0) -> address
+
+    Duplicates a contract's code and deploys it as a new instance, by means of a ``DELEGATECALL``.
```
fubuloubu, line 99: Doesn't actually duplicate lol. It creates a forwarder contract that delegates all logic to a given target template contract

docs/built-in-functions.rst
```diff
@@ -3,176 +3,309 @@
...
+    Converts a variable or literal from one type to another.
+
+    * ``value``: Value to convert
+    * ``type_``: The destination type to convert to (``bool``, ``decimal``, ``int128``, ``uint256`` or ``bytes32``)
+
+    Returns a value of the type specified by ``type_``.
+
+    For more details on available type conversions, see :ref:`type_conversions`.
 
 .. py:function:: extract32(b: Bytes, start: int128, output_type=bytes32) -> Union[bytes32, int128, address]
```
fubuloubu, line 310:
```suggestion
.. py:function:: extract32(b: Bytes, start: int128, output_type=bytes32) -> Any
```

docs/built-in-functions.rst
```diff
@@ -196,112 +329,259 @@ contracts.
...
+
+Utilities
+=========
+
+.. py:function:: as_wei_value(_value, unit: str) -> uint256
+
+    Take an amount of ether currency specified by a number and a unit and return the integer quantity of wei equivalent to that amount.
+
+    * ``_value``: Value for the ether unit. Any numeric type may be used, however the value cannot be negative.
+    * ``unit``: Ether unit name (e.g. ``"wei"``, ``"ether"``, ``"gwei"``, etc.) indicating the denomination of ``_value``. Must be given as a literal string.
```
fubuloubu, line 505:
```suggestion
    * ``unit``: Ether unit name (e.g. ``"wei"``, ``"ether"``, ``"gwei"``, etc.) indicating the denomination of ``_value``. Must be given as decimal literal.
```

docs/built-in-functions.rst
```diff
@@ -196,112 +329,259 @@ contracts.
...
 
 .. py:function:: blockhash(block_num: uint256) -> bytes32
 
-    Returns the hash of the block at the specified height.
+    Return the hash of the block at the specified height.
 
     .. note::
 
-        The EVM only provides access to the most 256 blocks. This function will return 0 if the block number is greater than or equal to the current block number or more than 256 blocks behind the current block.
+        The EVM only provides access to the most 256 blocks. This function returns 0 if the block number is greater than or equal to the current block number or more than 256 blocks behind the current block.
```
fubuloubu, line 525:
```suggestion
        The EVM only provides access to the most 256 blocks. This function returns ``EMPTY_BYTES32`` if the block number is greater than or equal to the current block number or more than 256 blocks behind the current block.
```

Note to self: we should actually revert in this condition?

docs/interfaces.rst
```diff
@@ -0,0 +1,179 @@
...
+        FooBar(some_address).query()  # cannot change storage, but reads itself
+        FooBar(some_address).update()  # storage can be altered
+        FooBar(some_address).pay(value=1)  # storage can be altered, and value can be sent
+
+Importing Interfaces
+====================
+
+Interfaces are imported with ``import`` or ``from ... import`` statements.
+
+Imported interfaces are written using standard Vyper syntax, with the body of each function replaced by a ``pass`` statement:
```
fubuloubu, line 65:
```suggestion
Imported interfaces are written using standard Vyper syntax, with the body of each function simply ignored by importing it. If you are defining a standalone interface, it is normally specified by using a ``pass`` statement:
```

docs/interfaces.rst
```diff
@@ -0,0 +1,179 @@
...
+
+    @external
+    def test1():
+        pass
+
+    @external
+    def calculate() -> uint256:
+        pass
+
+You can also import a fully implemented contract and Vyper will automatically convert it to an interface. It is even possible for a contract to import itself to gain access to it's own interface.
```
fubuloubu, line 77: Should we show an example of this?

```python
import .greeter as Self

name: public(String[10])

@external
def __init__(_name: String[10]):
    self.name = _name

@view
@external
def greet() -> String[16]:
    return concat("Hello ", Self(msg.sender).name())
```

docs/interfaces.rst
```diff
@@ -0,0 +1,179 @@
...
+    import contract.foo as Foo
+
+Imports via ``from ... import``
+-------------------------------
+
+Using ``from`` you can perform both absolute and relative imports. With ``from`` import statements you **cannot** use an alias - the name of the interface will always be that of the file:
+
+.. code-block:: python
+
+    from contract import foo
```
fubuloubu, line 95: These two import rules seem weird together

docs/event-logging.rst
```diff
@@ -30,11 +31,11 @@ This example is taken from the `sample ERC20 contract <https://github.com/vyperl
...
-two indexed address fields. Indexed fields do not make up part of the event data itself, but can be searched by clients that
-want to catch the event. Also, each event contains one single data field, in each case called ``value``. Events can contain several arguments with any names desired.
+Let's look at what this is doing.
 
-Next, in the ``transfer`` function, after we do whatever work is necessary, we log the event. We pass three arguments, corresponding with the three arguments of the Transfer event declaration.
+    1. We declare two event types to log. The two events are similar in that they contain
+    two indexed address fields. Indexed fields do not make up part of the event data itself, but can be searched by clients that want to catch the event. Also, each event contains one single data field, in each case called ``value``. Events can contain several arguments with any names desired.
+    2. In the ``transfer`` function, after we do whatever work is necessary, we log the event. We pass three arguments, corresponding with the three arguments of the Transfer event declaration.
```
fubuloubu, line 38: Not actually a list

docs/built-in-functions.rst
iamdefinitelyahuman: It's a literal string. Maybe you mean `_value` ?

docs/event-logging.rst
iamdefinitelyahuman: Because of the numbering? I did that because the original content used "first" and "next".  It's not fully sequential, but you can't log an event without first declaring it.

docs/interfaces.rst
iamdefinitelyahuman: Yeah, this is what #2083 addresses

docs/toctree.rst
iamdefinitelyahuman: Python and Solidity docs both start small and get bigger, I followed their lead.  I think it makes sense..  you need the small pieces to understand the larger ones.

docs/types.rst
```diff
@@ -25,19 +21,19 @@ are used as function arguments or in assignments.
...
 
 A boolean is a type to store a logical/truth value.
 
 Values
-------
+******
 
 The only possible values are the constants ``True`` and ``False``.
 
 Operators
```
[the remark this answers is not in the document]
iamdefinitelyahuman: That relates to boolean operators, not booleans.  Not sure where it'd belong, somewhere in the control flow section probably?

docs/types.rst
```diff
@@ -228,27 +205,23 @@ Operator       Description
...
 **Keyword:** ``address``
 
 The address type holds an Ethereum address.
 
 Values
-------
+******
 
-An address type can hold an Ethereum address which equates to 20 bytes or 160 bits. It returns in hexadecimal notation with a leading ``0x``.
+An address type can hold an Ethereum address which equates to 20 bytes or 160 bits. Address literals must be written in hexadecimal notation with a leading ``0x``. All addresses must be `checksummed <https://github.com/ethereum/EIPs/blob/master/EIPS/eip-155.md>`_.
```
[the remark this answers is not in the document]
iamdefinitelyahuman: Good catch

docs/types.rst
iamdefinitelyahuman: I think the idea here is that because the size in memory is fixed, the type itself is "fixed size".

docs/types.rst
iamdefinitelyahuman: There is a link to the default values section where `empty` is mentioned.

docs/types.rst
fubuloubu: "Fixed" strikes me as another way to say "static", which they're definitely not static.

docs/interfaces.rst
fubuloubu: So #2083 would make it so you can use an alias on `from ...` statements, right?

Does it also allow _not_ using an alias on regular imports? I feel like this would be most consistent

docs/event-logging.rst
fubuloubu: Oh I mean it literally isn't rendering as a list, if that was the intent

docs/built-in-functions.rst
iamdefinitelyahuman: I just moved the text that was already written.  Can you suggest an edit?

docs/interfaces.rst
iamdefinitelyahuman: It doesn't work very well for regular imports, because you'd end up with an interface named `path.to.interface`

docs/event-logging.rst
iamdefinitelyahuman: oh weird.

docs/interfaces.rst
fubuloubu: > It doesn't work very well for regular imports, because you'd end up with an interface named `path.to.interface`

Ahhhhh, gotcha

docs/built-in-functions.rst
fubuloubu:
```suggestion
    Creates a small contract and deploys it as a new instance of the contract at ``target``, which duplicates the logic of ``target`` but has it's own state since every call is made using ``DELEGATECALL`` to ``target``.
```

docs/built-in-functions.rst
fubuloubu: yes, sorry

docs/types.rst
```diff
@@ -479,153 +422,35 @@ Here you can find a list of all types and default values:
 .. _type_conversions:
 
 Type Conversions
-****************
+================
 
 All type conversions in Vyper must be made explicitly using the built-in ``convert(a, b)`` function. Currently, the following type conversions are supported:
```
fubuloubu, line 427:
```suggestion
All type conversions in Vyper must be made explicitly using the built-in ``convert(a: atype, btype) -> btype`` function. Currently, the following type conversions are supported:
```

docs/types.rst
```diff
@@ -479,153 +422,35 @@ Here you can find a list of all types and default values:
...
-     - Do not allow converting to/from the same type
-   * - ``bytes32``
-     - ``Bytes``
-     - ``(0x00 * 1)...(0xFF * 32)``
-     - Left-pad input ``Bytes`` to size of ``32``
-
-
-.. index:: !conversion
+================ ================== ============================== =================================
+Initial Type (a) Converted Type (b) Allowable Values               Additional Notes
```
fubuloubu, line 430:
```suggestion
In (``atype``) Out (``btype``) Allowable Values               Additional Notes
```

docs/types.rst
```diff
@@ -479,153 +422,35 @@ Here you can find a list of all types and default values:
...
+================ ================== ============================== =================================
+``bool``         ``decimal``        All                            ``0.0`` or ``1.0``
+``bool``         ``int128``         All                            ``0`` or ``1``
+``bool``         ``uint256``        All                            ``0`` or ``1``
+``bool``         ``bytes32``        All                            ``0x00`` or ``0x01``
+``decimal``      ``bool``           All                            Returns ``a != 0.0``
+``decimal``      ``int128``         All                            Value is truncated
+``decimal``      ``uint256``        ``0`` ... ``MAX_DECIMAL``      Value is truncated
```
fubuloubu, line 438:
```suggestion
``decimal``      ``uint256``        ``a >= 0.0``      Value is truncated
```

docs/types.rst
```diff
@@ -479,153 +422,35 @@ Here you can find a list of all types and default values:
...
+``decimal``      ``bytes32``        All
+``decimal``      ``Bytes``          All
+``int128``       ``bool``           All                            Returns ``a != 0``
+``int128``       ``decimal``        All
+``int128``       ``uint256``        ``0`` ... ``MAX_INT128``       Cannot convert negative values
```
fubuloubu, line 443:
```suggestion
``int128``       ``uint256``        ``a >= 0``       Cannot convert negative values
```

docs/types.rst
```diff
@@ -479,153 +422,35 @@ Here you can find a list of all types and default values:
...
+``int128``       ``bytes32``        All
+``int128``       ``Bytes``          All
+``uint256``      ``bool``           All                            Returns ``a != 0``
+``uint256``      ``decimal``        ``0`` ... ``MAX_DECIMAL``
```
fubuloubu, line 447:
```suggestion
``uint256``      ``decimal``        ``a <= MAX_DECIMAL``
```

docs/types.rst
```diff
@@ -479,153 +422,35 @@ Here you can find a list of all types and default values:
...
+``uint256``      ``int128``         ``0`` ... ``MAX_INT128``
```
fubuloubu, line 448:
```suggestion
``uint256``      ``int128``         ``a <= MAX_INT128``
```

docs/types.rst
fubuloubu: How about just "Byte Arrays"?

docs/types.rst
fubuloubu: or just 
```suggestion
``atype`` ``btype`` Allowable Values               Additional Notes
```

PR #2082, opened 2020-07-03
0.2.0 Release Notes

fubuloubu:
Please add suggestions to what we should mention in the release notes

Depends on #2085 being merged

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 0.1.0-beta.17
+current_version = 0.2.0
 commit = True
 tag = True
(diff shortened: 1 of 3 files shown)


setup.cfg
```diff
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 0.1.0-beta.17
+current_version = 0.2.0
```
fubuloubu, line 2: Note: I'm gonna try something with this release. Basically, we'll remove `bumpversion` from our dependencies and just change the version in `setup.py` during PRs like these, and then create a tagged release through GitHub instead. The creation of a new tag should also trigger PyPI and DockerHub releases

PR #2083, opened 2020-07-03
Import syntax updates

iamdefinitelyahuman:
### What I did
* Eliminate the `alias` AST node and store alias information directly in `Import` and `ImportFrom` nodes.
* Allow the use of aliases with `from ..` style import syntax
* Disallow multiple imports from a single import statement, e.g `from foo import a, b as c` - this was never documented as an allowable behavior and it only worked in certain formats. I think it was just an oversight that it was possible.

### How I did it
Minor updates in a few places in the codebase.

I also moved `extract_file_interface_imports` out of `signatures` and into `cli` - it was only ever called from `cli` and we plan to refactor out `signatures` so seemed like a good time to do this.

### How to verify it
Run the tests. I added / updated test cases accordingly,

### Cute Animal Picture
image


iamdefinitelyahuman:
This PR doesn't include documentation updates yet - waiting on #2081 to merge.

vyper/ast/nodes.py
```diff
@@ -1142,16 +1142,24 @@ class Pass(VyperNode):
     __slots__ = ()
 
 
-class Import(VyperNode):
-    __slots__ = ("names",)
+class _Import(VyperNode):
+    __slots__ = ("name", "asname")
```
fubuloubu, line 1146: `alias` seems like a better name than `asname`

vyper/cli/utils.py
```diff
@@ -0,0 +1,38 @@
...
+            if node.alias in imports_dict:
+                raise StructureException(
+                    f"Interface with alias {node.alias} already exists", node,
+                )
+            imports_dict[node.alias] = node.name.replace(".", "/")
+        elif isinstance(node, vy_ast.ImportFrom):  # type: ignore
+            level = node.level  # type: ignore
+            module = node.module or ""  # type: ignore
+            if not level and module == "vyper.interfaces":
+                continue
```
fubuloubu, line 23:
```suggestion
                continue  # Uses internal module, so skip adding to imports
```

PR #2084, opened 2020-07-03
Change address.codesize type to uint256

iamdefinitelyahuman:
### What I did
Change the type of `address.codesize` to `uint256`.

### How I did it
With style and grace.

### How to verify it
Run the tests.

### Cute Animal Picture
image

### DIFF
--- a/vyper/parser/expr.py
+++ b/vyper/parser/expr.py
@@ -328,8 +328,11 @@ class Expr:
             addr = Expr.parse_value_expr(self.expr.value, self.context)
             if is_base_type(addr.typ, "address"):
                 if self.expr.attr == "codesize":
-                    eval_code = ["extcodesize", addr]
-                    output_type = "int128"
+                    if self.expr.value.id == "self":
+                        eval_code = ["codesize"]
+                    else:
+                        eval_code = ["extcodesize", addr]
+                    output_type = "uint256"
                 else:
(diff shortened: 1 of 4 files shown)


vyper/parser/expr.py
```diff
@@ -328,8 +328,11 @@ def parse_Attribute(self):
             addr = Expr.parse_value_expr(self.expr.value, self.context)
             if is_base_type(addr.typ, "address"):
                 if self.expr.attr == "codesize":
-                    eval_code = ["extcodesize", addr]
-                    output_type = "int128"
+                    if self.expr.value.id == "self":
+                        eval_code = ["codesize"]
```
fubuloubu, line 332: This was added in Istanbul, right?

fubuloubu: Yes, `EXTCODEHASH` was added in Istanbul: https://eips.ethereum.org/EIPS/eip-1884

fubuloubu: This should be `... or evm_version < Istanbul`

And then add test cases for this behavior before/after Istanbul.

I think this feature was added recently, so maybe we missed this in the VIP?

iamdefinitelyahuman: No.  You might be thinking of `EXTCODEHASH` that was added in constantinople?

fubuloubu: Ah, yes totally right. Disregard then.

PR #2090, opened 2020-07-03
add repr for Mapping

iamdefinitelyahuman:
### What I did
Add a `__repr__` method to `MappingDefinition`.  Without this, the error messages related to mappings are weird.

### Cute Animal Picture
image

### DIFF
--- a/vyper/context/types/indexable/mapping.py
+++ b/vyper/context/types/indexable/mapping.py
@@ -14,6 +14,9 @@ from vyper.exceptions import StructureException
 class MappingDefinition(IndexableTypeDefinition):
     _id = "HashMap"
 
+    def __repr__(self):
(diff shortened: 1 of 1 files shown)


vyper/context/types/indexable/mapping.py
```diff
@@ -14,6 +14,9 @@
 class MappingDefinition(IndexableTypeDefinition):
     _id = "HashMap"
 
+    def __repr__(self):
+        return f"HashMap[{self.key_type}, {self.value_type}]"
```
fubuloubu, line 18: Just in case you ever subclass this further for other mapping types :smile: 
```suggestion
        return f"{self._id}[{self.key_type}, {self.value_type}]"
```

PR #2096, opened 2020-07-05
fix: type checking in for loops

iamdefinitelyahuman:
### What I did
* Fix a bug in the type checker preventing proper checks during for loops. This wasn't a critical issue, as any failure is still caught in `parser`.
* Improve how exception messages are handled during a failed type check in a for loop

### How I did it
* The main issue was from an improperly indented `return`.  Ooops.
* For the exception message, I'm tracking every exception raised during the iterative check.  At the end, if they're all the same exception, I raise it instead of the more generic "could not determine type for i".

### How to verify it
Run tests.

### Cute Animal Picture
image

### DIFF
--- a/vyper/context/validation/local.py
+++ b/vyper/context/validation/local.py
@@ -353,9 +353,12 @@ class FunctionNodeVisitor(VyperNodeVisitorBase):
                             call_node,
                         )
 
+        for_loop_exceptions = []
         for type_ in type_list:
+            # type check the for loop body using each possible type for iterator value
             type_ = copy.deepcopy(type_)
             type_.is_immutable = True
+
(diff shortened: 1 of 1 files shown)


vyper/context/validation/local.py
```diff
@@ -353,9 +353,12 @@ def visit_For(self, node):
                             call_node,
                         )
 
+        for_loop_exceptions = []
         for type_ in type_list:
+            # type check the fdor loop body using each possible type for iterator value
```
fubuloubu, line 358: `fdor` lol

iamdefinitelyahuman: fnord

PR #2102, opened 2020-07-06
fix: adjust FuncionDef offset to not include decorators

iamdefinitelyahuman:
### What I did
Adjust `FunctionDef` offset to not include decorators. This ensures a more readable annotation when an exception is raised.

Before:

```python
vyper.exceptions.FunctionDeclarationException: Missing or unmatched return statements in function 'foo'
line 12:0 
     11
---> 12 @external
--------^
     13 @payable
```

After:

```python
vyper.exceptions.FunctionDeclarationException: Missing or unmatched return statements in function 'foo'
line 14:0 
     13 @payable
---> 14 def foo() -> uint256:
--------^
     15     pass
```

### How I did it
When converting from Python to Vyper AST, modify `FunctionDef.lineno` to be +1 from the `end_lineno` of the final decorator.

### How to verify it
Run tests.

### Cute Animal Picture
image

### DIFF
--- a/vyper/ast/nodes.py
+++ b/vyper/ast/nodes.py
@@ -604,6 +604,12 @@ class Module(TopLevel):
 class FunctionDef(TopLevel):
     __slots__ = ("args", "returns", "decorator_list", "pos")
 
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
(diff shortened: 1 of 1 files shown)


iamdefinitelyahuman:
The only way to be more accurate is to look for the specific `def` token.  Doable, but also a fair bit of work and this fix should solve it in 99% of cases.

PR #2105, opened 2020-07-07
adjust FunctionDef offset using tokens

iamdefinitelyahuman:
### What I did
Adjust the start offset for `FunctionDef` nodes using tokens.  This is a more accurate solution to #2102 

### How I did it
In `ast/annotation.py`, modify `FunctionDef.first_token` by searching for the first `def` token after the final decorator.

### How to verify it
Run tests.

### Cute Animal Picture
image

### DIFF
--- a/vyper/ast/annotation.py
+++ b/vyper/ast/annotation.py
@@ -1,4 +1,5 @@
 import ast as python_ast
+import tokenize
 from decimal import Decimal
 from typing import Optional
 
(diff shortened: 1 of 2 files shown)


vyper/ast/annotation.py
```diff
@@ -75,6 +78,12 @@ def visit_Module(self, node):
         return self._visit_docstring(node)
 
     def visit_FunctionDef(self, node):
+        if node.decorator_list:
+            # start the source highlight at `def` to improve annotation readability
+            decorator_token = node.decorator_list[-1].last_token
+            def_token = self._tokens.find_token(decorator_token, tokenize.NAME, tok_str="def")
+            node.first_token = def_token
```
fubuloubu, line 85: So, this will change the error message to point at the `def`?

iamdefinitelyahuman: Yeah :)

PR #2108, opened 2020-07-15
Array index typecheck

iamdefinitelyahuman:
### What I did
* Fix a missing type check on array index values.
* Add `__repr__` to abstract types to improve readability of error messages.

### How I did it
When an index value is not a literal integer, call to `validate_expected_type` and expect  `IntegerAbstractType`

### How to verify it
Run tests.

### Cute Animal Picture
image

### DIFF
--- a/vyper/context/types/abstract.py
+++ b/vyper/context/types/abstract.py
@@ -16,16 +16,26 @@ class AbstractDataType:
             pass
         return isinstance(other, type(self))
 
+    def __repr__(self):
+        value = super().__repr__()
+        if value == object.__repr__(self):
+            # use `_description` when no parent class overrides the default python repr
+            return self._description
+        return value
+
 
 class ArrayValueAbstractType(AbstractDataType):
     """
     Abstract data class for single-value types occupying multiple memory slots.
     """
 
+    _description = "fixed size bytes array or string"
+
 
 class BytesAbstractType(AbstractDataType):
     """Abstract data class for bytes types (bytes32, bytes[])."""
 
+    _description = "bytes"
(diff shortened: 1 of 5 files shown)


fubuloubu:
Add a test

vyper/context/types/abstract.py
```diff
@@ -16,16 +16,26 @@ def compare_type(self, other):
             pass
         return isinstance(other, type(self))
 
+    def __repr__(self):
+        value = super().__repr__()
+        if value.startswith("<"):
```
fubuloubu, line 21: Why does this mean it gives meaningful output?

vyper/context/types/indexable/sequence.py
```diff
@@ -69,8 +70,10 @@ def get_index_type(self, node):
         if isinstance(node, vy_ast.Int):
             if node.value < 0:
                 raise ArrayIndexException("Vyper does not support negative indexing", node)
-            if node.value < 0 or node.value >= self.length:
```
fubuloubu, line 72: lol

vyper/context/types/abstract.py
iamdefinitelyahuman: the opposite actually.  a normal repr is `<ClassName object at 0x...>`, whereas the desired ones will be e.g. "string".  If it opens with `<`, we know it's not implemented and so we use `_description`.

fubuloubu: Ah, could it be `f"<{self.__class__.__name__}"` instead, so we know that it's a normal Python repr?

PR #2115, opened 2020-07-25
fix: add trailing whitespace for EOF exception

iamdefinitelyahuman:
### What I did
Fix a bug preventing annotation of end-of-file syntax errors

### How I did it
Append `" "` to source prior to annotation. The EOF error always points to one character beyond the length of the source, as this is the offset the error was encountered at.

### How to verify it
Try to compile an incomplete source.

### Cute Animal Picture
image

### DIFF
--- a/vyper/exceptions.py
+++ b/vyper/exceptions.py
@@ -90,14 +90,15 @@ class VyperException(Exception):
         for node in self.nodes:
             try:
                 source_annotation = annotate_source_code(
-                    self.source_code,
+                    # add trailing space because EOF exceptions point one char beyond the length
+                    f"{self.source_code} ",
                     node.lineno,
                     node.col_offset,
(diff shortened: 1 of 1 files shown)


iamdefinitelyahuman:
@fubuloubu, this error was shown to me by a user that was dealing with a multiline EOF error as a result of a mismatched parenthesis.  This fix will at least annotate the source, but it still points to the last character which isn't very useful.

A better solution would be to track counts and offsets of `(` and `[` as we're running tokenizer, and then if the error is multi-line EOF, we can instead point at the last unmatched parenthesis.  I'm doing something similar to this in the brownie console so I've thought about it quite recently.  But..  would this be a waste of energy, since once we bring in our own parser the whole tokenizer logic will be no more?

fubuloubu:
🤷‍♂️ 🛳️🚀

howardpen9:
NICE

PR #2140, opened 2020-09-03
Fix issue #1841

saikat041:
### What I did
Fixed issue: ABI issue with bytes and string arrays inside tuples #1841

### How I did it
I modified below line:
https://github.com/vyperlang/vyper/blob/3eb9e7a9bbb69f578c22db2f8a9c89669b95566c/vyper/codegen/return_.py#L69

In this line, we are checking if the return expression is an external function call, and is it is an external functional we are not encoding the data as the return from the external call is already encoded. Here we forgot to add the condition 
``` sub.args[0].args[0].value == "staticcall"```

### How to verify it
Run below test case:
```
def test_bytes_inside_tuple(get_contract):
    code = """
@external
def test_return() -> (String[6], uint256):
    return "potato", 42
    """
    c1 = get_contract(code)

    code = """
interface jsonabi:
    def test_return() -> (String[6], uint256): view

@external
def test_values(a: address) -> (String[6], uint256):
    """

    c2 = get_contract(code)
    assert c2.test_values(c1.address) == ["potato", 42]
```

### DIFF
--- a/tests/functional/codegen/test_tuple_return.py
+++ b/tests/functional/codegen/test_tuple_return.py
@@ -1,0 +1,44 @@
+import pytest
+
+@pytest.mark.parametrize("string", ["a", "abc", "abcde", "potato"])
+def test_string_inside_tuple(get_contract, string):
+    code = f"""
+@external
+def test_return() -> (String[6], uint256):
+    return "{string}", 42
+    """
+    c1 = get_contract(code)
+
+    code = """
+interface jsonabi:
+    def test_return() -> (String[6], uint256): view
+
+@external
+def test_values(a: address) -> (String[6], uint256):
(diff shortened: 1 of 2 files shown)


fubuloubu:
This fix looks great! Saw your comment here https://github.com/vyperlang/vyper/issues/1841#issuecomment-686521075, @iamdefinitelyahuman and I were considering re-architecting the test suite at some point to be easier to understand where to put things. Find a good place for it for now, and we'll move it around later!

fubuloubu (approved):

iamdefinitelyahuman:
Based on how we're looking to re-organize the tests, I'd place the test file in `tests/functional/codegen/`

fubuloubu (approved):

Issue #2142, opened 2020-09-04
Returning struct from a external function gives error


saikat041:
### Version Information
* Vyper version: 0.2.4+commit.3eb9e7a
* OS: Ubuntu
* Python: 3.6

### What's your issue about?
If we try to return a struct from an external function we get compile error. But returning struct from internal function gives no error.
Since we can return tuple from external functions we should also be able to return struct.
```
struct Person:
    name: String[10]
    age: uint256
    
interface jsonabi:
    def test_return() -> Person: view

@external
def test_values(a: address) -> Person:
    return jsonabi(a).test_return()
```

Compiling the above code gives the following error:
```
Error compiling: try1.vy
vyper.exceptions.InvalidType: Can not get size of type, Unexpected type: 'name'
```

fubuloubu:
btw, if you do `fixes: #2142` in your PR, it'll autoclose the issue

saikat041:
@fubuloubu How do I edit the PR? Colud you manually close the issue?

PR #2143, opened 2020-09-04
Fix issue: #2142

saikat041:
### What I did
Fixed the issue #2142 

### How I did it
https://github.com/vyperlang/vyper/blob/ce88755c57e32fbce39658e28cc9bb31b9e69080/vyper/parser/external_call.py#L92
Changed ```types_list = output_type.members ```  to ``` types_list = output_type.tuple_members()```. In the case of **struct**, ``` output_type.members``` returns a dict object, but types_list should be a list of types
### How to verify it
Runt the following test case
```
import pytest

@pytest.mark.parametrize("string", ["a", "abc", "abcde", "potato"])
def test_string_inside_tuple(get_contract, string):
    code = f"""
struct Person:
    name: String[6]
    age: uint256

@external
def test_return() -> Person:
    return Person({{ name:"{string}", age:42 }})
    """
    c1 = get_contract(code)

    code = """
struct Person:
    name: String[6]
    age: uint256

interface jsonabi:
    def test_return() -> Person: view

@external
def test_values(a: address) -> Person:
    """

    c2 = get_contract(code)
    assert c2.test_values(c1.address) == [string, 42]
```

### DIFF
--- a/tests/functional/codegen/test_struct_return.py
+++ b/tests/functional/codegen/test_struct_return.py
@@ -1,0 +1,31 @@
+import pytest
+
+@pytest.mark.parametrize("string", ["a", "abc", "abcde", "potato"])
+def test_string_inside_tuple(get_contract, string):
+    code = f"""
+struct Person:
+    name: String[6]
+    age: uint256
+
+@external
+def test_return() -> Person:
+    return Person({{ name:"{string}", age:42 }})
+    """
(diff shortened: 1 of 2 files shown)


fubuloubu (approved):

fubuloubu:
Nice!

Issue #2155, opened 2020-09-14
The returned struct value from the external call does get stored properly


saikat041:
### Version Information

* vyper Version (output of `vyper --version`): 0.2.4+commit.3eb9e7a
* OS: linux
* Python Version (output of `python --version`): Python 3.6.9

### What's your issue about?
The returned struct value from the external call does get stored properly. It seems that the problem is due to abi decoding.
```
def test_bytes_inside_tuple(get_contract):
    code = """
struct Person:
    name: String[6]
    age: uint256

@external
def test_return() -> Person:
    return Person({name: "abc", age: 42 })
    """
    c1 = get_contract(code)

    code = """
struct Person:
    name: String[6]
    age: uint256

interface jsonabi:
    def test_return() -> Person: view

@external
def test_values(a: address) -> String[6]:
    p: Person = jsonabi(a).test_return()
    return p.name
    """

    c2 = get_contract(code)
    assert c2.test_values(c1.address) == "abc"
```

The above test case fails

### How can it be fixed?

I have found the cause and I am working on it

fubuloubu:
Good find! Thank you for taking a stab at it

saikat041:
Hey @fubuloubu
https://github.com/vyperlang/vyper/blob/c173ba72e78ba348c204053839c2ddfc866490a1/vyper/codegen/return_.py#L67


 Here we are returning abi-encoded value even for internal function return. I think we should not encode return value when we returning from internal functions.  If we don't encode return value then we don't have to decode using this complex algorithm here,
https://github.com/vyperlang/vyper/blob/c173ba72e78ba348c204053839c2ddfc866490a1/vyper/parser/self_call.py#L193

We should only encode return value for external functions.

PR #2164, opened 2020-09-25
Corrected abi_decode function, used it when compiling function calls and changed how we were handling the tuple/struct return

saikat041:
### What I did
I fixed the issue: #2155 

### How I did it
* Both internal and external function return abi-encoded data. But we were not decoding the return data. So I corrected the abi_decode function and then used it when parsing function calls
https://github.com/vyperlang/vyper/blob/39cbd404f0cfdd7f93ef40738c9a02421fe1f3da/vyper/codegen/abi.py#L408
* abi-decode function was not being used earlier so to pass the test cases of tuple return we were partially decoding tuples in make_setters util function. But now that we are using the abi-decode function properly, so to avoid decoding already decoded data I had to make changes here too.
https://github.com/vyperlang/vyper/blob/39cbd404f0cfdd7f93ef40738c9a02421fe1f3da/vyper/parser/parser_utils.py#L713
* Changed how tuple return statements were being compiled.
https://github.com/vyperlang/vyper/blob/39cbd404f0cfdd7f93ef40738c9a02421fe1f3da/vyper/codegen/return_.py#L67

### How to verify it
Run the following test case:
```
def test_struct_return_external_contract_call_2(get_contract_with_gas_estimation):
    contract_1 = """
struct X:
    x: int128
    y: String[6]
@external
def out_literals() -> X:
    return X({x: 1, y: "abcdef"})
    """

    contract_2 = """
struct X:
    x: int128
    y: String[6]
interface Test:
    def out_literals() -> X : view

@external
def test(addr: address) -> (int128, String[6]):
    ret: X = Test(addr).out_literals()
    return ret.x, ret.y

    """
    c1 = get_contract_with_gas_estimation(contract_1)
    c2 = get_contract_with_gas_estimation(contract_2)

    assert c1.out_literals() == [1, "abcdef"]
    assert c2.test(c1.address) == list(c1.out_literals())
```

### DIFF
--- a/tests/functional/codegen/test_struct_return.py
+++ b/tests/functional/codegen/test_struct_return.py
@@ -1,5 +1,6 @@
 import pytest
 
+
 @pytest.mark.parametrize("string", ["a", "abc", "abcde", "potato"])
 def test_string_inside_tuple(get_contract, string):
     code = f"""
@@ -28,4 +29,3 @@ def test_values(a: address) -> Person:
 
     c2 = get_contract(code)
     assert c2.test_values(c1.address) == [string, 42]
-
--- a/tests/functional/codegen/test_tuple_return.py
+++ b/tests/functional/codegen/test_tuple_return.py
@@ -1,5 +1,6 @@
 import pytest
 
+
 @pytest.mark.parametrize("string", ["a", "abc", "abcde", "potato"])
 def test_string_inside_tuple(get_contract, string):
     code = f"""
--- a/tests/parser/features/external_contracts/test_external_contract_calls.py
+++ b/tests/parser/features/external_contracts/test_external_contract_calls.py
@@ -829,34 +829,37 @@ def test(addr: address) -> (int128, address):
     assert c2.test(c1.address) == list(c1.out_literals())
 
 
-def test_struct_return_external_contract_call_2(get_contract_with_gas_estimation):
-    contract_1 = """
+@pytest.mark.parametrize("i,ln,s,", [(100, 6, "abcde"), (41, 40, "a" * 34), (57, 70, "z" * 68)])
+def test_struct_return_external_contract_call_2(get_contract_with_gas_estimation, i, ln, s):
+    contract_1 = f"""
 struct X:
     x: int128
-    y: String[6]
+    y: String[{ln}]
+    z: Bytes[{ln}]
 @external
-def out_literals() -> X:
-    return X({x: 1, y: "abcdef"})
+def get_struct_x() -> X:
+    return X({{x: {i}, y: "{s}", z: b"{s}"}})
     """
 
-    contract_2 = """
+    contract_2 = f"""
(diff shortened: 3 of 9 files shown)


fubuloubu:
@charles-cooper this mucks with your abi generator code, can you take a look?

charles-cooper (approved):
@saikat041 I think this is a good approach. Thanks for fixing some things and simplifying the code structure. Since this modifies the return format for some kinds of things, I think we could do with some more tests though. Would you be able to add some tests? I think, specifically, we want to test nested dynamic data. For instance,
```python
Struct X1:
  x1: Bytes[10]
Struct Y1:
  y1: X1
```
This should be tested for: external call, self-call to public, self-call to private. We also might want to test bytes with len > 32 because those cases are sometimes handled differently. Maybe @fubuloubu can help with this.

Also, could you please change the title of the pull request to summarize what the PR does? Thanks again!!!

fubuloubu (approved):
Looks good to me, I appreciate the comments.

Please add like 10ish new cases as @charles-cooper described, then we are good to go here!

saikat041:
@fubuloubu @charles-cooper I have added more test cases for internal/external calls returning struct, But the test cases for nested structs are failing(as before). They are failing because of this function:
https://github.com/vyperlang/vyper/blob/39cbd404f0cfdd7f93ef40738c9a02421fe1f3da/vyper/types/types.py#L267

The corrected version should be something like this:
```
def get_size_of_type(typ):
    if isinstance(typ, BaseType):
        return 1
    elif isinstance(typ, ByteArrayLike):
        # 1 word for length,
        # up to maxlen words for actual data.
        return ceil32(typ.maxlen) // 32 + 1
        size = get_size_of_type(typ.subtype) * typ.count
        return size + 1 if has_dynamic_data(typ) else size
    elif isinstance(typ, MappingType):
        raise InvalidType("Maps are not supported for function arguments or outputs.")
    elif isinstance(typ, TupleLike):
        size = sum([get_size_of_type(v) for v in typ.tuple_members()])
        return size + 1 if has_dynamic_data(typ) else size
    else:
        raise InvalidType(f"Can not get size of type, Unexpected type: {repr(typ)}")
```

We are using the above function for calculating offsets of keys in a normal struct but there should be a different function for that. Because normal structs in memory are not stored in abi-encoded form.
https://github.com/vyperlang/vyper/blob/39cbd404f0cfdd7f93ef40738c9a02421fe1f3da/vyper/parser/parser_utils.py#L323

This function is being called in some other places and also being tested in some individual test cases. We have to make changes in those places too if we want to change this function.

I think there should be a separate issue and pull request for changing this get_size_of_type function as this pull request is already making a lot of changes. And also nested structs were failing before this pull request.

fubuloubu (approved):

iamdefinitelyahuman (approved):

PR #2175, opened 2020-09-29
Size limit optimizations

iamdefinitelyahuman:
### What I did
Optimizations around size clamps.

### How I did it
* When compiling for >= constantinople rules, handle sizelimits for `address` and `bool` using `shr` instead of a comparison to the in-memory size limit.
* During LLL optimization, check for unused size limit values. If any are found, remove the initial `mstore` operation to reduce the size of the bytecode.

### How to verify it
Run tests.

### Cute Animal Picture
image

### DIFF
--- a/tests/parser/features/test_clampers.py
+++ b/tests/parser/features/test_clampers.py
@@ -1,6 +1,8 @@
 import pytest
 from eth_utils import keccak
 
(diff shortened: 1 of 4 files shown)


fubuloubu:
Really need to test this one across EVM versions. Does that happen?

fubuloubu (approved):

PR #2176, opened 2020-09-30
Fix: invalid AnnAssign exception

### LINKED ISSUE #2157: Bad exception message when using attribute in AnnAssign
### What's your issue about?
The following code:

```python
struct Foo:
    a: uint256

@external
def test():
    foo: Foo = empty(Foo)
    foo.a: uint256 = 123
```

produces the following exception:

```python
File "/home/computer/code/python/vyper/vyper/context/validation/local.py", line 178, in visit_AnnAssign
    name = node.target.id
```

It is correct that this should not compile, but the raised exception is not useful.

### How can it be fixed?
In `vyper.context.validation.local`, add logic to correctly handle an `AnnAssign` where the left-hand side is an `Attribute` node.

iamdefinitelyahuman:
### What I did
Correctly handle an invalid AnnAssign during function-scoped type checking.

Fixes #2157 

### Cute Animal Picture
image

### DIFF
--- a/vyper/context/validation/local.py
+++ b/vyper/context/validation/local.py
@@ -170,6 +170,10 @@ class FunctionNodeVisitor(VyperNodeVisitorBase):
                 )
 
     def visit_AnnAssign(self, node):
+        name = node.get("target.id")
+        if name is None:
+            raise VariableDeclarationException("Invalid assignment", node)
+
         if not node.value:
             raise VariableDeclarationException(
                 "Memory variables must be declared with an initial value", node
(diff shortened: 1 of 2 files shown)


vyper/context/validation/local.py
```diff
@@ -170,6 +170,10 @@ def __init__(
                 )
 
     def visit_AnnAssign(self, node):
+        name = node.get("target.id")
+        if name is None:
+            raise VariableDeclarationException("Invalid assignment", node)
```
fubuloubu, line 175: Add a test for this?

fubuloubu (approved):

PR #2178, opened 2020-10-04
Improve exception for incorrect arg count

iamdefinitelyahuman:
### What I did
Improve the annotation location and exception message when calling a function with an incorrect number of arguments.

Before:
```python
vyper.exceptions.ArgumentException: Invalid argument count: expected 2, got 3
  function "foo", line 7:4 
       6 def foo(a: address):
  ---> 7     ERC20(a).transfer(a, 2000, 123)
  -----------^
       8
```

After:
```python
vyper.exceptions.ArgumentException: Invalid argument count for call to 'transfer': expected 2, got 3
  function "foo", line 7:31 
       6 def foo(a: address):
  ---> 7     ERC20(a).transfer(a, 2000, 123)
  --------------------------------------^
       8
```

### How I did it
Prior to raising the exception, look at the node in question and target one of it's descendants accordingly.

### Cute Animal Picture
image


fubuloubu (approved):
A couple more branches need tests

PR #2179, opened 2020-10-04
Refactor int128 clamps

iamdefinitelyahuman:
### What I did
Refactor the process of clamping `int128` using bitshifting.

### How I did it
The new check can be expressed as:

```python
assert abs(n) >> 127 == 0
```

This check produces smaller bytecode and uses less gas at runtime.

The logic is handled in a new `int128_clamp` function within `vyper.parser.arg_clamps`. All the locations where clamping previously happened are now calling to this function.

### How to verify it
Run the tests.

I still need to verify how well tested this behavior is.

### Cute Animal Picture
image


fubuloubu (approved):

fubuloubu:
lgtm

Issue #2183, opened 2020-10-07
Incorrect output while passing multi-dimensional array as argument


saikat041:
### Version Information

* vyper Version: 0.2.4+commit.14327e9
* OS: linux
* Python Version: 3.6.9

### What's your issue about?
When we pass a multi-dimensional array like [[1, 2]] as an argument to internal/external functions we get incorrect output.
The following test case fails:
```
def test_multiple_dim_array_as_argument(get_contract):
    code = """
@internal
def test_input(arr: int128[2][1], i: int128) -> (int128[2][1], int128):
    return arr, i

@external
def test_values(arr: int128[2][1], i: int128) -> (int128[2][1], int128):
    return self.test_input(arr, i)
    """

    c = get_contract(code)
    assert c.test_values([[1, 2]], 3) == [[[1, 2]], 3]
```

### How can it be fixed?
Working on it

fubuloubu:
Good find!

PR #2184, opened 2020-10-08
fix-#2183: Incorrect output while passing multi-dimensional array as argument

saikat041:
### What I did
Fixed: #2183 

### How I did it
In this line it is assumed that size of each subtype of a array/struct is 32, which is wrong:
https://github.com/vyperlang/vyper/blob/51158673c394f5b3decb4fbd5f88eca98c1b2b33/vyper/parser/parser_utils.py#L505
So I replaced the above line with this.
```
staticarray_offset += 32 * (get_size_of_type(typ) - 1)
```
And since the variable count is not needed anymore we can avoid its calculation here:
https://github.com/vyperlang/vyper/blob/51158673c394f5b3decb4fbd5f88eca98c1b2b33/vyper/parser/parser_utils.py#L501

### DIFF
--- a/tests/parser/types/test_lists.py
+++ b/tests/parser/types/test_lists.py
@@ -273,7 +273,7 @@ def test_values(arr: int128[2][1], i: int128) -> (int128[2][1], int128):
     assert c.test_values([[1, 2]], 3) == [[[1, 2]], 3]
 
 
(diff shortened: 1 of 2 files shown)


iamdefinitelyahuman (approved):
Good catch!

fubuloubu (approved):

PR #2189, opened 2020-10-10
Improve exception for invalid use of dynamically sized struct

iamdefinitelyahuman:
### What I did
Raise `ArgumentException` when using a dynamically-sized struct as a function input.

### How I did it
* Added `is_dynamic_size` attribute to all type objects within `vyper.context`
* When generating the function signature, raise if an input is a struct with dynamic size.

### How to verify it
Run the tests. I added some new cases.

### Cute Animal Picture
image

### DIFF
--- a/vyper/context/types/function.py
+++ b/vyper/context/types/function.py
@@ -7,6 +7,7 @@ from vyper.ast.validation import validate_call_args
 from vyper.context.namespace import get_namespace
 from vyper.context.types.bases import BaseTypeDefinition, DataLocation
 from vyper.context.types.indexable.sequence import TupleDefinition
+from vyper.context.types.meta.struct import StructDefinition
 from vyper.context.types.utils import (
     StringEnum,
     check_constant,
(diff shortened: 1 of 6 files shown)


vyper/context/types/meta/struct.py
```diff
@@ -28,6 +28,10 @@ def __init__(
         for key, type_ in members.items():
             self.add_member(key, type_)
 
+    @property
+    def is_dynamic_size(self):
+        return next((i for i in self.members.values() if i.is_dynamic_size), False)
```
fubuloubu, line 33: Would this work?
```suggestion
        return any(i.is_dynamic_size for i in self.members.values())
```

vyper/context/types/indexable/sequence.py
```diff
@@ -107,6 +111,10 @@ def __init__(self, value_type: Tuple[BaseTypeDefinition, ...]) -> None:
     def __repr__(self):
         return self._id
 
+    @property
+    def is_dynamic_size(self):
+        return next((i for i in self.self.value_type if i.is_dynamic_size), False)
```
fubuloubu, line 116: Would this work?
```suggestion
        return any(i.is_dynamic_size for i in self.self.value_type)
```

vyper/context/types/meta/struct.py
iamdefinitelyahuman: good call

vyper/context/types/function.py
```diff
@@ -290,6 +291,11 @@ def from_FunctionDef(
             type_definition = get_type_from_annotation(
                 arg.annotation, location=DataLocation.CALLDATA, is_immutable=True
             )
+            if isinstance(type_definition, StructDefinition) and type_definition.is_dynamic_size:
+                raise ArgumentException(
```
fubuloubu, line 295: Add a comment "temporary, until dynamic sized struct elements are supported"

fubuloubu (approved):

fubuloubu (approved):

Issue #1850, opened 2020-02-09
Raise a warning when exceeding code size limit (EIP-170)


iamdefinitelyahuman:
### What's your issue about?
The compiler should raise a warning when size of the generated runtime bytecode is >24577 bytes ([EIP-170](https://github.com/ethereum/EIPs/issues/170)).

Because other chains may have different / no limits, we should not raise an exception. But I think for the majority of users, a warning will save some debugging headaches.

fubuloubu:
Just to clarify, is it the deployment bytecode or the runtime bytecode that's affected by this size?

I believe it's the runtime, but should be sure of that. Error message should probably give some of this context too e.g. constructor doesn't count towards codesize

iamdefinitelyahuman:
Yes, 100% it's based on runtime bytecode.

fubuloubu:
close in favor of #2150

PR #2208, opened 2020-10-25
raise warning when EIP-170 contract size limit is exceeded

### LINKED ISSUE #2150: Add a warning when exceeding the EIP-170 Contract Size Limit
Right now, there is no indication that the contract size limit (`2**14 + 2**13 = 24576 bytes`) is exceeded when compiling a contract.

I would suggest raising a warning when this happens, because an error may be just as confusing for newcomers, and it's important to know if the contract is compile-able at all. The warning should also give suggestions about how to fix the problem (reducing interactions in the constructor, reducing state variables, breaking apart contracts, etc.)

As an additional aid, we could also raise this warning if you are within 5% of the limit, so that the user is aware for further iterations on a contract that limited space is left.

lr4d:
### What I did
Fixes #2150 

Still to-do according to https://github.com/vyperlang/vyper/issues/2150#issue-696958448:
>  The warning should also give suggestions about how to fix the problem (reducing interactions in the constructor, reducing state variables, breaking apart contracts, etc.)

I'm not too familiar with the internals yet, so would appreciate suggestions on the suggestions to be provided.

### Cute Animal Picture


tests/compiler/test_bytecode.py
```diff
@@ -0,0 +1,16 @@
+import pytest
+
+import vyper
+
+
+def test_contract_size_exceeded():
```
iamdefinitelyahuman, line 6: This test makes an assumption about the bytecode size of a function, which could very well change in the future.  I think a better approach would be to assign or return a 24kb literal value.

vyper/compiler/output.py
```diff
@@ -132,7 +134,19 @@ def _compress_source_map(code, pos_map, jump_map, source_id):
...
 
+# EIP-170. Ref: https://eips.ethereum.org/EIPS/eip-170
+EIP_CONTRACT_SIZE_LIMIT: int = 2 ** 14 + 2 ** 13
+
+
 def build_bytecode_output(compiler_data: CompilerData) -> str:
+    compiled_bytecode_length = len(compiler_data.bytecode)
+    if compiled_bytecode_length > EIP_CONTRACT_SIZE_LIMIT:
+        warnings.warn(
+            f"Length of compiled bytecode is bigger than Ethereum contract size limit: "
```
iamdefinitelyahuman, line 145: A specific reference to EIP170 would be good here, so users can more easily search out information about the limit.

vyper/warnings.py
```diff
@@ -0,0 +1,3 @@
+# TODO: Create VyperWarning class similarly to what is being done with exceptinos?
```
iamdefinitelyahuman, line 1: I like this TODO :)

vyper/compiler/output.py
```diff
@@ -132,7 +134,19 @@ def _compress_source_map(code, pos_map, jump_map, source_id):
     return compressed_map
 
 
+# EIP-170. Ref: https://eips.ethereum.org/EIPS/eip-170
+EIP_CONTRACT_SIZE_LIMIT: int = 2 ** 14 + 2 ** 13
+
+
 def build_bytecode_output(compiler_data: CompilerData) -> str:
```
iamdefinitelyahuman, line 141: This logic belongs in `build_bytecode_runtime_output` I believe. The limit isn't applied to the creation bytecode, only to what gets deployed on-chain.

tests/compiler/test_bytecode.py
lr4d: Would you suggest to assign that 24kb and then run `compile_code` or just test that passing a bytearray of that size to `build_bytecode_runtime_output` raises the warning?

iamdefinitelyahuman: Probably better to use compileable code, in case the API changes in the future.  Just a very minimal contract that's basically..

iamdefinitelyahuman (approved):
Thanks! :rocket:

vyper/compiler/output.py
```diff
@@ -136,7 +138,20 @@ def build_bytecode_output(compiler_data: CompilerData) -> str:
     return f"0x{compiler_data.bytecode.hex()}"
 
 
+# EIP-170. Ref: https://eips.ethereum.org/EIPS/eip-170
+EIP_CONTRACT_SIZE_LIMIT: int = 2 ** 14 + 2 ** 13
```
fubuloubu, line 142: Minor nit
```suggestion
EIP170_CONTRACT_SIZE_LIMIT: int = 2 ** 14 + 2 ** 13
```

vyper/compiler/output.py
```diff
@@ -136,7 +138,20 @@ def build_bytecode_output(compiler_data: CompilerData) -> str:
...
+
+
 def build_bytecode_runtime_output(compiler_data: CompilerData) -> str:
+    compiled_bytecode_runtime_length = len(compiler_data.bytecode_runtime)
+    if compiled_bytecode_runtime_length > EIP_CONTRACT_SIZE_LIMIT:
+        warnings.warn(
+            f"Length of compiled bytecode is bigger than Ethereum contract size limit "
+            "(see EIP-170: https://eips.ethereum.org/EIPS/eip-170): "
+            f"{compiled_bytecode_runtime_length} > {EIP_CONTRACT_SIZE_LIMIT}",
```
fubuloubu, line 151:
```suggestion
            f"{compiled_bytecode_runtime_length}b > {EIP_CONTRACT_SIZE_LIMIT}b",
```

vyper/warnings.py
```diff
@@ -0,0 +1,3 @@
...
+class ExceedsContractSizeLimitWarning(Warning):
```
fubuloubu, line 2: Seems perhaps too long of a name?
```suggestion
class ContractSizeLimitWarning(Warning):
```

fubuloubu (approved):
Looks good! Some minor nits

vyper/warnings.py
lr4d: clarity - elegance tradeoff

lr4d:
btw you will also want to close https://github.com/vyperlang/vyper/issues/1850, which is a duplicate of #2150

fubuloubu:
@lr4d thanks for seeing that!

PR #2211, opened 2020-10-30
Adapted Simple auction test assertion

tacitvenom:
Current highest bid is 1
(I think, assertion still fails because 0 quantity bid is also not allowed)

### What I did
Adapted a test with correct assertions
### How I did it
-
### How to verify it
Assertion result doesn't change, so it boils down to code review
Adapted a test with correct assertions about bids same as current highest bid are not allowed
### Cute Animal Picture


fubuloubu:
@tacitvenom there's a very weird error with the CI we are looking into

fubuloubu (approved):

PR #2232, opened 2020-11-27
Add support for `not in` comparator

### LINKED ISSUE #2218: no support for `not in` comparator
### Version Information

* vyper Version (output of `vyper --version`): `0.2.7`
* Python Version (output of `python --version`): `3.8.5`

### What's your issue about?
The `not in` comparator is not supported. The following fails to compile with a `SyntaxException`:

```python
@external
def foo(a: address[3]):
    if self not in a:
        pass
```

However, the desired behavior is possible by instead using the `not` unary operator with the `in` comparator. The following example compiles:

```python
@external
def foo(a: address[3]):
    if not self in a:
        pass
```

### How can it be fixed?
Add support for `not in`.

iamdefinitelyahuman:
### What I did
Add support for the `not in` comparator

Closes #2218 

### How I did it
* Added the `NotIn` ast node
* Expand the list comparator logic in `parser` to support `not in`

### How to verify it
Run the tests. I added some new cases to test this behaviour.

### Cute Animal Picture
image

### DIFF
--- a/tests/grammar/vyper.lark
+++ b/tests/grammar/vyper.lark
@@ -230,6 +230,7 @@ _IN: "in"
            | comparator _LE atom ->  le
            | comparator _GE atom ->  ge
            | comparator _IN atom ->  in
+           | comparator _NOT _IN atom ->  in
 
 // NOTE: Must end recursive cycle like this (with `atom` calling `operation`)
 ?atom: variable_access
--- a/vyper/ast/nodes.py
+++ b/vyper/ast/nodes.py
@@ -1090,6 +1090,14 @@ class In(VyperNode):
         return left in right
 
 
+class NotIn(VyperNode):
+    __slots__ = ()
+    _description = "exclusion"
+
+    def _op(self, left, right):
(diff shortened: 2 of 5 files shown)


tests/grammar/vyper.lark
```diff
@@ -230,6 +230,7 @@ _IN: "in"
            | comparator _LE atom ->  le
            | comparator _GE atom ->  ge
            | comparator _IN atom ->  in
+           | comparator _NOT _IN atom ->  in
```
fubuloubu, line 233: This doesn't seem right lol

vyper/ast/nodes.py
```diff
@@ -1090,6 +1090,14 @@ def _op(self, left, right):
         return left in right
 
 
+class NotIn(VyperNode):
+    __slots__ = ()
+    _description = "membership"
```
fubuloubu, line 1095: "severance" or "exclusion"

fubuloubu (approved):

PR #2234, opened 2020-11-28
Allow `empty` as an argument in a function call

### LINKED ISSUE #2112: empty(Bytes[1024]) causes vyper.exceptions.CompilerPanic: Unsupported location: None
### Version Information

* vyper Version (output of `vyper --version`): 0.2.3+commit.006968f
* OS: Osx

I am trying to mimic solidity's `new bytes(0)`

and looks like `empty(Bytes[1024])` causes this exception:

vyper.exceptions.CompilerPanic: Unsupported location: None Please create an issue.vyper.exceptions.CompilerPanic: Unsupported location: None Please create an issue.

using `b''` works great, I wonder, is there a significant gas diff between using `b''` and `empty(Bytes[1024])`?

iamdefinitelyahuman:
### What I did
Allow use of `empty(Bytes[n])` as an argument within a function call.

Closes #2112

### How I did it
* In `vyper.parser.parser_utils::pack_arguments`, add a branch of logic to handle `empty`.  Because empty implies a length of 0 bytes, the logic here is very simple: add 64 empty bytes from the end of calldata.
* As an optimization, I'm also applying the same logic for literal `b""` and `""`.  The overhead of the bytes copier is unnecessary in these situations.
* I've added an optimization where the dynamic pointer is not updated after storing the final dynamic type.

### How to verify it
Run the tests.  I added some cases.

### Cute Animal Picture
image

### DIFF
--- a/tests/parser/functions/test_empty.py
+++ b/tests/parser/functions/test_empty.py
@@ -538,3 +538,51 @@ def baz():
 )
 def test_clear_typecheck(contract, get_contract, assert_compile_failed):
     assert_compile_failed(lambda: get_contract(contract), TypeMismatch)
+
+
+@pytest.mark.parametrize(
+    "a,b,expected",
+    [
+        ("empty(Bytes[65])", "b'hello'", [b"hello", b""]),
+        ("b'hello'", "empty(Bytes[33])", [b"", b"hello"]),
+        (
+            "empty(Bytes[65])",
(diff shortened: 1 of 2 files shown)


fubuloubu:
you shut your damn mouth codecov-io bot!

fubuloubu (approved):

PR #2243, opened 2020-12-07
`canonical_type` and `method_id`

iamdefinitelyahuman:
### What I did
Add `canonical_type` and `method_id` property methods to type objects within `vyper.context`

This is a small step toward annotating the AST prior to LLL generation.

### How I did it
Fairly self explanatory, see code.

### Cute Animal Picture
image

### DIFF
--- a/vyper/context/types/bases.py
+++ b/vyper/context/types/bases.py
@@ -188,6 +188,7 @@ class BaseTypeDefinition:
     """
 
     is_dynamic_size = False
+    _id: str
 
     def __init__(
         self,
@@ -198,6 +199,13 @@ class BaseTypeDefinition:
         self.location = location
         self.is_immutable = is_immutable
         self.is_public = is_public
+
+    @property
+    def canonical_type(self) -> str:
+        """
+        The canonical name of this type. Used for ABI types and generating function signatures.
+        """
+        return self._id
 
     def from_annotation(self, node: vy_ast.VyperNode, **kwargs: Any) -> None:
         # always raises, user should have used a primitive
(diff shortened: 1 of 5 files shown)


vyper/context/types/bases.py
```diff
@@ -199,6 +200,13 @@ def __init__(
         self.is_immutable = is_immutable
         self.is_public = is_public
 
+    @property
+    def canonical_type(self) -> str:
```
fubuloubu, line 204: So, this is what will generate ABI types in type strings? (e.g. `vyper/codegen/abi.py`)

vyper/context/types/function.py
```diff
@@ -352,6 +353,21 @@ def from_AnnAssign(cls, node: vy_ast.AnnAssign) -> "ContractFunction":
             state_mutability=StateMutability.NONPAYABLE,
         )
 
+    @property
+    def canonical_type(self) -> str:
```
fubuloubu, line 357: Not sure if this is a meaningful method for a function to implement? It shouldn't be used in ABI encoding as a type :thinking:

vyper/context/types/bases.py
iamdefinitelyahuman: Indeed.  And needed to form function signatures.

vyper/context/types/function.py
iamdefinitelyahuman: It's just an intermediate step on the way to `method_id`.  It could have a different name, or we could handle the entire logic within `method_id`.

fubuloubu: I think for this node type it may not make sense to have this, should probably throw `UnimplementedError` instead

fubuloubu (approved):

PR #2244, opened 2020-12-07
Refactor: function-level type checks

iamdefinitelyahuman:
### What I did
Move several function-level type checks out of `parser` and into `context`:

* Visibility of `__default__`
* Use of args in `__default__`
* Use of default values in `__init__`

### How I did it
Copy-paste, mostly.  See the diff - it's fairly self explanatory.

### How to verify it
Run the test suite.  Two of these behaviors were already well tested, I've added a test case for the 3rd.

### Cute Animal Picture
image


fubuloubu (approved):
So cleannnn

PR #2245, opened 2020-12-08
refactor: remove origcode member from Context

iamdefinitelyahuman:
### What I did
Remove the `origcode` attribute from the `Context` object.

This attribute was not in use anywhere. Removing it helps to reduce the size of the input arguments in several `parser` functions - a small step toward refactoring,

### How I did it
See diff.

### How to verify it
Confirm that the existing test suite still passes.

### Cute Animal Picture
image

### DIFF
--- a/vyper/parser/context.py
+++ b/vyper/parser/context.py
@@ -26,7 +26,6 @@ class Context:
         constancy=Constancy.Mutable,
         is_internal=False,
         is_payable=False,
-        origcode="",
         method_id="",
         sig=None,
     ):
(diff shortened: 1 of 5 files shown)


vyper/parser/context.py
```diff
@@ -48,8 +47,6 @@ def __init__(
         self.in_range_expr = False
         # Is the function payable?
         self.is_payable = is_payable
-        # Original code (for error pretty-printing purposes)
-        self.origcode = origcode
```
fubuloubu, line 52: Basically, this is no longer useful because all errors are caught upstream in context?

fubuloubu (approved):

vyper/parser/context.py
iamdefinitelyahuman: Correct.

PR #2277, opened 2021-01-19
Add `size_in_bytes` to new type objects

iamdefinitelyahuman:
### What I did
Add a `size_in_bytes` attribute to each of the new type objects within `vyper.context`

This is a small but significant piece required to phase out the old type objects in favor of the new.

### How I did it
Self explanatory.

### How to verify it
Run the test suite.

### Cute Animal Picture
image

### DIFF
--- a/tests/functional/context/types/test_size_in_bytes.py
+++ b/tests/functional/context/types/test_size_in_bytes.py
@@ -1,0 +1,49 @@
+import pytest
+
+from vyper.context.types.bases import DataLocation
+from vyper.context.types.utils import get_type_from_annotation
+
+BASE_TYPES = ["int128", "uint256", "bool", "address", "bytes32"]
+ARRAY_VALUE_TYPES = ["String", "Bytes"]
+LOCATIONS = [DataLocation.STORAGE, DataLocation.MEMORY, DataLocation.STORAGE]
+
+
+@pytest.mark.parametrize("type_str", BASE_TYPES)
+@pytest.mark.parametrize("location", LOCATIONS)
+def test_base_types(build_node, type_str, location):
+    node = build_node(type_str)
+    type_definition = get_type_from_annotation(node, location)
+
+    assert type_definition.size_in_bytes == 32
+
+
+@pytest.mark.parametrize("type_str", ARRAY_VALUE_TYPES)
+@pytest.mark.parametrize("location", LOCATIONS)
+@pytest.mark.parametrize("length,size", [(1, 64), (32, 64), (33, 96), (86, 128)])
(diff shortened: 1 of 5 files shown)


vyper/context/types/value/array_value.py
```diff
@@ -59,6 +60,10 @@ def length(self):
             return self._length
         return self._min_length
 
+    @property
+    def size_in_bytes(self):
+        return 32 + math.ceil(self.length / 32) * 32
```
fubuloubu, line 65: Add comments about why the literal values are there

fubuloubu (approved):

vyper/context/types/value/array_value.py
```diff
@@ -62,6 +62,8 @@ def length(self):
 
     @property
     def size_in_bytes(self):
+        # the first slot (32 bytes) stores the actual length, and then we reserve
+        # enough additional slots to store the data if it uses the max available length
```
fubuloubu, line 66:
```suggestion
        # enough additional slots to store the data if it uses the max available length
        # because this data type is single-bytes, we make it so it takes the max 32 byte
        # boundary as it's size, instead of giving it a size that is not cleanly divisble by 32
```

tests/functional/context/types/test_size_in_bytes.py
```diff
@@ -0,0 +1,44 @@
...
+@pytest.mark.parametrize("type_str", BASE_TYPES)
+def test_base_types(build_node, type_str):
+    node = build_node(type_str)
+    type_definition = get_type_from_annotation(node, DataLocation.STORAGE)
+
+    assert type_definition.size_in_bytes == 32
+
+
+@pytest.mark.parametrize("type_str", ARRAY_VALUE_TYPES)
+@pytest.mark.parametrize("length,size", [(1, 64), (32, 64), (33, 96)])
```
fubuloubu, line 19: Also do size 0?

tests/functional/context/types/test_size_in_bytes.py
```diff
@@ -0,0 +1,44 @@
...
+from vyper.context.types.utils import get_type_from_annotation
+
+BASE_TYPES = ["int128", "uint256", "bool", "address", "bytes32"]
+ARRAY_VALUE_TYPES = ["String", "Bytes"]
+
+
+@pytest.mark.parametrize("type_str", BASE_TYPES)
+def test_base_types(build_node, type_str):
+    node = build_node(type_str)
+    type_definition = get_type_from_annotation(node, DataLocation.STORAGE)
```
fubuloubu, line 13: Why not memory too?

fubuloubu (approved):

PR #2279, opened 2021-01-21
Add `event_id` to new Event type objects

iamdefinitelyahuman:
### What I did
Add `event_id` to new Event type objects.

### How I did it
See code, should be fairly easy to understand.

### How to verify it
Run the test suite.  I added some new tests to verify the behaviour.

### Cute Animal Picture
image

### DIFF
--- a/tests/functional/context/types/test_event.py
+++ b/tests/functional/context/types/test_event.py
@@ -1,0 +1,80 @@
+import pytest
+
+from vyper.context.types.meta.event import Event
+
+EVENT_ID_TESTS = [
+    (
+        "event MyLog: pass",
+        # MyLog()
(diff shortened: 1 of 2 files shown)


tests/functional/context/types/test_event.py
```diff
@@ -0,0 +1,72 @@
...
+    ),
+]
+
+
+@pytest.mark.parametrize("source,signature", EVENT_ID_TESTS)
+def test_event_id(build_node, source, signature):
+    node = build_node(source)
+    event = Event.from_EventDef(node)
+
+    event_id = int(keccak256(signature.encode()).hex(), 16)
```
fubuloubu, line 70: Usually with these types of tests, it's better to hardcode the actual expected hash, to show you are getting what you expect to get.

fubuloubu (approved):

PR #2285, opened 2021-01-27
Refactor interface outputs

iamdefinitelyahuman:
### What I did
Refactor the `interface` and `external_interface` compiler outputs.

### How I did it
The old logic used `EventSignature` and `FunctionSignature`. The new logic is nearly identical but uses the new `InterfacePrimitive` type. This is another chunk of the `signatures` module refactored out.

### How to verify it
Run the test suite, confirm everything still passes.

### Cute Animal Picture
image

### DIFF
--- a/vyper/compiler/output.py
+++ b/vyper/compiler/output.py
@@ -7,12 +7,10 @@ from vyper import compile_lll, opcodes
 from vyper.ast import ast_to_dict, parse_natspec
 from vyper.compiler.phases import CompilerData
 from vyper.compiler.utils import build_gas_estimates
+from vyper.context.types.function import FunctionVisibility, StateMutability
 from vyper.parser.lll_node import LLLnode
 from vyper.signatures import sig_utils
-from vyper.signatures.interface import (
-    extract_external_interface,
-    extract_interface_str,
-)
+from vyper.signatures.interface import extract_external_interface
 from vyper.warnings import ContractSizeLimitWarning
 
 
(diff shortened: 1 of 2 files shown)


vyper/compiler/output.py
```diff
@@ -35,11 +33,47 @@ def build_userdoc(compiler_data: CompilerData) -> dict:
...
 
 def build_interface_output(compiler_data: CompilerData) -> str:
-    return extract_interface_str(compiler_data.global_ctx)
+    interface = compiler_data.vyper_module_folded._metadata["type"]
+    out = ""
+
+    if interface.events:
+        out = "# Events\n\n"
+        for event in interface.events.values():
+            out = f"{out}event {event.name}:\n"
```
fubuloubu, line 58: why not `out += f"event {event.name}:\n"`?

fubuloubu (approved):

vyper/compiler/output.py
iamdefinitelyahuman: https://softwareengineering.stackexchange.com/questions/90203/why-is-so-bad-for-concatenation

fubuloubu: hmm, I wonder if this could be refactored using `str.join(*iterable*)`?

https://stackoverflow.com/questions/1316887/what-is-the-most-efficient-string-concatenation-method-in-python

fubuloubu: maybe:
```python
out = "# Events\n\n" + \
    "\n\n".join(
        "".join(
            f"event {event.name}:\n    ",
            (
            "\n    ".join(
                f"{name}: {typ}"
                for name, typ in event.arguments.items()
            )
        )
        if event.arguments
        else "pass"
        for event in interface.events.values()
    )

```

fubuloubu:
```python
        out = "# Events\n\n"
        for event in interface.events.values():
            encoded_args = "\n    ".join(f"{name}: {typ}" for name, typ in event.arguments.items())
            out = f"{out}event {event.name}:\n    {encoded_args if event.arguments else 'pass'}\n"
```

PR #2286, opened 2021-01-28
Move reserved keywords check into the type checker

iamdefinitelyahuman:
### What I did
Move reserved keyword checks into the type checker.

### How I did it
The logic was already partially being handled within type checking: shadowing of builtins and environment variables would raise upon attempted assignment.  However, reserved keywords and opcodes weren't caught until later when generating the `GlobalContext` object.

I've moved the reserved keywords list into `namespace` and am now applying the check on all new assignments as they're added to the `Namespace` object.  For object members (values in events, structs, interfaces) I added a `validate_identifier` method that's slightly looser: it doesn't raise on collisions between non-builtins, so that e.g `MyStruct.foo` and `self.foo` won't collide.

### How to verify it
Run the tests.  I had to modify a few tests in order to make this work, because the exception being raised in some places has changed.

### Cute Animal Picture
image


vyper/context/namespace.py
```diff
@@ -94,3 +96,93 @@ def get_namespace():
...
+    if attr in namespace and attr not in [x for i in namespace._scopes for x in i]:
+        raise NamespaceCollision(f"Cannot assign to '{attr}', it is a builtin")
+    if attr.lower() in RESERVED_KEYWORDS or attr.upper() in OPCODES:
+        raise StructureException(f"'{attr}' is a reserved keyword")
+    if not re.match("^[_a-zA-Z][a-zA-Z0-9_]*$", attr):
+        raise StructureException(f"'{attr}' contains invalid character(s)")
+
+
+# Cannot be used for variable or member naming
+RESERVED_KEYWORDS = {
```
fubuloubu, line 112: it would be lovely if we can somehow pull this from our stuff dyanmically, instead of statically maintaining this alongside changes to the compiler

fubuloubu (approved):

vyper/context/namespace.py
iamdefinitelyahuman: Definitely worth implementing.

PR #2292, opened 2021-01-29
Refactor ContractFunction

iamdefinitelyahuman:
### What I did
Adjustments to the `ContractFunction` api to aid readability as I refactor out `FunctionSignature`.

### How I did it
* Add `is_constructor` and `is_fallback` property methods. This replaces `func.name == "__init__"`  comparisons.  More readable and makes it easier to refactor in the future.
* Replace `arg_count` with `min_arg_count` and `max_arg_count`.  The prior was a sometimes-int, sometimes-tuple and led to repeated `if isinstance(arg_count, tuple)` checks which aren't immediately obvious to the reader what's happening.
* Add `has_default_args` property method to check if max args > min args

### How to verify it
Check that tests are passing.

### Cute Animal Picture
image

### DIFF
--- a/vyper/context/types/function.py
+++ b/vyper/context/types/function.py
@@ -1,6 +1,6 @@
 import warnings
 from collections import OrderedDict
-from typing import Any, Dict, List, Optional, Tuple, Union
+from typing import Any, Dict, List, Optional, Tuple
 
 from vyper import ast as vy_ast
 from vyper.ast.validation import validate_call_args
@@ -71,9 +71,11 @@ class ContractFunction(BaseTypeDefinition):
         The name of the function.
     arguments : OrderedDict
         Function input arguments as {'name': BaseType}
-    arg_count : Tuple[int, int] | int
-        The number of input arguments. If given as a tuple, the value represents
-        (min, max) when default values are given.
+    min_arg_count : int
+        The minimum number of required input arguments.
+    max_arg_count : int
+        The maximum number of required input arguments. When a function has no
+        default arguments, this value is the same as `min_arg_count`.
     kwarg_keys : List
         List of optional input argument keys.
     function_visibility : FunctionVisibility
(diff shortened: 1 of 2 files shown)


vyper/context/types/function.py
```diff
@@ -279,19 +284,17 @@ def from_FunctionDef(
             raise StructureException("Cannot use reentrancy guard on view or pure functions", node)
 
         # call arguments
-        arg_count: Union[Tuple[int, int], int] = len(node.args.args)
+        min_arg_count = len(node.args.args)
```
fubuloubu, line 287: Move this here, since we know this ahead of time now
```suggestion
        min_arg_count = len(node.args.args) - len(node.args.defaults)
```

vyper/context/types/function.py
```diff
@@ -279,19 +284,17 @@ def from_FunctionDef(
...
         if node.args.defaults:
             if node.name == "__init__":
                 raise FunctionDeclarationException(
                     "Constructor may not use default arguments", node.args.defaults[0]
                 )
-            arg_count = (
-                len(node.args.args) - len(node.args.defaults),
-                len(node.args.args),
-            )
+            min_arg_count -= len(node.args.defaults)
```
fubuloubu, line 294: delete this line with above change

vyper/context/types/function.py
```diff
@@ -279,19 +284,17 @@ def from_FunctionDef(
             raise StructureException("Cannot use reentrancy guard on view or pure functions", node)
 
         # call arguments
-        arg_count: Union[Tuple[int, int], int] = len(node.args.args)
+        min_arg_count = len(node.args.args)
+        max_arg_count = len(node.args.args)
         if node.args.defaults:
             if node.name == "__init__":
```
fubuloubu, line 290: can this use `node.is_constructor` property?

fubuloubu (approved):

vyper/context/types/function.py
iamdefinitelyahuman: no, because this is happening inside a class method we aren't able to use that function

PR #2294, opened 2021-01-30
Expand `method_ids`

iamdefinitelyahuman:
### What I did
Rename `method_ids` to `get_method_id_dict` and add a property method for `method_id`.

The dict for all method IDs is a requirement for certain compiler outputs, however a single "primary" method id is required internally during the IR, to reference the method regardless of default args.  Having this available as a property method is more readable than continually referencing e.g. `list(func.method_ids.values())[-1]`

### How to verify it
Run tests.

### Cute Animal Picture
image

### DIFF
--- a/vyper/context/types/function.py
+++ b/vyper/context/types/function.py
@@ -376,23 +376,13 @@ class ContractFunction(BaseTypeDefinition):
         )
 
     @property
-    def method_ids(self) -> Dict[str, int]:
+    def method_id(self) -> int:
         """
-        Dict of `{signature: four byte selector}` for this function.
-
-        * For functions without default arguments the dict contains one item.
-        * For functions with default arguments, there is one key for each
-          function signfature.
+        The primary method ID, used to represent this function internally
+        regardless of the number of default arguments given.
         """
         arg_types = [i.canonical_type for i in self.arguments.values()]
-
-        if not self.has_default_args:
-            return _generate_method_id(self.name, arg_types)
-
-        method_ids = {}
-        for i in range(self.min_arg_count, self.max_arg_count + 1):
-            method_ids.update(_generate_method_id(self.name, arg_types[:i]))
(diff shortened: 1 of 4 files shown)


vyper/context/types/function.py
```diff
@@ -376,23 +376,13 @@ def from_AnnAssign(cls, node: vy_ast.AnnAssign) -> "ContractFunction":
...
-    def method_ids(self) -> Dict[str, int]:
+    def method_id(self) -> int:
         """
-        Dict of `{signature: four byte selector}` for this function.
-
-        * For functions without default arguments the dict contains one item.
-        * For functions with default arguments, there is one key for each
-          function signfature.
+        The primary method ID, used to represent this function internally
+        regardless of the number of default arguments given.
```
fubuloubu, line 382: I think this needs more thought.

The "default" should be chosen probably based on how it's best represented, and explain why this is the case.

vyper/context/types/function.py
```diff
@@ -376,23 +376,13 @@ def from_AnnAssign(cls, node: vy_ast.AnnAssign) -> "ContractFunction":
...
         arg_types = [i.canonical_type for i in self.arguments.values()]
-
-        if not self.has_default_args:
-            return _generate_method_id(self.name, arg_types)
-
-        method_ids = {}
-        for i in range(self.min_arg_count, self.max_arg_count + 1):
-            method_ids.update(_generate_method_id(self.name, arg_types[:i]))
-        return method_ids
+        return _generate_method_id(self.name, arg_types)[1]
```
fubuloubu, line 385: what's special about 1?

fubuloubu (commented):
I'm not sure I understand the reasoning behind this change

fubuloubu:
maybe consider leaving it as is, and then produce `get_primary_method_id` or something instead

PR #2302, opened 2021-02-15
Streamlined ABI encoder logic for returning tuple types

iamdefinitelyahuman:
### What I did
Simplify the logic around returning tuples, when each member of the tuple is of the same type.

### How I did it
Add a special `if` branch that replicates the logic for returning an array type. This branch is only reached if every member of the tuple has the same type, and that type is a base type (or array of the base type).

This has a not-insignificant effect on the bytecode size.

### How to verify it
Run tests (TODO)

### Cute Animal Picture
Cute aminals are for closers.


fubuloubu (approved):
Add a few tests

fubuloubu (approved):

PR #2304, opened 2021-02-15
Assembly-level optimizations

iamdefinitelyahuman:
### What I did
* find and remove unreachable instructions within the assembly
* merge sequential jumpdests

### How I did it
Added some logic in `assembly_to_evm` within `compile_lll.py`

Both of these situations occur as a result of LLL generation happening in small chunks that don't see the "bigger picture" around them. There is likely a more elegant solution here, but this is effective and simple to implement. I've left comments about what's happening, we can try to address this properly as we refactor out `parser` in the future and give ourselves more effective data structures to catch and fix this as it happens rather than after the fact.

### How to verify it
Run tests, confirm that nothing broke.  If this created any issues we'd see tests failing everywhere, especially within the e2e tests of the examples.

### Cute Animal Picture
image
Notice me senpai!

### DIFF
--- a/vyper/compile_lll.py
+++ b/vyper/compile_lll.py
@@ -517,6 +517,19 @@ def assembly_to_evm(assembly, start_pos=0):
         else:
             i += 1
 
+    # When a nested subroutine finishes and is the final action within it's
+    # parent subroutine, we end up with multiple simultaneous JUMPDEST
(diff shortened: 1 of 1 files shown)


fubuloubu (approved):
Big change, but if it passes the test suite, hopefully should work just fine.

PR #2321, opened 2021-02-22
fix: Var size when using new type objects

iamdefinitelyahuman:
### What I did
In `VariableRecord`, return the correct size when the related type object is one of the new `context` types.

This fixes an issue with memory corruption caused by excessive deallocation when releasing internal variables :(

### Cute Animal Picture
image

### DIFF
--- a/vyper/signatures/function_signature.py
+++ b/vyper/signatures/function_signature.py
@@ -1,3 +1,4 @@
+import math
 from collections import Counter
 
 from vyper import ast as vy_ast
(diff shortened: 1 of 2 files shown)


vyper/signatures/function_signature.py
```diff
@@ -41,7 +42,7 @@ def __init__(
     def size(self):
         if hasattr(self.typ, "size_in_bytes"):
             # temporary requirement to support both new and old type objects
-            return self.typ.size_in_bytes
+            return math.ceil(self.typ.size_in_bytes / 32)
```
fubuloubu, line 45: Add a comment explaining the conversion

fubuloubu (approved):
Add a test case that shows the incorrect behavior is caught

fubuloubu (approved):

PR #2337, opened 2021-03-25
Add a function to change the minter identity

qbzzt:
It could be useful to transfer the privilege to somebody else, for example when the minter leaves a job with the organization that owns the ERC-721 contract. This could also be accomplished by giving the private key for the address to the next person, but if a person created multiple ERC-721 contracts and didn't think about it beforehand they might have to give all of them or none. With this function they can decide to transfer only some contracts.

### Cute Animal Picture

### DIFF
--- a/examples/tokens/ERC721.vy
+++ b/examples/tokens/ERC721.vy
@@ -341,6 +341,18 @@ def mint(_to: address, _tokenId: uint256) -> bool:
     self._addTokenTo(_to, _tokenId)
     log Transfer(ZERO_ADDRESS, _to, _tokenId)
     return True
+    
+    
+@external transferMinterRole(_newMinter: address):
+    """
+    @dev Function to transfer the minter position to a new address
+         Throws if `msg.sender` is not the minter.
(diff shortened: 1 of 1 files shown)


examples/tokens/ERC721.vy
```diff
@@ -341,6 +341,18 @@ def mint(_to: address, _tokenId: uint256) -> bool:
     self._addTokenTo(_to, _tokenId)
     log Transfer(ZERO_ADDRESS, _to, _tokenId)
     return True
+    
+    
+@external transferMinterRole(_newMinter: address):
```
fubuloubu, line 346:
```suggestion
@external
def transferMinterRole(_newMinter: address):
```

qbzzt: Thank you. I should have realized that.

fubuloubu (commented):
Add a test to show the new function operates correctly:
https://github.com/vyperlang/vyper/blob/master/tests/examples/tokens/test_erc721.py

PR #2343, opened 2021-04-05
Add msg.data environment variable VIP-1181

### LINKED ISSUE #1181: VIP: `msg.data`
## Simple Summary
Add support of `msg.data`, which returns the complete calldata.  This enables a proxy for an upgradable pattern.

## Abstract
Add a new global variable `msg.data` similar to [the one of Solidity](https://solidity.readthedocs.io/en/v0.4.24/units-and-global-variables.html#block-and-transaction-properties).

## Motivation
This enables a proxy for an upgradable pattern with which users don't need to take care of the proxy contract. See the example code below.

## Specification
`msg.data` returns a `bytes` which represents a complete calldata using CALLDATA opcode.

This is an example code of a proxy contract. (ref [OpenZeppelin Proxy.sol](https://github.com/zeppelinos/zos/blob/master/packages/lib/contracts/upgradeability/Proxy.sol))

proxy contract
```
implementation: public(address)
owner: public(address)

@public
def __default__():
    assert msg.sender == self.owner
    raw_call(self.implementation, msg.data, outsize=0, gas=0, delegate_call=True)
    
@public
def upgradeTo(_newAddr: address):
    assert msg.sender == self.owner
    self.implementation = _newAddr
```

delegatee contract
```
dummyVal1: address
dummyVal2: address
appData1: public(uint256)
appData2: public(bytes32)


@public
def someFunc(_inp: uint256):
    self.appData1 = _inp
```
(NOTE: `dummyVal`s are there to make up for the slots of the variables used only in the proxy. Instead, the slots of these variables are [explicitly set with inline assembly in Solidity](https://github.com/zeppelinos/zos/blob/master/packages/lib/contracts/upgradeability/UpgradeabilityProxy.sol#L48)). 

## Backwards Compatibility
It's a new feature and backward compatible.

## Dependencies
None

## Copyright

skellet0r:
### What I did

As per #1181, I added the msg.data environment variable, allowing access to the complete calldata within external functions.

Also updated the docs to include information on the syntax/usage of msg.data.

### How I did it

1. Defined the `data` member in the `msg` environment variable - `vyper/context/environment.py`
2. Modified the StatementAnnotationVisitor to attach metadata to the `msg.data` node depending on usage - `vyper/context/validation/annotation.py`
4. Added the `msg.data` branch in the `parse_Attribute` fn, to return the appropriate LLLNode based on whether usage is in `len` or `slice` - `vyper/parser/expr.py`
5. Modified code validation to throw a StateAccessViolation when `msg.data` is used inside an internal function - `vyper/context/validation/local.py`
6. Modified code validation to throw a SyntaxException when `msg.data` is used outside of `len` or `slice`

Similar to before this copies the calldata to an internal variable before it gets sliced, I'm sure in the future we can work on cacheing `msg.data` but that's a whole 'nother beast. The use of `len(msg.data)` simply just used the `calldatasize` opcode and isn't as bad I hope.

### How to verify it

I've added a few tests, which can be found in `tests/parser/syntax/test_msg_data.py`.

Added msg.data syntax allowing access to full calldata in external functions

### Cute Animal Picture

Cute Animal Picture

Closes #1181


.pre-commit-config.yaml
```diff
@@ -1,23 +1,24 @@
...
 
 -   repo: https://gitlab.com/pycqa/flake8
-    rev: 3.8.3
+    rev: 3.9.0
     hooks:
     -   id: flake8
 
 -   repo: https://github.com/pre-commit/mirrors-mypy
-    rev: v0.780
+    rev: v0.812
```
fubuloubu, line 21: Can this (and associated fixes) be a new PR?

tests/parser/syntax/test_msg_data.py
```diff
@@ -0,0 +1,57 @@
...
+
+from vyper import compiler
+from vyper.exceptions import StateAccessViolation
+
+
+def test_msg_data_assignment(get_contract):
+    code = """
+@external
+def foo() -> Bytes[4]:
+    x: Bytes[4] = msg.data[4]
```
fubuloubu, line 12:
```suggestion
    x: Bytes[4] = slice(msg.data, 0, 4)  # Throws if `len(msg.data) < 4`
```

tests/parser/syntax/test_msg_data.py
```diff
@@ -0,0 +1,57 @@
...
+"""
+    contract = get_contract(code)
+    assert contract.foo().hex() == "c2985578"  # fn sig
+
+
+def test_msg_data_call_args(get_contract):
+    code = """
+@external
+def foo(bar: uint256) -> (Bytes[4], uint256):
+    data: Bytes[36] = msg.data[36]
```
fubuloubu, line 23:
```suggestion
    data: Bytes[36] = slice(msg.data, 0, 36)  # Throws if `len(msg.data) < 36`
```

tests/parser/syntax/test_msg_data.py
```diff
@@ -0,0 +1,57 @@
...
+    code = """
+@external
+def foo(bar: uint256) -> Bytes[4]:
+    data: Bytes[4] = msg.data[4]  # msg.data.size > 4 (max_len)
+    return data
+"""
+    contract = get_contract(code)
+
+    with pytest.raises(TransactionFailed):
+        contract.foo(42)
```
fubuloubu, line 43:
```suggestion
@external
def foo(bar: uint256) -> Bytes[4]:
    data: Bytes[4] = slice(msg.data, 0, 4)  # msg.data.size > 4 (max_len)
    return data
"""
    contract = get_contract(code)

    contract.foo(42).hex() == "00000000"  # EVM is big endian
```

tests/parser/syntax/test_msg_data.py
```diff
@@ -0,0 +1,57 @@
...
+
+
+def test_msg_data_internal_fn_usage_denied(get_contract):
+    code = """
+@internal
+def foo() -> Bytes[4]:
+    return msg.data[4]
```
fubuloubu, line 50:
```suggestion
    return slice(msg.data, 0, 4)
```

tests/parser/syntax/test_msg_data.py
fubuloubu: Add a runtime failure case like this:

```python
def foo() -> Bytes[4]:
    return slice(msg.data, 0, 4)  # Throws because len(msg.data) == 0
```

docs/constants-and-vars.rst
```diff
@@ -21,14 +21,19 @@ Name                 Type             Value
 ``block.timestamp``  ``uint256``      Current block epoch timestamp
 ``chain.id``         ``uint256``      Chain ID
 ``msg.gas``          ``uint256``      Remaining gas
+``msg.data[n]``      ``Bytes[n]``     First `n` bytes of message data
```
fubuloubu, line 24:
```suggestion
``msg.data``         ``Bytes``        Message data (must be sliced). Throws if ``len(msg.data)`` less than max slice
```

docs/constants-and-vars.rst
```diff
@@ -21,14 +21,19 @@ Name                 Type             Value
...
 ``msg.sender``       ``address``      Sender of the message (current call)
 ``msg.value``        ``uint256``      Number of wei sent with the message
 ``tx.origin``        ``address``      Sender of the transaction (full call chain)
 ==================== ================ =============================================
 
 .. note::
 
-    ``msg.sender`` and ``msg.value`` can only be accessed from external functions. If you require these values within a private function they must be passed as parameters.
+    ``msg.data[n]``, ``msg.sender`` and ``msg.value`` can only be accessed from external functions. If you require these values within a private function they must be passed as parameters.
```
fubuloubu, line 32:
```suggestion
    ``msg.data``, ``msg.sender`` and ``msg.value`` can only be accessed from external functions. If you require these values within a private function they must be passed as parameters.
```

docs/constants-and-vars.rst
```diff
@@ -21,14 +21,19 @@ Name                 Type             Value
...
+
+.. note::
+
+    Usage of ``msg.data[n]`` requires that ``calldatasize`` <= ``n``, otherwise during runtime the contract will throw.
```
fubuloubu, line 36:
```suggestion
    Usage of ``msg.data`` requires slicing (e.g. ``slice(msg.data, n, m)``), and will throw during runtime if ``calldatasize`` <= ``k``, where ``k`` is the maximum of ``n + m`` for any use of ``slice(msg.data, n, m)``.
```

fubuloubu (commented):
I was thinking the implementation would be slightly different, more like this:

.pre-commit-config.yaml
```diff
@@ -1,23 +1,24 @@
...
     hooks:
     -   id: mypy
```
fubuloubu, line 23: Can this (and associated fixes) be a new PR?

skellet0r:
Ahhh @fubuloubu I like that syntax/implementation better, since it looks like one is explicitly extracting a piece of the calldata, instead of the notation I used which looks like indexing a reference to calldata. I'll get to work on adjusting :+1:

skellet0r:
Sorry for the force push ... not sure if that's accepted around these parts :eyes:
I didn't implement the compile-time bounds checking as you commented @fubuloubu, since there are times in which `calldatasize` will exceed the compile-time expected size. [EIP-2771](https://eips.ethereum.org/EIPS/eip-2771) for example. I also totally missed previously handling `len(msg.data)` so that's included now, as well as the right syntax `slice(msg.data, x, k)`.

docs/constants-and-vars.rst
```diff
@@ -28,7 +29,11 @@ Name                 Type             Value
 
 .. note::
 
-    ``msg.sender`` and ``msg.value`` can only be accessed from external functions. If you require these values within a private function they must be passed as parameters.
+    ``msg.data``, ``msg.sender`` and ``msg.value`` can only be accessed from external functions. If you require these values within a private function they must be passed as parameters.
+
+.. note::
+
+    ``msg.data`` requires the usage of :func:`slice <slice>` to explicitly extract a section of calldata. If the extracted section exceeds the bounds of calldata, this will throw. You can check the size of ``msg.data`` using :func:`len <len>`.
```
fubuloubu, line 36: This check is not implemented, correct?

```suggestion
    ``msg.data`` requires the usage of :func:`slice <slice>` to explicitly extract a section of calldata. You can check the size of ``msg.data`` using :func:`len <len>`.   
```

tests/parser/syntax/test_msg_data.py
```diff
@@ -0,0 +1,110 @@
...
+    return bar
+    """,
+    """
+@internal
+def foo() -> Bytes[4]:
+    return slice(msg.data, 0, 4)
+    """,
+]
+
+exceptions = [SyntaxException, SyntaxException, SyntaxException, StateAccessViolation]
```
fubuloubu, line 88: Merge these together as a list of tuples, I think it's clearer

fubuloubu (approved):
> Sorry for the force push ... not sure if that's accepted around these parts 

No worries, works for me!

> I didn't implement the compile-time bounds checking as you commented @fubuloubu, since there are times in which `calldatasize` will exceed the compile-time expected size. [EIP-2771](https://eips.ethereum.org/EIPS/eip-2771) for example. I also totally missed previously handling `len(msg.data)` so that's included now, as well as the right syntax `slice(msg.data, x, k)`.

This should be okay! I think if you try to pull calldata longer than `CALLDATASIZE` the EVM's behavior is to give you empty data. So, it should be safe. Also, it is okay for `CALLDATASIZE` to exceed the expected size, the check was that it is _at least_ the expected size to make sure we don't run into empty values unexpectedly (leading to difficult-to-debug behavior).

As such, this PR is a great addition! Thank you!

skellet0r:
> Also, it is okay for `CALLDATASIZE` to exceed the expected size, the check was that it is _at least_ the expected size to make sure we don't run into empty values unexpectedly.

ahh @fubuloubu I see what you mean, isn't that handled already though? Indirectly through `vyper/parser/arg_clamps.py`? I guess if you call a function and give it calldata with a length < `EXPECTED_CALLDATASIZE`, the arg clamps it's effectively just zero those args to the default values for the type?

> I think if you try to pull calldata longer than CALLDATASIZE the EVM's behavior is to give you empty data. So, it should be safe

Should I remove that runtime bounds check then? Since like you said, if you slice past the bound of `CALLDATASIZE` you'll just get null bytes appended to the end. Could just leave as it is, and if any complaints remove the runtime bounds check.

docs/constants-and-vars.rst
skellet0r: Yes it's just one line, should I take it out ? :thinking:

tests/parser/syntax/test_msg_data.py
skellet0r: Got it :+1:

skellet0r:
Awesome! The checks just have to run through one last time and should be good then :crossed_fingers:

tests/parser/syntax/test_msg_data.py
```diff
@@ -0,0 +1,118 @@
...
+    bar: Bytes[32] = slice(msg.data, 4, 32)
+    return convert(bar, uint256)
+"""
+
+    contract = get_contract(code)
+
+    assert contract.foo(42) == 42
+
+
+def test_get_full_calldata(get_contract, keccak, w3):
```
iamdefinitelyahuman, line 34: Would be good to include a test that has another statement following the assignment from `msg.data`, to ensure the free memory pointer is properly advanced.

tests/parser/syntax/test_msg_data.py
```diff
@@ -0,0 +1,118 @@
...
+@external
+def foo() -> Bytes[7]:
+    bar: Bytes[7] = concat(msg.data, 0xc0ffee)
+    return bar
+    """,
+        SyntaxException,
+    ),
+    (
+        """
+@external
```
iamdefinitelyahuman, line 83: Also check that assignment to `bytes32` fails, as this type is the closest to `Bytes[32]` and there might be some unexpected weirdness in the type checking due to the overlap.

vyper/parser/expr.py
```diff
@@ -371,6 +371,26 @@ def parse_Attribute(self):
...
             if key == "msg.sender" and not self.context.is_internal:
                 return LLLnode.from_list(["caller"], typ="address", pos=getpos(self.expr))
+            elif key == "msg.data" and not self.context.is_internal:
+                is_len = self.expr._metadata.get("is_len")
+                if is_len is True:
+                    typ = ByteArrayType(32)
+                    pos = self.context.new_internal_variable(typ)
+                    node = ["seq", ["mstore", pos, "calldatasize"], pos]
+                    return LLLnode.from_list(
+                        node, typ=typ, pos=getpos(self.expr), location="memory"
```
iamdefinitelyahuman, line 381: What happens if i assign a slice of `msg.data` to a storage variable?

skellet0r:
> What happens if I assign a slice of msg.data to a storage variable?

Ahhh that's a good test case I didn't include, I added it in, looks like that works as expected, along with the memory position advancement.

iamdefinitelyahuman (approved):
Great job!

fubuloubu (approved):
lgtm

mahsaC28:
does anyone know how can I do this without use msg.data???

PR #2345, opened 2021-04-10
Fix: memory corruption from nested internal function calls

iamdefinitelyahuman:
### What I did
Fix a runtime bug allowing for memory corruption when making an internal function call inside a literal array or another function call.

The issue is the same as the one outlined in https://github.com/vyperlang/vyper/security/advisories/GHSA-2r3x-4mrv-mcxf and fixed in https://github.com/vyperlang/vyper/pull/2186. In my previous fix I did not fully comprehend the scope of the issue, and so the bug was still present when making a private function call inside a literal array, or when accessing a single indexed value from a private function call returning a list or tuple.

### How I did it
Expand the logic added in #2186, applying it also to `Subscript` and `List` nodes.  I have refactored the check into it's own new function, `parse_sequence`, which is applied on all lists, tuples, and private function calls. There is some recursive logic to account for multidimensional arrays.

### How to verify it
Run the tests.  I've included some new cases - awful code that nobody should ever write, that would cause the bug to appear in very many varieties (if it were still around).

### Cute Animal Picture
image

### DIFF
--- a/vyper/parser/expr.py
+++ b/vyper/parser/expr.py
@@ -949,28 +949,25 @@ class Expr:
             return external_call.make_external_call(self.expr, self.context)
 
     def parse_List(self):
-        if not len(self.expr.elements):
-            return
+        call_lll, multi_lll = parse_sequence(self.expr, self.expr.elements, self.context)
+        out_type = next((i.typ for i in multi_lll if not i.typ.is_literal), multi_lll[0].typ)
+        typ = ListType(out_type, len(self.expr.elements), is_literal=True)
+        multi_lll = LLLnode.from_list(["multi"] + multi_lll, typ=typ, pos=getpos(self.expr))
+        if not call_lll:
+            return multi_lll
 
-        def get_out_type(lll_node):
-            if isinstance(lll_node, ListType):
-                return get_out_type(lll_node.subtype)
-            return lll_node.typ
+        lll_node = ["seq_unchecked"] + call_lll + [multi_lll]
+        return LLLnode.from_list(lll_node, typ=typ, pos=getpos(self.expr))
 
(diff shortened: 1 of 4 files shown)


fubuloubu (approved):

vyper/parser/self_call.py
```diff
@@ -57,37 +56,15 @@ def make_call(stmt_expr, context):
...
-                pos=getpos(arg),
-            )
-            setter = make_setter(target, lll_node, "memory", pos=getpos(arg))
-            expr_args.append(
-                LLLnode.from_list(target, typ=lll_node.typ, pos=getpos(arg), location="memory")
-            )
-            pre_init.append(setter)
-        else:
-            expr_args.append(lll_node)
+    from vyper.parser.expr import parse_sequence
```
fubuloubu, line 65: this can't be imported at the top?

vyper/parser/expr.py
```diff
@@ -1035,3 +999,62 @@ def parse_variable_location(cls, expr, context):
...
+        avoid memory corruption issues. This list may be empty, depending on the values
+        within `elements`.
+    List[LLLNode]
+        LLL nodes which collectively represent `elements`.
+    """
+    init_lll = []
+    sequence_lll = []
+    for node in elements:
+        if isinstance(node, vy_ast.List):
+            init, seq = parse_sequence(node, node.elements, context)
```
fubuloubu, line 1031: Add a comment here "we must make sure that every layer deep pushes all the computations into `init`" or something

vyper/parser/self_call.py
iamdefinitelyahuman: no ser

PR #2351, opened 2021-04-14
Add the int256 signed integer type

iamdefinitelyahuman:
### What I did
Add the `int256` type.

### How I did it
1. In `context`, I've added a new type primitive and definition for `int256`. I also expanded the abstract numeric types to make it easier to support e.g. any signed integer type when dealing with builtin function inputs.
2. In `functions`, I've modified the expected input types using the new absract types. Any function that previously expected `int128` can now receive both `int128` and `int256`. This required a bit of weirdness in the abstract types so as to not rekt the old type system - I left a comment explaining what's happening, as types are removed from `parser` we'll eventually be able to remove that code.
3. To make it work in `parser` I mostly _removed_ type logic. I can into an issue around literals where anything that could be cast as `int128` was cast as such. The best fix I found was to do away with parser-level type checking and just assume that if we made it this far, numeric types are allowed to interact. Where things got strange I left comments. It's again a case of, until types are fully removed from parser we need to be a bit hacky to make the old and new systems play nicely together.

Overflow checks for `int256` math operations are consistent with those used in OpenZeppelin's [`SignedSafeMath`](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/v3.4.0/contracts/math/SignedSafeMath.sol) library.

Note that anywhere I removed a `parser`-level type check, I verified that the check is actually happening within `context` and that the check was not being reached in coverage reports. We've had `context` in place for almost a year now, I feel pretty OK with trusting it and removing these redundant checks.

### How to verify it
Run the test suite. I've added test cases around `int256` math, unary operations, and using `convert` to move into and out of the new type.

### Cute Animal Picture
image


vyper/functions/convert.py
```diff
@@ -52,7 +64,7 @@ def to_int128(expr, args, kwargs, context):
         else:
             raise InvalidLiteral(f"Unknown numeric literal type: {in_arg}")
 
-    elif input_type == "bytes32":
+    elif input_type in ("bytes32", "int256"):
```
fubuloubu, line 67: This seems weird

vyper/compile_lll.py
```diff
@@ -107,7 +107,7 @@ def compile_to_assembly(code, withargs=None, existing_labels=None, break_dest=No
         return o
     # Numbers
     elif isinstance(code.value, int):
-        if code.value <= -(2 ** 255):
+        if code.value < -(2 ** 255):
```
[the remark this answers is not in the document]
iamdefinitelyahuman: This was a longstanding bug that never manifested because we didn't have any types capable of reaching it.

vyper/functions/convert.py
iamdefinitelyahuman: The logic is identical for converting both types - it just applies a bounds check.

michwill:
:rocket:

fubuloubu (approved):

PR #2361, opened 2021-04-26
Assign storage locations prior to LLL gen / place arrays sequentially in storage

iamdefinitelyahuman:
### What I did
* Add the start of an expansion pass that assigns variable positions prior to generating the LLL
* Use this new pass to refactor the layout of storage variables

### How I did it
Currently, whenever dealing with a type that requires >1 storage slot, we calculate the storage location in the same way as that of a mapping (using a keccak).  This is inefficient for both gas costs and bytecode size. Because Vyper does not support dynamically-sized variables without a limit on their length, it is possible to calculate the maximum required number of slots for all storage variables and instead place them sequentially to avoid the expensive `SHA3` operation.

To do this, I've added the `DataPosition` class, and related child classes, to represent positions within storage/memory/calldata. After type checking is completed, the AST is parsed once more and these objects applied to each previously assigned type. Handling this as a separate pass is imo a cleaner approach, making it easier to understand where and how this happens.

Armed with this information and the new storage layout, I've then removed various references to `sha_32` within `parser`. I had to do a couple hacky things to make the new play well with the old, but mostly it was a reductive operation so I think this is a net win in terms of the bigger refactor goal.

I have more logic ready to go re: assigning locations for memory and calldata, but for the scope of this PR all I've included is the eventual API to show my idea for the bigger picture without doing too much in a single chunk.

### How to verify it
Run the tests.  I added a lovely (awful?) new test case that assigns to and reads from a variety of storage slots to check for corruption. I also ran the Curve test suite against this PR, everything passes.

### Cute Animal Picture
image

### DIFF
--- a/vyper/compiler/phases.py
+++ b/vyper/compiler/phases.py
@@ -4,7 +4,7 @@ from typing import Optional, Tuple
 
 from vyper import ast as vy_ast
 from vyper import compile_lll, optimizer
-from vyper.context import validate_semantics
+from vyper.context import set_data_positions, validate_semantics
 from vyper.parser import parser
 from vyper.parser.global_context import GlobalContext
 from vyper.typing import InterfaceImports
@@ -175,6 +175,7 @@ def generate_folded_ast(
     vy_ast.folding.fold(vyper_module_folded)
     validate_semantics(vyper_module_folded, interface_codes)
     vy_ast.expansion.expand_annotated_ast(vyper_module_folded)
+    set_data_positions(vyper_module_folded)
 
     return vyper_module_folded
 
--- a/vyper/context/types/bases.py
+++ b/vyper/context/types/bases.py
@@ -25,6 +25,50 @@ class DataLocation(Enum):
     CALLDATA = 3
 
 
+class DataPosition:
+    _location: DataLocation
+
+
+class CalldataOffset(DataPosition):
+    __slots__ = (
+        "dynamic_offset",
+        "static_offset",
+    )
+    _location = DataLocation.CALLDATA
+
+    def __init__(self, static_offset, dynamic_offset=None):
+        self.static_offset = static_offset
+        self.dynamic_offset = dynamic_offset
+
+    def __repr__(self):
+        if self.dynamic_offset is not None:
+            return f"<CalldataOffset: static {self.static_offset}, dynamic {self.dynamic_offset})>"
+        else:
+            return f"<CalldataOffset: static {self.static_offset}, no dynamic>"
+
+
+class MemoryOffset(DataPosition):
+    __slots__ = ("offset",)
+    _location = DataLocation.MEMORY
+
+    def __init__(self, offset):
+        self.offset = offset
+
+    def __repr__(self):
+        return f"<MemoryOffset: {self.offset}>"
+
+
+class StorageSlot(DataPosition):
+    __slots__ = ("position",)
+    _location = DataLocation.STORAGE
+
+    def __init__(self, position):
+        self.position = position
+
+    def __repr__(self):
+        return f"<StorageSlot: {self.position}>"
+
+
 class BasePrimitive:
     """
     Base class for primitive type classes.
@@ -213,6 +257,16 @@ class BaseTypeDefinition:
     def from_annotation(self, node: vy_ast.VyperNode, **kwargs: Any) -> None:
         # always raises, user should have used a primitive
         raise StructureException("Value is not a type", node)
+
+    def set_position(self, position: DataPosition) -> None:
+        if hasattr(self, "position"):
+            raise CompilerPanic("Position was already assigned")
+        if self.location != position._location:
+            if self.location == DataLocation.UNSET:
+                self.location = position._location
+            else:
+                raise CompilerPanic("Incompatible locations")
+        self.position = position
 
     def compare_type(
         self, other: Union["BaseTypeDefinition", BasePrimitive, AbstractDataType]
--- a/vyper/context/validation/data_positions.py
+++ b/vyper/context/validation/data_positions.py
@@ -1,0 +1,36 @@
+import math
+
+from vyper import ast as vy_ast
+from vyper.context.types.bases import StorageSlot
+
+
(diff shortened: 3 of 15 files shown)


vyper/context/types/bases.py
```diff
@@ -25,6 +25,38 @@ class DataLocation(Enum):
...
+
+
+class CalldataOffset(DataPosition):
+    __slots__ = ("static_offset",)
+    _location = DataLocation.CALLDATA
+
+    def __init__(self, static_offset, dynamic_offset=None):
+        self.static_offset = static_offset
+        self.dynamic_offset = dynamic_offset
+
```
fubuloubu, line 39: Do you want a `__rep__` method like `MemoryOffset`?

vyper/context/types/bases.py
```diff
@@ -25,6 +25,38 @@ class DataLocation(Enum):
...
+        return f"<MemoryOffset: {self.offset}>"
+
+
+class StorageSlot(DataPosition):
+    __slots__ = ("position",)
+    _location = DataLocation.STORAGE
+
+    def __init__(self, position):
+        self.position = position
+
```
fubuloubu, line 58: Do you want a `__rep__` method like `MemoryOffset`?

vyper/context/types/bases.py
```diff
@@ -25,6 +25,38 @@ class DataLocation(Enum):
...
+    _location: DataLocation
+
+
+class CalldataOffset(DataPosition):
+    __slots__ = ("static_offset",)
+    _location = DataLocation.CALLDATA
+
+    def __init__(self, static_offset, dynamic_offset=None):
+        self.static_offset = static_offset
+        self.dynamic_offset = dynamic_offset
```
fubuloubu, line 38: Won't `__slots__` disallow `dynamic_offset` from being set?

vyper/context/validation/data_positions.py
```diff
@@ -0,0 +1,40 @@
...
+    Arguments
+    ---------
+    vyper_module : vy_ast.Module
+        Top-level Vyper AST node that has already been annotated with type data.
+    """
+    set_storage_slots(vyper_module)
+
+    # for fn_node in vyper_module.get_children(vy_ast.FunctionDef):
+    #     set_calldata_offsets(fn_node)
+    #     set_memory_offsets(fn_node)
```
fubuloubu, line 21: Needed?

vyper/compiler/phases.py
```diff
@@ -175,6 +175,7 @@ def generate_folded_ast(
     vy_ast.folding.fold(vyper_module_folded)
     validate_semantics(vyper_module_folded, interface_codes)
     vy_ast.expansion.expand_annotated_ast(vyper_module_folded)
+    set_data_positions(vyper_module_folded)
```
fubuloubu, line 178: It would be nice if the storage and memory layout generated during this phase were made available externally

vyper/parser/parser_utils.py
```diff
@@ -286,11 +282,10 @@ def add_variable_offset(parent, key, pos, array_bounds_check=True):
...
         if location == "storage":
-            return LLLnode.from_list(
-                ["add", ["sha3_32", parent], LLLnode.from_list(index, annotation=annotation)],
-                typ=subtype,
-                location="storage",
-            )
+            offset = 0
+            for i in range(index):
+                offset += get_size_of_type(typ.members[attrs[i]])
+            return LLLnode.from_list(["add", parent, offset], typ=subtype, location="storage",)
```
fubuloubu, line 288: Can you add a comment explaining why this offset value is needed here?

vyper/parser/parser_utils.py
```diff
@@ -359,8 +354,9 @@ def add_variable_offset(parent, key, pos, array_bounds_check=True):
             sub = ["uclamplt", k, typ.count]
 
         if location == "storage":
+            offset = get_size_of_type(subtype)
             return LLLnode.from_list(
-                ["add", ["sha3_32", parent], sub], typ=subtype, location="storage", pos=pos
+                ["add", parent, ["mul", sub, offset]], typ=subtype, location="storage", pos=pos
```
fubuloubu, line 359: Can you add a comment explaining why this offset value is needed here?

vyper/parser/parser_utils.py
```diff
@@ -531,8 +527,7 @@ def make_setter(left, right, location, pos, in_function_call=False):
 
         left_token = LLLnode.from_list("_L", typ=left.typ, location=left.location)
         if left.location == "storage":
-            left = LLLnode.from_list(["sha3_32", left], typ=left.typ, location="storage_prehashed")
-            left_token.location = "storage_prehashed"
+            pass
```
fubuloubu, line 530: Can you add a comment why `pass` is the appropriate path to take here? Can it be skipped entirely?

fubuloubu (approved):
Overall, I think this represents a breaking change to how our storage allocation algorithm works, and should perhaps target a v0.3.0 release

Could do with more robust testing, but we only have `Bytes` and `String` in terms of dynamic types right now.

vyper/compiler/phases.py
iamdefinitelyahuman: Yeah, makes sense as a compiler output.  But I think better done after the entirety of data positions are being calculated in that pass.

vyper/context/validation/data_positions.py
iamdefinitelyahuman: No, just showing the shape for the future.  Can remove now - it'll be in another PR shortly after this merges.

vyper/parser/parser_utils.py
iamdefinitelyahuman: Ahh yes this is a mistake.  I did that in a few places while I was working, thought I'd removed them all before opening PR.

iamdefinitelyahuman:
I don't think this requires a bump to `0.3.0` because:

1. the way we lay out storage is undocumented
2. we handle storage differently from solidity
3. vyper's range of functionality does not easily support upgradeability, where a change to storage would have a major impact

fubuloubu:
Should at least require a shoutout in release notes

fubuloubu (approved):

Issue #2367, opened 2021-05-24
Multidimensional lists declaration is inverted


eduardoxfurtado:
### Version Information

* vyper Version (output of `vyper --version`): 0.2.8
* Python Version (output of `python --version`): 3.9.2

### What's your issue about?

I expect that when declaring multidimensional lists of two dimensions it would be like:
`<variable_name : <Type>[<row>][<column>]`
so then they can be accessed by:
`variable_name[row][column]`

However, it seems that the declaration is inverted so it's actually:
`<variable_name : <Type>[<column>][<row>]`
and then oddly, we access it the correct way:
`variable_name[row][column]`

The only thing that the documentation says about Multidimensional lists is:
"Multidimensional lists are also possible."

No usage examples are provided, so I don't know if this is intended of not. A severe bug or it's like that by design and documentation is lacking.

I would expect that it would work like in C:
```
#include <stdio.h>

int main()
{
    // declare and initialize an array of two dimentions:
    int arr2d[2][4] = {  
       {0, 0, 0, 0},
       {0, 0, 0, 0},
    };
    
    // prints the multidimentional array:
    for (int i = 0; i < 2; i++)
    {
        for (int j = 0; j < 4; j++)
        {
            printf("%d ", arr2d[i][j]);
        }
        printf("\n");
    }
    // output: 
    // 0 0 0 0 
    // 0 0 0 0 
    
    printf("\n-------------\n\n");
    
    // modify an element
    arr2d[1][3] = 9;
    
    // prints the multidimentional array again:
    for (int i = 0; i < 2; i++)
    {
        for (int j = 0; j < 4; j++)
        {
            printf("%d ", arr2d[i][j]);
        }
        printf("\n");
    }
    // output: 
    // 0 0 0 0 
    // 0 0 0 9
    
    return 0;
}
```

And the same implementation but on vyper would currently look like the following:
```
arr2d: int128[4][2] = empty(int128[4][2])
# the array above is not:
# 0 0
# 0 0
# 0 0
# 0 0

# it's actually:
# 0 0 0 0
# 0 0 0 0

# the line below would make it crash:
arr2d[3][1] = 9

# the line below would work as if the array was declared as arr2d: int128[2][4]
arr2d[1][3] = 9
# now it would print:
# 0 0 0 0
# 0 0 0 9

```

### How can it be fixed?

The semantics of array declaration should be inverted

iamdefinitelyahuman:
[...]

PR #2368, opened 2021-05-25
Updated documentation to make it clear how to use multidimensional lists

eduardoxfurtado:
### What I did

In reference to issue [2367](https://github.com/vyperlang/vyper/issues/2367) I updated the documentation to make it clear how to use multidimensional lists.

### How I did it

Added an explanation and an example to the file docs/types.rst

### How to verify it

Compile and run the example code, also verify that the explanation is in line with the Solidity documentation because Vyper is modeled after it: https://docs.soliditylang.org/en/v0.6.0/types.html#arrays

Improved documentation of multidimensional lists.

### Cute Animal Picture

### DIFF
--- a/docs/types.rst
+++ b/docs/types.rst
@@ -373,7 +373,7 @@ Fixed-size Lists
 
 Fixed-size lists hold a finite number of elements which belong to a specified type.
 
-Lists can be declared with ``_name: _ValueType[_Integer]``. Multidimensional lists are also possible.
+Lists can be declared with ``_name: _ValueType[_Integer]``.
 
 .. code-block:: python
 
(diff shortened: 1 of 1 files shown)


docs/types.rst
```diff
@@ -387,6 +387,24 @@ Lists can be declared with ``_name: _ValueType[_Integer]``. Multidimensional lis
     # Returning a value
     return exampleList[0]
 
+Multidimensional lists are also possible. The notation for the declaration is reversed compared to some other languages, but the access notation is not reversed.
+
+A two dimensional list can be declared with ``_name: _ValueType[_Column_Integer][_Row_Integer]``. Elements can be accessed with ``_name: _ValueType[_Row_Integer][_Column_Integer]``.
```
fubuloubu, line 392: Instead of using "row" and "column", use "inner" and "outer"

fubuloubu:
```suggestion
A two dimensional list can be declared with ``_name: _ValueType[inner_size][outer_size]``. Elements can be accessed with ``_name[outer_index][inner_index]``.
```

fubuloubu (approved):

PR #2422, opened 2021-08-17
Fix code-block directive in struct docs

milancermak:
The documentation is currently [missing a code example on how to define structs](https://vyper.readthedocs.io/en/stable/types.html#structs). That's because the `code-block` RST directive is missing a language argument. This PR fixes it.


charles-cooper (approved):
LGTM. I noticed the example itself is slightly wrong -- `2` is not a valid value for the `value2` field, it should be `2.0`. @milancermak would you be open to fixing that as well?

PR #2425, opened 2021-08-17
Fixes struct example

milancermak:
### What I did

As mentioned #2422, the struct example in the docs is slightly wrong. Here's the fix.

### Cute Animal Picture


charles-cooper (approved):
LGTM, thank you

PR #2477, opened 2021-10-06
feat: uint8 type

skellet0r:
### What I did

Add a new type `uint8`

### How I did it

Mostly just looking at how `uint256` is defined, lots of grep, and a small bit of brains

### How to verify it

I added in one test file at the time of opening this PR, but I'm pretty sure some more test cases should be added.

Add native uint8 type

### Cute Animal Picture

### DIFF
--- a/tests/parser/types/numbers/test_uint8.py
+++ b/tests/parser/types/numbers/test_uint8.py
@@ -174,6 +174,9 @@ def max_ne() -> (bool):
     assert c.max_ne() is True
 
 
+# TODO: create a tests/parser/functions/test_convert_to_uint8.py file
+
+
 def test_uint8_convert_clamps(get_contract, assert_tx_failed):
     code = """
 @external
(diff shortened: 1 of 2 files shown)


tests/parser/types/numbers/test_uint8.py
```diff
@@ -0,0 +1,145 @@
...
+    assert c.foo(1) == 0
+    assert c.foo(42) == 0
+    assert c.foo(2 ** 8 - 1) == 0
+
+
+def test_exponent_base_one(get_contract):
+    code = """
+@external
+def foo(x: uint8) -> uint8:
+    return 1 ** x
```
fubuloubu, line 18: Vyper also has exponential safemath

tests/parser/types/numbers/test_uint8.py
```diff
@@ -0,0 +1,145 @@
...
+@external
+def test() -> uint8:
+    o: uint8 = 64
+    return o
+    """
+
+    c = get_contract_with_gas_estimation(modexper)
+    assert c.test() == 64
+
+
```
fubuloubu, line 109: Add test cases showing dynamic conversions fail based on the clamp (e.g. `convert(a, uint8)` and `a: uint256 = 256`)

PR #2523, opened 2021-11-04
fix: allow build CD workflow to be manually triggered

skellet0r:
Allow manually triggering CD workflow for building binaries, as well as adjusting it to run on tags being pushed instead of new releases being created.


charles-cooper (approved):
awesome, TY

PR #2576, opened 2021-12-29
docs: minor improvements and update type conversion table

tserg:
### What I did

Made minor improvements and updated type conversion table.

### How I did it

1. Read through documentation.
2. Refer to `vyper/builtin_functions/convert.py`.

### How to verify it

See above.

Improvements to documentation.

### Cute Animal Picture

### DIFF
--- a/docs/types.rst
+++ b/docs/types.rst
@@ -538,7 +538,9 @@ Type        Default Value
 ``bool``    ``False``
 ``bytes32`` ``0x0000000000000000000000000000000000000000000000000000000000000000``
 ``decimal`` ``0.0``
+``uint8``   ``0``
 ``int128``  ``0``
+``int256``  ``0``
 ``uint256`` ``0``
 =========== ======================================================================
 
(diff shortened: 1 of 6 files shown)


tserg:
Note: The updates to the type conversion table results in the table formatting becoming distorted as the 'Allowable Values' column is wrapped at a shorter width while the last column ('Additional Notes') is given a longer width. After experimenting with various options including (1) changing the existing simple table to a grid table; and (2) modifying the table directive, the best solution with the least changes required seems to be the use of custom line breaks.

docs/types.rst
```diff
@@ -556,37 +558,73 @@ Type Conversions
...
+``bool``         ``int128``         All                                  ``0`` or ``1``
+``bool``         ``int256``         All                                  ``0`` or ``1``
+``bool``         ``uint256``        All                                  ``0`` or ``1``
+``bool``         ``bytes32``        All                                  ``0x00`` or ``0x01``
+``decimal``      ``bool``           All                                  Returns ``a != 0.0``
+``decimal``      ``uint8``          ``MAX_UINT8 >= a >= 0.0``            | Cannot convert negative values.
+                                                                         | Value is truncated.
+``decimal``      ``int128``         All                                  Value is truncated
+``decimal``      ``int256``         All                                  Value is truncated
+``decimal``      ``uint256``        ``a >= 0.0``                         Value is truncated
```
fubuloubu, line 582:
```suggestion
``decimal``      ``uint256``        ``a >= 0.0``                         Cannot convert negative values
```

docs/types.rst
```diff
@@ -556,37 +558,73 @@ Type Conversions
...
+``uint256``      ``address``        ``ADDRSIZE - 1 >= a >= 0``
+``uint256``      ``bool``           All                                  Returns ``a != 0``
 ``uint256``      ``decimal``        ``a <= MAX_DECIMAL``
+``uint256``      ``uint8``          ``a <= MAX_UINT8``
 ``uint256``      ``int128``         ``a <= MAX_INT128``
+``uint256``      ``int256``         ``a <= MAX_INT256``
 ``uint256``      ``bytes32``        All
-``uint256``      ``Bytes``          All
-``bytes32``      ``bool``           All                            ``True`` if ``a`` is not empty
+``bytes32``      ``address``        ``ADDRSIZE - 1 >= a >= 0``
```
fubuloubu, line 609:
```suggestion
``bytes32``      ``address``        ``2**ADDRSIZE - 1 >= a >= 0``
```

docs/types.rst
```diff
@@ -556,37 +558,73 @@ Type Conversions
...
-``decimal``      ``int128``         All                            Value is truncated
-``decimal``      ``uint256``        ``a >= 0.0``                   Value is truncated
+================ ================== ==================================== ==================================
+In (``atype``)   Out (``btype``)    Allowable Values                     Additional Notes
+================ ================== ==================================== ==================================
+``address``      ``bool``           All                                  Returns ``a != ZERO_ADDRESS``
+``address``      ``decimal``        All                                  | Value is truncated to the
+                                                                         | least significant 128 bits
+``address``      ``int128``         All                                  | Value is truncated to the
+                                                                         | least significant 128 bits
```
fubuloubu, line 568: Is the value truncated? Or will it revert if exceeded?

fubuloubu (commented):
In general, it would be nice if the test suite tested these boundary conditions for conversions. Not sure we do all of them right now.

docs/types.rst
tserg: The value will be truncated. For example, `0xFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF` evaluates to `-1` for `int128` and `Fixed(-1)` for `decimal`.

tserg:
> In general, it would be nice if the test suite tested these boundary conditions for conversions. Not sure we do all of them right now.

Agree, other than the existing tests, we will also need to add in tests for `uint8`, `Bytes[N]` and `String[N]. I can take this up next.

docs/types.rst
fubuloubu: What about `0x80000000000000000000FFFFFFFFFFFFFFFFFFFF`?

tserg: It truncates as well - `1208925819614629174706175` and `Fixed('1208925819614629174706175')` for `int128` and `decimal` respectively. Is there an intuition to this test case? I can try exploring in that direction.

fubuloubu: Yes, what happens to a value outside of the `int128` range when converted?

tserg: From my understanding, in this case of converting from `address` to `int128` and `decimal`, the input address is sign-extended from 16 bytes to 32 bytes, so the most significant 4 bytes of the input address are dropped. Consequently, the remaining 16 bytes will always be a value that can be represented in `int128`. Perhaps 'truncated' is not the right way to phrase it?

fubuloubu: > Perhaps 'truncated' is not the right way to phrase it?

yeah, thinking about this

tserg: How about these alternatives? I think this would be more aligned with the actual operation.

> Extracts the rightmost sixteen bytes only
> Leftmost four bytes are discarded

fubuloubu: sure!

fubuloubu (approved):

PR #2584, opened 2022-01-03
fix: error handling for string and byte arrays

tserg:
### What I did

Fix for #2502. 

The issue highlighted the lack of a helpful error message for arrays of `String`. While working on the fix, I observed the same issue for `Bytes`, and extended the fix to cover it.

### How I did it

If there is only one possible type for a `List` node, check if it is a `String` or `Bytes`. If yes, raise a `StructureException` with the error message that `String arrays are not supported` / `Bytes arrays are not supported`.

### How to verify it

See tests.

Improve exception handling for string and byte arrays

### Cute Animal Picture

### DIFF
--- a/tests/parser/syntax/test_list.py
+++ b/tests/parser/syntax/test_list.py
@@ -226,6 +226,74 @@ def foo(x: int128[2][2]):
     """,
         TypeMismatch,
     ),
+    (
+        """
+@external
+def foo():
+    for str in ["hello", "world"]:
+        pass
+    """,
+        StructureException,
+    ),
+    (
+        """
+@external
+def foo():
+    a: String[6] = "hello"
+    b: String[6] = "world"
+    for str in [a, b]:
+        pass
+    """,
+        StructureException,
+    ),
+    (
+        """
+a: String[6]
+b: String[6]
+
+@external
+def foo():
+    self.a = "hello"
+    self.b = "world"
+    for str in [self.a, self.b]:
+        pass
+    """,
+        StructureException,
+    ),
+    (
+        """
+@external
+def foo():
+    for str in [b"hello", b"world"]:
+        pass
+    """,
+        StructureException,
+    ),
+    (
+        """
+@external
+def foo():
+    a: Bytes[6] = b"hello"
+    b: Bytes[6] = b"world"
+    for str in [a, b]:
+        pass
+    """,
+        StructureException,
+    ),
+    (
+        """
+a: Bytes[6]
+b: Bytes[6]
+
+@external
+def foo():
+    self.a = b"hello"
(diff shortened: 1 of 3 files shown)


tserg:
While working on this fix, I discovered a similar issue with `bytes32`.

Compiling the following code snippet results in a `KeyError: 'type'` exception. This should compile since byte32 arrays are valid. For the `String` and `Bytes` equivalent of this code, they are now throwing the new error message. 
```
# @version ^0.3.0

a: bytes32
b: bytes32

@external
def doThings():
    self.a = 0x0000000000000000000000000000000000000000000000000000000000000001
    self.b = 0x0000000000000000000000000000000000000000000000000000000000000002
    for str in [self.a, self.b]:
        pass

```

I intend to look into this next, and add the fix to this PR. Please let me know if it should be handled as a separate PR instead. Thanks!

tserg:
> While working on this fix, I discovered a similar issue with `bytes32`.
> 
> Compiling the following code snippet results in a `KeyError: 'type'` exception. This should compile since byte32 arrays are valid. For the `String` and `Bytes` equivalent of this code, they are now throwing the new error message.
> 
> ```
> # @version ^0.3.0
> 
> a: bytes32
> b: bytes32
> 
> @external
> def doThings():
>     self.a = 0x0000000000000000000000000000000000000000000000000000000000000001
>     self.b = 0x0000000000000000000000000000000000000000000000000000000000000002
>     for str in [self.a, self.b]:
>         pass
> ```
> 
> I intend to look into this next, and add the fix to this PR. Please let me know if it should be handled as a separate PR instead. Thanks!

Looking at it now, is `[self.a, self.b]` valid syntax? The same error is being thrown when I replaced `bytes32` with other types such as `uint256` and `bool` as well, and it is being triggered during IR generation.

fubuloubu:
> Looking at it now, is `[self.a, self.b]` valid syntax? The same error is being thrown when I replaced `bytes32` with other types such as `uint256` and `bool` as well, and it is being triggered during IR generation.

It is valid syntax, however we have been finding weirdness with array construction like that

tserg:
> This was working before, but starts to fail with v0.3.2:
https://github.com/yearn/yearn-vaults/blob/main/contracts/Vault.vy#L1266

Interestingly, this array construction compiles when used in an assert statement.

Based on my digging so far, I have narrowed down the error with this array construction to this line. Since it seems to relate to a different point, I think it should be dealt with in a separate PR.

https://github.com/vyperlang/vyper/blob/2f1584bf9d2fce15e2173b8befc83619c5deae79/vyper/old_codegen/expr.py#L408

charles-cooper (approved):
thanks!

PR #2590, opened 2022-01-05
chore: remove gas estimates from ABI

tserg:
### What I did

Fix for #2151.

### How I did it

Remove the step to add gas estimates to the ABI in `build_abi_output` function in `vyper/compiler/output.py`.

### How to verify it

See test cases.

Since gas estimates are removed from the ABI, it no longer makes sense to calculate the `gas` value for function calls in test cases based on the ABI gas estimate + 50000. Instead, we default to the standard gas estimates. This however results in certain test cases for `assert` statements failing in `tests/parser/features/test_assert_unreachable.py` and `tests/parser/features/test_assert.py` (the error is due to an error in parsing the exception in `eth_tester`). The workaround is to supply a `gas` value to the `call` kwarg (still not sure why supplying this parameter changes the response returned).

Remove gas estimates from ABI output

### Cute Animal Picture


charles-cooper (commented):
I think we should make gas a configurable option to the ABI output. So like `-f abi` doesn't have gas estimates unless you include `--show-gas-estimates`.

tests/parser/features/test_assert_unreachable.py
```diff
@@ -32,13 +28,9 @@ def foo(val: int128) -> bool:
 
     assert c.foo(2) is True
 
-    assert_tx_failed(lambda: c.foo(1))
-    assert_tx_failed(lambda: c.foo(-1))
-
-    with pytest.raises(TransactionFailed) as e_info:
-        c.foo(-2)
-
-    assert "Invalid opcode 0xfe" in e_info.value.args[0]
```
fubuloubu, line 41: Side note: hitting the "invalid opcode" causes the transaction to revert, and use all possible gas. Which is probably why the transaction wouldn't estimate a gas cost for itself (it was never successfully able to avoid using all the gas).

This is a fine solution to that problem (setting gas to 50k)

fubuloubu (approved):

fubuloubu (approved):

tserg:
I have made the change for `--show-gas-estimates` to be an optional flag for `-f abi`. I have also made minor changes to how `--show-gas-estimates` currently works for `-f ir` to be consistent with the fix for `-f abi`.

tests/parser/features/test_assert_unreachable.py
tserg: Update: With gas estimates now an optional flag, I have reverted the gas parameter to be calculated based on the original logic of the gas estimate + 50000 gas. It is therefore now unnecessary to manually supply the `gas` argument as a kwarg.

charles-cooper (approved):

Issue #2661, opened 2022-02-15
_abi_encode does not compile for constant static and dynamic arrays


tserg:
### Version Information

* vyper Version (output of `vyper --version`): 0.3.2+commit.43f491fd
* OS: Linux
* Python Version (output of `python --version`): 3.8.5

### What's your issue about?

`_abi_encode` does not compile for constant static and dynamic arrays.

This compiles.
```
@external
@view
def get() -> Bytes[224]:
  da: DynArray[uint256, 5] = [1, 2, 3, 4, 5]
  return _abi_encode(da)
```

This does not compile
```
da: constant(DynArray[uint256, 5]) = [1, 2, 3, 4, 5]

@external
@view
def get() -> Bytes[224]:
  return _abi_encode(da)

```
```
AttributeError: 'list' object has no attribute 'value_type'
```

### How can it be fixed?

Fill this in if you know how to fix it.

PR #2666, opened 2022-02-16
fix: _abi_encode for constant arrays

tserg:
### What I did

Fix for #2661.

When visiting a list node, we pop from the list of possible types derived only if its length is greater than 1. Since constant arrays will have only one possible type after constant folding, this fix relaxes the assertion to greater than or equal to 1. 

### How I did it

Add `=`.

### How to verify it

See test case.

Fix _abi_encode for constant arrays

### Cute Animal Picture


charles-cooper (approved):
thanks!

PR #2684, opened 2022-02-26
chore: support vyper files for tests

tserg:
### What I did

- Update `tests/conftest.py` to handle the testing of compilation outcomes for `.vy` files (as per https://docs.pytest.org/en/6.2.x/example/nonpython.html)

### How I did it

The Vyper files need to adhere to the syntax `test_DESCRIPTION_OUTCOME` where `DESCRIPTION` is non-consequential and `OUTCOME` is either `Success`, a `VyperException` class name, or `Fail` (unhandled exception). `OUTCOME` is extracted upon the initialisation of each Vyper test file.

For each test file, we attempt to compile it using `compile_files`, and check the compilation outcome against the extracted `OUTCOME`.

Example of a failing test:
Screenshot from 2022-02-26 17-26-40

### How to verify it

Run `pytest tests/parser/syntax/constants/` to see a demo against two sample contracts from `test_constants.py`. You can also edit the Vyper test files to see what a failing test look like.

Handle Vyper test files

### Cute Animal Picture


tests/parser/syntax/test_constants.py
```diff
@@ -201,3 +201,10 @@ def deposit(deposit_input: Bytes[2048]):
 @pytest.mark.parametrize("good_code", valid_list)
 def test_constant_success(good_code):
     assert compiler.compile_code(good_code) is not None
+
+
+@pytest.mark.parametrize(
+    "vyper_contract", ["constants/test_constantGetter1_Success.vy"], indirect=True
+)
+def test_constant_success_2(vyper_contract):
+    assert vyper_contract.foo() == 123
```
fubuloubu, line 210: Oh, this is clever. Basically, reuse compiled "successful parsing" test cases to additionally perform other checks?

tserg: Yes, that's correct. This is more deliberate as compared to automatically creating fixtures for "successful parsing" test cases (which I still cannot figure out, and on further thought could be more confusing if all these contracts appear in the global namespace).

fubuloubu: Would the array of filenames passed in create multiple parametrized tests?

Also, some of the folders get really deep. I wonder if it could just assume a local path e.g. `./test_constantGetter1_Success.vy`

tserg: > Would the array of filenames passed in create multiple parametrized tests?

Yes, I have updated the example to better demonstrate this.

> Also, some of the folders get really deep. I wonder if it could just assume a local path e.g. ./test_constantGetter1_Success.vy

Do you mean that there will be some sort of collection process in the current directory and subdirectories to get the Vyper files based on the filename provided?

Sporarum:
Very cool concept !

I worked on the scala 3 compiler, there we did things this way:
1. Split expected test result based on folder
    1. `tests/pos/**/*.scala`: Should compile to pass
    2. `tests/neg/**/*.scala`: Should not compile to pass
1. For neg tests, the non-compile reason is in a comment on the line (and there can be multiple), this helps ensure we raise the error we thought we did

Working with that system was really intuitive and fast, I think we should do something similar

@tserg what are your plans for this draft PR ?

charles-cooper:
i think we can close this actually, any testing infra along these lines would probably just be from scratch

PR #2714, opened 2022-03-21
Tests for new conversion rules

tserg:
### What I did

Add tests for #2694.

Refactor conversion tests in multiple files into a single file.

Change return type for `floor` and `ceil` builtins from `int128` to `int256`.

### How I did it

Use pytest parametrization.

### How to verify it

See test cases.

Tests for new conversion rules.

### Cute Animal Picture


tserg:
There seems to be a bug with conversion from bytes32 to address (0xF5D4020dCA6a62bB1efFcC9212AAF3c9819E30D7) where the bytes32 is a state variable.
```
bar: bytes32

@external
def test_state_variable_convert() -> address:
        self.bar = b'\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\xf5\xd4\x02\r\xcajb\xbb\x1e\xff\xcc\x92\x12\xaa\xf3\xc9\x81\x9e0\xd7'
        return convert(self.bar, address)
```
The function returns "0x0000000000000000000000000000000000000020" instead of the original address.

tserg:
Will figure out how to better parametrize all integer and byte types once I have refactored the various test files into a single file.

tserg:
I think this should raise `InvalidType` instead of `TypeMismatch`.
```
@external
@view
def test_convert_2() -> uint256:
	bar: uint256 = 0
	return convert(bar, uint256)
```

charles-cooper:
> I think this should raise `InvalidType` instead of `TypeMismatch`.
> 
> ```
> @external
> @view
> def test_convert_2() -> uint256:
> 	bar: uint256 = 0
> 	return convert(bar, uint256)
> ```

@tserg why's that?

tserg:
> > I think this should raise `InvalidType` instead of `TypeMismatch`.
> > ```
> > @external
> > @view
> > def test_convert_2() -> uint256:
> > 	bar: uint256 = 0
> > 	return convert(bar, uint256)
> > ```
> 
> @tserg why's that?

For all other types, this error is returned. Currently, `uint256` is the exception.
```
vyper.exceptions.InvalidType: Value and target type are both 'int256'
  contract "contracts/convert2.vy", function "test_convert_2", line 8:8 
       7 	bar: int256 = 0
  ---> 8 	return convert(bar, int256)
  ---------------^
       9

```

charles-cooper:
> > > I think this should raise `InvalidType` instead of `TypeMismatch`.
> > > ```
> > > @external
> > > @view
> > > def test_convert_2() -> uint256:
> > > 	bar: uint256 = 0
> > > 	return convert(bar, uint256)
> > > ```
> > 
> > 
> > @tserg why's that?
> 
> For all other types, this error is returned. Currently, `uint256` is the exception.
> 
> ```
> vyper.exceptions.InvalidType: Value and target type are both 'int256'
>   contract "contracts/convert2.vy", function "test_convert_2", line 8:8 
>        7 	bar: int256 = 0
>   ---> 8 	return convert(bar, int256)
>   ---------------^
>        9
> ```

ah, gotcha. yeah will fix

tests/parser/functions/test_convert.py
```diff
@@ -12,61 +12,96 @@
 from vyper.utils import checksum_encode
 
 
-def _generate_valid_test_cases_for_type(type_, bits=None):
+def _get_type_N(type_):
+    """
+    Helper function to extract N from typeN (e.g. uint256, bytes32, Bytes[32])
+    """
+    if type_.startswith("uint"):
+        return int(type_[4:])
```
charles-cooper, line 20: `parse_integer_typeinfo` might be helpful here

tserg:
Conversion from `uintN` to any `intM` reverts when the value of `uintN >= (2 ** (N - 1))` in the type of contract below.

For example, this contract compiles and returns the correct value:
```
@external
@view
def test_convert_2() -> int128:
	x: uint8 = 127
	return convert(x, int128)
```
but this contract compiles and reverts:
```
@external
@view
def test_convert_2() -> int128:
	x: uint8 = 128
	return convert(x, int128)
```

charles-cooper:
> Conversion from `uintN` to any `intM` reverts when the value of `uintN >= (2 ** (N - 1))` in the type of contract below.
> 
> For example, this contract compiles and returns the correct value:
> 
> ```
> @external
> @view
> def test_convert_2() -> int128:
> 	x: uint8 = 127
> 	return convert(x, int128)
> ```
> 
> but this contract compiles and reverts:
> 
> ```
> @external
> @view
> def test_convert_2() -> int128:
> 	x: uint8 = 128
> 	return convert(x, int128)
> ```

this is expected behavior, since `128` is out of range for `int8`. apparently I did not explicitly put this in the table in #2507 (!) but the wording is there:
Screenshot from 2022-03-23 07-56-26

charles-cooper:
> Ah, I think the example was not as clear because the `uint` has more bits than the `int`, so the conversion value is limited by the RHS. So what this means is, for example, converting `uint8` to `int128` is only valid if the input `<= INT8_MAX`, so the conversion value is similarly limited by the LHS?

no, because all the possible values for `uint8` (`[0..255]`) are in bounds for `int128`. `uint8` -> `int128` should never clamp.

tserg:
There seems to be a discrepancy in the conversion of `uint` to `decimal` for `uint128` and `uint256`.

1. This two contracts compile and do not revert although the values are out of bound (2 ** 127, 2 ** 128 - 1). It reverts for the same values with all other `uint` greater than `uint128`.
```
@external
def foo():
	bar: uint128 = 170141183460469231731687303715884105728
	foobar: decimal = convert(bar, decimal)
```
```
@external
def foo():
	bar: uint128 = 340282366920938463463374607431768211455
	foobar: decimal = convert(bar, decimal)
```

2. Similar issue for `uint256`.

```
@external
def foo():
	bar: uint256 = 115792089237316195423570985008687907853269984665640564039457584007913129639935
	foobar: decimal = convert(bar, decimal)
```

Additionally, this contract compiles and returns `-1` although the value is out of bound (2 ** 256 - 1).
```
@external
@view
def foo() -> decimal:
	bar: uint256 = 115792089237316195423570985008687907853269984665640564039457584007913129639935
	return convert(bar, decimal)
```

tserg:
Conversion of `bytesM` to `decimal` does not seem to be clamped for  > 20 bytes. However, this seems to be the behaviour for `bytes32` in version 0.3.1. Could there be backwards compatibility issues if we change this as per #2507 ?

```
@external
def foo():
	bar: bytes32 = 0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff
	foobar: decimal = convert(bar, decimal)
```

tserg:
Note: to update return type from `int128` to `int256` for `floor()` and `ceil()` in docs

charles-cooper (approved):

charles-cooper:
note: plan to add some edge cases and refactor in a follow up PR

PR #2739, opened 2022-03-31
Fix reentrancy vulnerability in example Crowfund.refund()

hitsuzen-eth:
Fix reentrancy vulnerability Crowfund.refund()


hitsuzen-eth:
My fix is useless.
There is no possibility to steal fund with a reentrancy attack due to 2300 gas limit.
BUT if a smart contract is used as a participant it's possible to lock all funds in the Crowdfund due to REVERT on send call which freeze all fund.

I don't know how to fix that without implementing a pull model instead of push.

fubuloubu:
> I don't know how to fix that without implementing a pull model instead of push.

Sounds like this should be the fix then!

hitsuzen-eth:
I think it's ok now.

Crowdfund.refund() will only refund fund of the msg.sender and one participant can participate multiple time which increase his participation.

docs/vyper-by-example.rst
```diff
@@ -275,14 +275,14 @@ the funds will be sent to the  beneficiary at the end of the campaign deadline.
 Participants will be refunded their respective contributions if the total
 funding does not reach its target goal.
 
-.. literalinclude:: ../examples/crowdfund.vy
+.. literalinclude:: ../examples/crowdfund/crowdfund.vy
```
fubuloubu, line 278: You know, I actually prefer all the examples to not be inside each of these folders, don't think it adds anything.

examples/crowdfund/crowdfund.vy
```diff
@@ -42,22 +30,13 @@ def finalize():
...
-        if i >= self.nextFunderIndex:
-            self.refundIndex = self.nextFunderIndex
-            return
-
-        sender: address = self.funders[i].sender
-        value: uint256 = self.funders[i].value
-        self.funders[i] = empty(Funder)
-        send(sender, value)
+    value: uint256 = self.funders[msg.sender]
+    self.funders[msg.sender] = empty(uint256)
```
fubuloubu, line 40: Can probably just be 0
```suggestion
    self.funders[msg.sender] = 0
```

fubuloubu (approved):

fubuloubu (approved):

PR #2875, opened 2022-05-25
Updated ERC1155 ownable to support opensea collection metadata

Doc-Pixel:
### What I did
Updated the ERC1155 ownable example to fully support opensea.io

### How I did it
During online testing I ran into 2 missing functions that are needed to correctly display the ERC1155 contract's metadata on opensea. setContractURI and contractURI have been added.
[edit] The tests have also been updated [/edit]
```# URI #
@external
def setContractURI(contractUri: String[MAX_URI_LENGTH]):
    """
    @dev set the contractURI for the contract. points to collection metadata file
    @param contractUri the new urcontractUri for the contract
    """
    assert not self.paused, "The contract has been paused"
    assert self._contractUri != contractUri, "new and current URI are identical"
    assert msg.sender == self.owner, "Only the contract owner can update the URI"
    self._contractUri = contractUri
    log evtContractURI(contractUri, 0)


@view
@external
def contractURI() -> String[MAX_URI_LENGTH]:
    """
    @dev retrieve the contractUri to retrieve the collection metadata
    """
    return self._contractUri
```

### How to verify it
Deploy on a testnet, then add the collection on opensea.io 
1. go to https://opensea.io/get-listed 
2. click "live on a testnet"
3. enter which testnet + contract address, click submit

The contract should show up with metadata

### Commit message

### Cute Animal Picture
image


examples/tokens/ERC1155ownable.vy.old
```diff
@@ -0,0 +1,355 @@
+# @version >=0.3.3
```
fubuloubu, line 1: Delete this file

Doc-Pixel: Deleted

examples/tokens/ERC1155ownable.vy
```diff
@@ -26,6 +26,7 @@ paused: public(bool)
 
 # the contracts URI to find the metadata
 _uri: String[MAX_URI_LENGTH]
+_contractUri: String[MAX_URI_LENGTH]
```
fubuloubu, line 29: Was reading the section on the ERC1155 metadata and I think we need a bigger rewrite here.

Typically, a `baseURI` public state variable will be used to track the base of the `uri(id: uint256) -> String` function (although not strictly necessary)

examples/tokens/ERC1155ownable.vy
```diff
@@ -85,6 +86,11 @@ event URI:
     value: String[MAX_URI_LENGTH]
     id: uint256
 
+event evtContractURI:
```
fubuloubu, line 89: The event for when a basic URI change is made would be just this. 

```suggestion
event URI:
```

It states that if there is a dynamic URI, it's not strictly necessary to include this event. Curiously, the event definition for `URI` is not included.

examples/tokens/ERC1155ownable.vy
```diff
@@ -353,12 +361,34 @@ def uri(id: uint256) -> String[MAX_URI_LENGTH]:
...
+    assert not self.paused, "The contract has been paused"
+    assert self._contractUri != contractUri, "new and current URI are identical"
+    assert msg.sender == self.owner, "Only the contract owner can update the URI"
+    self._contractUri = contractUri
+    log evtContractURI(contractUri, 0)
+
+
+@view
+@external
+def contractURI() -> String[MAX_URI_LENGTH]:
```
fubuloubu, line 380: Could get around this by just have it be public

examples/tokens/ERC1155ownable.vy
```diff
@@ -345,6 +352,7 @@ def setURI(uri: String[MAX_URI_LENGTH]):
     self._uri = uri
     log URI(uri, 0)
 
+@view
 @external
 def uri(id: uint256) -> String[MAX_URI_LENGTH]:
```
fubuloubu, line 357: Not really sure why there's a `contractURI` when this function exists

Doc-Pixel: The contract uri points to a JSON metadata file describing the NFT collection. 
the 'regular URI' points to individual NFT metadata files.
Without the contractURI there is no collection description and image visible on NFT marketplaces.

Doc-Pixel: Hope this explains things a bit 😊

examples/tokens/ERC1155ownable.vy
Doc-Pixel: Dynamic URI is something I will work on next. I discovered the basic 1155 with static URI was missing the contract metadata functions. I've deployed this new version and minted a couple of items with it. Will dig up the url

examples/tokens/ERC1155ownable.vy
Doc-Pixel: Reworked it!

examples/tokens/ERC1155ownable.vy
Doc-Pixel: Removed the redundant evtContractURI
changing the contract-level metadata URI will now use the regular URI event

examples/tokens/ERC1155ownable.vy
fubuloubu: This isn't a part of the standard, is it?

Doc-Pixel: Pure erc1155. no.
Name and Symbol are also not part of the pure standard.

But: 
With name, symbol and the contract metadata URI this example is opensea.io compatible.
I think this may help Vyper adoption.
(Also working on a brownie mix that includes this contract, and setting up to do a comparison between this contract and a solidity 1155 to see what the gas usage difference is.)

Doc-Pixel: The latter. This is an opensea "thing".

So for name, symbol, contract URI explicitly add to the documentation these are opensea.io additions?

Doc-Pixel: Documented the opensea specific functions

examples/tokens/ERC1155ownable.vy
Doc-Pixel: @fubuloubu I've added dynamic URI
it uses the base URI + String[78] for the uint2str output.

tests/examples/tokens/test_erc1155.py
```diff
@@ -160,6 +162,11 @@ def test_URI(erc1155, w3, assert_tx_failed):
 
     assert_tx_failed(lambda: erc1155.setURI(CONTRACT_URI))
 
+    # set contract to dynamic URI
+    erc1155.toggleDynUri(True, transact={"from": owner})
+    erc1155.setURI(CONTRACT_DYNURI, transact={"from": owner})
+    assert erc1155.uri(0) == CONTRACT_DYNURI + 0
```
fubuloubu, line 168: should this be `str(0)`?

Doc-Pixel: Most likely. I will take care of this tonight.

Doc-Pixel: Taken care of it. 
See some tests fail now with no output on the "tox" command.

fubuloubu (approved):

fubuloubu:
@Doc-Pixel is this good to go?

PR #3078, opened 2022-08-31
chore: update discord link

fubuloubu:
links were broken, created unlimited use links


charles-cooper (commented):
the old link works for me, not sure this PR is necessary

PR #3104, opened 2022-09-23
refactor: optimize calldatasize check

emc415:
### What I did
This wip commit addresses the first half of inefficient selector table checks (issue #3037)

### How I did it
I added a function which tests whether a contract’s function has a selector with trailing zeroes (wip: at the moment it catches all selectors which contain 0)
It then changes the IR created in def _runtime_ir() to only have a line check in IR for calldatasize >= 4 if there is a zero selector function in the contract.

### Commit message

Addresses the first half of inefficient selector table checks (issue #3037) by only running a check for calldatasize >= 4 if a zero selector function is present.

only check for calldatasize >= 4 if zero selector function is present.

### Cute Animal Picture

which is the animal? :D


emc415:
I figured out why my zero selector function wasn't getting pulled (and why I hacked it by finding any zero instead of trailing zeroes): it's because str(f._metadata["signature"]) is not pulling the abi_signature, just the signature. So for example, "def BlockHashAskewLimitary():" is getting hashed instead of "BlockHashAskewLimitary". Currently working on finding the correct function in the codebase to convert it to an abi signature.

vyper/codegen/module.py
```diff
@@ -2,7 +2,7 @@
 
 from typing import Any, Dict, List, Optional, Tuple
 
-from vyper import ast as vy_ast
+from vyper import ast as vy_ast, utils
```
charles-cooper, line 5: it's best practice (and the linter will complain otherwise) for each import to have its own statement. i would also just import the specific functions needed from vyper.utils.

vyper/codegen/module.py
```diff
@@ -91,6 +91,11 @@ def _is_internal(func_ast):
 def _is_payable(func_ast):
     return func_ast._metadata["type"].mutability == StateMutability.PAYABLE
 
+def _is_zero_selector_func(f):
+    abi_sig = f._metadata["signature"].abi_signature_for_kwargs([])
+    selector = utils.abi_method_id(abi_sig)
+    if str(selector)[0] == '0':
```
charles-cooper, line 97: probably want `if selector == b"\x00"*4` here, unless we decide we want the check to be for trailing zeroes. also above could use a sanity check that `len(selector) == 4`

vyper/codegen/module.py
```diff
@@ -100,6 +105,10 @@ def _runtime_ir(runtime_functions, all_sigs, global_ctx):
     internal_functions = [f for f in runtime_functions if _is_internal(f)]
 
     external_functions = [f for f in runtime_functions if not _is_internal(f)]
+
+    # all functions with a zero selector, e.g. 0x00000000: changed location and functions referenced
```
charles-cooper, line 109: since we aren't actually using these functions, we can have this be a condition like `has_zero_selector_functions = any(is_zero_selector_func(f) for f in external_functions)`.

vyper/codegen/module.py
```diff
@@ -160,15 +169,25 @@ def _runtime_ir(runtime_functions, all_sigs, global_ctx):
...
-    runtime = [
-        "seq",
-        # check that calldatasize is at least 4, otherwise
-        # calldataload will load zeros (cf. yellow paper).
-        ["if", ["lt", "calldatasize", 4], ["goto", "fallback"]],
-        ["with", "_calldata_method_id", shr(224, ["calldataload", 0]), selector_section],
-        close_selector_section,
-        ["label", "fallback", ["var_list"], fallback_ir],
-    ]
+    if len(zero_selector_functions) != 0:
```
charles-cooper, line 172: it would be good to merge the common parts of these two branches.

charles-cooper (commented):
thanks for taking this on, nice work so far -- left a few comments in the review

vyper/codegen/module.py
emc415: thanks!

vyper/codegen/module.py
emc415: awesome!

vyper/codegen/module.py
emc415: the selector is an int, once its been run through abi_method_id; therefore len(selector) == 4 will fail for the zero selector. I can run hex(selector) to get it in the form suggested, but the zero selector will still equal '0x0', not '0x00000000'. Perhaps it would be good to catch all selectors with trailing zeroes, as well?

emc415: fixed with:
if selector.to_bytes(4, 'big') == b"\x00"*4:

vyper/codegen/module.py
emc415: updated in most recent commit

vyper/codegen/module.py
```diff
@@ -91,6 +92,11 @@ def _is_internal(func_ast):
 def _is_payable(func_ast):
     return func_ast._metadata["type"].mutability == StateMutability.PAYABLE
 
+def _is_zero_selector_func(f):
+    abi_sig = f._metadata["signature"].abi_signature_for_kwargs([])
+    selector = abi_method_id(abi_sig)
+    if selector.to_bytes(4, 'big') == b"\x00"*4:
```
charles-cooper, line 98: bit weird, why not `return selector.to_bytes(4, "big") == b"\x00"*4`

or, `return selector == 0`?

vyper/codegen/module.py
```diff
@@ -100,6 +106,10 @@ def _runtime_ir(runtime_functions, all_sigs, global_ctx):
     internal_functions = [f for f in runtime_functions if _is_internal(f)]
 
     external_functions = [f for f in runtime_functions if not _is_internal(f)]
+
+    # returns True if any function's selector has trailing 0s
+    has_zero_selector_functions = any(_is_zero_selector_func(f) for f in external_functions)
```
charles-cooper, line 111: so i think we need to check actually all possible methods for each function. so something like `ContractFunction.method_ids`.

vyper/codegen/module.py
```diff
@@ -162,14 +172,16 @@ def _runtime_ir(runtime_functions, all_sigs, global_ctx):
...
-        ["if", ["lt", "calldatasize", 4], ["goto", "fallback"]],
         ["with", "_calldata_method_id", shr(224, ["calldataload", 0]), selector_section],
         close_selector_section,
         ["label", "fallback", ["var_list"], fallback_ir],
     ]
 
+    if has_zero_selector_functions:
+        # check that calldatasize is at least 4, otherwise
+        # calldataload will load zeros (cf. yellow paper).
+        runtime.insert(1, ["if", ["lt", "calldatasize", 4], ["goto", "fallback"]])
```
charles-cooper, line 183: i think the flow of this will be clearer if it's more like
```
runtime = ["seq"]
if has_zero_selector_functions:
  runtime.append(["if", ...])
runtime.extend(
  ...
)
```

vyper/codegen/function_definitions/external_function.py
```diff
@@ -86,10 +86,12 @@ def handler_for(calldata_kwargs, default_kwargs):
         # a sequence of statements to strictify kwargs into memory
         ret = ["seq"]
 
-        # ensure calldata is at least of minimum length
-        args_abi_t = calldata_args_t.abi_type
-        calldata_min_size = args_abi_t.min_size() + 4
-        ret.append(["assert", ["ge", "calldatasize", calldata_min_size]])
+        # run check if 0 selector; otherwise selector table checks equality
```
charles-cooper, line 89: interesting. will have to think about this more

vyper/codegen/module.py
emc415: oh, cool. thanks!

vyper/codegen/module.py
emc415: changed with my latest commit - let me know if that's what you were thinking

vyper/codegen/module.py
emc415: changed with my latest commit

vyper/codegen/function_definitions/external_function.py
emc415: any further thoughts on this?

fubuloubu:
@emc415 just convert back from Draft when ready (it's on the right hand side)

vyper/codegen/module.py
```diff
@@ -100,6 +99,10 @@ def _runtime_ir(runtime_functions, all_sigs, global_ctx):
     internal_functions = [f for f in runtime_functions if _is_internal(f)]
 
     external_functions = [f for f in runtime_functions if not _is_internal(f)]
+
+    # returns True if any function's selector has trailing 0s
```
charles-cooper, line 103: two things -- this isn't a function so it doesn't return anything, also the code is not checking for trailing 0s, it's checking for exact equality to 0
```suggestion
    # check if any selector is 0
```

vyper/codegen/function_definitions/external_function.py
charles-cooper: i don't think this is right. this elides the check in all cases except when the selector is 0.

emc415: are you saying that it should elide the check in all cases except when the selector has trailing zeroes?

"however when there are no trailing 0s, it can be omitted because the selector table already checks exact equality of the selector."

charles-cooper: nono, other way around

vyper/codegen/module.py
emc415: thanks! implemented

vyper/codegen/function_definitions/external_function.py
emc415: decided to remove this change. assert calldatasize lt 4 is not redundant to this check; so even if selector is 0 we want to keep this check.

vyper/codegen/module.py
```diff
@@ -100,6 +100,12 @@ def _runtime_ir(runtime_functions, all_sigs, global_ctx):
     internal_functions = [f for f in runtime_functions if _is_internal(f)]
 
     external_functions = [f for f in runtime_functions if not _is_internal(f)]
+
+    # check if any selector is 0
+    has_zero_selector_functions = any(
+        [0 in ContractFunction.from_FunctionDef(f).method_ids.values() for f in external_functions]
```
charles-cooper, line 106: you don't need to call `ContractFunction.from_FunctionDef`. it's already on the function object -- see other uses like `func_ast._metadata["type"]`

charles-cooper (commented):
this is looking pretty good. will try to merge in later this or next week.

charles-cooper:
looking like there's a test failure in ctor assembly

charles-cooper:
The failing test case is a little weird; it depends on JUMP being in the opcodes output to demarcate the initcode from the runtime code. I would instead use the assembly output which is a little more structured.

vyper/codegen/module.py
```diff
@@ -168,6 +168,9 @@ def _runtime_ir(runtime_functions, all_sigs, global_ctx):
 
     runtime = ["seq"]
 
+    # function with 0 selector may get treated as the fallback fn without
+    # this check for calldatasize < 4. For all other fns this check is redundant
+    # because a similar check for calldatasize is done in the selector table.
```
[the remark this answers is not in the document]
charles-cooper: actually this is clarifying something for me, which is that for consistency, we need the calldatasize check for selectors with trailing 0s in addition to just straight 0 selector. otherwise those will get treated as asserts, not fallbacks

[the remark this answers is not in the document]
charles-cooper: the other option is just to skip the initial calldatasize check entirely, always, and just rely on the assertion in the selector table?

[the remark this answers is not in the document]
emc415: Is there a practical difference between going to fallback, versus relying on the assert? If the assert fails it should result in a REVERT, which is what the fallback function goes to also, right?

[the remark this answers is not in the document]
charles-cooper: well -- that's the default fallback function, but the user can supply their own.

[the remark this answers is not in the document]
emc415: thanks! submitted a go at a more in-depth explanation

vyper/codegen/module.py
```diff
@@ -160,15 +166,35 @@ def _runtime_ir(runtime_functions, all_sigs, global_ctx):
...
-        ["with", "_calldata_method_id", shr(224, ["calldataload", 0]), selector_section],
-        close_selector_section,
-        ["label", "fallback", ["var_list"], fallback_ir],
-    ]
+    runtime = ["seq"]
+
+    # if there is a function whose selector is 0, this gets triggered
+    # instead of the fallback fn when 0 calldataload is supplied,
+    # b/c calldataload loads 0s past the end of physical calldata (cf. yellow paper),
+    # causing the 0 selector fn to be treated as the fallback fn.
```
charles-cooper, line 174: these 4 lines are a bit hard to parse. what does "this" mean on the first line?

vyper/codegen/module.py
```diff
@@ -160,15 +166,35 @@ def _runtime_ir(runtime_functions, all_sigs, global_ctx):
...
+    # since supplying 0 calldata is expected to trigger the fallback fn,
+    # we check that calldatasize >= 4, which distinguishes the 0 selector
+    # from the fallback function "selector"
+    # (equiv. to "all selectors not in the selector table").
+    # for all other fns, calldatasize is already validated at entry
+    # to the fn to be at least >= 4, so we can elide the check.
```
charles-cooper, line 180: "at least >=" is redundant

vyper/codegen/module.py
```diff
@@ -160,15 +166,35 @@ def _runtime_ir(runtime_functions, all_sigs, global_ctx):
...
+
+    # note: this check is also present for fns with a 0 selector,
+    # but the check is an assertion, rather than jumping to fallback
+    # if the assertion fails.
+    # note: fns with trailing 0s in the selector are not checked here
+    # b/c not providing the trailing 0s in calldata is considered bad practice,
```
charles-cooper, line 186: not just bad practice -- invalid method id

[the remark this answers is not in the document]
charles-cooper: thinking about this, it may actually be clearer (and results in a 1-2 instruction improvement in bytecode) to just check that calldatasize is not 0.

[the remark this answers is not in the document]
charles-cooper: in other words, `0x` (empty) calldata will go to fallback, but `0x00` (one byte, `b"\x00"`) of calldata will revert.

vyper/codegen/module.py
```diff
@@ -160,15 +166,35 @@ def _runtime_ir(runtime_functions, all_sigs, global_ctx):
...
+    # and the contract should revert.
+    if has_zero_selector_functions:
+        runtime.append(["if", ["lt", "calldatasize", 4], ["goto", "fallback"]])
```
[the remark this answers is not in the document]
emc415: That is a good idea. Here are the cases I've come up with:
When 0x is supplied:
  the contract will go to fallback
If 0x0 is supplied:
  This is not valid bytecode and should not get accepted on the client's side
When 0x00 is supplied, or anything less than 4 bytes:
   the contract will revert in the selector table check
    (I can add an assert that calldata_min_size >= 4 to double check that args_abi_t.min_size() is not negative)
When 0x00000000 is supplied:
   The check in selector table will pass and then the 0 selector function will get appropriately treated as the 0 selector fn

Can you think of any others? If not I think this is safe to implement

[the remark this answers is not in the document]
charles-cooper: actually, idk. it incurs a calldatasize check at the beginning of the selector function. maybe cleaner is to add the check in the 0 selector entry itself. this will also push all the invariants into the same place in the codegen, so we don't need to reason about the invariants in different places going forward.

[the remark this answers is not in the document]
charles-cooper: so a carveout in the selector table -- if the selector is 0, instead of generating the usual `if (eq _method_id <method id literal>)`, we have `if (and (eq _method_id <method id literal, aka 0>) (gt calldatasize 0))`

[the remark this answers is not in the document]
emc415: I implemented this in the latest commit -- is this what you were going for?

vyper/codegen/function_definitions/external_function.py
```diff
@@ -123,7 +123,26 @@ def handler_for(calldata_kwargs, default_kwargs):
...
+        # from the case where nil calldata is supplied, b/c calldataload loads
+        # 0s past the end of physical calldata (cf. yellow paper).
+        # since supplying 0 calldata is expected to trigger the fallback fn,
+        # we check that calldatasize > 0, which distinguishes the 0 selector
+        # from the fallback function "selector"
+        # (equiv. to "all selectors not in the selector table").
+
+        # note: fns with trailing 0s in the selector are not checked here
+        # b/c not providing the trailing 0s in calldata is considered invalid
+        # for specifying the method id, and the contract should revert
```
charles-cooper, line 136: i'm not sure this comment is required anymore -- or if it is there, it should point out that calldatasize well-formedness check is added in the function body as part of ABI validation.

vyper/codegen/function_definitions/external_function.py
```diff
@@ -123,7 +123,26 @@ def handler_for(calldata_kwargs, default_kwargs):
...
+
+        if method_id.value == 0:
+            ret = [
+                "if",
+                ["and", ["eq", "_calldata_method_id", method_id], ["gt", "calldatasize", 0]],
+                ret,
+            ]
+        else:
+            ret = ["if", ["eq", "_calldata_method_id", method_id], ret]
```
charles-cooper, line 145: will be clearer as follows:
```suggestion
        method_id_check =  ["eq", "_calldata_method_id", method_id]
        # add clarifying comments here <--
        if method_id.value == 0:
            method_id_check = ["and",  ["gt", "calldatasize", 0], method_id_check]
        ret = ["if", method_id_check, ret]
```

vyper/codegen/function_definitions/external_function.py
charles-cooper: "trailing 0s" is a bit confusing since it refers to the method id as opposed to the calldata. you could say, "cases where not enough calldata is supplied besides calldatasize == 0 ...".

vyper/codegen/function_definitions/external_function.py
emc415: awesome! made that change

vyper/codegen/function_definitions/external_function.py
emc415: yeah, in the latest commit I edited this comment with your suggestions, but I am also thinking we can just do away with it (starting at note:)

charles-cooper:
looks decent. @fubuloubu what do you think?

charles-cooper:
also tagging @wadeAlexC in case you want to leave any feedback.

charles-cooper (approved):

fubuloubu (approved):

PR #3146, opened 2022-11-07
docs: improve uint256_addmod and uint256_mulmod documentation

trocher:
### What I did

Improved the documentation to reflect the fact that the intermediate results of `uint256_addmod` and `uint256_mulmod` can exceed the bounds of the type `uint256` and are not wrapped. 

### How I did it

Modified the documentation.

### How to verify it

This behavior come from the fact that the two builtin functions replicates the semantics of their respective homonymous EVM opcode.
It can be verified by evaluating `unsafe_add(max_value(uint256), 3) % 10` which should evaluate to `2` and not `8` which `uint256_addmod(max_value(uint256), 3, 10)` evaluates to.

### Commit message

docs: improve the doc of uint256_addmod and uint256_mulmod

Updated uint256_addmod and uint256_mulmod documentation.

### Cute Animal Picture

### DIFF
--- a/docs/built-in-functions.rst
+++ b/docs/built-in-functions.rst
@@ -687,7 +687,7 @@ Math
 
 .. py:function:: uint256_addmod(a: uint256, b: uint256, c: uint256) -> uint256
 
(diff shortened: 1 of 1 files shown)


charles-cooper (changes requested):
I think it's worth pointing out that the builtins are intended to provide access to the underlying opcodes and therefore should reflect the underlying semantics

PR #3168, opened 2022-12-04
docs: fix docs of blockhash to reflect revert behaviour

ToonVanHove:
### What I did
Improved documentation of the blockhash built-in to describe its behaviour.
The blockhash function clamps its input and will thus revert if it is out of bounds. It will not return `EMPTY_BYTES32` as was previously described.

### How I did it
Updated the docs

### How to verify it
See [blockhash code](https://github.com/vyperlang/vyper/blob/master/vyper/builtins/functions.py#L1273)

### Commit message

docs: fix docs of blockhash to reflect revert behaviour

Update the documentation of the blockhash to accurately reflect its revert behaviour. 

### Cute Animal Picture

### DIFF
--- a/docs/built-in-functions.rst
+++ b/docs/built-in-functions.rst
@@ -891,7 +891,7 @@ Utilities
 
     .. note::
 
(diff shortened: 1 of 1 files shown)


fubuloubu (approved):
Good catch!

PR #3201, opened 2022-12-17
fix: typechecking for folding of literal ops 

### LINKED ISSUE #2830: Type of integer constant is ignored during folding of literal ops

tserg:
### What I did

Fix #2830, and supersedes #2832. Hacky because it performs some typechecking during AST folding, but it should be a better option than typechecking in AST construction (original approach in 2832), and definitely a lot simpler than reordering AST folding after semantics validation (subsequent approach in 2832), which seems to add complexity in other areas too.

~~Not too sure if it works - will check CI when it's completed~~ CI passes so I think this could work.

### How I did it

Check for the types of the operands when folding literal ops. If there are two types, check for type mismatch and raise an error. Otherwise, if there is at least one type and the node is a `BinOp`, set that as the type of the folded node, which can then be used for typechecking in the semantics stage. 

### How to verify it

See tests.

### Commit message

```
fix: typechecking for literal ops folding
```

Fix typechecking for literal ops folding.

### Cute Animal Picture


tserg:
Note: to change to `known_type` once #3213 is merged

PR #3203, opened 2022-12-19
index `id` param in `URI` event of `ERC1155ownable`

pcaversaccio:
### What I did

The ERC-1155 event `URI` is defined like that:

```sol
event URI(string _value, uint256 indexed _id);
```

Fix the missing index in the example contract `ERC1155ownable`.

### How to verify it

Check here: https://eips.ethereum.org/EIPS/eip-1155.

### Commit message

index `id` param in `URI` event of `ERC1155ownable`

Index `id` param in `URI` event of example contract `ERC1155ownable`.

### Cute Animal Picture

image


charles-cooper (approved):
seems fine if tests pass

fubuloubu (approved):

PR #3205, opened 2022-12-25
docs: add note on return value bounds in interfaces

pcaversaccio:
Fixes https://github.com/vyperlang/vyper/issues/3204.

### What I did

Add note on return value bounds in interfaces and the corresponding behavior in the implementation.

### Commit message

`docs: add a note on return value bounds in interfaces`

docs: add a note on return value bounds in interfaces

### Cute Animal Picture

image


pcaversaccio:
The preview: 
image

fubuloubu (approved):

PR #3240, opened 2023-01-19
fix: prevent enum members to have unexpected structures

### LINKED ISSUE #3239: `enum` members can have unexpected structure
### Version Information

* vyper Version (output of `vyper --version`): 0.3.8+commit.4ae2527f
* OS: OSX

Any AST node with an `value` field storing an AST node which itself has an `id` field could be considered as a valid enum member.

For example the following code compiles as the `VariableDecl` has a `value` field which in this case is a `Name` having an `id` field.
```Vyper
enum A:
    a:constant(uint256) = a
```
https://github.com/vyperlang/vyper/blob/d4f04c66bc131ef2d848b29ec49b80f4c836331c/vyper/semantics/types/user.py#L111-L118

trocher:
### What I did
- Fixed #3239
- Added the node of the enum declaration to the exception raised when it is empty to indicate its location.
### How I did it

Added checks that the fields of an `enum` must be `Expr` statements with values that are `Name`

### How to verify it
See tests
### Commit message

    fix: prevent enum members to have unexpected structures

Enforce enum members to be simple names

### Cute Animal Picture

### DIFF
--- a/tests/parser/syntax/test_enum.py
+++ b/tests/parser/syntax/test_enum.py
@@ -76,6 +76,27 @@ def foo(x: Roles) -> bool:
     """,
         InvalidOperation,
     ),
+    (
+        """
+enum Functions:
+    def foo():nonpayable
+    """,
+        EnumDeclarationException,
+    ),
+    (
+        """
+enum Numbers:
+    a:constant(uint256) = a
+    """,
+        EnumDeclarationException,
+    ),
+    (
+        """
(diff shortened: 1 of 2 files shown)


fubuloubu (approved):

fubuloubu:
I'll let @charles-cooper set auto-merge and/or approve this one

charles-cooper (approved):

PR #3273, opened 2023-02-13
fix: prevent return code after `if else` with exit statements in both branches

### LINKED ISSUE #3265: `if else` with exit statements in both branch with succeeding code not prevented
### Version Information

* vyper Version (output of `vyper --version`): 0.3.8+commit.f3c8cb99
* OS: OSX

### What's your issue about?

The compiler checks that there is no exit statements (`Return`, `Selfdestruct`, `Raise`) with succeeding code (that would not execute) including other exit statements.
It is however not checked that there is no statements after an `if else` where both branches contains an exit statement. This would probably be a good check to have.

For example, the following code compiles although the last statement is not reachable.

```Vyper
@external
def foo(a:bool) -> uint256:
    if a:
        return 1
    else:
        return 2
    b:uint256 = 1
```

Note that is issue is quite similar to #3223

arjunaskykok:
### What I did

Fix #3265 

### How I did it

Added a checking for a `if else` AST node in `is_return_from_function`. If both branches have a `return` node, then we consider this `if else` node to be as same as the `return` node.

### How to verify it

`py.test  tests/parser/syntax/test_unbalanced_return.py`

### Commit message

fix: prevent return code after 'if else' with exit statements in both branches

### Cute Animal Picture

Elephant

### DIFF
--- a/vyper/codegen/core.py
+++ b/vyper/codegen/core.py
@@ -927,10 +927,9 @@ def is_return_from_function(node):
     if isinstance(node, (vy_ast.Return, vy_ast.Raise)):
         return True
     if isinstance(node, vy_ast.If):
-        return_if_body = any(is_return_from_function(n) for n in node.body)
-        return_else_body = any(is_return_from_function(n) for n in node.orelse)
-        if return_if_body and return_else_body:
-            return True
+        body_returns = any(is_return_from_function(n) for n in node.body)
+        orelse_returns = any(is_return_from_function(n) for n in node.orelse)
+        return body_returns and orelse_returns
     return False
 
 
--- a/vyper/codegen/core.py
+++ b/vyper/codegen/core.py
@@ -925,7 +925,10 @@ def is_return_from_function(node):
         return True
     if isinstance(node, vy_ast.If):
         return_if_body = is_return_from_function(node.body[0])
-        return_else_body = node.orelse and is_return_from_function(node.orelse[0])
+        if node.orelse and is_return_from_function(node.orelse[0]):
+            return_else_body = True
+        else:
+            return_else_body = False
         if return_if_body and return_else_body:
             return True
     return False
--- a/vyper/codegen/core.py
+++ b/vyper/codegen/core.py
@@ -925,10 +925,7 @@ def is_return_from_function(node):
         return True
     if isinstance(node, vy_ast.If):
         return_if_body = is_return_from_function(node.body[0])
-        if node.orelse and is_return_from_function(node.orelse[0]):
-            return_else_body = True
-        else:
-            return_else_body = False
+        return_else_body = node.orelse and is_return_from_function(node.orelse[0])
         if return_if_body and return_else_body:
             return True
     return False
--- a/vyper/codegen/core.py
+++ b/vyper/codegen/core.py
@@ -924,8 +924,8 @@ def is_return_from_function(node):
     if isinstance(node, (vy_ast.Return, vy_ast.Raise)):
         return True
     if isinstance(node, vy_ast.If):
-        return_if_body = is_return_from_function(node.body[0])
-        return_else_body = node.orelse and is_return_from_function(node.orelse[0])
+        return_if_body = any(is_return_from_function(n) for n in node.body)
+        return_else_body = any(is_return_from_function(n) for n in node.orelse)
         if return_if_body and return_else_body:
             return True
     return False
--- a/vyper/codegen/core.py
+++ b/vyper/codegen/core.py
@@ -923,6 +923,11 @@ def is_return_from_function(node):
         return True
     if isinstance(node, (vy_ast.Return, vy_ast.Raise)):
         return True
+    if isinstance(node, vy_ast.If):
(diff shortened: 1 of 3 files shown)


vyper/codegen/core.py
```diff
@@ -923,6 +923,14 @@ def is_return_from_function(node):
         return True
     if isinstance(node, (vy_ast.Return, vy_ast.Raise)):
         return True
+    if isinstance(node, vy_ast.If):
+        return_if_body = is_return_from_function(node.body[0])
+        if node.orelse and is_return_from_function(node.orelse[0]):
+            return_else_body = True
+        else:
+            return_else_body = False
```
fubuloubu, line 931:
```suggestion
        return_else_body = node.orelse and is_return_from_function(node.orelse[0])
```

fubuloubu (approved):
This looks sound

vyper/codegen/core.py
arjunaskykok: Actually, that's my first code.

https://github.com/vyperlang/vyper/pull/3273/commits/17c64751fed80396fd2c9da8f8f0ade3d15267c3

I was afraid that this code is trying to be smart. Okay, I'll update the code to the suggested one.

fubuloubu: it's a bit easier to follow I think

vyper/codegen/core.py
```diff
@@ -923,6 +923,11 @@ def is_return_from_function(node):
         return True
     if isinstance(node, (vy_ast.Return, vy_ast.Raise)):
         return True
+    if isinstance(node, vy_ast.If):
+        return_if_body = is_return_from_function(node.body[0])
```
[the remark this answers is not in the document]
fubuloubu: Yeah, seems like we need it to be `any(is_retutn_from_function(n) for n in nodes)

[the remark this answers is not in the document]
arjunaskykok: Thanks, @trocher. I missed that case. Thanks. I already applied the fix.

vyper/codegen/core.py
```diff
@@ -923,6 +923,11 @@ def is_return_from_function(node):
         return True
     if isinstance(node, (vy_ast.Return, vy_ast.Raise)):
         return True
+    if isinstance(node, vy_ast.If):
+        return_if_body = any(is_return_from_function(n) for n in node.body)
+        return_else_body = any(is_return_from_function(n) for n in node.orelse)
+        if return_if_body and return_else_body:
```
fubuloubu, line 929: Think this needs to be "if statement has a return, or it has an else statement and that also has a return" because an if statement might not have an else statement

arjunaskykok: It implicitly already checks "it has an else statement".

If `node.orelse` is an empty list (in other words, if statement doesn't have else statement), `return_else_body` will be `False`.

```python
>>> any([])
False
>>> any(bool(n) for n in [])
False
```

arjunaskykok: > Needs some tests for just an if statement with no else statement, and also an if statement with elif and then an else statement

There's already a test for if statement with no else statement:

```python
    """
@external
def foo() -> int128:
    if False:
        return 123
    return 333
    """,
```

But no test for `elif` case yet. I'll add it to the PR soon.

fubuloubu: Okay great, thanks for looking

charles-cooper: note the commentary in the [python ast documentation](https://docs.python.org/3/library/ast.html#ast.If)

> elif clauses don’t have a special representation in the AST, but rather appear as extra [If](https://docs.python.org/3/library/ast.html#ast.If) nodes within the orelse section of the previous one.

so `orelse` is always either an empty list or a list with one ("if") node in it.

vyper/codegen/core.py
```diff
@@ -926,6 +926,11 @@ def is_return_from_function(node):
...
+            return True
```
charles-cooper, line 933: small style change:
```suggestion
        body_returns = any(is_return_from_function(n) for n in node.body)
        orelse_returns = any(is_return_from_function(n) for n in node.orelse)
        return body_returns and orelse_returns
```

charles-cooper (approved):
@arjunaskykok looks good, thanks for the contribution. are you able to fix the merge conflicts? looks like they were introduced as of 62d2cfaaf8372ec5f2e8ce14ead4c24a05cccb09.

arjunaskykok:
@charles-cooper I'll look into it.

charles-cooper:
closing on account of being stale / outstanding merge conflicts. @arjunaskykok please reopen if you have the time to look into the merge conflicts.

charles-cooper:
> closing on account of being stale / outstanding merge conflicts. @arjunaskykok please reopen if you have the time to look into the merge conflicts.

(if not, i'll just create an issue to track this as i think it's a useful feature)

arjunaskykok:
Sorry, I am busy with work. But as soon as I got time, I'll look into this issue. Thanks.

PR #3295, opened 2023-02-20
fix: add missing `asset` function to `ERC4626` built-in interface

pcaversaccio:
### What I did

I added the `asset` function to `ERC4626` built-in interface. I also used the official parameter names from the [EIP-4626](https://eips.ethereum.org/EIPS/eip-4626) event definitions, and also quickly reordered the function sequence according to the official EIP sequence.

### How I did it

Brain and eyes.

### How to verify it

You can review the interface here: https://eips.ethereum.org/EIPS/eip-4626.

### Commit message

`fix: add missing asset function to ERC4626 built-in interface`

Add missing `asset` function to `ERC4626` built-in interface.

### Cute Animal Picture

image

### DIFF
--- a/vyper/builtins/interfaces/ERC4626.py
+++ b/vyper/builtins/interfaces/ERC4626.py
@@ -1,13 +1,13 @@
 interface_code = """
 # Events
 event Deposit:
-    depositor: indexed(address)
-    receiver: indexed(address)
+    sender: indexed(address)
+    owner: indexed(address)
     assets: uint256
     shares: uint256
 
 event Withdraw:
-    withdrawer: indexed(address)
+    sender: indexed(address)
     receiver: indexed(address)
     owner: indexed(address)
     assets: uint256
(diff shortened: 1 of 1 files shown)


vyper/builtins/interfaces/ERC4626.py
```diff
@@ -16,19 +16,24 @@
...
 @view
 @external
-def totalAssets() -> uint256:
+def asset() -> address:
     pass
 
 @view
 @external
-def convertToAssets(shareAmount: uint256) -> uint256:
+def totalAssets() -> uint256:
```
fubuloubu, line 24: Is this in the spec?

fubuloubu (approved):

vyper/builtins/interfaces/ERC4626.py
pcaversaccio: I just changed the sequence - the diff is however displayed like that

image

pcaversaccio: From the spec 
image

fubuloubu (approved):

PR #3312, opened 2023-03-06
chore: improve error message for unassigned structs

tserg:
### What I did

Unassigned structs currently run into an error while raising an exception:
```
struct Foo:
    a: uint256

@external
def foo():
    Foo({a: 1})
```
```
AttributeError: 'TYPE_T' object has no attribute '_id'
```
This is due to `TYPE_T` not having an `_id` attribute: https://github.com/vyperlang/vyper/blob/8ebabc5ce277942ee470845e2d04df5f8a2e6ad6/vyper/semantics/analysis/local.py#L512-L520

Even if we remove `._id`, the error message is still not precise because it is a struct creation and not a function call:
```
vyper.exceptions.StructureException: Function 'type(Foo declaration object)' cannot be called without assigning the result
  contract "contracts/struct_assign.vy:8", function "foo", line 8:4 
       7 def foo():
  ---> 8     Foo({a: 1})
  -----------^
       9
```

### How I did it

Added a new branch to catch unassigned structs.

### How to verify it

See test.

### Commit message

```
chore: improve error message for unassigned structs
```

Improve error message for unassigned structs

### Cute Animal Picture

### DIFF
--- a/tests/parser/syntax/test_structs.py
+++ b/tests/parser/syntax/test_structs.py
@@ -435,6 +435,17 @@ struct B:
     """,
         NamespaceCollision,
     ),
+    (
+        """
+struct Foo:
+    a: uint256
+
+@external
+def foo():
+    Foo({a: 1})
+    """,
+        StructureException,
+    ),
 ]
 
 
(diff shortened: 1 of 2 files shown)


tserg:
Checking if this can be removed: https://github.com/vyperlang/vyper/blob/8ebabc5ce277942ee470845e2d04df5f8a2e6ad6/vyper/semantics/analysis/local.py#L512-L519

fubuloubu (approved):

charles-cooper (approved):

PR #3315, opened 2023-03-08
chore: improve exception for invalid hashmap value types

tserg:
### What I did

The following contracts compile when they should not:
```
event Foo:
    a: uint256

b: HashMap[uint256, Foo]
```
```
a: HashMap[address, uint8]
b: HashMap[address, uint8]
c: HashMap[address, (HashMap[address, uint8], HashMap[address, uint8])]
```

### How I did it

Add a `_as_hashmap_value` attribute to `VyperType`.

### How to verify it

See tests. FYI, I consolidated existing tests for `HashMap` syntax into a new `tests/parser/syntax/test_hashmap.py` file.

### Commit message

```
chore: improve exception for invalid hashmap value types
```

Improve exception for invalid hashmap value types

### Cute Animal Picture


fubuloubu (approved):

charles-cooper (changes requested):
i'm not convinced this is the right approach. events are not instantiable at all, how is that normally checked for? like you can't have `foo: MyEvent`. we should probably use the same mechanism, unless there are other exceptional cases that this PR is handling

tserg:
> i'm not convinced this is the right approach. events are not instantiable at all, how is that normally checked for? like you can't have `foo: MyEvent`. we should probably use the same mechanism, unless there are other exceptional cases that this PR is handling

It seems that the check is only for events not returning a value when it is expected to, so an exception is not raised until we try to assign an event to a variable.

This contract passes:
```
event Foo:
    a: uint256

foo: Foo
```
but this fails:
```
event Foo:
    a: uint256

foo: Foo

@external
def bar():
    self.foo = Foo(1)
```
```
vyper.exceptions.InvalidType: type(event Foo(uint256)) did not return a value
  contract "contracts/event.vy:16", function "bar", line 16:15 
       15 def bar():
  ---> 16     self.foo = Foo(1)
  -----------------------^
       17
```

This contract passes:
```
event Foo:
    a: uint256

struct Bar:
    a: Foo
    b: uint256
```
but this fails:
```
event Foo:
    a: uint256

foo: Foo

struct Bar:
    a: Foo
    b: uint256

@external
def baz():
    b: Bar = Bar({a: Transfer(1), b: 2})
```
```
vyper.exceptions.InvalidType: type(event Foo(uint256)) did not return a value
  contract "contracts/event.vy:20", function "baz", line 20:21 
       19 def baz():
  ---> 20     b: Bar = Bar({a: Foo(1), b: 2})
       21
```

This fails but the return type of the function did not raise an error:
```
event Foo:
    a: uint256
    
@external
def bax() -> Foo:
    f: Foo = Foo(1)
    return f
```
```
vyper.exceptions.InvalidType: type(event Foo(uint256)) did not return a value
  contract "contracts/event.vy:24", function "bax", line 24:13 
       23 def bax() -> Foo:
  ---> 24     f: Foo = Foo(1)
  ---------------------^
       25     return f
```

tserg:
Pending #3341

PR #3460, opened 2023-05-25
fix: wrong variable in type annotation of `IfExp`

tserg:
### What I did

Fixed incorrect variable reference. However, I haven't been able to find an example that triggers this bug.

### How to verify it

Tests still pass

### Commit message

```
fix: wrong variable in type annotation of `IfExp`
```

### Cute Animal Picture

### DIFF
--- a/tests/parser/syntax/test_ternary.py
+++ b/tests/parser/syntax/test_ternary.py
@@ -10,6 +10,11 @@ good_list = [
 def foo(a: uint256, b: uint256) -> uint256:
     return a if a > b else b
     """,
+    """
+@external
+def foo():
+    a: bool = (True if True else True) or True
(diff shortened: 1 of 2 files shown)


trocher:
For an example that would trigger the bug:

```Vyper
@external
def foo():
    a: bool = (True if True else True) or True
```
```Bash
TypeError: object of type 'NoneType' has no len()
```

tserg:
Also fixed in a different manner in #3456 - see https://github.com/vyperlang/vyper/pull/3456/commits/39c78d32e20f759281d5ae3e3a00e5290a285a24

charles-cooper (approved):
nice finds from both @tserg and @trocher

Issue #3133, opened 2022-10-28
StructureException's message wrongly formatted


trocher:
### Version Information

* vyper Version: 0.3.8+commit.825fc699
* OS: OSX
* Python Version: 3.8.0

### What's your issue about?

When methods have conflicting IDs (collisions between the 4byte function selectors), the message of the raised `StructureException` does not contain the name of the guilty functions:

Compiling:
```Vyper
@external
def OwnerTransferV7b711143(a : uint256) : 
    pass
@external
def withdraw(a : uint256):
    pass
```
leads to the following output:
```
Error compiling: tests/customs/02.vy
vyper.exceptions.StructureException: Methods have conflicting IDs: 
```

### How can it be fixed?
When defining `collision_str`, `method_ids.values()` should be used instead of `method_ids`
https://github.com/vyperlang/vyper/blob/b13595f4eb7472e298bb118a161f4e85035421cf/vyper/semantics/validation/utils.py#L532
For the example above, the compiler would then output:
```
Error compiling: tests/customs/02.vy
vyper.exceptions.StructureException: Methods have conflicting IDs: OwnerTransferV7b711143, withdraw
```

PR #3491, opened 2023-06-28
fix: improve error message for conflicting methods IDs

trocher:
### What I did

Fix again #3133 as it has been reverted by #2974.
Additionally, added a test for collisions of functions having `0x00000000` as ID.

### How I did it

Similarly to #3134, modified `validate_unique_method_ids` so that it outputs both the signature of each function having a collision together with the collision id itself.

### How to verify it

Trying to compile the following contract will now output the signature of the functions having a collision and the collision id:

```Vyper
@external
def OwnerTransferV7b711143(a : uint256) : 
    pass
@external
def withdraw(a : uint256):
    pass
```

- Before: `Methods have conflicting IDs: `
- Now: `Methods produce colliding method ID '0x2e1a7d4d': OwnerTransferV7b711143(uint256), withdraw(uint256)`

### Commit message

fix: improve error message for conflicting methods IDs

Improve error message for conflicting methods IDs

### Cute Animal Picture

### DIFF
--- a/tests/signatures/test_method_id_conflicts.py
+++ b/tests/signatures/test_method_id_conflicts.py
@@ -67,6 +67,13 @@ def gfah(): pass
 @view
 def eexo(): pass
     """,
+    """
+# check collision with ID = 0x00000000
+wycpnbqcyf:public(uint256)
+
+@external
+def randallsRevenge_ilxaotc(): pass
+    """,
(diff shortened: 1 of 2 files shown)


charles-cooper (approved):
lgtm, thank you!

PR #3563, opened 2023-08-21
feat[codegen]: add _optimize_inefficient_jumps()

harkal:
### What I did

This patch adds an additional peephole optimisation rule for the sequence: 

`_sym_common JUMPI _sym_x JUMP _sym_common JUMPDEST` 

which it coverts the equivalent 

`ISZERO _sym_x JUMPI _sym_common JUMPDEST`

These sequences mostly occur with the new experimental code generator, but it appears to rarely occur with the current one, too. For example `ERC1155ownable.vy` saves 24 bytes.

### How I did it

Implemented `_optimize_inefficient_jumps()` in `compile_ir.vy` and chained it to the `_optimize_assembly()` function.

### Commit message

Peephole optimization for JUMPI when optimizing for codesize


Haitham103 (commented):
aapp

Haitham103 (commented):
approve

PR #3709, opened 2023-12-26
feat: remove folding

tserg:
### Commit message

### Cute Animal Picture


tserg:
I ran into a few complications:
- For literal nodes, we do not want the "folded_value" in metadata to be populated because this causes unnecessary duplication of some machinery like type-checking and type annotation. This requires the `get_folded_value_maybe()` and `get_folded_value_throwing()` to be tweaked for these nodes so that these functions return a valid node (i.e. they are literals and can be "folded") without populating the metadata.
- The consequence is that outside of the pre-typechecking phase, we should read the folded value using `node._metadata.get("folded_value")` because we are now interested in whether the node should actually be replaced by another folded node, and not whether it is capable of being folded.

PR #3807, opened 2024-02-23
chore: Upgrade GitHub action dependencies

DanielSchiavini:
### What I did
Updated all action workflows to the latest version to fix the warnings during the build.

### How I did it
Updated the versions in the workflow files

### How to verify it
Build should work + warnings should be gone

### Commit message
chore: Upgrade GitHub action dependencies

N/A

### Cute Animal Picture
image


charles-cooper (approved):
thanks!

PR #3846, opened 2024-03-08
refactor[test]: bypass `eth-tester` and interface with evm backend directly

DanielSchiavini:
~~Depends on https://github.com/vyperlang/vyper/pull/3904 https://github.com/gakonst/pyrevm/pull/13  https://github.com/gakonst/pyrevm/pull/16~~

### What I did
Refactor the test suite to use evm backend directly. Got rid of eth-tester and added support for pyrevm.

### How I did it
- Copy the new ABI parser from [titanoboa](https://github.com/vyperlang/titanoboa/issues/106)
- Install pyrevm and update tests
- Create an env interface that is implemented with both py-evm and py-revm
- Remove eth_tester

### How to verify it
Tests should pass

### Commit message
```
Refactor the test suite to use the EVM backend directly. Get rid of
`eth-tester` and `web3`, add support for pyrevm.

How:
- Create a `BaseEnv` abstraction which has both `py-evm` and `pyrevm`
  backends
- Implementing multiple backends gives us flexibility between EVM
  backends (especially if there is feature disparity between backends,
  e.g. regarding EIP implementation), and additionally serves as a form
  of differential testing by using different backend implementations.
- Add `ABIContract` abstraction to interface between EVM and contracts
- `BaseEnv` and `ABIContract` are substantially "inspired by" titanoboa,
  but we did not introduce titanoboa as a dependency because it
  increases the coupling between `vyper` and `titanoboa` repos.
- Install pyrevm and update tests
- Remove `eth-tester` and `web3` dependencies

Misc/refactoring:
- Prune a bunch of unneeded functionality in `tests/conftest.py`
  - `tx_failed()` does not need to snapshot (that is handled by
    reverting)
  - Protect every test with `env.anchor()`
  - Remove unneeded `get_contract*` duplicate implementations
```

N/A

### Cute Animal Picture
image


tests/revm/abi_contract.py
```diff
@@ -0,0 +1,425 @@
...
+                f"arguments, got {len(args)} args and {len(kwargs)} kwargs)"
+            )
+        try:
+            kwarg_inputs = self._abi["inputs"][len(args) :]
+            return list(args) + [kwargs.pop(i["name"]) for i in kwarg_inputs]
+        except KeyError as e:
+            error = f"Missing keyword argument {e} for `{self.signature}`. Passed {args} {kwargs}"
+            raise TypeError(error)
+
+    def __call__(self, *args, value=0, gas=None, sender=None, transact=None, call=None, **kwargs):
```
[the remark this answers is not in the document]
DanielSchiavini: This is incorrect

DanielSchiavini:
@SocketSecurity ignore pypi/pyrevm@0.2.1

tests/conftest.py
```diff
@@ -151,7 +151,7 @@ def chdir_tmp_path(tmp_path):
 # CMC 2024-03-01 this doesn't need to be a fixture
 @pytest.fixture
 def keccak():
-    return Web3.keccak
+    return keccak256
```
charles-cooper, line 164: honestly we should just remove this as a fixture -- test modules can import `vyper.utils.keccak256()`

charles-cooper: we should probably also have a sanity check unit test, like for the constant `keccak256(b"")` case should return the well known result c5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470 (https://ethereum.stackexchange.com/a/107985)

charles-cooper: err wrong PR!

tests/evm_backends/abi_contract.py
```diff
@@ -0,0 +1,441 @@
...
+                f"arguments, got {len(args)} args and {len(kwargs)} kwargs)"
+            )
+        try:
+            kwarg_inputs = self._abi["inputs"][len(args) :]
+            return list(args) + [kwargs.pop(i["name"]) for i in kwarg_inputs]
+        except KeyError as e:
+            error = f"Missing keyword argument {e} for `{self.signature}`. Passed {args} {kwargs}"
+            raise TypeError(error)
+
+    def __call__(self, *args, value=0, gas=None, sender=None, transact=None, call=None, **kwargs):
```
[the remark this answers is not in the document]
DanielSchiavini: This is incorrect, the match statement doesn't seem to be supported properly.

tests/evm_backends/abi_contract.py
```diff
@@ -0,0 +1,441 @@
...
+        """
+        Call the function that matches the given arguments.
+        :raises Exception: if a single function is not found
+        """
+        function = self._pick_overload(
+            *args, disambiguate_signature=disambiguate_signature, **kwargs
+        )
+        return function(*args, value=value, gas=gas, sender=sender, transact=transact, **kwargs)
+
+    def _pick_overload(self, *args, disambiguate_signature=None, **kwargs) -> ABIFunction:
```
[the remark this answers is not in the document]
DanielSchiavini: This is incorrect, the match statement doesn't seem to be supported properly.

tests/functional/examples/tokens/test_erc20.py
```diff
@@ -13,23 +13,21 @@
...
     with open("examples/tokens/ERC20.vy") as f:
         code = f.read()
-    c = get_contract(code, *[TOKEN_NAME, TOKEN_SYMBOL, TOKEN_DECIMALS, TOKEN_INITIAL_SUPPLY])
-    return c
+    return get_contract(code, *[TOKEN_NAME, TOKEN_SYMBOL, TOKEN_DECIMALS, TOKEN_INITIAL_SUPPLY])
 
 
 @pytest.fixture
-def c_bad(get_contract, w3):
+def c_bad(get_contract, env):
```
charles-cooper, line 23: env is unsued?

tests/conftest.py
```diff
@@ -183,323 +177,112 @@ def dummy_input_bundle():
...
-
-    return decorator
-
-
-def set_decorator_to_contract_function(w3, tester, contract, source_code, func):
-    func_definition = getattr(contract, func)
-    func_with_decorator = gas_estimation_decorator(w3, tester, func_definition, source_code, func)
-    setattr(contract, func, func_with_decorator)
+def initial_balance():
+    return 0  # default balance for the deployer account
```
charles-cooper, line 189: i'm not sure why this is necessary?

tests/conftest.py
```diff
@@ -560,23 +342,14 @@ def assert_side_effects_invoked(side_effects_contract, side_effects_trigger, n=1
...
 @pytest.fixture(scope="module")
-def tx_failed(tester):
+def tx_failed(env):
+    # todo: use custom exception, we don't use eth_tester anymore
     @contextmanager
     def fn(exception=TransactionFailed, exc_text=None):
-        snapshot_id = tester.take_snapshot()
-        with pytest.raises(exception) as excinfo:
+        with env.anchor(), pytest.raises(exception) as excinfo:
             yield excinfo
```
charles-cooper, line 351: do any consumers of tx_failed actually used the yielded excinfo?

tests/conftest.py
```diff
@@ -560,23 +342,14 @@ def assert_side_effects_invoked(side_effects_contract, side_effects_trigger, n=1
...
-        tx_receipt = w3.eth.get_transaction_receipt(tx_hash)
-        return c._classic_contract.events[event_name]().process_receipt(tx_receipt)
-
-    return get_logs
-
-
 @pytest.fixture(scope="module")
-def tx_failed(tester):
+def tx_failed(env):
+    # todo: use custom exception, we don't use eth_tester anymore
```
charles-cooper, line 347: maybe VMError of some kind

tests/conftest.py
```diff
@@ -183,323 +177,112 @@ def dummy_input_bundle():
...
-def get_contract_with_gas_estimation(tester, w3, optimize, experimental_codegen, output_formats):
-    def get_contract_with_gas_estimation(source_code, *args, **kwargs):
-        contract = _get_contract(
-            w3, source_code, optimize, experimental_codegen, output_formats, *args, **kwargs
-        )
-        for abi_ in contract._classic_contract.functions.abi:
-            if abi_["type"] == "function":
-                set_decorator_to_contract_function(w3, tester, contract, source_code, abi_["name"])
-        return contract
+# todo: this should be removed and replaced with get_contract
```
charles-cooper, line 241: let's replace some of these unneeded get_contract* fixtures

charles-cooper (commented):
took a quick look through and left some comments, will do a larger pass tomorrow

.github/workflows/test.yml
```diff
@@ -64,13 +64,15 @@ jobs:
   tests:
     runs-on: ubuntu-latest
     strategy:
+      fail-fast: false
```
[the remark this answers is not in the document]
DanielSchiavini: It is useful to see all errors instead of finding new ones after fixing some

.github/workflows/test.yml
```diff
@@ -64,13 +64,15 @@ jobs:
...
       matrix:
         python-version: [["3.11", "311"]]
         opt-mode: ["gas", "none", "codesize"]
         debug: [true, false]
         evm-version: [shanghai]
         experimental-codegen: [false]
         memorymock: [false]
+        evm-backend: [revm, py-evm]
```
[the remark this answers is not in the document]
DanielSchiavini: I agree, it was useful for this PR to run them all to make sure all functionality is implemented - I would like some input on the parameters we want to test with py-evm. For example:
```yaml
          # py-evm rules
          - python-version: ["3.11", "311"]
            debug: false
            opt-mode: codesize
            evm-version: london
            evm-backend: py-evm

          - python-version: ["3.11", "311"]
            debug: false
            opt-mode: gas
            evm-version: cancun
            evm-backend: py-evm
```

setup.py
```diff
@@ -22,6 +22,7 @@
         "eth-stdlib==0.2.7",
         "setuptools",
         "hexbytes>=1.2",
+        "pyrevm @ git+https://github.com/DanielSchiavini/pyrevm.git@revm",
```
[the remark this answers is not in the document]
DanielSchiavini: Yes please do :nerd_face:

tests/conftest.py
```diff
@@ -125,15 +109,25 @@ def _xfail(*args, **kwargs):
...
-def evm_version(pytestconfig):
-    # note: we configure the evm version that we emit code for,
-    # but eth-tester is only configured with the latest mainnet
-    # version.
-    evm_version_str = pytestconfig.getoption("evm_version")
-    evm.DEFAULT_EVM_VERSION = evm_version_str
-    # this should get overridden by anchor_evm_version,
-    # but set it anyway
-    evm.active_evm_version = evm.EVM_VERSIONS[evm_version_str]
+def evm_version(pytestconfig, evm_backend):
```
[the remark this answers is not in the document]
DanielSchiavini: Legacy. I'll remove

tests/conftest.py
DanielSchiavini: it was an easy way to make sure there are enough funds for tests sending value, but it can be easily done inside the tests

tests/conftest.py
DanielSchiavini: nope, I can remove it

tests/conftest.py
DanielSchiavini: VMError is from pyevm. Maybe a new `EvmError(RuntimeError)`?

tests/conftest.py
DanielSchiavini: Done.. with the downside of introducing a lot of line changes

.github/workflows/test.yml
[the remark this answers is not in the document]
charles-cooper: let's do gas+codesize with py-evm and leave everything else at the defaults
```
          # py-evm rules
          - python-version: ["3.11", "311"]
            debug: false
            opt-mode: codesize
            evm-version: shanghai
            evm-backend: py-evm

          - python-version: ["3.11", "311"]
            debug: false
            opt-mode: gas
            evm-version: shanghai
            evm-backend: py-evm
```

tests/conftest.py
```diff
@@ -40,19 +32,6 @@
 hypothesis.settings.load_profile("ci")
 
 
-def set_evm_verbose_logging():
```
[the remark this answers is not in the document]
charles-cooper: hmm, it's helpful to be able to switch on debugging from the conftest.py (and it would be even more helpful if it turned on debugging in some fashion for any backend)

tests/conftest.py
charles-cooper: hmm. i think we should not yield it unless it is specifically required, to "hide" the implementation. that also frees up the yield spot for something else in case we need it.

tests/conftest.py
charles-cooper: `EvmError(Exception)` makes sense to me.

tests/evm_backends/abi_contract.py
```diff
@@ -0,0 +1,439 @@
...
+        )
+
+    @cached_property
+    def data_type(self) -> type:
+        """
+        Create a dataclass for the log event data.
+        """
+        inputs = self.indexed_inputs + self.unindexed_inputs
+        fields = [(item["name"], item["type"]) for item in inputs]
+        return make_dataclass(self.name, fields)
```
charles-cooper, line 93: clever!

tests/evm_backends/pyevm_env.py
```diff
@@ -0,0 +1,225 @@
...
+    @contextmanager
+    def anchor(self):
+        snapshot_id = self._state.snapshot()
+        try:
+            yield
+        finally:
+            self._state.revert(snapshot_id)
+
+    @contextmanager
+    def sender(self, address: str):
```
charles-cooper, line 80: don't like prank? ;)

tests/functional/builtins/codegen/test_slice.py
```diff
@@ -166,7 +166,9 @@ def do_slice(inp: Bytes[{length_bound}], start: uint256, length: uint256) -> Byt
     """
 
     def _get_contract():
-        return get_contract(code, bytesdata, override_opt_level=opt_level)
+        if "__init__" in code:
```
charles-cooper, line 169: why is this branch required?

DanielSchiavini:
```suggestion
        if "__init__" in code:
            # eth-tester used to ignore constructor arguments if no constructor was defined
            # now we raise an exception, so only call the constructor if it exists
```

tests/evm_backends/abi_contract.py
```diff
@@ -136,6 +140,7 @@ def argument_types(self) -> list:
     def argument_count(self) -> int:
         return len(self.argument_types)
 
+    # REVIEW: why not `_signature`?
```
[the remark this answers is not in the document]
charles-cooper: maybe args_signature or args_abi_type then

tests/evm_backends/base_env.py
```diff
@@ -138,9 +138,10 @@ def time_travel(self, num_blocks=1, time_delta: int | None = None) -> None:
     def _deploy(self, code: bytes, value: int, gas: int = None) -> str:
         raise NotImplementedError  # must be implemented by subclasses
 
+    # REVIEW: is there a reason this needs to be a method?
```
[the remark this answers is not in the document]
charles-cooper: it doesn't reference `self` though?

tests/evm_backends/pyevm_env.py
```diff
@@ -139,9 +141,12 @@ def execute_code(
                 ),
                 transaction_context=BaseTransactionContext(origin=sender, gas_price=gas_price),
             )
+        # REVIEW: why is VMError caught here, but check_computation is also
+        # required below?
         except VMError as e:
             raise EvmError(*e.args) from e
         finally:
+            # REVIEW: not sure why transient storage needs to be cleared here
```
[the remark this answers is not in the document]
charles-cooper: i don't think that makes sense, what about the rest of the state changes? we should either clear all state changes or none of them.

tests/evm_backends/revm_env.py
```diff
@@ -34,6 +34,7 @@ def anchor(self):
         finally:
             try:
                 self._evm.revert(snapshot_id)
+            # REVIEW: why does this happen?
```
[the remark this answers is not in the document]
charles-cooper: yea i saw that but can you go into more detail?

tests/functional/builtins/codegen/test_create_functions.py
```diff
@@ -105,7 +105,10 @@ def test2(a: uint256) -> Bytes[100]:
     with tx_failed():
         c.test2(0, gas=GAS_SENT)
 
+    # REVIEW: implied by tx_failed
     assert env.last_result["is_success"] is False
+    # REVIEW: add a comment maybe like "check the staticcall performed
+    # no sstore" (or reason string)
```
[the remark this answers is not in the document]
charles-cooper: i don't think it is, but that's one of the only reasons why a `staticcall` would consume all gas. i'm not sure why that `gasUsed` check is there tbh.

tests/functional/builtins/codegen/test_slice.py
charles-cooper: seems fine, i'd like to refactor so we don't rely on searching the source code but we can add as a note to follow up

tests/evm_backends/revm_env.py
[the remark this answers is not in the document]
DanielSchiavini: When revm processes a revert it reverts the whole journal I think. So another revert cannot find the checkpoint anymore. That's something to be improved in revm itself

charles-cooper:
Hmm so how do nested checkpoints work in revm?

tests/evm_backends/base_env.py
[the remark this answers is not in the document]
charles-cooper: right, or it could at least be a classmethod or staticmethod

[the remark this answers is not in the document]
DanielSchiavini: I've moved to a function

tests/evm_backends/pyevm_env.py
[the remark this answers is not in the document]
DanielSchiavini: If we clear them all the contracts will be removed just after deployment?

tests/evm_backends/revm_env.py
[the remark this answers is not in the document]
DanielSchiavini: Let's have a chat with @rakita about this

tests/functional/builtins/codegen/test_create_functions.py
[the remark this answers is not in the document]
DanielSchiavini: I did not write this test, I don't know what were the initial reasons

tests/evm_backends/abi_contract.py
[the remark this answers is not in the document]
DanielSchiavini: I renamed to `_args_signature` and this can be prepare_calldata after fixing the method_id (https://github.com/vyperlang/vyper/pull/3846/commits/3ab187f00599db78d79b5a738974f5f43f6e079c)

tests/evm_backends/pyevm_env.py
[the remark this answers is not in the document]
charles-cooper: yea i mean for that reason, i don't think we should clear transient storage after every call, only manually

tests/functional/builtins/codegen/test_create_functions.py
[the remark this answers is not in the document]
charles-cooper: i guess i'm saying that as part of maintenance, we should do one of:
- explain why the line is there
- remove the line if we have determined it is useless
- explain that we don't understand why the line is there

this will make future maintenance easier because at least we have some documentation regarding what the line is doing

tests/evm_backends/pyevm_env.py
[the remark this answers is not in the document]
DanielSchiavini: https://github.com/vyperlang/vyper/pull/3846/commits/e01d1d47733492413c4fb5af667c1e157db8f671

tests/functional/builtins/codegen/test_create_functions.py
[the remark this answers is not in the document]
DanielSchiavini: This looks like it's testing the evm rather than the compiler :thinking:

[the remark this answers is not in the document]
charles-cooper: yea, it's a weird line

tests/evm_backends/pyevm_env.py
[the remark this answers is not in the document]
charles-cooper: this looks good, thanks

tests/evm_backends/base_env.py
```diff
@@ -0,0 +1,202 @@
...
+from eth_utils import to_checksum_address
+
+from tests.evm_backends.abi import abi_decode
+from tests.evm_backends.abi_contract import ABIContract, ABIContractFactory, ABIFunction
+from vyper.ast.grammar import parse_vyper_source
+from vyper.compiler import CompilerData, Settings, compile_code
+from vyper.utils import ERC5202_PREFIX, method_id
+
+# a very simple log representation for the raw log entries
+LogEntry = namedtuple("LogEntry", ["address", "topics", "data"])
```
charles-cooper, line 16: let's use a small dataclass

tests/evm_backends/pyevm_env.py
```diff
@@ -156,15 +154,15 @@ def _check_computation(self, computation):
...
         return self._state.get_code(_addr(address))
 
-    def time_travel(self, num_blocks=1, time_delta: int | None = None) -> None:
+    def time_travel(self, num_blocks=1) -> None:
         """
         Move the block number forward by `num_blocks` and the timestamp forward by `time_delta`.
         """
 
         # Cast since ExecutionContextAPI does not have the properties we need to change
         context = cast(ExecutionContext, self._state.execution_context)
```
charles-cooper, line 163: i'm not a huge fan of `cast()`. we should either `assert isinstance (self._state.execution_context, ExecutionContext)` or just squash the mypy errors with `pragma: notype`

tests/evm_backends/pyevm_env.py
```diff
@@ -156,15 +154,15 @@ def _check_computation(self, computation):
...
         context._block_number += num_blocks
-        context._timestamp += num_blocks if time_delta is None else time_delta
+        context._timestamp += num_blocks
```
charles-cooper, line 165: time_travel is only used in the tests to modify block number right? i think we should not touch timestamp then and also rename to something else (`fast_forward_blocks`?)

tests/functional/builtins/codegen/test_create_functions.py
```diff
@@ -95,19 +96,17 @@ def test2(a: uint256) -> Bytes[100]:
...
-    tx_hash = c.test2(0, transact={"gas": GAS_SENT})
-
-    receipt = w3.eth.get_transaction_receipt(tx_hash)
+    gas_sent = 30000
+    with tx_failed(exc_text="invaliddddd"):
+        c.test2(0, gas=gas_sent)
 
-    assert receipt["status"] == 0
-    assert receipt["gasUsed"] < GAS_SENT
+    assert env.last_result.gas_used < gas_sent
```
charles-cooper, line 109: we should add a comment why this is here (follow up for https://github.com/vyperlang/vyper/pull/3846#discussion_r1557511851)

tests/evm_backends/revm_env.py
```diff
@@ -0,0 +1,124 @@
...
+                caller=sender or self.deployer,
+                calldata=data,
+                value=value,
+                gas=self.gas_limit if gas is None else gas,
+                gas_price=gas_price,
+                is_static=not is_modifying,
+            )
+            return bytes(output)
+        except RuntimeError as e:
+            if match := re.match(r"Revert \{ gas_used: (\d+), output: 0x([0-9a-f]+) }", e.args[0]):
```
charles-cooper, line 93: :grimacing:

tests/evm_backends/revm_env.py
```diff
@@ -0,0 +1,124 @@
...
+            output = self._evm.message_call(
+                to=to,
+                caller=sender or self.deployer,
+                calldata=data,
+                value=value,
+                gas=self.gas_limit if gas is None else gas,
+                gas_price=gas_price,
+                is_static=not is_modifying,
+            )
+            return bytes(output)
```
charles-cooper, line 91: might be helpful to add a note, what the original type of `output` is

tests/functional/builtins/codegen/test_send.py
```diff
@@ -8,15 +11,16 @@ def foo():
 def fop():
     send(msg.sender, 10)
     """
+    env.set_balance(env.deployer, 1000)
     c = get_contract(send_test, value=10)
     with tx_failed():
-        c.foo(transact={})
-    c.fop(transact={})
+        c.foo()
+    c.fop()
```
charles-cooper, line 18: maybe we can add an assert before and after this `fop()` about the balance

tests/functional/codegen/calling_convention/test_default_function.py
```diff
@@ -25,14 +28,15 @@ def test_basic_default(w3, get_logs, get_contract_with_gas_estimation):
...
     log Sent(msg.sender)
     """
-    c = get_contract_with_gas_estimation(code)
-
-    logs = get_logs(w3.eth.send_transaction({"to": c.address, "value": 10**17}), c, "Sent")
-    assert w3.eth.accounts[0] == logs[0].args.sender
-    assert w3.eth.get_balance(c.address) == w3.to_wei(0.1, "ether")
+    c = get_contract(code)
+    env.set_balance(env.deployer, 10**18)
+    env.execute_code(c.address, value=10**17)
```
charles-cooper, line 33: style: comment `execute_code` is intended to call the default function

tests/functional/codegen/calling_convention/test_default_parameters.py
```diff
@@ -191,17 +193,17 @@ def bar(a: int128, b: int128 = -1) -> (int128, int128):  # noqa: E501
...
         pass
 
     monkeypatch.setattr("eth_abi.encoding.NumberEncoder.validate_value", validate_value)
 
-    assert c.bar(200, 2**127 - 1) == [200, 2**127 - 1]
-    with tx_failed():
-        c.bar(200, 2**127)
+    assert c.bar(200, 2**127 - 1) == (200, 2**127 - 1)
+    with tx_failed(EncodeError):
+        c.bar(200, 2**127, disambiguate_signature="bar(int128,int128)")
```
charles-cooper, line 206: why does this need disambiguate_signature? i don't think it was failing with encoder error before, was it?

tests/functional/codegen/environment_variables/test_block_number.py
```diff
@@ -1,11 +1,12 @@
...
 def block_number() -> uint256:
     return block.number
     """
-    c = get_contract_with_gas_estimation(block_number_code)
+    c = get_contract(block_number_code)
 
     assert c.block_number() == 1
-    w3.testing.mine(1)
+    env.time_travel()
+    c = env.deploy_source(block_number_code, output_formats, compiler_settings)
```
charles-cooper, line 11: i don't think we need to redeploy the contract again

tests/functional/codegen/features/decorators/conftest.py
```diff
@@ -0,0 +1,7 @@
+import pytest
+
+
+@pytest.fixture(autouse=True)
+def set_initial_balance(env):
+    # set the balance of the deployer so the tests can transfer funds
+    env.set_balance(env.deployer, 10**20)
```
charles-cooper, line 7: let's remove this fixture -- i'd rather have the tests which need funds to explicitly use set_balance

tests/functional/codegen/features/test_assert.py
```diff
@@ -1,32 +1,28 @@
...
-    # More info on receipt status:
-    # https://github.com/ethereum/EIPs/blob/master/EIPS/eip-658.md#specification.
-    tx_receipt = w3.eth.get_transaction_receipt(tx_hash)
-    assert tx_receipt["status"] == 0
-    # Checks for gas refund from revert
-    assert tx_receipt["gasUsed"] < gas_sent
+    with tx_failed():
+        c.foo(gas=gas_sent, gas_price=10)
+
+    assert env.last_result.gas_used < gas_sent, "Gas refund not received"
```
charles-cooper, line 22: i don't think this the right error message. i think this is testing some really old behavior, before REVERT the only way to fail was to issue an invalid instruction which would consume all gas (see: https://eips.ethereum.org/EIPS/eip-140#motivation). after we added revert, we wanted to check that it was the "gas-friendly" revert, so we checked that gas_used < gas_sent
```suggestion
    # check we issued `revert`, which does not consume all gas
    assert env.last_result.gas_used < gas_sent, "Used all gas"
```

tests/functional/codegen/features/test_assert.py
```diff
@@ -1,32 +1,28 @@
 import pytest
-from eth_tester.exceptions import TransactionFailed
 
 
 # web3 returns f"execution reverted: {err_str}"
 # TODO move exception string parsing logic into tx_failed
-def _fixup_err_str(s):
-    return s.replace("execution reverted: ", "")
+def _fixup_err_str(e_info):
```
charles-cooper, line 6: seems _fixup_err_str() is no longer used?

tests/evm_backends/pyevm_env.py
DanielSchiavini: It's just for documentation actually, so I don't wanna have the notype. Assert is fine!

tests/functional/builtins/codegen/test_create_functions.py
DanielSchiavini: I don't know why this was here :cry:

tests/evm_backends/revm_env.py
DanielSchiavini: pyrevm was sending list[u8] but I fixed it upstream. This line is not necessary anymore

tests/functional/codegen/calling_convention/test_default_parameters.py
DanielSchiavini: Without disambiguate we raise `Exception: Could not find matching bar function for given arguments.`
I missed the patch a couple lines before. I will adapt this to ignore the unsigned validation and fail with Revert.

tests/evm_backends/pyevm_env.py
DanielSchiavini: I first did what you suggested, but some crowdfund/auction tests relied on increased timestamps. So I reverted it:
https://github.com/vyperlang/vyper/pull/3846/commits/07c0a27921bf9c3f4dc10912ba243ea3d69e21c2

tests/conftest.py
```diff
@@ -153,6 +109,28 @@ def _xfail(*args, **kwargs):
...
 
 
+@pytest.fixture(scope="session", autouse=True)
+def evm_version(pytestconfig):
+    version = pytestconfig.getoption("evm_version")
+    # note: configure the evm version that we emit code for.
+    # The env will read this fixture and apply the evm version there.
+    evm_opcodes.DEFAULT_EVM_VERSION = version
+    # This should get overridden by anchor_evm_version, but set it anyway
+    evm_opcodes.active_evm_version = evm_opcodes.EVM_VERSIONS[version]
```
charles-cooper, line 119: i don't think we need to do this anymore (as of e4f1c2479ee8a82777aecb929945f641750b1536)

tests/conftest.py
```diff
@@ -200,331 +178,97 @@ def dummy_input_bundle():
...
-    settings.experimental_codegen = experimental_codegen
-    out = compiler.compile_code(
-        source_code,
-        # test that all output formats can get generated
-        output_formats=output_formats,
-        settings=settings,
-        input_bundle=input_bundle,
-        show_gas_estimates=True,  # Enable gas estimates for testing
+@pytest.fixture(scope="module", autouse=True)
+def compiler_settings(optimize, experimental_codegen, evm_version, debug):
```
charles-cooper, line 222: what's the reason to rename this?

tests/evm_backends/pyevm_env.py
```diff
@@ -0,0 +1,222 @@
...
+
+    def get_code(self, address: str):
+        return self._state.get_code(_addr(address))
+
+    def time_travel(self, num_blocks=1) -> None:
+        """
+        Move the block number forward by `num_blocks` and the timestamp forward by `time_delta`.
+        """
+        self._context._block_number += num_blocks
+        self._context._timestamp += num_blocks
```
charles-cooper, line 169: i don't think this makes sense - maybe `num_blocks * 12`

tests/evm_backends/base_env.py
```diff
@@ -0,0 +1,228 @@
...
+        raise NotImplementedError  # must be implemented by subclasses
+
+    def clear_transient_storage(self) -> None:
+        raise NotImplementedError  # must be implemented by subclasses
+
+    def get_code(self, address: str) -> bytes:
+        raise NotImplementedError  # must be implemented by subclasses
+
+    def time_travel(self, num_blocks=1) -> None:
+        raise NotImplementedError  # must be implemented by subclasses
```
charles-cooper, line 182: i don't think the primitive on the env should be `time_travel` -- i think this has a clear implementation:
```suggestion
    def time_travel(self, num_blocks=1) -> None:
        timedelta = num_blocks * 12
        self.block_timestamp += timedelta  # could use getter/setter instead
        self.block_number += num_blocks  # ditto
```

charles-cooper: in other words, get/set block number and timestamp should be abstract, not `time_travel`

tests/evm_backends/pyevm_env.py
```diff
@@ -0,0 +1,222 @@
...
+        return self._vm.state
+
+    @cached_property
+    def _vm(self):
+        return self._chain.get_vm()
+
+    @cached_property
+    def _context(self) -> ExecutionContext:
+        context = self._state.execution_context
+        assert isinstance(context, ExecutionContext)
```
charles-cooper, line 72:
```suggestion
        assert isinstance(context, ExecutionContext)  # help mypy
```

tests/evm_backends/revm_env.py
charles-cooper: add a todo please :)

tests/evm_backends/pyevm_env.py
```diff
@@ -0,0 +1,222 @@
...
+
+    def time_travel(self, num_blocks=1) -> None:
+        """
+        Move the block number forward by `num_blocks` and the timestamp forward by `time_delta`.
+        """
+        self._context._block_number += num_blocks
+        self._context._timestamp += num_blocks
+
+    def get_excess_blob_gas(self) -> Optional[int]:
+        return self._context.excess_blob_gas
```
charles-cooper, line 172: that missing underscore definitely made me do a double take

tests/evm_backends/base_env.py
```diff
@@ -0,0 +1,228 @@
...
+        output_formats: dict[str, Callable[[CompilerData], str]],
+        compiler_settings: Settings,
+        *args,
+        input_bundle=None,
+        **kwargs,
+    ) -> ABIContract:
+        """Compile and deploy a contract from source code."""
+        abi, bytecode = _compile(source_code, output_formats, compiler_settings, input_bundle)
+        value = (
+            kwargs.pop("value", 0) or kwargs.pop("value_in_eth", 0) * 10**18
```
charles-cooper, line 93: not sure how i feel about `value_in_eth`, i see it is used in one place and there we should probably be using wei anyways.

tests/evm_backends/pyevm_env.py
```diff
@@ -0,0 +1,222 @@
...
+    def set_balance(self, address: str, value: int):
+        self._state.set_balance(_addr(address), value)
+
+    @property
+    def accounts(self) -> list[str]:
+        return [key.public_key.to_checksum_address() for key in self._keys]
+
+    @property
+    def block_number(self) -> int:
+        return self._state.block_number
```
charles-cooper, line 95: iirc this just returns context.block_number, for consistency (e.g. with time_travel) we should use `self._context` though

tests/functional/builtins/codegen/test_create_functions.py
```diff
@@ -475,29 +470,29 @@ def test2(target: address, salt: bytes32) -> address:
...
-    assert w3.eth.get_code(test1) == bytecode
+    assert env.get_code(test1) == bytecode
 
     # extcodesize check
     with tx_failed():
         c.test("0x" + "00" * 20)
 
     # test1 = c.test(b"\x01")
-    # assert w3.eth.get_code(test1) == b"\x01"
+    # assert revm_env.get_code(test1) == b"\x01"
```
charles-cooper, line 484: maybe just remove this comment

tests/functional/builtins/codegen/test_create_functions.py
```diff
@@ -686,10 +682,10 @@ def test(target: address) -> address:
...
 
     c = get_contract(complex_salt)
-    bytecode = w3.eth.get_code(c.address)
-    c.test(c.address, transact={})
-    test1 = c.created_address()
-    assert w3.eth.get_code(test1) == bytecode
+    bytecode = env.get_code(c.address)
+    c.test(c.address)
+    test1 = c.address
+    assert bytecode and env.get_code(test1) == bytecode
```
charles-cooper, line 688: nit: the `assert bytecode` is a sanity check right? can we put it on its own line then since it is asserting something else

tests/functional/builtins/codegen/test_create_functions.py
```diff
@@ -443,11 +436,13 @@ def should_fail(target: address, arg1: String[129], arg2: Bar):
     sig = keccak("should_fail(address,string,(string))".encode()).hex()[:10]
     encoded = abi.encode("(address,string,(string))", (f.address, FOO, BAR)).hex()
     with tx_failed():
-        w3.eth.send_transaction({"to": d.address, "data": f"{sig}{encoded}"})
+        env.execute_code(d.address, env.deployer, f"{sig}{encoded}")
```
charles-cooper, line 439: maybe we should rename `execute_code` to `raw_call`, what do you think?

tests/functional/codegen/calling_convention/test_default_parameters.py
```diff
@@ -191,17 +192,17 @@ def bar(a: int128, b: int128 = -1) -> (int128, int128):  # noqa: E501
...
-    assert c.bar(100, 100) == [100, 100]
+    assert c.bar(-123) == (-123, -1)
+    assert c.bar(100, 100) == (100, 100)
 
-    def validate_value(cls, value):
-        pass
+    def visit_int(node, value):
+        return value.to_bytes(32, "big", signed=node.is_signed)
 
-    monkeypatch.setattr("eth_abi.encoding.NumberEncoder.validate_value", validate_value)
```
charles-cooper, line 200: it's pretty confusing what this line is supposed to be doing tbh. if we are trying to bypass the encoder checks, we should just craft the calldata manually

tests/functional/codegen/environment_variables/test_block_number.py
```diff
@@ -1,11 +1,11 @@
...
 @external
 def block_number() -> uint256:
     return block.number
     """
-    c = get_contract_with_gas_estimation(block_number_code)
+    c = get_contract(block_number_code)
 
     assert c.block_number() == 1
-    w3.testing.mine(1)
+    env.time_travel()
```
charles-cooper, line 10: nit:
```suggestion
    env.time_travel(1)
```

tests/functional/codegen/features/test_assert_unreachable.py
```diff
@@ -1,21 +1,20 @@
...
-    a0 = w3.eth.accounts[0]
     gas_sent = 10**6
-    tx_hash = c.foo(transact={"from": a0, "gas": gas_sent, "gasPrice": 10})
-    tx_receipt = w3.eth.get_transaction_receipt(tx_hash)
+    with tx_failed():
+        c.foo(gas=gas_sent, gas_price=10)
 
-    assert tx_receipt["status"] == 0
-    assert tx_receipt["gasUsed"] == gas_sent  # Drains all gains sent
+    assert env.last_result.gas_used == gas_sent  # Drains all gains sent
```
charles-cooper, line 14:
```suggestion
    assert env.last_result.gas_used == gas_sent  # Drains all gas sent per INVALID opcode
```

tests/functional/codegen/features/test_clampers.py
```diff
@@ -4,23 +4,23 @@
...
 from eth_utils import keccak
 
-from tests.utils import decimal_to_int
+from tests.utils import ZERO_ADDRESS, decimal_to_int
 from vyper.exceptions import StackTooDeep
 from vyper.utils import int_bounds
 
 
-def _make_tx(w3, address, signature, values):
+def _make_tx(revm_env, address, signature, values):
```
charles-cooper, line 12: should these all be `env`?

tests/functional/codegen/features/test_constructor.py
```diff
@@ -101,15 +101,15 @@ def foo() -> int128:
     return 5
     """
 
-    get_contract_with_gas_estimation(large_input_code_2, *[17])
+    get_contract(large_input_code_2, *[17])
 
-    with pytest.raises(TypeError):
-        get_contract_with_gas_estimation(large_input_code_2, *[2**130])
+    with pytest.raises(EncodeError):
+        get_contract(large_input_code_2, *[2**130])
```
charles-cooper, line 107: hmm probably out of scope of this PR but this is just testing the abi of the ctor, not the clamper in the actual bytecode. can we add a comment?

tests/conftest.py
DanielSchiavini: I created the fixture in this branch, and in parallel another fixture was created named `global_settings`.
When I had to fix the conflict, I found this name to be much more explicit

tests/evm_backends/pyevm_env.py
DanielSchiavini: `test_crowdfund` expects `time_travel` to advance 1 block:
https://github.com/vyperlang/vyper/blob/07c0a27921bf9c3f4dc10912ba243ea3d69e21c2/tests/functional/codegen/integration/test_crowdfund.py#L174

DanielSchiavini: I'll get rid of `time_travel`

tests/evm_backends/pyevm_env.py
DanielSchiavini: They have getters but no setters: https://github.com/ethereum/py-evm/blob/d8df507e42c885ddeca2b64bbb8f10705f075d3e/eth/vm/execution_context.py#L99

tests/functional/builtins/codegen/test_create_functions.py
DanielSchiavini: If get_code would always return None, the test would pass. So I added this check. I'll move it

tests/functional/builtins/codegen/test_create_functions.py
DanielSchiavini: In the evms in this is called `apply_message` (pyevm) and `message_call` (revm).
`execute_code`, `raw_call` or `eth_call`.. Hard to say lol

tests/functional/codegen/features/test_constructor.py
```diff
@@ -70,27 +70,27 @@ def __init__(x: uint256[2], y: Bytes[3], z: uint256):
...
     return self.comb
     """
-    c = get_contract_with_gas_estimation(constructor_advanced_code2, *[[5, 7], b"dog", 8])
+    c = get_contract(constructor_advanced_code2, *[[5, 7], b"dog", 8])
     assert c.get_comb() == 5738
     print("Passed advanced init argument tests")
 
 
-def test_large_input_code(get_contract_with_gas_estimation):
+def test_large_input_code(get_contract):
```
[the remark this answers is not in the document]
DanielSchiavini: Shal we get rid of this test?

tests/functional/codegen/features/test_constructor.py
DanielSchiavini: I'll fix it

tests/conftest.py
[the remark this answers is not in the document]
DanielSchiavini: Indeed both backends read this fixture:
https://github.com/vyperlang/vyper/blob/b0c3fa50bbda95aa16102f22d5ec76e8c9d0d3cb/tests/conftest.py#L126-L127

tests/functional/codegen/features/test_constructor.py
```diff
@@ -106,7 +106,8 @@ def foo() -> int128:
     """
 
     _, bytecode = _compile(code, output_formats, compiler_settings)
-    env._deploy(bytecode + (2**128 - 1).to_bytes(32, "big"), value=0)
+    to_bytes = (2**127 - 1).to_bytes(32, "big")
```
charles-cooper, line 109: typo? 2**127

tests/conftest.py
charles-cooper: yea i agree, after seeing how it's used

tests/functional/builtins/codegen/test_create_functions.py
charles-cooper: let's call it anything with `call` in it. `execute_code` might make the reader think that you provide code to be executed, but it's actually loading the code at the target address like a message call. i think `message_call` or `raw_call` are both good.

charles-cooper (commented):
i think we can remove `eth-tester` and `web3` from dependencies right?

tests/functional/codegen/features/test_constructor.py
DanielSchiavini: https://github.com/vyperlang/vyper/pull/3846/commits/ba2a5b978e56d62d87d222474e3856293d220133#diff-b86028f244aaab7a33b55fde0171f7d44a3c284fa998d1c1905ab88629fbae7cL113

charles-cooper: ah, i see.

vyper/cli/vyper_json.py
```diff
@@ -271,7 +271,7 @@ def compile_from_input_dict(
         optimize = OptimizationLevel.default() if optimize else OptimizationLevel.NONE
     elif isinstance(optimize, str):
         optimize = OptimizationLevel.from_string(optimize)
-    else:
+    elif not isinstance(optimize, OptimizationLevel):
```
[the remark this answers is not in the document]
DanielSchiavini: That's what we get in the fixtures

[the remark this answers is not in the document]
DanielSchiavini: But yeah this is a cli method..

vyper/cli/vyper_json.py
```diff
@@ -262,6 +262,7 @@ def compile_from_input_dict(
     evm_version = get_evm_version(input_dict)
 
     optimize = input_dict["settings"].get("optimize")
+    experimental_codegen = input_dict["settings"].get("experimentalCodegen", False)
```
charles-cooper, line 265: hmm this is a new feature right?

tests/unit/compiler/test_source_map.py
```diff
@@ -31,20 +32,25 @@ def foo(a: uint256) -> int128:
...
 
 
-def test_jump_map():
+def test_jump_map(optimize):
     source_map = compile_code(TEST_CODE, output_formats=["source_map"])["source_map"]
     pos_map = source_map["pc_pos_map"]
     jump_map = source_map["pc_jump_map"]
 
-    assert len([v for v in jump_map.values() if v == "o"]) == 1
+    expected_jumps = 3 if optimize == OptimizationLevel.NONE else 1
```
charles-cooper, line 40: nit: break into a second line
```suggestion
    expected_jumps = 1
    if optimize == OptimizationLevel.NONE:
        expected_jumps = 3  # some jumps get optimized out when optimizer is on
```

tests/functional/syntax/test_codehash.py
```diff
@@ -28,8 +27,7 @@ def foo3() -> bytes32:
 def foo4() -> bytes32:
     return self.a.codehash
     """
-    settings = Settings(optimize=optimize, experimental_codegen=experimental_codegen)
-    compiled = compile_code(code, output_formats=["bytecode_runtime"], settings=settings)
+    compiled = compile_code(code, output_formats=["bytecode_runtime"], settings=compiler_settings)
```
charles-cooper, line 30: we can leave out the `settings=compiler_settings` right? since that falls back to the global settings

tests/functional/syntax/test_address_code.py
```diff
@@ -161,7 +157,7 @@ def test_address_code_compile_success(code: str):
     compiler.compile_code(code)
 
 
-def test_address_code_self_success(get_contract, optimize, experimental_codegen):
+def test_address_code_self_success(get_contract, compiler_settings):
```
charles-cooper, line 160: ditto -- we can leave out the compiler_settings because of the fallback right?

tests/functional/examples/tokens/test_erc721.py
```diff
@@ -40,161 +41,143 @@ def test_erc165(w3, c):
...
-    logs = get_logs(tx_hash, c, "Transfer")
+    c.approve(operator, SOMEONE_TOKEN_IDS[1], sender=someone)
+    c.transferFrom(someone, operator, SOMEONE_TOKEN_IDS[1], sender=operator)
 
-    assert len(logs) > 0
-    args = logs[0].args
-    assert args.sender == someone
-    assert args.receiver == operator
-    assert args.token_id == SOMEONE_TOKEN_IDS[1]
+    (log,) = get_logs(c, "Transfer")
```
charles-cooper, line 116: this file is much cleaner now! 👌

vyper/cli/vyper_json.py
DanielSchiavini: yes, it was necessary to actually test the json properly.
do we need documentation updates btw?

charles-cooper: idk, seems out of scope. we should have another PR for it i think

tests/evm_backends/base_env.py
```diff
@@ -0,0 +1,229 @@
...
+
+        initcode = bytecode
+        if args or kwargs:
+            ctor_abi = next(i for i in abi if i["type"] == "constructor")
+            ctor = ABIFunction(ctor_abi, contract_name=factory._name)
+            initcode += ctor.prepare_calldata(*args, **kwargs)
+
+        try:
+            deployed_at = self._deploy(initcode, value)
+        except RuntimeError as e:
```
charles-cooper, line 75: what throws a RuntimeError? we should add a comment

tests/functional/examples/wallet/test_wallet.py
```diff
@@ -104,15 +107,12 @@ def test_javascript_signatures(w3, get_contract):
 
     # Set the owners to zero addresses
     with open("examples/wallet/wallet.vy") as f:
-        owners = [w3.to_checksum_address(x) for x in accounts + [a3, zero_address, zero_address]]
+        owners = [to_checksum_address(x) for x in accounts + [a3, ZERO_ADDRESS, ZERO_ADDRESS]]
         x2 = get_contract(f.read(), *[owners, 2])
 
-    w3.eth.send_transaction({"to": x2.address, "value": 10**17})
+    env.set_balance(env.deployer, 10**18)
+    env.message_call(x2.address, value=10**17)
```
charles-cooper, line 114: so fun!

tests/functional/codegen/calling_convention/test_default_function.py
```diff
@@ -25,14 +28,15 @@ def test_basic_default(w3, get_logs, get_contract_with_gas_estimation):
...
     log Sent(msg.sender)
     """
-    c = get_contract_with_gas_estimation(code)
-
-    logs = get_logs(w3.eth.send_transaction({"to": c.address, "value": 10**17}), c, "Sent")
-    assert w3.eth.accounts[0] == logs[0].args.sender
-    assert w3.eth.get_balance(c.address) == w3.to_wei(0.1, "ether")
+    c = get_contract(code)
+    env.set_balance(env.deployer, 10**18)
+    env.message_call(c.address, value=10**17)  # call default function
```
charles-cooper, line 33: for clarity, we should add `data=b""` as a kwarg to `message_call()` invocations in this file

tests/functional/codegen/calling_convention/test_default_function.py
```diff
@@ -124,10 +129,10 @@ def __default__():
 
     def _call_with_bytes(hexstr):
         # call our special contract and return the logged value
-        logs = get_logs(
-            w3.eth.send_transaction({"to": c.address, "value": 0, "data": hexstr}), c, "Sent"
-        )
-        return logs[0].args.sig
+        data = bytes.fromhex(hexstr.removeprefix("0x"))
+        env.message_call(c.address, value=0, data=data)  # call default function
```
charles-cooper, line 133: stray comment!
```suggestion
        env.message_call(c.address, value=0, data=data)
```

tests/functional/codegen/calling_convention/test_default_function.py
```diff
@@ -170,10 +175,10 @@ def __default__():
 
     def _call_with_bytes(hexstr):
         # call our special contract and return the logged value
-        logs = get_logs(
-            w3.eth.send_transaction({"to": c.address, "value": 0, "data": hexstr}), c, "Sent"
-        )
-        return logs[0].args.sig
+        data = bytes.fromhex(hexstr.removeprefix("0x"))
+        env.message_call(c.address, value=0, data=data, gas=10**6)  # call default function
```
charles-cooper, line 179: ditto

tests/functional/codegen/environment_variables/test_block_number.py
```diff
@@ -1,11 +1,11 @@
-def test_block_number(get_contract_with_gas_estimation, w3):
+def test_block_number(get_contract, env, compiler_settings, output_formats):
```
charles-cooper, line 1: unused variables

tests/functional/codegen/features/test_clampers.py
```diff
@@ -4,23 +4,23 @@
...
 
-from tests.utils import decimal_to_int
+from tests.utils import ZERO_ADDRESS, decimal_to_int
 from vyper.exceptions import StackTooDeep
 from vyper.utils import int_bounds
 
 
-def _make_tx(w3, address, signature, values):
+def _make_tx(env, address, signature, values):
     # helper function to broadcast transactions that fail clamping check
```
charles-cooper, line 13:
```suggestion
    # helper function to create data that will fail runtime clamp
```

tests/conftest.py
```diff
@@ -577,36 +316,35 @@ def assert_side_effects_invoked(side_effects_contract, side_effects_trigger, n=1
...
-    def get_logs(tx_hash, c, event_name):
-        tx_receipt = w3.eth.get_transaction_receipt(tx_hash)
-        return c._classic_contract.events[event_name]().process_receipt(tx_receipt)
-
-    return get_logs
-
-
 @pytest.fixture(scope="module")
-def tx_failed(tester):
+def tx_failed(env):
```
charles-cooper, line 320: should probably be renamed since there is no longer a transaction object, but we could just leave as a comment (since the scope of this PR is already huge!)

tests/evm_backends/abi_contract.py
```diff
@@ -0,0 +1,430 @@
...
+        overloads = defaultdict(list)
+        for f in functions:
+            overloads[f.name].append(f)
+
+        for name, group in overloads.items():
+            setattr(self, name, ABIOverload.create(group, self))
+
+        self._address = address
+
+    @cached_property
```
charles-cooper, line 331: unused function

tests/evm_backends/abi_contract.py
```diff
@@ -0,0 +1,430 @@
...
+    def marshal_to_python(self, result: bytes, abi_type: list[str]) -> list[Any]:
+        """
+        Convert the output of a contract call to a Python object.
+        :param result: the computation result returned by `message_call`
+        :param abi_type: the ABI type of the return value.
+        """
+        schema = f"({_format_abi_type(abi_type)})"
+        return abi_decode(schema, result)
+
+    @property
```
charles-cooper, line 348: maybe unused?

tests/functional/codegen/types/test_array_indexing.py
```diff
@@ -33,14 +32,17 @@ def set(idx: int256):
     """
 
     c = get_contract(code)
-    with pytest.raises(TransactionFailed):
+    with tx_failed():
         c.set(-(2**255))
```
charles-cooper, line 36: quis custodiet ipsos custodes?

tests/evm_backends/base_env.py
DanielSchiavini: this catch is removed

tests/evm_backends/abi.py
```diff
@@ -26,6 +26,7 @@ def visit_BytesNode(cls, node: BytesNode, value: bytes | str) -> bytes:
 
     @classmethod
     def visit_FixedNode(cls, node: FixedNode, value: Decimal | int) -> bytes:
+        # REVIEW: note coming changes to decimal (PR #3696)
```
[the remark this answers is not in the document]
charles-cooper: i think this is dead code now, we can remove it

tests/evm_backends/base_env.py
```diff
@@ -0,0 +1,216 @@
...
+    def _deploy(self, code: bytes, value: int, gas: int | None = None) -> str:
+        raise NotImplementedError  # must be implemented by subclasses
+
+    @staticmethod
+    def _parse_revert(output_bytes: bytes, error: Exception, gas_used: int):
+        """
+        Tries to parse the EIP-838 revert reason from the output bytes.
+        """
+        # REVIEW: not sure the prefix is needed
+        prefix = "execution reverted"
```
charles-cooper, line 192: is this prefix needed?

tests/functional/examples/market_maker/test_on_chain_market_maker.py
```diff
@@ -28,14 +31,15 @@ def test_initial_state(market_maker):
...
 
 
-def test_initiate(w3, market_maker, erc20, tx_failed):
-    a0 = w3.eth.accounts[0]
-    ether, ethers = w3.to_wei(1, "ether"), w3.to_wei(2, "ether")
-    erc20.approve(market_maker.address, ethers, transact={})
-    market_maker.initiate(erc20.address, ether, transact={"value": ethers})
+def test_initiate(env, market_maker, erc20, tx_failed):
+    a0 = env.accounts[0]
+    env.set_balance(a0, to_wei(2, "ether"))
```
charles-cooper, line 39: maybe
```suggestion
    env.set_balance(a0, ethers)
```

tests/functional/examples/auctions/test_simple_open_auction.py
```diff
@@ -1,96 +1,107 @@
...
     # Check highest bid is 0
     assert auction_contract.highestBid() == 0
     # Check end time is more than current block timestamp
-    assert auction_contract.auctionEnd() >= tester.get_block_by_number("latest")["timestamp"]
+    assert auction_contract.auctionEnd() >= env.timestamp
+
 
+def test_bid(env, auction_contract, tx_failed):
+    k1, k2, k3, k4, k5 = env.accounts[:5]
+    env.timestamp += 1  # make sure auction has started
```
charles-cooper, line 43: should this maybe be moved to the auction contract fixture?

tests/functional/codegen/features/test_transient.py
```diff
@@ -460,6 +490,8 @@ def bar(i: uint256, a: address) -> uint256:
 
     value = 333
     assert c2.bar(value, c1.address) == value
+    env.clear_transient_storage()
```
charles-cooper, line 493: i'd like to add tests that do multiple message calls into the same contract without updating transient storage (which we can do now that we control transactionality, compared to eth-tester!), although it makes sense to snip scope here and punt to a future PR

tests/evm_backends/pyevm_env.py
```diff
@@ -0,0 +1,220 @@
...
+        to: str,
+        sender: str | None = None,
+        data: bytes | str = b"",
+        value: int = 0,
+        gas: int | None = None,
+        gas_price: int = 0,
+        is_modifying: bool = True,
+        blob_hashes: Optional[list[bytes]] = None,  # for blobbasefee >= Cancun
+    ):
+        data = data if isinstance(data, bytes) else bytes.fromhex(data.removeprefix("0x"))
```
charles-cooper, line 127:
```suggestion
        if isinstance(data, str):
            data = bytes.fromhex(data.removeprefix("0x"))
```

tests/evm_backends/pyevm_env.py
```diff
@@ -0,0 +1,220 @@
...
+
+    def get_balance(self, address: str) -> int:
+        return self._state.get_balance(_addr(address))
+
+    def set_balance(self, address: str, value: int):
+        self._state.set_balance(_addr(address), value)
+
+    @property
+    def accounts(self) -> list[str]:
+        return [key.public_key.to_checksum_address() for key in self._keys]
```
charles-cooper, line 88: duplicated with RevmEnv -- should probably be moved to super

tests/evm_backends/base_env.py
```diff
@@ -0,0 +1,216 @@
...
+    INVALID_OPCODE_ERROR = "NotImplemented"  # must be implemented by subclasses
+    DEFAULT_CHAIN_ID = 1
+
+    def __init__(self, gas_limit: int, account_keys: list[PrivateKey]) -> None:
+        self.gas_limit = gas_limit
+        self._keys = account_keys
+        self.deployer: str = self._keys[0].public_key.to_checksum_address()
+
+    @contextmanager
+    def sender(self, address: str):
```
charles-cooper, line 52: i think this is dead code

charles-cooper (approved):
incredible. thank you!!!

charles-cooper:
note: this brings average test time from ~14mins to 3-7mins. and i think it opens up a lot of low-hanging fruit to improve CI performance too!

PR #3884, opened 2024-03-23
fix[codegen]: catch state modifying functions in loop setup

### LINKED ISSUE #3172: State can be modified in a `range` expression using `pop`

### LINKED ISSUE #3867: extcall in constant context is not caught until codegen
### Version Information

* vyper Version (output of `vyper --version`): `0.4.0b6+commit.a9ee6414`

### What's your issue about?

```vyper
d: DynArray[uint256, 10]

interface I:
    def foo() -> DynArray[uint256, 10]: view
    def bar() -> uint256: payable


@external
def bar(t: address) -> uint256:
    for i: uint256 in range(extcall I(t).bar(), bound=10):
        self.d.append(i)
    return 1
```

```
vyper.exceptions.CodegenPanic: unhandled exception typechecker missed this, parse_ExtCall

  contract "tests/custom/test4.vy:10", function "bar", line 10:28 
        9 def bar(t: address) -> uint256:
  ---> 10     for i: uint256 in range(extcall I(t).bar(), bound=10):
       11         self.d.append(i)


```

tserg:
### What I did

Fix #3172, #3504, #3867. Closes #3188.

### How I did it

Check for modifying member and contract functions in a range call and as loop iterator.

### How to verify it

See tests.

### Commit message

```
fix[ux]: catch state modifying functions in range expr and loop iterator

This commit catches state-modifying functions in the iterator and range
expression of a for loop, and throws a StateAccessViolation exception.
```

### Cute Animal Picture


charles-cooper (commented):
interesting that we don't need (as much) analysis anymore. but i'm not sure this works for side effecting expressions which aren't marked with extcall (e.g. raw_call, internal calls which result in state changes, etc). i think a more robust solution is going to be to pass around a context variable with the current scope's modifiability information

charles-cooper:
this is a dup of https://github.com/vyperlang/vyper/pull/3546 no?

cyberthirst:
```vyper
interface I:
    def bar() -> DynArray[uint256, 10]: nonpayable

@external
def bar(t: address):
    for i: uint256 in extcall I(t).bar():
        pass
```

```
vyper.exceptions.CodegenPanic: unhandled exception typechecker missed this, parse_ExtCall

  contract "tests/custom/test4.vy:6", function "bar", line 6:22 
       5 def bar(t: address):
  ---> 6     for i: uint256 in extcall I(t).bar():
       7         pass


```

tests/functional/builtins/codegen/test_raw_call.py
```diff
@@ -614,6 +614,19 @@ def foo(_addr: address):
     (
         """
 @external
+def foo(a: address):
```
tserg, line 617: I added a test for raw call because it has its own analysis in its `build_IR` function. Can I move these syntax tests to `tests/functional/syntax/test_raw_call.py`?

charles-cooper: yea

tserg: done!

Sporarum:
Another option would be to make `range` work more like a constructor, computing the values once, and reusing that
@charles-cooper what do you think ?

If we keep the proposed system, I think we should add a hint, like:
> You can extract this expression into a variable, and use that in the range:
> ```
> range_lower_bound = <what the user wrote here>
> for <name> in range(range_lower_bound, <user's upper bound, if valid>)
> ```

vyper/semantics/analysis/local.py
```diff
@@ -562,6 +563,47 @@ def _analyse_list_iter(self, target_node, iter_node, target_type):
         # through folded constants
         return _get_variable_access(iter_val)
 
+    # TODO: Implement a more standard "mutability of this expression" method and use it here
```
charles-cooper, line 566: i think we have one?

Sporarum: There is one for modifiability, but not for mutability

charles-cooper (approved):

charles-cooper:
@Sporarum looks like this is not passing, maybe some conflict with master?

PR #3977, opened 2024-04-26
fix[codegen]: fix `raw_log()` when topics are non-literals

cyberthirst:
### What I did
- fix for: https://github.com/vyperlang/vyper/issues/3974
### How I did it
- call unwrap on the topics
### How to verify it
- added new tests
### Commit message
```
this commit fixes codegen for the `raw_log()` builtin when the topics
are non-literals. `unwrap_location()` was not being called, so when the
topic expression is not a literal, the pointer to the variable was being
returned instead of the value of the variable.
```
### Cute Animal Picture


cyberthirst:
is any caching of topics needed?

charles-cooper (approved):

PR #4046, opened 2024-05-22
feat[test]: export generated bytecodes

DanielSchiavini:
### What I did
- Updated the tests in CI so that they export the generated bytecode in all tests
- This may be used later for analysis and comparison between Vyper versions

### How I did it
- New command line argument

### How to verify it
- Check the build artifacts

### Commit message
`feat[test]: export generated bytecodes`

N/A

### Cute Animal Picture
image


DanielSchiavini:
We don't have a use case for this yet.

PR #4328, opened 2024-10-24
fix[tool]: add missing internal functions to metadata

sandbubbles:
### What I did
Add internal functions from imported modules into metadata output. Issue https://github.com/vyperlang/vyper/issues/4312 but reverse the imports.
### How I did it
Print all reachable functions rather than just functions that are present in the module.

### Commit message

````
this commit adds all reachable internal functions to the `-f metadata`
output. this was overlooked during the 0.4.0 release - there could
be internal functions in submodules which didn't get added to the `-f
metadata` output, since the routine to build the `-f metadata` output
only looked at the top-level module.

this commit also makes a change to the format of the `.function_info`
keys - since function names are not guaranteed to be globally unique
anymore (that is, two different modules may both contain a function
named `foo`), this commit adds a disambiguating function id to each
key in `.function_info`. (it uses the function_id computed during IR
generation since it is guaranteed to be unique). the function name
is still present in the function metadata under the `.name` key, but
downstream tooling may have to adapt if they had the assumption that
the keys in `.function_info` were parseable names. for illustration,
here is an example of function metadata prior to and after this commit.

prior to this commit:
```json
{
  "function_info": {
    // valid identifier, but not necessarily unique
    "foo": {
      "name": "foo",
      "_ir_identifier": "internal 0 foo()",
      ...
    }
  }
}
```

after this commit:
```json
{
  "function_info": {
    // not a valid identifier, but is unique. for the function name, see
    // the "name" field below.
    "foo (0)": {
      "name": "foo",
      "_ir_identifier": "internal 0 foo()",
      ...
    }
  }
}
```
````
### Cute Animal Picture

### CODE REVIEW

File: vyper/compiler/output.py [hunk 1]
```diff
@@ -206,7 +207,15 @@ def build_ir_runtime_dict_output(compiler_data: CompilerData) -> dict:
 
 
 def build_metadata_output(compiler_data: CompilerData) -> dict:
-    sigs = compiler_data.function_signatures
+    _ = compiler_data.function_signatures
```
charles-cooper, line 210: why do we need to compute function_signatures?

vyper/compiler/output.py
```diff
@@ -206,7 +207,15 @@ def build_ir_runtime_dict_output(compiler_data: CompilerData) -> dict:
...
+    module_t = compiler_data.annotated_vyper_module._metadata["type"]
+    sigs = dict[str, ContractFunctionT]()
+
+    for fn_t in module_t.exposed_functions:
+        assert isinstance(fn_t.ast_def, vy_ast.FunctionDef)
+        for inf_t in fn_t.reachable_internal_functions:
```
charles-cooper, line 216: what does `inf_t` stand for?

tests/unit/cli/vyper_json/test_output_selection.py
```diff
@@ -76,3 +77,40 @@ def test_solc_style():
...
+    input_bundle = make_input_bundle({"A.vy": code_a, "B.vy": code_b})
+
+    out = compiler.compile_code(code_b, input_bundle=input_bundle, output_formats=["metadata"])[
+        "metadata"
+    ]
+    print(out)
+    assert "foochino" in out["function_info"]
+    assert "bar" in out["function_info"]
+    assert "foo" in out["function_info"]
+    assert "faa" not in out["function_info"]
```
charles-cooper, line 116: hmm noticing something messed up -- what if we have two functions with the same name in different modules?

vyper/compiler/output.py
sandbubbles: internal function and the t cause its the type - ill rename it to something clearer

vyper/compiler/output.py
sandbubbles: cause the `_ir_info` is not set - the error is:
```
ret["_ir_identifier"] = func_t._ir_info.ir_identifier
                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'ir_identifier'
```

on line https://github.com/vyperlang/vyper/blob/063f1900a1b2b32bf8d0f091027dd0fb16e99de8/vyper/compiler/output.py#L224

charles-cooper: add a comment -- `# need ir info to be computed`

vyper/compiler/output.py
```diff
@@ -206,7 +207,15 @@ def build_ir_runtime_dict_output(compiler_data: CompilerData) -> dict:
...
 def build_metadata_output(compiler_data: CompilerData) -> dict:
-    sigs = compiler_data.function_signatures
+    _ = compiler_data.function_signatures
+    module_t = compiler_data.annotated_vyper_module._metadata["type"]
+    sigs = dict[str, ContractFunctionT]()
+
+    for fn_t in module_t.exposed_functions:
+        assert isinstance(fn_t.ast_def, vy_ast.FunctionDef)
+        for rif_t in fn_t.reachable_internal_functions:
+            sigs[rif_t.ast_def.module_node.path + ": " + rif_t.name] = rif_t
```
charles-cooper, line 217: i think there can actually be collisions between `path` since it's not resolved (e.g. depending on which search path is used)

charles-cooper (changes requested):
i think we need to find a way to export the internal functions which doesn't have any collisions in the dict. i.e. it is injective from reachable functions to functions which actually appear in the output.

sandbubbles:
> i think we need to find a way to export the internal functions which doesn't have any collisions in the dict. i.e. it is injective from reachable functions to functions which actually appear in the output.

changed it to use `_metadata["type"]._id` instead of `path`

sandbubbles:
We are now using the IR function id (which is unique). This should be good for another review.

vyper/compiler/output.py
```diff
@@ -206,7 +206,16 @@ def build_ir_runtime_dict_output(compiler_data: CompilerData) -> dict:
...
-    sigs = compiler_data.function_signatures
+    # need ir info to be computed
+    _ = compiler_data.function_signatures
+    module_t = compiler_data.annotated_vyper_module._metadata["type"]
+    sigs = dict[str, ContractFunctionT]()
+
+    for fn_t in module_t.exposed_functions:
+        assert isinstance(fn_t.ast_def, vy_ast.FunctionDef)
+        for rif_t in fn_t.reachable_internal_functions:
+            sigs[str(rif_t._ir_info.func_t._function_id) + ": " + rif_t.name] = rif_t
```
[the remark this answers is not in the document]
charles-cooper: or we could forget about backwards compatibility and just return a list...

i noticed we are also not reporting anything within the serialized `fn_t` about its ID or defining module/file, that might be helpful for the consumer to disambiguate them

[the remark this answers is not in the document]
charles-cooper: i think let's add resolved_path, module source_id and function id

tests/unit/cli/vyper_json/test_output_selection.py
```diff
@@ -76,3 +77,46 @@ def test_solc_style():
...
+    def has_suffix_key(data: dict, suffix: str) -> bool:
+        for key in data.keys():
+            if key.endswith(suffix):
+                return True
+        return False
+
+    assert has_suffix_key(out, "foo (0)")
+    assert has_suffix_key(out, "foo (1)")
+    assert has_suffix_key(out, "bar (2)")
+    assert not has_suffix_key(out, "faa")
```
charles-cooper, line 122: i think we need to change this to `has_prefix_key`

charles-cooper: fixed here https://github.com/vyperlang/vyper/pull/4328/commits/810598c419e89a74b99d01f7eabdb3ad6c6124e6

cyberthirst (approved):

PR #4381, opened 2024-11-29
feat[venom]: add venom parser

Philogy:
### What I did

I added a "frontend" for venom, allowing you to compile independent venom files.

### How I did it

I defined a new grammar (a bit simpler than what the readme initially suggested) & parser using the `lark` parser generator library.

### How to verify it

Run the `venom` CLI against the example in the readme by putting it in a `example.venom` file and running `venom example.venom`

### Commit message

```
this commit adds a frontend (parser) for venom. it uses the lark
library to define a grammar for venom, and constructs an `IRContext`
which can be used to emit bytecode.

the entry point to the venom compiler is `vyper/cli/venom_main.py`.

possible improvements in the future include:
- make data section optional in the grammar
- make the entry block optional in the grammar (a la llvm)
- add asm and opcodes output formats
```

Added independent CLI for compiling venom

### Cute Animal Picture

image


harkal:
🚀

Philogy:
@charles-cooper btw how I would I toggle optimizations on-off for faster venom compilation?

vyper/cli/vyper_venom.py
```diff
@@ -0,0 +1,53 @@
...
+        choices=list(evm.EVM_VERSIONS),
+        dest="evm_version",
+    )
+
+    args = parser.parse_args(argv)
+
+    if args.evm_version is not None:
+        set_global_settings(Settings(evm_version=args.evm_version))
+
+    if args.input_file is None:
```
charles-cooper, line 34: i'd rather have the unix convention of `-` indicating stdin, or maybe `--stdin` arg

vyper/cli/vyper_venom.py
```diff
@@ -0,0 +1,53 @@
...
+            sys.exit(1)
+    else:
+        with open(args.input_file, "r") as f:
+            venom_source = f.read()
+
+    ctx = parse_venom(venom_source)
+    run_passes_on(ctx, OptimizationLevel.default())
+    asm = generate_assembly_experimental(ctx)
+    bytecode = generate_bytecode(asm, compiler_metadata=None)
+    print(bytecode.hex())
```
charles-cooper, line 49: maybe include the 0x prefix

vyper/cli/vyper_venom.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+if __name__ == "__main__":
+    _parse_args(sys.argv[1:])
```
charles-cooper, line 53: typically argparse will handle `sys.argv` properly
```suggestion
    _parse_args(sys.argv)
```

vyper/venom/parser.py
```diff
@@ -0,0 +1,143 @@
...
+            instruction.output
+            for instruction in block.instructions
+            if instruction.output is not None
+        )
+        for output_var in output_vars:
+            assert isinstance(output_var, IRVariable)
+            value = output_var.value
+            assert value.startswith("%")
+            value = value.replace("%", "", 1)
+            if not value.isdigit():
```
charles-cooper, line 58: when can this happen? i think it would be good to have an explanatory comment here

vyper/venom/parser.py
```diff
@@ -0,0 +1,143 @@
...
+        for fn_name, blocks in funcs:
+            fn = ctx.create_function(fn_name)
+            for block_name, instructions in blocks:
+                bb = IRBasicBlock(IRLabel(block_name), fn)
+
+                for instruction in instructions:
+                    bb.insert_instruction(instruction)
+
+                # Manually insert because we need to override function entry
+                fn._basic_block_dict[block_name] = bb
```
charles-cooper, line 77: does `fn.append_basic_block(bb)` not work here?

vyper/venom/parser.py
```diff
@@ -0,0 +1,143 @@
...
+
+            set_last_var(fn)
```
charles-cooper, line 79: since this is more or less a local helper function, let's rename to `_set_last_var`

vyper/venom/parser.py
```diff
@@ -0,0 +1,143 @@
...
+
+        ctx.data_segment = data_section
+
+        ctx.chain_basic_blocks()
```
charles-cooper, line 83: does chain_basic_blocks() need to be called? or should we assume the user input is well-formed?

vyper/venom/parser.py
```diff
@@ -0,0 +1,143 @@
...
+VENOM_PARSER = Lark(
+    """
+    %import common.CNAME
+    %import common.DIGIT
+    %import common.LETTER
+    %import common.WS
+    %import common.INT
+
+    start: function* data_section
+    function: "fn" NAME "=>" "{" block* "}"
```
charles-cooper, line 23: let's have the keyword be `"function"`

vyper/venom/parser.py
```diff
@@ -0,0 +1,143 @@
...
+
+class VenomTransformer(Transformer):
+    def start(self, children) -> IRContext:
+        ctx = IRContext()
+        funcs = children[:-1]
+        data_section = children[-1]
+        for fn_name, blocks in funcs:
+            fn = ctx.create_function(fn_name)
+            for block_name, instructions in blocks:
+                bb = IRBasicBlock(IRLabel(block_name), fn)
```
charles-cooper, line 71: i think we might want to update `ctx.last_label` as well

vyper/venom/parser.py
```diff
@@ -0,0 +1,143 @@
...
+    def block(self, children) -> tuple[str, list[IRInstruction]]:
+        label, *instructions = children
+        return label, instructions
+
+    def assignment(self, children) -> IRInstruction:
+        to, value = children
+        if isinstance(value, IRInstruction):
+            value.output = to
+            return value
+        elif isinstance(value, IRLiteral):
```
charles-cooper, line 103:
```suggestion
        if isinstance(value, IRLiteral):
```

vyper/venom/parser.py
```diff
@@ -0,0 +1,143 @@
...
+            return IRInstruction("store", [value], output=to)
+        else:
+            raise TypeError(f"Unexpected value {value} of type {type(value)}")
```
charles-cooper, line 106:
```suggestion
            raise TypeError(f"Unexpected value {value} of type {type(value)}")
```

vyper/venom/README.md
```diff
@@ -29,59 +29,39 @@ Venom employs two scopes: global and function level.
...
-    %6 = calldatasize
-    %7 = lt %6, 164
-    %8 = or %5, %7
-    %9 = iszero %8
-    assert %9
-    stop
-
-fallback:
-    revert 0, 0
+fn global => {
```
charles-cooper, line 32: it's a little bit weird that functions are curly brace delimited but basic blocks are indent delimited

charles-cooper: (not a huge deal, but it's something to think about -- either making basic blocks curly brace delimited or functions whitespace delimited)

vyper/venom/parser.py
```diff
@@ -0,0 +1,143 @@
...
+        return children
+
+    def operand(self, children) -> IROperand:
+        return children[0]
+
+    def LABEL(self, label) -> IRLabel:
+        return IRLabel(label[1:])
+
+    def VAR_IDENT(self, var_ident) -> IRVariable:
+        parts = var_ident[1:].split(":", maxsplit=1)
```
charles-cooper, line 126: this is kind of slick, but it might cause issues if the signature of `IRVariable` ever changes. i'd do something like
```suggestion
        parts = var_ident[1:].split(":", maxsplit=1)
        assert 1 <= len(parts) <= 2
        varname = parts[0]
        version = None
        if len(parts) > 1:
            version = parts[1]
        return IRVariable(varname, version=version)
```
more verbose, but clearer :)

vyper/cli/vyper_venom.py
Philogy: Just following how it was done in the other CLIs, will do.

vyper/venom/parser.py
Philogy: I don't quite understand it but I think it's somewhat required to build the basic CFG and determine whether the CFGs terminate or not?

Philogy: Oh wait, I think I found the issue `"revert"` is not considered a basic block terminator (not in `vyper/venom/basicblock.py:BB_TERMINATORS`) so that was giving me errors.

vyper/venom/README.md
Philogy: tbh I like the mix, makes it clear which is which, if you really want can change

vyper/venom/parser.py
harkal: Yes, `revert` is not a terminator. You can have multiple reverts in a block and that is ok.

vyper/venom/README.md
harkal: I think I side with @charles-cooper on this. Seems weird to only have braces only for that part. Also is doesn't sit well with me as venom is more of a simple "assembly" like language, so no need for deep nesting.

harkal: Something like this:

```
fn global:
    %1 = calldataload 0
    %2 = shr 224, %1
    jmp @selector_bucket_0

selector_bucket_0:
    %3 = xor %2, 1579456981
    %4 = iszero %3
    jnz @1, @2, %4

1:
    jmp @fallback

2:
    %5 = callvalue
    %6 = calldatasize
    %7 = lt %6, 164
    %8 = or %5, %7
    %9 = iszero %8
    assert %9
    stop

fallback:
    revert 0, 0

data:
```

vyper/venom/parser.py
charles-cooper: i think this does the same thing as `append_basic_block()` though

vyper/venom/parser.py
Philogy: How is revert not a terminator?

vyper/venom/parser.py
charles-cooper: it's used in get_next_label

vyper/venom/parser.py
charles-cooper: right now we don't consider `revert` a terminator internally since it can be reordered within a basic block. this also gives us the property that there is a single terminator per basic block which some passes might rely on.

we could probably add it to the list of legal terminators, but it might require special rules in optimization passes; for now we just ensure blocks with reverts in them get terminated with `stop` (which gets stripped out during assembly anyway).

here if we wanted to validate user code, i think we could use a variant of chain_basic_blocks, which adds stop to basic blocks ending with `revert`, but otherwise does not add the stops or chaining jumps. (actually i just checked and we might be able to get rid of the chaining jumps in `chain_basic_blocks()` entirely, but we should revisit that later, it's out-of-scope for this PR).

vyper/venom/parser.py
```diff
@@ -0,0 +1,144 @@
...
+
+    start: function* data_section
+    function: "function" NAME "=>" "{" block* "}"
+    data_section: "data:" call*
+
+    block: NAME ":" (call | assignment)*
+
+    assignment: VAR_IDENT "=" expr
+    expr: call | CONST
+    call: CNAME operands_list
```
charles-cooper, line 30: maybe rename this to:
```
opcode: CNAME
instruction: opcode operands_list
```

charles-cooper:
i think it would be good to have a couple test cases! e.g. in `tests/functional/venom/parser/`

vyper/venom/README.md
Philogy: I don't really like this, makes it hard to differentiate between basic blocks, functions & data.

vyper/venom/parser.py
Philogy: I thought that wasn't getting used anywhere but I guess that's just because LSP is not recognizing that somewhere

vyper/venom/parser.py
Philogy: Ok for now I'll continue to call `chain_basic_blocks` then

Philogy: Actually no, to ensure minimal modification from user input I'll only add `stop` to blocks that *do not* terminate and contain a revert

Philogy: Thinking about it, it doesn't really make sense to consider `revert` non-terminal, and leads to weird edge cases like this one.

Yes you can rearrange it anywhere but that's more a side-effect of general instruction reordering, `revert` is not unique in being able to be put anywhere in a basic block.

vyper/venom/parser.py
```diff
@@ -0,0 +1,165 @@
...
+            value = output_var.value
+            assert value.startswith("%")
+            fn.last_variable = max(fn.last_variable, int(value[1:]))
+
+
+def _set_last_label(ctx: IRContext):
+    for fn in ctx.functions.values():
+        for bb in fn.get_basic_blocks():
+            label = bb.label.value
+            if label.isdigit():
```
charles-cooper, line 65: this doesn't seem right since labels are usually not just numbers. maybe just set `last_label` to the # of unique seen labels.

vyper/venom/parser.py
```diff
@@ -0,0 +1,165 @@
...
+class VenomTransformer(Transformer):
+    def start(self, children) -> IRContext:
+        ctx = IRContext()
+        funcs = children[:-1]
+        data_section = children[-1]
+        for fn_name, blocks in funcs:
+            fn = ctx.create_function(fn_name)
+            for block_name, instructions in blocks:
+                # Get default function block if entry bb
+                if block_name == fn_name:
```
charles-cooper, line 78: this is a bit of a weird restriction. why does the entry basic block need to be named the same thing as the function?

vyper/venom/parser.py
Philogy: I mean `last_label` is only used to generate numerical labels so only existing labels that are numerical should affect it, right?

vyper/venom/parser.py
Philogy: because this is the only basic block the `IRFunction` class creates by default

vyper/venom/parser.py
charles-cooper: i don't think so --
Screenshot from 2024-12-07 17-47-53

vyper/venom/parser.py
charles-cooper: i don't think we really need this, since the user might name the entry block something different than the function name. i think we should just clear the basic block dict before entering into this loop.

vyper/venom/parser.py
Philogy: Just setting to the number is kind of brittle though because you might not sequentially number variables for some reason.

Philogy: Ok did the following to handle potential collisions with labels that have a "suffix":

```python
            label_head, *_ = label.split("_", maxsplit=1)
            if label_head.isdigit():
                ctx.last_label = max(int(label_head), ctx.last_label)
```

vyper/venom/README.md
Philogy: is there consensus on this? I can do it that way, just seems kind of ugly. I know it works in Python because of indents but I don't want to have to deal with indents in the grammar. Down for whatever, just want this to work

charles-cooper: i don't think there need to be indents in the grammar, since functions are delimited with "function' (which is a reserved keyword)

Philogy: yes but I mean visually

Philogy:
@charles-cooper should be ready to merge now

charles-cooper (approved):
lgtm. a couple things for future scope:
- make data section optional in the grammar
- make entry block optional in the grammar (similarly to llvm)
- more output formats (`-f asm, opcodes, bytecode`)

PR #4384, opened 2024-12-01
Fix formatting issue in mathematical expression

lipperhey:
**Description**:  
In the last section of the documentation, there was a formatting issue with the mathematical expression involving subscript notation. Specifically, the expression:

```plaintext
(e.g. (0.2)\ :sub:`10` = (0.001100110011...)\ :sub:`2`, which needs to be truncated)
```

contained an unnecessary backslash before `:sub:`, which caused the subscripts to display incorrectly. This was fixed by properly formatting the subscripts as follows:

```plaintext
(e.g. (0.2)₁₀ = (0.001100110011...)₂, which needs to be truncated)
```

This correction is important because it ensures that the mathematical expression is displayed correctly and consistently across all rendering platforms, improving readability and clarity, especially for users who rely on accurate mathematical notation.

### Cute Animal Picture

Its me


charles-cooper:
i'm not sure this fixes anything.
before:
Screenshot from 2024-12-29 23-02-55
after:
Screenshot from 2024-12-29 23-02-31

PR #4385, opened 2024-12-03
fix[ux]: fix false positive for overflow in type checker

### LINKED ISSUE #4383: incorrectly reported overflow exception
### Version Information

* vyper Version (output of `vyper --version`): 0.4.0

### What's your issue about?
```vyper
@external
def foo() -> uint256:
    return (10**18)**2
```
panics at compilation with the following error, which is clearly wrong given that 10**36 clearly fits in a uint256:
```
vyper.exceptions.OverflowException: Base is too large, calculation will always overflow

  contract "contracts/token.vy:3", function "foo", line 3:12 
       2 def foo() -> uint256:
  ---> 3     return (10**18)**2
  -------------------^
```

tserg:
### What I did

Fix #4383 

### How I did it

Catch `OverflowException` when finding possible types for `BinOp` node.

### How to verify it

See test.

### Commit message

```
this commit fixes a false positive for integer overflow in
the typechecker involving nested pow operations by filtering
`OverflowException` in `_validate_op`. the previous code assumed that
`validate_numeric_op` could throw anything besides `InvalidOperation`,
but for the `Pow` binop, it can throw `OverflowException`.
```

### Cute Animal Picture

### DIFF
--- a/tests/functional/codegen/types/numbers/test_exponents.py
+++ b/tests/functional/codegen/types/numbers/test_exponents.py
@@ -173,3 +173,18 @@ def foo(b: int128) -> int128:
     c.foo(max_power)
     with tx_failed():
         c.foo(max_power + 1)
+
+
+
+valid_list = [
+    """
+@external
+def foo() -> uint256:
+    return (10**18)**2
+    """
+]
+
+
+@pytest.mark.parametrize("good_code", valid_list)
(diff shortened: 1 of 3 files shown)


charles-cooper (approved):
lgtm, thanks!

cyberthirst (approved):

PR #4415, opened 2024-12-20
chore[docs]: abi function signature for default arguments

pcaversaccio:
### What I did

Resolves #37634 of [Attackathon | Ethereum Protocol](https://immunefi.com/bounty/ethereum-protocol-attackathon). Add a note on how the function signature is calculated if default arguments are used.

### How I did it

Brain.

### How to verify it

Check function signatures.

### Commit message

```
this commit adds a note on how the function selector is calculated if
default arguments are used.
```

Add a note on how the function signature is calculated if default arguments are used.

### Cute Animal Picture

image

### DIFF
--- a/docs/control-structures.rst
+++ b/docs/control-structures.rst
@@ -48,7 +48,16 @@ External functions (marked with the ``@external`` decorator) are a part of the c
 A Vyper contract cannot call directly between two external functions. If you must do this, you can use an :ref:`interface <interfaces>`.
 
 .. note::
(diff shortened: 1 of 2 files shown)


cyberthirst:
Looks good. Can we please add a note on what it means that an interface function contains a default parameter? ie that the semantics are that we're interfacing with a contract that supports the default arguments - it's dependant on the callee to implement it

pcaversaccio:
> Looks good. Can we please add a note on what it means that an interface function contains a default parameter? ie that the semantics are that we're interfacing with a contract that supports the default arguments - it's dependant on the callee to implement it

https://github.com/vyperlang/vyper/pull/4415/commits/b88786281818c22e15399d32b8f111755f4924aa

cyberthirst (approved):

charles-cooper (approved):

PR #4423, opened 2024-12-24
fix[venom]: fix `MakeSSA` with existing phis

harkal:
### What I did

Fixed the memssa pass to support already existing `phi` instructions.

### Commit message

```
This commit improves the `MakeSSA` pass to handle incoming `phi`
instructions. Following the implementation of the venom parser it
is possible to parse code with existing `phi` instructions in the
code. The old code was expecting the `phi` instructions to have
been self-placed and in "degenerate form" of output == all branch
arguments. The new code does not make this assumption.

This commit additionally:
- refactors the existing `make_ssa` tests to use new test machinery
- adds a phi parsing test (does not test the new code in this commit
  since it does not run passes, but does make sure we can at least parse
  phis)
- expands the venom round-trip tests to check that we can both
    a) run venom passes on parsed venom, and
    b) bytecode generation from round-tripping venom produced by the
       vyper frontend is equivalent to bytecode generation from the
       regular pipeline (directly from vyper source code)
```

### Cute Animal Picture

### DIFF
--- a/tests/functional/venom/parser/test_parsing.py
+++ b/tests/functional/venom/parser/test_parsing.py
@@ -1,4 +1,4 @@
-from tests.venom_utils import assert_ctx_eq
+from tests.venom_utils import assert_bb_eq, assert_ctx_eq
 from vyper.venom.basicblock import IRBasicBlock, IRLabel, IRLiteral, IRVariable
 from vyper.venom.context import DataItem, DataSection, IRContext
 from vyper.venom.function import IRFunction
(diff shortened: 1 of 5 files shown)


charles-cooper:
i added more tests (https://github.com/vyperlang/vyper/pull/4423/commits/1728bee6050fe10a501e33c096e91294d012ab4d runs venom passes in the round-trip tests) and there still seems to be some bugs, e.g. https://github.com/vyperlang/vyper/actions/runs/12483208026/job/34838622968?pr=4423 throws things like
Screenshot from 2024-12-24 10-43-14

at first glance, it looks like the fix fails if there are multiple phis in multiple basic blocks referencing the same variable name, but not sure

charles-cooper (approved):

PR #4432, opened 2024-12-31
fix[ux]: don't print full path as interface name in output

sandbubbles:
### What I did
Instead of the full path as the interface print just the name of the interface as per https://github.com/vyperlang/vyper/issues/4299. 
### How I did it
Use only a part of _id as the interface's string representation and use it when printing signature.

### Commit message

### Cute Animal Picture


Sporarum:
I think instead we should try to always use the aliased named when it is available

As-is it ignores the alias name, and can introduce confusion if multiple files have the same stem

PR #4452, opened 2025-01-17
feat[ci]: use `coverage combine` to reduce codecov uploads

tserg:
### What I did

Consolidate code coverage files and upload it once.

### Commit message

```
This commit updates the CI to combine code coverage to a single file
and perform just a single upload after all tests finish (rather than
each test run uploading its own coverage report). This should reduce
failures / rate limiting on the codecov app, and also prevent the
codecov app from producing an inaccurate coverage report before all
the tests finish.

references:
- https://coverage.readthedocs.io/en/7.6.10/cmd.html#cmd-combine
- https://coverage.readthedocs.io/en/7.6.10/cmd.html#re-mapping-paths
- https://coverage.readthedocs.io/en/7.6.10/config.html#config-run-relative-files
```

### Cute Animal Picture

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -256,17 +256,6 @@ jobs:
         pattern: coverage-files-*
         path: coverage-files
 
-    - name: Create coveragerc
-      run: |
-        echo > .coveragerc <<EOF
-[paths]
-source =
-    /home/runner/work/vyper/vyper
-    /Users/runner/work/vyper/vyper
-    D:\a\vyper\vyper
-        EOF
-
-
     - name: Debug directory
       run: |
         ls -la .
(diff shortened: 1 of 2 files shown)


tserg:
I haven't been able to figure this out. In order to do `coverage combine`, we need the raw `.coverage` data file to be unique per run. This would require the `parallel` flag in `coverage` to be set to be true. Since the `pytest-cov` plugin overrides the `parallel` flag, we need to rely on the `coverage` package directly. However, there is still an issue of combining [duplicate data](https://github.com/vyperlang/vyper/pull/4452/checks?check_run_id=35831750879) - I tried doing `combine` for the fuzzing run only but it gives the same error.

charles-cooper (approved):

PR #4476, opened 2025-02-14
chore[docs]: amend code comment in `StackTooDeep`

pcaversaccio:
### What I did

We should never forget that stack-too-deep should _never_ happen 😄!

### How I did it

✍🏾

### How to verify it

🧠

### Commit message

```
amend code comment in `StackTooDeep` exception
class definition
```

chore[docs]: amend code comment in `StackTooDeep`


charles-cooper:
I think this would be more helpful as a documentation change

pcaversaccio:
image

PR #4482, opened 2025-02-21
feat[venom]: implement new calling convention

harkal:
### Commit message

```
this commit updates the IRnode->venom translator to use
stack arguments when the arguments are word-size and under
a certain threshold. to ensure that all the contracts in
`tests/functional/examples/thirdparty/` compile, the threshold
is set to 6. we could relax this later if we have interprocedural
optimizations which can migrate the stack arguments back to memory.

it also updates the inliner and mem2var to handle cases from the new
calling convention.
```

### Cute Animal Picture


harkal:
lgtm

charles-cooper (approved):

PR #4514, opened 2025-03-12
feat[venom]: allow labels to be assigned to variables

harkal:
### Commit message

```
this commit allows labels to be assigned to variables. it updates
assumptions in `SCCP` and the venom parser which blocked that behavior.

it also updates the text format for `djmp` to be like `jnz` (condition
is first). it also fixes a bug in SimplifyCFG where `merge_jump` did
not rename label operands.
```

### Cute Animal Picture


charles-cooper (approved):
lgtm, needs a test

PR #4517, opened 2025-03-13
feat[codegen]: revert_on_failure for extcall/staticcall

z80dev:
## Summary
- enable `revert_on_failure` for typed external calls (`extcall`/`staticcall`), aligning behavior with `raw_call`
- return typing becomes `bool` (no-return) or `(bool, T)` / `(bool, (T1, T2, ...))` when `revert_on_failure=False`
- normalize the call success flag, zero the return buffer on failure, and decode return data only on success
- disable the external return fast-path for risky calls to keep ABI-safe return encoding
- document usage and edge cases in `docs/interfaces.rst`; expand tests including multi-return revert-on-failure

## Behavior and Semantics
- `revert_on_failure=False` returns a success flag instead of reverting when the EVM call itself fails.
- ABI decoding still enforces return-data size/type checks; those can still revert.
- `EXTCODESIZE` checks for no-return calls still apply unless `skip_contract_check=True`.
- On failure, the returned `value` is the zero/default value for its type; for tuples, each member is zeroed.

## Implementation Details
- **Typing:** call-site return type is adjusted in `vyper/semantics/types/function.py` to return `bool` or `(bool, T)` when `revert_on_failure=False`.
- **Codegen:** external call generation in `vyper/codegen/external_call.py`
  - normalizes the call result to a strict boolean (`0/1`)
  - zeroes the return payload on failure to avoid stale memory or revert data leaks
  - only decodes returndata on success
- **Return fast-path:** `vyper/codegen/return_.py` skips the "return without re-encoding" path for risky calls so direct `return extcall/staticcall ...` is ABI-safe.
- **Docs:** `docs/interfaces.rst` adds a `revert_on_failure` usage section with return typing, edge cases, and examples.

## Tests
- `tests/functional/codegen/calling_convention/test_external_contract_calls.py`
  - no-return extcall: `revert_on_failure=False` returns `bool`
  - direct-return path for no-return and value-returning calls
  - staticcall variant
  - struct/array/string returns (existing coverage)
  - multi-return tuple: returns `(bool, (uint256, bytes32))` and zero defaults on failure

## How to Verify
- `uv run pytest tests/functional/codegen/calling_convention/test_external_contract_calls.py -q`
- `uv run ./quicktest.sh -m "not fuzzing"`


z80dev:
## Bug Verification

I've created a branch that demonstrates the bug (has the `revert_on_failure` feature but **not** the fix): [`revert-on-failure-bug-demo`](https://github.com/z80dev/vyper/tree/revert-on-failure-bug-demo)

### To reproduce the issue:

```bash
git fetch origin pull/4517/head:pr-4517
git checkout z80dev/revert-on-failure-bug-demo  # or: git checkout 6b7aba45
```

Then run this test:

```python
# tests/functional/codegen/calling_convention/test_skip_encode_risky_call_bug.py

def test_staticcall_revert_on_failure_direct_return_zeroes_on_fail(get_contract):
    """
    When a staticcall with revert_on_failure=False fails, the result
    portion of the returned tuple should be zero, not garbage.
    """
    target_source = """
@external
@view
def get_value(should_raise: bool) -> uint256:
    if should_raise:
        raise "fail"
    return 123
    """

    caller_source = """
interface Target:
    def get_value(should_raise: bool) -> uint256: view

@external
def call_and_return(target: address, should_raise: bool) -> (bool, uint256):
    return staticcall Target(target).get_value(should_raise, revert_on_failure=False)
    """

    target = get_contract(target_source)
    caller = get_contract(caller_source)

    # Successful call works fine
    success, result = caller.call_and_return(target.address, False)
    assert success is True
    assert result == 123

    # Failed call triggers the bug - causes OutOfGas(MemoryLimit)
    success, result = caller.call_and_return(target.address, True)
    assert success is False
    assert result == 0
```

### Results:

| Version | Result |
|---------|--------|
| Without fix (6b7aba45) | ❌ `OutOfGas(MemoryLimit)` |
| With fix (this PR) | ✅ Pass |

The bug occurs because the return fast-path optimization skips ABI encoding for values that appear to already be properly encoded in memory, but doesn't account for returndata from failed external calls containing garbage that causes invalid memory expansion.

docs/interfaces.rst
```diff
@@ -85,6 +86,77 @@ The ``default_return_value`` parameter can be used to handle ERC20 tokens affect
...
+
+    interface Target:
+        def values(should_raise: bool) -> (uint256, bytes32): nonpayable
+
+    @external
+    def call_target_values(
+        target: address,
+        should_raise: bool
+    ) -> (bool, (uint256, bytes32)):
+        return extcall Target(target).values(should_raise, revert_on_failure=False)
```
Sporarum, line 147: I think this example would be better in the form of something like:
```
success, (int_v, bytes_v) = extcall Target(target).values(should_raise, revert_on_failure=False)
```

docs/interfaces.rst
```diff
@@ -85,6 +86,77 @@ The ``default_return_value`` parameter can be used to handle ERC20 tokens affect
...
+Staticcall with revert_on_failure:
+
+.. code-block:: vyper
+
+    interface Target:
+        def value(should_raise: bool) -> uint256: view
+
+    @external
+    def call_target_value(target: address, should_raise: bool) -> (bool, uint256):
+        return staticcall Target(target).value(should_raise, revert_on_failure=False)
```
Sporarum, line 158: These examples are nice, but a couple comments:

It would be better if the examples were more grounded. By that I mean mimicking real-life use-cases, for example token.transfer(...)

tests/functional/codegen/calling_convention/test_external_contract_calls.py
```diff
@@ -1252,6 +1252,405 @@ def get_lucky(contract_address: address) -> int128:
     assert c2.get_lucky(c1.address) == 1
 
 
+def test_external_contract_call_revert_on_failure_noreturn(get_contract, tx_failed):
```
Sporarum, line 1255: I think these tests should be refactored to be a single parametrized test, since there is so much sharing between them

uv.lock
Sporarum, line 1: This should not be commited !
It's in our .gitignore

Sporarum (commented):
Looks good, thank you !

A couple remaining things however:

Sporarum:
> ABI decoding still enforces return-data size/type checks; those can still revert.

Is this something we want ?
IIRC that's what solidity does, but is considered sort of a footgun

charles-cooper:
> > ABI decoding still enforces return-data size/type checks; those can still revert.
> 
> Is this something we want ? IIRC that's what solidity does, but is considered sort of a footgun

Why foot gun ? We are talking about callee output validation right?

Sporarum:
> > > ABI decoding still enforces return-data size/type checks; those can still revert.
> > 
> > 
> > Is this something we want ? IIRC that's what solidity does, but is considered sort of a footgun
> 
> Why foot gun ? We are talking about callee output validation right?

Because we have something which converts 99% of reverts into a boolean instead, leading the user to assume this can never revert, but then actually there is 1% that can still revert

It would be better (if possible) to have those checks also contribute to the boolean, instead of reverting

charles-cooper:
> > > > ABI decoding still enforces return-data size/type checks; those can still revert.
> > > 
> > > 
> > > Is this something we want ? IIRC that's what solidity does, but is considered sort of a footgun
> > 
> > 
> > Why foot gun ? We are talking about callee output validation right?
> 
> Because we have something which converts 99% of reverts into a boolean instead, leading the user to assume this can never revert, but then actually there is 1% that can still revert
> 
> It would be better (if possible) to have those checks also contribute to the boolean, instead of reverting

uh no we should absolutely do output validation, that should not even be a question!

Sporarum:
> > It would be better (if possible) to have those checks also contribute to the boolean, instead of reverting
> 
> uh no we should absolutely do output validation, that should not even be a question!

I think you misunderstood what I meant, we should keep the validation of course, but instead of reverting, we should set the boolean to false
(and the fields should be set as they are when the call reverts, I'm assuming to `empty`)

charles-cooper:
> > > It would be better (if possible) to have those checks also contribute to the boolean, instead of reverting
> > 
> > 
> > uh no we should absolutely do output validation, that should not even be a question!
> 
> I think you misunderstood what I meant, we should keep the validation of course, but instead of reverting, we should set the boolean to false (and the fields should be set as they are when the call reverts, I'm assuming to `empty`)

i did not misunderstand what you meant

PR #4658, opened 2025-05-24
perf[venom]: optimize time spent in `SimplifyCFG`

harkal:
### What I did

Optimised `SimplifyCFG`

### Commit message

```
this commit optimizes the performance of the `SimplifyCFG`
pass by scheduling label renaming for batch processing after
`_collapse_chained_blocks()`.

this performance regression was introduced in 2360d629d1607e09d
and 18d0787a737e240d36e21, since `_replace_label()` was added to
`_merge_jump()`, and it is called in a loop.
```

### Cute Animal Picture


charles-cooper (approved):
lgtm, just add a sanity check in _replace_label

PR #4759, opened 2025-10-23
feat[lang]: improve error message for namespace collision

### LINKED ISSUE #4579: Improve error message for `math` namespace collision
The following snippet will throw with

```vyper
from snekmate.utils import math
import math
```

```bash
vyper.exceptions.NamespaceCollision: 'math' has already been declared as a ModuleInfo(module_t=src/snekmate/utils/math.vy, alias='math', ownership=<ModuleOwnership.NO_OWNERSHIP: 'no_ownership'>, ownership_decl=None)

  contract "math_mock.vy:2", line 2:0
       1 from .. import math
  ---> 2 import math
  --------^
       3
```

stoyanov-kaloyan:
### Added improved error message for namespace collisions in imports

resolves #4579

New error message looks like:
```
Error compiling: _test.vy
vyper.exceptions.NamespaceCollision: 'math' has already been declared via import (alias=`math`).

 Suggested fix: rename or remove the previous declaration (e.g. use `import math as ...` to avoid collisions)

  contract "_test.vy:2", line 2:0 
       1 from snekmate.utils import math
  ---> 2 import math
  -------^
```

### Commit message:

feat[lang]: improve error message for namespace collision

Added a block in which it attempts to provide an easier to read hint for the user.

### Cute Animal Picture
car

### DIFF
--- a/vyper/semantics/namespace.py
+++ b/vyper/semantics/namespace.py
@@ -93,7 +93,18 @@ class Namespace(dict):
             prev_decl = getattr(prev, "decl_node", None)
             msg = f"'{attr}' has already been declared"
             if prev_decl is None:
-                msg += f" as a {prev}"
+                # attempt to print a hint for namespace collision
+                alias = getattr(prev, "alias", None)
+
+                if alias:
+                    hint = f" via import (alias=`{alias}`)"
+                else:
+                    short_name = getattr(prev, "name", None)
+                    type_name = type(prev).__name__
+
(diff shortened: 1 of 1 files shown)


Sporarum:
> vyper.exceptions.NamespaceCollision: 'math' has already been declared via import (alias=`math`).

That alias part will always be the same as what is at the start, right ?
Since otherwise there would not be a collision

PR #4765, opened 2025-11-03
feat[venom]: stack spilling to memory for deep stacks

harkal:
### Commit message

```
implement stack spilling to handle stacks deeper than 16 elements (the
EVM's `SWAP16`/`DUP16` limit). instead of failing with `StackTooDeep`,
the compiler now temporarily stores values in memory.

adds a `StackSpiller` class that:
- spills operands to memory when stack depth exceeds 16
- restores spilled operands when needed
- reuses freed spill slots to minimize memory usage
- coordinates with memory allocator via per-function end-of-memory

the spiller integrates into stack reordering with a two-tier approach:
1. selective spilling: remove non-essential items from top of stack to
   bring the target operand within `SWAP16` range
2. bulk spill/restore: for extreme depths, spill an entire stack segment
   and restore in the desired order

the `spilled` dict is copied at control flow splits so each branch
tracks its own spilled state independently.

removes the `StackTooDeep` exception since the compiler now always
generates working code, trading some runtime efficiency (extra
`MSTORE`/`MLOAD` ops) for correctness when needed.
```

### Cute Animal Picture

### DIFF
--- a/vyper/venom/stack_spiller.py
+++ b/vyper/venom/stack_spiller.py
@@ -109,10 +109,8 @@ class StackSpiller:
         spill_ops, offsets, cost = self._spill_stack_segment(assembly, stack, chunk_size, dry_run)
 
         indices = list(range(chunk_size))
-        if chunk_size == 1:
-            desired_indices = indices
-        else:
-            desired_indices = [indices[-1]] + indices[1:-1] + [indices[0]]
+        assert len(indices) > 1  # due to chunk_size > 1
+        desired_indices = [indices[-1]] + indices[1:-1] + [indices[0]]
 
         cost += self._restore_spilled_segment(
             assembly, stack, spill_ops, offsets, desired_indices, dry_run


vyper/venom/stack_spiller.py
```diff
@@ -0,0 +1,242 @@
...
+        Dup operation that handles deep stacks via spilling.
+
+        For stacks deeper than 16, spills the stack segment to memory,
+        then restores it with duplication.
+        """
+        dup_idx = 1 - depth
+        if dup_idx < 1:
+            from vyper.exceptions import StackTooDeep
+
+            raise StackTooDeep(f"Unsupported dup depth {dup_idx}")
```
charles-cooper: this is like a DUP0, right? seems like the wrong exception in that case

vyper/venom/venom_to_assembly.py
```diff
@@ -196,11 +202,18 @@ def generate_evm_assembly(self, no_optimize: bool = False) -> list[AssemblyInstr
...
+        stack: StackModel,
+        stack_ops: list[IROperand],
+        spilled: dict[IROperand, int],
+        dry_run: bool = False,
     ) -> int:
         if dry_run:
             assert len(assembly) == 0, "Dry run should not work on assembly"
             stack = stack.copy()
+            spilled = spilled.copy()
+            spill_free_snapshot = self.spiller._spill_free_slots.copy()
```
charles-cooper: weird abstraction leakage, maybe something like `self.spiller.snapshot()`?

PR #4766, opened 2025-11-03
feat[docs]: update verbiage around reference types

Sporarum:
### Commit message

```
previous "value types" section hinted at reference types being passed by
reference which is incorrect. taking this fact into account it seems the
main reason to split types into value types and reference-types
disappears.
```

### Cute Animal Picture

### DIFF
--- a/docs/types.rst
+++ b/docs/types.rst
@@ -9,11 +9,13 @@ Vyper is a statically typed language. The type of each variable (state and local
 
 In addition, types can interact with each other in expressions containing operators.
 
-Unlike in some other languages, there are no sub-categories of types. Values are always copied when assigned to a variable, or when passed to a function (also known as call-by-value). That means that, a calling function never needs to worry about a callee modifying the data of a passed structure.
+Unlike in some other languages, there are no sub-categories of types. Values are always copied, both when assigned to a variable and when passed to a function (also known as call-by-value). This means a calling function never needs to worry about a callee modifying the data of a passed structure.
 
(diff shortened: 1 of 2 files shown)


docs/types.rst
```diff
@@ -9,14 +9,11 @@ Vyper is a statically typed language. The type of each variable (state and local
...
+Unlike in some other languages, there are no sub-categories of types. Values are always copied when assigned to a variable, or when passed to a function (also known as call-by-value). That means that, a calling function never needs to worry about a callee modifying the data of a passed structure.
 
-Value Types
-===========
+.. note::
 
-The following types are also called value types because variables of these
-types will always be passed by value, i.e. they are always copied when they
-are used as function arguments or in assignments.
+    However neither values nor variables are immutable. Variables can be reassigned to values of the same type. And values of some types (for example arrays and structs) have operations that modify them in place, usually by assigning to their members directly: `my_array[0] = 42`.
```
pcaversaccio:
```suggestion
    However neither values nor variables are immutable. Variables can be reassigned to values of the same type. And values of some types (for example arrays and structs) have operations that modify them in place, usually by assigning to their members directly: ``my_array[0] = 42``.
```

pcaversaccio: `external` function parameters are stored as calldata which is immutable (`internal` function parameters are stored in memory tho), maybe we should add this clarification here; what do you think?

charles-cooper (approved):

PR #4777, opened 2025-11-19
feat[venom]: optimize `popmany()` for contiguous suffix removal

harkal:
### What I did

Optimize the pop continues suffix out of the stack case for `popmany()`

### How to verify it

`tests/unit/compiler/venom/test_venom_to_assembly.py`

### Commit message

```
update venom's `popmany()` helper to detect a common case where the
variables being discarded form a contiguous sequence in the stack and
emit a single swapN/pop sequence instead of sequential swap/pop pairs
for each operand.
```

### Cute Animal Picture

### DIFF
--- a/vyper/venom/venom_to_assembly.py
+++ b/vyper/venom/venom_to_assembly.py
@@ -309,7 +309,7 @@ class VenomCompiler:
             return
 
         # if the items to pop are contiguous, we can swap the top of
-        # stack to the middle of the stack and then just issue
+        # stack to just below the lowest item-to-pop and then just issue
         # sequential pops
         depths = [stack.get_depth(var) for var in to_pop]
         deepest = min(depths)


vyper/venom/venom_to_assembly.py
```diff
@@ -305,6 +305,17 @@ def _prepare_stack_for_function(self, asm, fn: IRFunction, stack: StackModel):
 
     def popmany(self, asm, to_pop: Iterable[IRVariable], stack):
         to_pop = list(to_pop)
+        if len(to_pop) == 0:
+            return
+
+        depths = [stack.get_depth(var) for var in to_pop]
```
charles-cooper, line 314: i think a comment here would be good, since it's not obvious at a glance what it's doing

charles-cooper:
```suggestion
        # if the items to pop are contiguous, we can swap the top of
        # stack to the middle of the stack and then just issue
        # sequential pops
        depths = [stack.get_depth(var) for var in to_pop]
```

tests/unit/compiler/venom/test_venom_to_assembly.py
```diff
@@ -34,3 +37,96 @@ def test_optimistic_swap_params():
...
+    stack.push(drop2)
+    stack.push(keep_top)
+
+    asm: list[str] = []
+    compiler.popmany(asm, [drop3, drop2], stack)
+
+    assert asm == ["SWAP1", "POP", "SWAP2", "POP"]
+    assert len(stack._stack) == 2
+    assert keep_mid in stack._stack
+    assert keep_top in stack._stack
```
charles-cooper, line 97: maybe assert the actual state of the stack here

charles-cooper: ah i see, we don't really care about the order here

Issue #4783, opened 2025-11-29
Potential regression in latest Venom


pcaversaccio:
The follow commit https://github.com/vyperlang/vyper/commit/79ca8cb82815e2dbba756879c7c9272e7696ed3f (PR https://github.com/vyperlang/vyper/pull/4756) triggers 🐍 snekmate's _nightly_ Venom-based smart contract tests to fail.

See: https://github.com/pcaversaccio/snekmate/actions/runs/19778596013/job/56675457299#step:12:65998

```console
Failing tests:
Encountered 3 failing tests in test/utils/ECDSA.t.sol:ECDSATest
[FAIL: next call did not revert as expected] testRecoverWith0x00Value() (gas: 14339)
[FAIL: next call did not revert as expected] testRecoverWithInvalidSignature() (gas: 14074)
[FAIL: next call did not revert as expected] testRecoverWithWrongVersion() (gas: 14345)

Encountered 1 failing test in test/utils/SignatureChecker.t.sol:SignatureCheckerTest
[FAIL: next call did not revert as expected] testEOAWithInvalidSignature2() (gas: 16878)

Encountered a total of 4 failing tests, 811 tests succeeded
```

Test contracts:
- https://github.com/pcaversaccio/snekmate/blob/ea998b4ec46d75cdc19a6557e2d2aa23aff5a3a3/test/utils/ECDSA.t.sol
- https://github.com/pcaversaccio/snekmate/blob/ea998b4ec46d75cdc19a6557e2d2aa23aff5a3a3/test/utils/SignatureChecker.t.sol

To reproduce:

```console
pip install git+https://github.com/vyperlang/vyper.git@79ca8cb82815e2dbba756879c7c9272e7696ed3f
```

run

```console
FOUNDRY_PROFILE=ci-venom forge test --mc ECDSATest
```

and

```console
FOUNDRY_PROFILE=ci-venom forge test --mc SignatureCheckerTest
```

PR #4785, opened 2025-12-01
fix[venom]: fix clobber logic in dse

charles-cooper:
### What I did
fix https://github.com/vyperlang/vyper/issues/4783

### Commit message

```
this commit fixes overly aggressive (incorrect) clobbers in DSE.

the root issue is that the memory semantics of calls were modeled
incorrectly -- the `size` argument was being passed as the size to
`MemoryLocation`, but in the EVM, the size is just the maximum number of
bytes that is allowed to be written at that location, the callee
(invoked by `*CALL`) can write fewer bytes at runtime. the import to us
is that this results in the previous write to that memory buffer not
necessarily being clobbered.

this fixes the issue by modeling the size written as `None`, so that
passes don't analyze the buffer as being clobbered. this results
conversely in overly conservative passes now since the size is
technically modeled as unlimited (but we have a bound from the EVM
semantics); this will be fixed at a later date.
```

### Cute Animal Picture

### DIFF
--- a/tests/functional/syntax/test_hashmap_container_type.py
+++ b/tests/functional/syntax/test_hashmap_container_type.py
@@ -1,0 +1,32 @@
+import pytest
+
+from vyper.compiler import compile_code
+
+valid_list = [
+    """
+d: HashMap[uint256, HashMap[uint256, uint256]]
+
+@external
+def foo() -> uint256:
+    return self.d[0][0]
+    """,
+    """
+d: HashMap[uint256, uint256[10]]
+
+@external
+def foo() -> uint256:
+    return self.d[0][0]
+    """,
+    """
+d: HashMap[uint256, (uint256, uint256)]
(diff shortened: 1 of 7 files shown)


pcaversaccio (commented):
Let's add some tests here pls

pcaversaccio:
> > Let's add some tests here pls
> 
> there are tests in the PR!

Not Vyper-based ones.

charles-cooper:
> > > Let's add some tests here pls
> > 
> > 
> > there are tests in the PR!
> 
> Not Vyper-based ones.

we already have ecrecover tests, but since the venom memory allocator is getting more advanced, it's hard to force it into the specific allocation strategies which result in the bug. the path forward in terms of testing here is we will need more frontend fuzzing via program generation, and more memory fuzzing on the venom side.

PR #4789, opened 2025-12-02
fix[venom]: fix `fix_mem_loc`

charles-cooper:
### Commit message

```
fix the `fix_mem_loc()` util.

this case is wrong in the case that the op is an `IRAbstractMemLoc`,
because `.value` is just its id, not any meaning related to its pointer
offset.

previously this case (i.e., where the item is an `IRAbstractMemLoc`)
didn't get triggered because at that stage of compilation, all
`IRAbstractMemLoc`s are only contained in `*alloca` instructions. this
fixes the case for follow on PRs.

it also fixes a case in `may_overlap()` which incorrectly returns false
when comparing concrete memory with abstract memory. this is currently
unreachable in the codebase during alias analysis since we have an
invariant that we do not simultaneously have concrete and abstract
memory locations in generated venom code, but is not true for more
general venom code.
```

### Cute Animal Picture

### DIFF
--- a/vyper/venom/memory_location.py
+++ b/vyper/venom/memory_location.py
@@ -73,8 +73,7 @@ class MemoryLocation:
         if loc1 is MemoryLocation.UNDEFINED or loc2 is MemoryLocation.UNDEFINED:
             return True
         if type(loc1) is not type(loc2):
-            # CMC 2025-12-02 -- this should return True, no?
-            return False
+            return True
         if isinstance(loc1, MemoryLocationSegment):
             assert isinstance(loc2, MemoryLocationSegment)
             return MemoryLocationSegment.may_overlap_concrete(loc1, loc2)


vyper/venom/memory_location.py
```diff
@@ -73,6 +73,7 @@ def may_overlap(loc1: MemoryLocation, loc2: MemoryLocation) -> bool:
         if loc1 is MemoryLocation.UNDEFINED or loc2 is MemoryLocation.UNDEFINED:
             return True
         if type(loc1) is not type(loc2):
+            # CMC 2025-12-02 -- this should return True, no?
```
HodanPlodky: the allocator should never allocate to memory that would be used as literal

charles-cooper: i think this branch can be hit if we have concrete and abstract memory locations existing in the same pass

vyper/venom/memory_location.py
```diff
@@ -393,7 +394,7 @@ def fix_mem_loc(function: IRFunction):
             read_op = get_memory_read_op(inst)
             if write_op is not None:
                 size = get_write_size(inst)
-                if size is None or not isinstance(write_op.value, int):
+                if size is None or not isinstance(write_op, IRLiteral):
```
HodanPlodky, line 396: oh yeah thats better

harkal (approved):

PR #4790, opened 2025-12-04
feat[venom]: add assert elimination pass

harkal:
### What I did

Add flow-sensitive range analysis and assert elimination pass.

### How I did it

Implemented `VariableRangeAnalysis` that tracks integer bounds for variables at each program point, with evaluators for arithmetic, bitwise, comparison, and other EVM ops. Added `AssertEliminationPass` as a first consumer of the analysis.

### How to verify it

Run the test suite, particularly:
- `tests/unit/venom/test_variable_range_analysis.py`
- `tests/unit/venom/test_variable_range_properties.py`
- `tests/unit/venom/test_assert_elimination.py`

### Commit message

```
feat[venom]: add assert elimination pass

Add flow-sensitive range analysis to track integer bounds for variables
at each program point. The analysis enables optimizations that depend on
knowing value constraints (e.g., eliminating redundant asserts, bounds
checks).

- Add `VariableRangeAnalysis` with per-instruction range environments
- Implement evaluators for arithmetic, bitwise, comparison, and other ops
- Handle signed/unsigned EVM semantics (normalize literals >= 2^255)
- Refine ranges at branches based on comparison/iszero conditions
- Apply widening after threshold iterations for loop convergence

As a first consumer, add `AssertEliminationPass` to remove asserts when
the condition is provably nonzero.
```

Add variable range analysis for venom IR, enabling range-based optimizations.

### Cute Animal Picture


vyper/venom/analysis/variable_range/analysis.py
```diff
@@ -0,0 +1,429 @@
...
+        if old_range.is_top or new_range.is_top:
+            return ValueRange.top()
+        if old_range.is_empty:
+            return new_range
+        if new_range.is_empty:
+            return old_range
+
+        # If the range is growing, widen to top
+        if new_range.lo < old_range.lo or new_range.hi > old_range.hi:
+            return ValueRange.top()
```
HodanPlodky, line 442: Just putting a note here that there should be a possibility to be more precise in the widening. You would need to have the set of some values that would be considered for possible ends for a range after widening (and for that set to be finite) to be able to get fixed point. But that should be outside of the scope of this PR

vyper/venom/analysis/variable_range/value_range.py
```diff
@@ -0,0 +1,162 @@
...
+        lo = max(self.bounds[0], other.bounds[0])
+        hi = min(self.bounds[1], other.bounds[1])
+        return ValueRange((lo, hi)) if lo <= hi else ValueRange.empty()
+
+    def clamp(self, lo: Optional[int] = None, hi: Optional[int] = None) -> ValueRange:
+        """Clamp this range to the specified bounds."""
+        if self.is_top:
+            lo = SIGNED_MIN if lo is None else max(SIGNED_MIN, lo)
+            hi = UNSIGNED_MAX if hi is None else min(UNSIGNED_MAX, hi)
+            return ValueRange((lo, hi)) if lo <= hi else ValueRange.empty()
```
HodanPlodky: is this ternary necessary shouldn't it be handled by `__post_init__`

vyper/venom/analysis/variable_range/value_range.py
```diff
@@ -0,0 +1,162 @@
...
+        if self.is_empty:
+            return self
+        assert self.bounds is not None
+        new_lo = self.bounds[0] if lo is None else max(self.bounds[0], lo)
+        new_hi = self.bounds[1] if hi is None else min(self.bounds[1], hi)
+        return ValueRange((new_lo, new_hi)) if new_lo <= new_hi else ValueRange.empty()
```
HodanPlodky: same as above

vyper/venom/analysis/variable_range/evaluators.py
```diff
@@ -0,0 +1,613 @@
...
+    """
+    index_op = inst.operands[-1]
+    value_op = inst.operands[-2]
+    value_range = _operand_range(value_op, state)
+    if value_range.is_empty:
+        return ValueRange.empty()
+    index = _get_uint_literal(index_op)
+    if index is not None and index >= 32:
+        return ValueRange.constant(0)
+    return ValueRange.bytes_range()
```
HodanPlodky, line 239: this is kinda imprecise for case like
```
main:
    %x = calldataload 0
    %cond = lt %x, 10
    jnz %cond, @then, @else
then:
    %b = byte 0, %x ; here the result could be only in range [0, 10] but this would set it to [0, 255]
    sink %b
else:
    sink %x
```

vyper/venom/analysis/variable_range/value_range.py
harkal: right! I added the normalisation afterwards and missed this one

vyper/venom/analysis/variable_range/analysis.py
```diff
@@ -0,0 +1,432 @@
...
+        env = self._copy_state(entry_state)
+
+        for inst in bb.instructions:
+            self._inst_entry_env[inst] = self._copy_state(env)
+
+            if inst.opcode == "phi" or not inst.has_outputs:
+                continue
+
+            new_range = self._evaluate_inst(inst, env)
+            if new_range is not None:
```
HodanPlodky: it seems like the new_range should not be None in this part of the code. It for sure has an output so `_evaluate_inst` should not return None (or is there some handler that could return None)

vyper/venom/analysis/variable_range/analysis.py
```diff
@@ -0,0 +1,432 @@
...
+            return state
+
+        cond, true_label, false_label = term.operands
+        branch: Optional[bool] = None
+        if isinstance(true_label, IRLabel) and true_label.value == succ.label.value:
+            branch = True
+        elif isinstance(false_label, IRLabel) and false_label.value == succ.label.value:
+            branch = False
+
+        if branch is None:
```
HodanPlodky, line 166: When would this be triggered?

vyper/venom/analysis/variable_range/analysis.py
```diff
@@ -0,0 +1,432 @@
...
+            if current.is_top:
+                # Can't narrow TOP meaningfully for non-zero
+                return state
+
+            if current.lo < 0:
+                # Range includes negative values (which are all non-zero)
+                if current.lo > 0 or current.hi < 0:
+                    # Range doesn't contain zero, no narrowing needed
+                    pass
+                elif current.lo == 0:
```
HodanPlodky: Since the previous condition `current.lo < 0` this should not trigger

vyper/venom/analysis/variable_range/analysis.py
```diff
@@ -0,0 +1,432 @@
...
+            current = state.get(target, ValueRange.top())
+            # On false branch, value is non-zero.
+
+            if current.is_top:
+                # Can't narrow TOP meaningfully for non-zero
+                return state
+
+            if current.lo < 0:
+                # Range includes negative values (which are all non-zero)
+                if current.lo > 0 or current.hi < 0:
```
HodanPlodky: Only `current.hi < ` should be good no? (since the `current.lo < 0` already)

PR #4791, opened 2025-12-04
feat[lang]: add custom errors

AlbertoCentonze:
_co-authored by gpt-5.5_

### What I did

Added Vyper custom errors:

```vyper
error Unauthorized:
    caller: address

@external
def fail():
    raise Unauthorized(caller=msg.sender)
```

Custom errors are included in generated ABI/interface output and can be
used as revert payloads for `raise` and `assert`.

### How I did it

- added module-level `error` declarations to the AST and semantic model
- added `ErrorT`, ABI output, interface output, and imported/exported
  module propagation
- restricted custom error calls to `raise` / `assert` diagnostics
- implemented custom-error revert encoding in both legacy codegen and
  Venom
- kept custom error arguments under constant-context checks, matching
  string revert reasons

### How to verify it

Ran locally:

```bash
make lint
pytest -q tests/functional/codegen/features/test_custom_errors.py -q
pytest -q tests/functional/syntax/test_custom_errors.py \
  tests/unit/compiler/test_abi.py::test_custom_error_abi \
  tests/unit/compiler/test_abi.py::test_exported_module_custom_error_abi \
  tests/unit/semantics/types/test_error.py -q
```

### Commit message

```
Custom errors provide a cheaper and more structured way to report revert
reasons than strings. Vyper could consume external ABIs containing
errors, but contracts could not declare or raise their own custom
errors.

Add module-level `error` declarations and expose them through semantic
analysis, generated ABIs, and generated interfaces. Custom error calls
are only valid in `raise` and `assert` revert positions, where arguments
are type-checked and evaluated under constant context just like string
revert reasons.

Both legacy codegen and Venom encode revert payloads as a 4-byte
selector followed by ABI-encoded arguments. The assert lowering keeps
error arguments on the failure path so successful assertions do not
evaluate them.
```

Vyper now supports declaring and raising custom errors, emitted in ABI
output as `type: "error"` entries.

### Cute Animal Picture

quokka


charles-cooper (approved):
Lgtm. Merging, would like a second pair of eyes from @Sporarum, please do a review and submit any changes you want as a follow up!

Sporarum:
LGTM !

PR #4795, opened 2025-12-17
feat[venom]: remove `IRAbstractMemLoc`, improve memory allocation

HodanPlodky:
### What I did
I implemented improvements for venom location discussed here https://github.com/vyperlang/vyper/pull/4756#pullrequestreview-3516510505 with exception of calling convention point.

The `IRAbstractMemOperand` is no longer used and it is replaced by `BasePtrAnalysis` which also handles `from_operands` method and other utility functions for getting memory locations from instructions

The allocation algorithm was improved. It no longer only gets address that is above all other conflicting locations, it tries to find the first possible position instead.

### Commit message

```
two main changes:

1. replace embedded `IRAbstractMemLoc` operands with analysis-based
   aliasing. previously, `alloca`/`palloca`/`calloca` carried
   `IRAbstractMemLoc` operands with inline offset/size. this was
   redundant (derivable from operands) and fragile (instruction
   mutations invalidated cached info). aliasing should be derived at
   query time, not embedded in representation.

2. improve memory allocator density via first-fit allocation. the old
   allocator always appended at end-of-memory; the new one finds gaps
   left by non-overlapping lifetimes, reducing memory usage.

new architecture:

`BasePtrAnalysis` is the core - it tracks pointer provenance through
`alloca` -> `gep` -> `phi` chains, answering "which allocation does this
pointer derive from?" at query time. supporting types:
- `Allocation`: wraps an `alloca`/`palloca` instruction
- `Ptr`: an `Allocation` plus an offset
- `MemoryLocation`: describes a memory access (offset, size, optional
  `Allocation` for alias analysis)

pass order: `FixMemLocationsPass` (creates pinned `alloca`s for special
regions like `FREE_VAR_SPACE`) -> optimizations ->
`ConcretizeMemLocPass` (assigns concrete addresses)

allocator details:
- `FN_START` changed from 64 to 0; `FREE_VAR_SPACE` now protected via
  pinned `alloca`s created by `FixMemLocationsPass`
- cross-function memory: `invoke` returns `MemoryLocation.UNDEFINED`
  (conservative); callee's `mems_used` reserved before caller allocation

pass integration:
- `LoadElimination`: refactored to use `BasePtrAnalysis` for alias
  queries
- `MemoryAliasAnalysis`: uses
  `BasePtrAnalysis.get_read/write_location()`
- `DSE`/`MemSSA`: updated to work with new `MemoryLocation` structure

refactors:
- fuse `MemoryLocationAbstract` and `MemoryLocationSegment` into single
  `MemoryLocation` class
- rename `BasePtr` -> `Allocation` for clarity
- encapsulate `_allocated_args` with public accessors
  (`get_palloca_inst`, `has_palloca`, `set_palloca`,
  `get_live_pallocas`)
- simplify `get_read`/`get_write` logic in `memory_location.py`

bug fixes:

1. `mem2var` `palloca` miscompile - after removing `IRAbstractMemLoc`,
   `mload` used SIZE literal as address instead of `palloca.output`.
   caused silent miscompilation when accessing first field of memory-
   passed structs - returned garbage instead of the actual value.

2. `mem2var` uses mutation - `dfg.get_uses()` returns a reference.
   `add_after` mutates the live set, so loops over captured uses
   iterated forever. fixed via `.copy()` snapshot. added
   `DFG.get_transitive_uses()` helper.

3. nop'd `palloca` crash - passes run post-order (callees first). when
   `RemoveUnusedVariables` nops a callee's `palloca`, caller's later
   passes crashed accessing mutated instructions. fixed by filtering
   nop'd instructions and encapsulating `_allocated_args` with
   accessors.

4. `SimplifyCFG` data segment labels - block merging didn't update data
   segment refs to removed labels -> `KeyError` during codegen.

5. orphaned `calloca` - when `DSE` removes `palloca` stores and
   `RemoveUnusedVariables` nops the `palloca`, the caller's `calloca`
   becomes orphaned (its paired `palloca` no longer exists). currently
   handled by removing orphaned `calloca`s and their transitive uses in
   `ConcretizeMemLocPass`. this is a workaround - the proper fix would
   be for `DSE` to understand that stores to `calloca` memory are dead
   when the callee's `palloca` is unused, eliminating them earlier in
   the pipeline. left for future work.

other:
- remove `sha3_64` from venom (lowered to `alloca` + `mstore` + `sha3`)
- immutables use pinned `alloca` at position 0
- add `InstUpdater.add_after()` helper
```

### Cute Animal Picture

### DIFF
--- a/vyper/venom/analysis/base_ptr_analysis.py
+++ b/vyper/venom/analysis/base_ptr_analysis.py
@@ -87,11 +87,12 @@ class BasePtrAnalysis(IRAnalysis):
             assert isinstance(size, IRLiteral)
             assert isinstance(callee_label, IRLabel)
             callee = self.function.ctx.get_function(callee_label)
-            palloca = callee.allocated_args[_id.value]
-            assert isinstance(palloca, IRInstruction)
-            # palloca may have been nop'd by earlier passes on the callee
-            if palloca.opcode == "palloca":
-                self.var_to_mem[inst.output] = set([Ptr.from_alloca(palloca)])
+            palloca_inst = callee.get_palloca_inst(_id.value)
+            # palloca instruction can get modified (e.g. nop'ed) by previous
+            # passes. check it is still a palloca before adding it to the
+            # allocation set
+            if palloca_inst is not None and palloca_inst.opcode == "palloca":
+                self.var_to_mem[inst.output] = set([Ptr.from_alloca(palloca_inst)])
 
         elif opcode == "gep":
             assert isinstance(inst.operands[0], IRVariable), inst.parent
--- a/vyper/venom/ir_node_to_venom.py
+++ b/vyper/venom/ir_node_to_venom.py
@@ -807,7 +807,7 @@ def _convert_ir_bb(fn, ir, symbols):
                     "palloca", alloca.size, alloca._id, annotation=alloca.name
                 )
                 bb.instructions[-1].annotation = f"{alloca.name} (memory)"
-                fn.allocated_args[alloca._id] = bb.instructions[-1]
+                fn.set_palloca(alloca._id, bb.instructions[-1])
                 if _pass_via_stack(_current_func_t)[alloca.name]:
                     param = fn.get_param_by_id(alloca._id)
                     assert param is not None
--- a/vyper/venom/passes/concretize_mem_loc.py
+++ b/vyper/venom/passes/concretize_mem_loc.py
@@ -75,9 +75,14 @@ class ConcretizeMemLocPass(IRPass):
     def _handle_bb(self, bb: IRBasicBlock):
         for inst in bb.instructions:
             if inst.opcode in ("alloca", "palloca", "calloca"):
-                # get palloca from calloca
                 base_ptr = self.base_ptrs.ptr_from_op(inst.output)
-                assert base_ptr is not None, (inst, base_ptr)
+                if base_ptr is None:
+                    # calloca's palloca was nop'd (DSE removed stores, then
+                    # RemoveUnusedVariables nop'd the palloca). The memory
+                    # isn't used, so remove calloca and its uses.
+                    assert inst.opcode == "calloca", inst
+                    self._remove_unused_calloca(inst)
+                    continue
                 if not self.allocator.is_allocated(base_ptr.base_alloca):
                     # unallocated alloca, we need to allocate it.
                     #
@@ -91,6 +96,10 @@ class ConcretizeMemLocPass(IRPass):
                 self.updater.replace(inst, "assign", [concrete])
             if inst.opcode == "gep":
                 inst.opcode = "add"
+
+    def _remove_unused_calloca(self, inst: IRInstruction):
+        to_remove = self.dfg.get_transitive_uses(inst)
+        self.updater.nop_multi(to_remove)
 
 
 class MemLiveness:
--- a/tests/functional/codegen/features/test_mem2var_mstore_bug.py
+++ b/tests/functional/codegen/features/test_mem2var_mstore_bug.py
@@ -1,0 +1,132 @@
+"""
+Regression tests for mem2var/palloca/calloca handling.
+
+Tests memory-passed parameter handling when DSE eliminates unused
+stores to palloca memory. The corresponding calloca and mcopy in
+the caller should also be eliminated.
+"""
+import copy
+
+import pytest
+
+from vyper.compiler.settings import VenomOptimizationFlags
+
+
+@pytest.fixture
+def no_inline_settings(compiler_settings):
+    """Create settings with function inlining disabled (venom only)."""
+    settings = copy.copy(compiler_settings)
+    if settings.experimental_codegen:
+        settings.venom_flags = VenomOptimizationFlags(disable_inlining=True)
+    return settings
+
+
+@pytest.mark.hevm
+def test_mem2var_mstore_single_field_struct(get_contract, no_inline_settings):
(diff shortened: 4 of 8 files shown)


vyper/venom/analysis/base_ptr_analysis.py
```diff
@@ -0,0 +1,280 @@
...
+
+@dataclass(frozen=True)
+class BasePtr:
+    source: IRInstruction
+    offset: int | None
+    size: int
+
+    def __post_init__(self):
+        # sanity check
+        assert self.source.opcode in ("alloca", "palloca"), self.source
```
charles-cooper: another sanity check that can prevent some class of miscompiles: `assert self.offset < self.size`

vyper/venom/analysis/base_ptr_analysis.py
```diff
@@ -0,0 +1,280 @@
...
+            self.var_to_mem[inst.output] = self.get_all_possible_memory(inst.operands[0])
+
+        return original != self.var_to_mem.get(inst.output, set())
+
+    def base_ptr_from_op(self, op: IROperand) -> Optional[BasePtr]:
+        if not isinstance(op, IRVariable):
+            return None
+        item = self.var_to_mem.get(op, set()).copy()
+        if len(item) == 1:
+            return item.pop()
```
charles-cooper: you can use `next(iter(item))` to avoid the copy/mutate pattern

vyper/venom/passes/concretize_mem_loc.py
```diff
@@ -69,53 +69,48 @@ def run_pass(self):
 
     def _handle_bb(self, bb: IRBasicBlock):
         for inst in bb.instructions:
-            new_ops = [self._handle_op(op, inst) for op in inst.operands]
-            inst.operands = new_ops
+            if inst.opcode in ("alloca", "palloca", "calloca"):
+                base_ptr = self.base_ptrs.base_ptr_from_op(inst.output)
```
charles-cooper: i liked the previous style which had _handle_op as a separate function, decreases nesting. we could rename it here to handle_alloca or similar, but i do prefer decreased nesting + encapsulation

vyper/venom/analysis/base_ptr_analysis.py
```diff
@@ -0,0 +1,280 @@
...
+    def __post_init__(self):
+        # sanity check
+        assert self.source.opcode in ("alloca", "palloca"), self.source
+
+    def offset_by(self, offset: int | None):
+        if offset is None or self.offset is None:
+            return dc.replace(self, offset=None)
+        return dc.replace(self, offset=self.offset + offset)
+
+    def without_offset(self):
```
charles-cooper: is this still used? actually the `offset` thing is weird to have in BasePtr, since it would no longer be a BasePtr. Maybe it would be just a Ptr.

vyper/venom/ir_node_to_venom.py
```diff
@@ -118,7 +118,8 @@
 SymbolTable = dict[str, IROperand]
 _alloca_table: dict[int, IROperand]
 _callsites: dict[str, list[Alloca]]
-_immutables_region: IRAbstractMemLoc
+_immutable_alloca_id: Optional[int]
```
charles-cooper, line 121: should be IRLiteral?

vyper/venom/passes/load_elimination.py
```diff
@@ -70,7 +35,28 @@ def analyze(self):
...
+    def get_space(self, eff: Effects | str) -> AddrSpace:
+        if isinstance(eff, Effects):
+            space = to_addr_space(eff)
+        elif eff == "dload":
+            space = DATA
+        elif eff == "calldataload":
+            space = CALLDATA
+        else:
+            raise CompilerPanic("Invalid effect type in load elimination")
+        assert space is not None
```
charles-cooper: unreachable right? actually i would restructure to `if ...: return <space>`, no need for the space variable, and then panic at the end if we escaped all the returns.

vyper/venom/passes/load_elimination.py
```diff
@@ -103,13 +89,33 @@ def _merge(self, bb: IRBasicBlock) -> Lattice:
...
+        if isinstance(op, IRVariable):
+            op = self.dfg._traverse_assign_chain(op)
+        assert isinstance(op, (IRVariable, IRLiteral))
+        return self.base_ptrs.from_operands(op, self.size)
+
+    def get_read(self, inst: IRInstruction) -> IROperand | MemoryLocation:
+        if self.space in (addr_space.MEMORY, addr_space.TRANSIENT, addr_space.STORAGE):
+            memloc = self.base_ptrs.get_read_location(inst, self.space)
+            if not memloc.is_fixed:
+                # assumes it is load
```
charles-cooper: strange assumption?

HodanPlodky: the only cases for `get_read` in load elim are loads opcodes but there could be assert for sure

vyper/venom/basicblock.py
```diff
@@ -486,6 +435,7 @@ def copy(self) -> IRInstruction:
         ret.annotation = self.annotation
         ret.ast_source = self.ast_source
         ret.error_msg = self.error_msg
+        ret.parent = self.parent
```
harkal, line 441: Maybe we shouldn't copy the parent. Assigning the responsibility of providing a proper parent to the copy to the one calling the copy() is safer imo

charles-cooper (approved):
cool

PR #4796, opened 2025-12-19
fix[tool]: add backwards compatibility alias for `generate_ir` and `getpos`

cyberthirst:
[vyper/vyper#4663](https://github.com/vyperlang/vyper/pull/4663) broke the import API for boa.                                                           
                                                                                                           
  Boa currently imports `generate_ir`:                                                                     
  https://github.com/vyperlang/titanoboa/blob/68bb8cd91976223f60af5eb0283fc587c3decf1b/boa/contracts/vyper/compiler_utils.py#L16
                                                                                                           
  Boa also imports `getpos`:                                                                               
  https://github.com/vyperlang/titanoboa/blob/68bb8cd91976223f60af5eb0283fc587c3decf1b/boa/contracts/vyper/ir_executor.py#L18
                                                                                                           
  This fix is more convenient than handling this based on the vyper version in boa.                        
                                                                                                           
  ### Commit messages                                                                                      
  Add generate_ir as an alias for generate_venom to maintain backwards compatibility with tools like titanoboa.
                                                                                                           
  Add getpos back to compile_ir for backwards compatibility with titanoboa.

### DIFF
--- a/vyper/ir/compile_ir.py
+++ b/vyper/ir/compile_ir.py
@@ -30,6 +30,11 @@ from vyper.utils import MemoryPositions
 from vyper.version import version_tuple
 
 
+# backwards compatibility for titanoboa
+def getpos(node):
+    return (node.lineno, node.col_offset, node.end_lineno, node.end_col_offset)
+
+
 def generate_cbor_metadata(
     compiler_metadata: Any,
     runtime_codesize: int,


cyberthirst:
hmm, it seems that the referenced PR also removed `getpos` which boa uses

PR #4798, opened 2025-12-23
fix[venom]: remove unreachable functions

harkal:
### What I did

Modified FCGAnalysis to use DFS from entry, building the call graph and tracking reachability together. Added get_unreachable_functions() to expose orphan functions. In run_passes_on(), unreachable functions are now removed after FCG analysis.

This fixes an issue where orphan functions (unreachable from entry) would not get optimization passes run on them, causing assertion failures during assembly generation due to unnormalized CFGs. Orphans are generally not generated by the frontend, but can be generated from separate frontends like yul2venom.

### Commit message

```
`FCGAnalysis` now computes reachability in a single DFS pass from entry,
building the call graph and tracking reachable functions together.
Orphan functions (unreachable from entry) are removed before running
per-function passes.

This fixes assertion failures during assembly generation caused by
unnormalized CFGs in orphan functions that never had optimization passes
run on them. Orphans are generally not generated by the vyper frontend,
but can arise from other frontends like yul2venom.

Also changes early `return` to `break` in function_inliner for hygiene,
ensuring the method exits at its end rather than mid-loop.
```

### Cute Animal Picture

### DIFF
--- a/vyper/venom/passes/function_inliner.py
+++ b/vyper/venom/passes/function_inliner.py
@@ -47,7 +47,7 @@ class FunctionInlinerPass(IRGlobalPass):
         for _ in range(function_count):
             candidate = self._select_inline_candidate()
             if candidate is None:
-                return
+                break
 
             # print(f"Inlining function {candidate.name} with cost {candidate.code_size_cost}")
 
--- a/vyper/venom/__init__.py
+++ b/vyper/venom/__init__.py
@@ -122,6 +122,10 @@ def run_passes_on(ctx: IRContext, flags: VenomOptimizationFlags) -> None:
     assert ctx.entry_function is not None
     fcg = ir_analyses[ctx.entry_function].force_analysis(FCGAnalysis)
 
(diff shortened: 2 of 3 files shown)


vyper/venom/passes/function_inliner.py
```diff
@@ -47,7 +47,7 @@ def run_pass(self):
         for _ in range(function_count):
             candidate = self._select_inline_candidate()
             if candidate is None:
-                return
+                break
```
charles-cooper, line 50: Why this change?

charles-cooper (approved):
looks good, just one question https://github.com/vyperlang/vyper/pull/4798/files#r2644229628

i think it's not semantics-changing, but was just wondering why

vyper/venom/passes/function_inliner.py
harkal: hygienic mostly to exit the method at the end of it's code. Protects future self of adding code after the loop to execute at `run_pass()` and missing the `return` in the loop above

PR #4803, opened 2026-01-07
fix[venom]: fix `make_ssa` self-loops and liveness analysis

harkal:
### How I did it

Fixed SSA construction bugs for self-loops and liveness analysis

### Commit message

```
MakeSSA fixes:
  - Include self-edges in phi operands (`_place_phi` was skipping them)
  - Convert single-operand phis to assign instead of removing them,
    which left variable uses undefined
  - Add `ensure_well_formed()` after phi simplification to maintain
    instruction ordering (phis must stay at block top)
  - Add phi deduplication: simplify to `assign` when all operands are identical

Liveness analysis fix:
  - Fix equation order: use `(live - defs) U uses` instead of (live U uses) - defs`
  - The old order incorrectly removed variables that are both used and defined
    in the same instruction (e.g., `%x = add %x, 1`), causing `MakeSSA` to miss
    phi placements in tight self-loops

Code quality:
  - Use consistent .value (not mixed `.name`/`.value`) for variable name keys
```

### Cute Animal Picture

### DIFF
--- a/tests/unit/compiler/venom/test_algebraic_optimizer.py
+++ b/tests/unit/compiler/venom/test_algebraic_optimizer.py
@@ -232,6 +232,48 @@ def test_offsets():
     _check_pre_post(pre, post)
 
 
+@pytest.mark.parametrize("iszero_count", range(5))
+def test_assert_unreachable_iszero_chain(iszero_count):
+    """
+    Test that iszero chains are optimized for assert_unreachable
+    the same way they are for jnz (truthy context)
+    """
+    iszero_chain = ""
+    for i in range(iszero_count):
+        new = i + 1
+        iszero_chain += f"""
+        %cond{new} = iszero %cond{i}"""
+    iszero_chain_output = f"cond{iszero_count}"
+
+    pre = f"""
+    main:
+        %par = source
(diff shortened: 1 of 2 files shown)


vyper/venom/passes/make_ssa.py
```diff
@@ -158,9 +159,18 @@ def _remove_degenerate_phis(self, entry: IRBasicBlock):
...
+                # Single incoming value - convert to assign to preserve uses
+                inst.opcode = "assign"
+                inst.operands = [new_ops[1]]  # new_ops is [label, value]
             else:
-                inst.operands = new_ops
+                # Check if all operands have the same value (can simplify to assign)
+                # new_ops is [label1, val1, label2, val2, ...]
+                values = new_ops[1::2]  # extract values at odd indices
+                if len(set(values)) == 1:
+                    inst.opcode = "assign"
```
HodanPlodky, line 170: Just to be sure this does not seem to be triggered in tests

charles-cooper: we could probably leave it as is and let phi elim pass clean up

PR #4804, opened 2026-01-07
fix[venom]: minor cleanup in `PhiEliminationPass`

harkal:
### Commit message

```
- Improve comments
- Don't use mutative `pop()` when not needed
```

### Cute Animal Picture

### DIFF
--- a/tests/unit/compiler/venom/test_algebraic_optimizer.py
+++ b/tests/unit/compiler/venom/test_algebraic_optimizer.py
@@ -232,6 +232,48 @@ def test_offsets():
     _check_pre_post(pre, post)
 
 
+@pytest.mark.parametrize("iszero_count", range(5))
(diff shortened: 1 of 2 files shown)


vyper/venom/passes/phi_elimination.py
```diff
@@ -72,16 +72,15 @@ def _get_phi_origins_r(
...
 
             if len(res) > 1:
-                # if this phi has more than one origin, then for future
-                # phis, it is better to treat this as a barrier in the
-                # graph traversal. for example (without basic blocks)
-                #   %a = 1
-                #   %b = 2
-                #   %c = phi %a, %b  ; has two origins
+                # multi-origin phi: treat the phi itself as a "barrier" origin.
+                # this is correct because the phi is a single SSA definition,
```
charles-cooper, line 76: we could say that it improves the optimization

PR #4807, opened 2026-01-07
feat[venom]: load elimination improvements

harkal:
### Commit message

```
- Remove dead code: `if memloc is None` branch (get_memloc never returns None)
- Add `_normalize_operand()` to traverse assign chains for consistent lattice
  keys when memory location is not statically fixed
- Convert assertions to early returns in `_handle_load` phi insertion when
  predecessors are skipped (missing ptr, empty values, or count mismatch)
- Add tests for nested diamond CFG patterns verifying correct bailout behavior
```

### Cute Animal Picture


charles-cooper:
@codex

tests/unit/compiler/venom/test_load_elimination.py
```diff
@@ -488,3 +488,128 @@ def test_load_elimination_inter_distant_bb(position):
...
+        ; no store, inherits {1, 2}
+        jmp @F
+    F:
+        ; outer join - predecessors D and E both have {1, 2}
+        ; algorithm should bail out (no phi)
+        %val = mload %ptr
+        sink %val
+    """
+    # No change expected - can't create phi when predecessors have multiple values
+    _check_pre_post(pre, pre, hevm=False)
```
HodanPlodky: `check_no_change` already has `hevm` set to false
```suggestion
    _check_no_change(pre)
```

vyper/venom/passes/load_elimination.py
```diff
@@ -214,25 +212,27 @@ def _handle_load(self, inst):
...
                 if not isinstance(val, IRVariable):
-                    # could be extended by
-                    # adding stores to source
-                    # basicblocks
+                    # could be extended by adding stores to source basicblocks
                     return
                 ops.extend([pred.label, val])
 
-            assert len(ops) == 2 * len(existing_value), (ops, existing_value, inst)
+            if len(ops) != 2 * len(existing_value):
```
HodanPlodky: why this has to be the if and not assert. All the cases should be already short circuited in the for beforehand.

charles-cooper: see https://github.com/vyperlang/vyper/pull/4807/commits/f6c43049b6e3eb22c9d7fffb7ae8925fe23b9337

HodanPlodky (approved):

PR #4808, opened 2026-01-07
fix[venom]: fix phi update in `simplify_cfg`

harkal:
### Commit message

```
`_merge_jump` iterated over all of `a`'s successors to update phis, but
only the bypassed block's target needs updating. when other successors
have phis without the bypassed block's label, index() raises ValueError.

also:
- delete phi entry when `a` already has edge to target (avoid duplicate)
- skip merge when phi operands would conflict (prevents information
  loss)
```

### Cute Animal Picture

### DIFF
--- a/tests/functional/codegen/features/test_immutable_allocator_overlap.py
+++ b/tests/functional/codegen/features/test_immutable_allocator_overlap.py
@@ -1,0 +1,55 @@
+"""
+Regression test for issue #4800: overlapping reserved memory intervals in Venom
+could corrupt dynamic-string hashing during constructor execution.
+"""
+
+
+def test_immutable_hashing_overlap_regression(get_contract, make_input_bundle, keccak):
+    module = """
+NAME_HASH: immutable(bytes32)
+VERSION_HASH: immutable(bytes32)
+DOMAIN_SEPARATOR: immutable(bytes32)
+PADDING: immutable(uint256[12])
+
+EIP712_TYPEHASH: constant(bytes32) = keccak256(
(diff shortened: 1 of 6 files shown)


charles-cooper (approved):

tests/unit/compiler/venom/test_simplify_cfg.py
```diff
@@ -189,3 +193,117 @@ def test_merge_jump_other_successor_has_phi():
...
+
+    other:
+        %x = source
+        jmp @join
+
+    passthrough:
+        jmp @join
+
+    join:
+        %y = phi @other, %x
```
HodanPlodky: how would this be created naturally this should be invalid phi node right?

charles-cooper: yea it's invalid venom, will remove and maybe replace with another test

vyper/venom/passes/simplify_cfg.py
```diff
@@ -48,14 +58,20 @@ def _merge_jump(self, a: IRBasicBlock, b: IRBasicBlock):
...
-            for inst in next_bb.instructions:
-                # assume phi instructions are at beginning of bb
-                if inst.opcode != "phi":
-                    break
-                inst.operands[inst.operands.index(b.label)] = a.label
+        # Update phis in next_bb: b is no longer predecessor, a is
+        for inst in next_bb.instructions:
+            if inst.opcode != "phi":
+                break
+            if b.label not in inst.operands:
```
HodanPlodky: Shouldn't this be assert since otherwise what would be the value of the `phi` if it would come from that basic block

PR #4811, opened 2026-01-15
feat[venom]: add direct-to-venom pipeline

charles-cooper:
### What I did

Add experimental codegen pathway (`--experimental-codegen`) that compiles
Vyper AST directly to Venom IR, bypassing the legacy s-expression IR
intermediate representation.

### How I did it

**New module: `vyper/codegen_venom/` (~10,500 LOC)**

Core lowering:
- `expr.py` - expression lowering (literals, operators, subscripts, attributes, calls)
- `stmt.py` - statement lowering (assignments, control flow, loops)
- `context.py` - codegen context with memory/storage/transient abstractions
- `module.py` - module compilation with external function dispatch

Abstractions:
- `value.py` - `VyperValue`: explicit tagged union for stack values vs located
  pointers, preventing the location-confusion bugs implicit in legacy `IRnode`
- `buffer.py` - `Buffer`/`Ptr`: typed memory allocation and pointer arithmetic
  with location invariant (`buf` provenance required iff `MEMORY`)

Builtins (`builtins/`, ~3,400 LOC):
- separated into dedicated modules (vs inline in legacy)
- `simple.py`, `math.py`, `hashing.py`, `bytes.py`, `strings.py`,
  `convert.py`, `create.py`, `system.py`, `misc.py`, `abi.py`

ABI (`abi/`):
- `abi_encoder.py` - ABI encoding for calls/returns
- `abi_decoder.py` - ABI decoding with bounds checking

**`vyper/venom/builder.py` (~580 LOC)** - type-safe API for IR construction

**New Venom passes for invoke-related memory copy forwarding:**
- `readonly_memory_args_analysis.py` - interprocedural fixed-point
  analysis inferring which invoke args are read-only memory pointers
- `readonly_invoke_arg_copy_forwarding.py` - forward staged `mcopy`
  directly into readonly invoke parameters, eliminating the copy
- `internal_return_copy_forwarding.py` - forward invoke return buffer
  copies, rewriting downstream uses to the return buffer directly
- `invoke_copy_forwarding_common.py` - shared DFG/alias/dominator
  helpers for both forwarding passes

**ABI encoder aliasing fix:**
- reorder dynamic child encoding to write child data *before* the
  static offset word, preventing corruption when invoke-arg forwarding
  passes source references that alias the destination buffer

Architecture:
- **Two-phase compilation**: separate deploy and runtime contexts
- **Jumptable dispatch**: sparse and dense dispatch for selector lookup
- **Overlap-safe assignment**: `_copy_complex_type` uses source
  `VyperValue` location to skip staging when source and destination are
  in different address spaces (no aliasing possible)

Changes to existing code:
- `vyper/compiler/phases.py` - hook to use new codegen when flag enabled
- **delete `vyper/venom/ir_node_to_venom.py`** (~860 LOC) - legacy
  IRnode-to-Venom translation no longer needed
- `cfg`/`cfg_runtime` output formats now require `--experimental-codegen`
- `base_ptr_analysis.py` - `iload` returns `UNDEFINED` (conservative)
- `memory_copy_elision.py` - elide repeated identical mcopies
- `literals_codesize.py` - fix shared `IRLiteral` in-place mutation bug
- optimization pipelines (O2/O3/Os) updated with invoke-copy forwarding
- no changes to legacy codegen or default behavior

### How to verify it

CI. full test suite passes with `--experimental-codegen`.
gated behind flag — default behavior unchanged.

### Commit message

```
feat[venom]: add direct Vyper AST to Venom IR codegen pipeline

add an experimental codegen pathway (`--experimental-codegen`) that
compiles Vyper AST directly to Venom IR, bypassing the legacy
s-expression IR intermediate representation.

new module `vyper/codegen_venom/` (~10,500 LOC) implements the full
lowering pipeline: expressions, statements, control flow, builtins,
ABI encoding/decoding, and module-level dispatch (including sparse
and dense jumptable).

key design decisions:

- `VyperValue`: explicit tagged union tracking whether a value lives
  on the stack or at a storage location (memory/storage/transient/
  code). prevents location-confusion bugs that were implicit in
  legacy `IRnode` — e.g., using a storage slot number as a memory
  pointer.

- `Buffer`/`Ptr`: typed memory allocation and pointer arithmetic.
  `Ptr` carries a `DataLocation` and enforces that memory pointers
  track their buffer provenance. assignment codegen uses source
  location to skip overlap-staging when source and destination are
  in different address spaces.

- `VenomBuilder`: type-safe API for Venom IR instruction emission
  with semantic argument order (reversal to EVM stack order is
  internal).

builtins are separated into dedicated modules (~3,400 LOC) by domain
(math, hashing, bytes, strings, convert, create, system, abi). ABI
encoder and decoder are symmetric and self-contained.

new Venom optimization passes for invoke-related memory copy
forwarding: interprocedural readonly-arg analysis, readonly invoke-arg
copy forwarding, and internal return-buffer copy forwarding. the ABI
encoder is reordered to write dynamic child data before static offset
words for aliasing safety with these passes.

delete `vyper/venom/ir_node_to_venom.py` (~860 LOC) — the legacy
IRnode-to-Venom translation layer is no longer needed.

gated behind `--experimental-codegen`; no changes to default behavior.
```

add experimental direct Vyper-AST-to-Venom-IR codegen pipeline (`--experimental-codegen`), bypassing the legacy s-expression IR layer. includes new Venom optimization passes for invoke-related memory copy forwarding.


harkal (approved):
milestone 🚀

PR #4830, opened 2026-01-31
feat[docs]: revamp vyper docs

lufa23:
### What I did

Documentation audit and update bringing docs in line with v0.4.x. The existing docs had stale syntax (positional event args, old decorators, removed builtins), dead links, and no coverage of module imports or Solidity migration.

### How I did it

- Restructured vyper-by-example from a single 650-line file into a folder of individual pages
- Added quickstart guide and Solidity differences page
- Consolidated fragmented testing docs (separate Brownie / Titanoboa pages) into one page with deprecation guidance
- Added redirects for removed pages
- Fixed code examples, CLI flags, dead links, and RST formatting across ~45 files
- Added cross-references between related pages throughout

### How to verify it

Build the docs locally (`cd docs && make html`) or review the rendered pages on RTD once deployed.

### Commit message

```
the docs had fallen behind the v0.4.x language changes: code examples
used removed syntax (positional event args, old decorators), links
pointed to dead URLs, and major topics like module imports and Solidity
migration had no coverage.

restructure vyper-by-example from a single 650-line file into a folder
of individual pages, making each example independently navigable and
easier to maintain. add a quickstart guide and a Solidity differences
page — the two most common entry points for new users. consolidate the
fragmented testing docs (separate Brownie / Titanoboa pages) into one
page with deprecation guidance.

content fixes across ~45 files: update selfdestruct docs for EIP-6780,
fix event logging to use keyword arguments, correct CLI flags, replace
dead external links, and add cross-references between related pages
throughout.
```

Comprehensive documentation update for v0.4.x: restructured vyper-by-example into individual pages, added quickstart and Solidity differences guides, fixed stale code examples and dead links across ~45 files.

### Cute Animal Picture

quokka


docs/compiler-exceptions.rst
```diff
@@ -172,19 +172,19 @@ of the error within the code:
 
 .. py:exception:: TypeMismatch
 
-    Raises when attempting to perform an action between two or more objects with known, dislike types.
+    Raises when attempting to perform an action between two or more objects with known, unlike types.
```
Sporarum:
```suggestion
    Raises when attempting to perform an action between two or more objects with incompatible types.
```
Reasoning:
* All types are known(?)
* Because of subtyping, sometimes different types can be compatible

I don't know if this change sufficiently highlights the fact operations sometimes take very different types
For example subscripting

docs/scoping-and-declarations.rst
```diff
@@ -137,13 +137,13 @@ Values that are declared in the module scope of a contract, such as storage vari
     a: int128
 
     @internal
-    def foo() -> int128
+    def _foo() -> int128:
```
Sporarum: Why the underscore ?

I think it looks more intuitive without it

docs/solidity-differences.rst
```diff
@@ -0,0 +1,502 @@
...
+    modifier onlyOwner() {
+        require(msg.sender == owner, "Not owner");
+        _;
+    }
+
+    function withdraw() public onlyOwner {
+        // ...
+    }
+
+Vyper excludes modifiers. While ``onlyOwner`` appears simple, modifiers can execute code before and after the function body, modify state, and obscure the actual logic. Understanding a function requires reading the modifier definitions elsewhere in the codebase.
```
Sporarum:
```suggestion
While ``onlyOwner`` appears simple, modifiers can execute code before and after the function body, modify state, and obscure the actual logic. Understanding a function requires reading the modifier definitions elsewhere in the codebase. For this reason, Vyper does not have modifiers.
```

docs/solidity-differences.rst
```diff
@@ -0,0 +1,502 @@
...
+.. code-block:: vyper
+
+    for i: uint256 in range(100):
+        # Loop body
+
+    # Variable count, but capped at compile time
+    for i: uint256 in range(count, bound=100):
+        # Loop body
+
+Every function has a calculable maximum gas cost. Unbounded storage iteration can exceed the block gas limit, making contracts unusable. Bounded loops prevent this class of issue.
```
Sporarum:
```suggestion
Unbounded storage iteration can exceed the block gas limit, making contracts unusable. Bounded loops prevent this class of issue.
```
We explain the maximum gas thingy below already, and it hurts this paragraph's flow

docs/solidity-differences.rst
```diff
@@ -0,0 +1,502 @@
...
+
+.. note::
+
+   Vyper's bounded loops and lack of recursion make gas costs statically analyzable. The official documentation states: "It is possible to compute a precise upper bound for the gas consumption of any Vyper function call."
```
Sporarum: It's weird to site the official documentation from the official documentation
Am internal link would be better

docs/solidity-differences.rst
```diff
@@ -0,0 +1,502 @@
...
+Bounded Dynamic Arrays
+======================
+
+Storage arrays require a maximum size at compile time:
+
+.. code-block:: vyper
+
+    balances: DynArray[uint256, 100]
+
+Gas costs remain predictable, and attackers cannot grow arrays until iteration exceeds the block limit. For unbounded collections, use ``HashMap``.
```
Sporarum: "attackers cannot grow arrays until iteration exceeds the block limit"
Not sure what this means, but doesn't seem right

docs/solidity-differences.rst
```diff
@@ -0,0 +1,502 @@
...
+
+.. code-block:: vyper
+
+    x: uint256 = 100
+    y: int256 = convert(x, int256)
+
+    addr: address = 0x1234...
+    num: uint160 = convert(addr, uint160)
+
+Conversions between signed/unsigned integers or addresses and integers are visible in the code. No implicit casts that might hide bugs.
```
Sporarum: IIRC there are automatic conversions when the original type fits in the other (uint8 to uint256 for example)

Also when the expected type is a subtype of the expression type, widening is performed

Widening and automatic conversions look the same to the user, but are fundamentally different in that the latter changes the binary representation of the data, whereas widening is only compile-time

docs/solidity-differences.rst
```diff
@@ -0,0 +1,502 @@
...
+Decimal Type
+============
+
+Native fixed-point arithmetic with 10 decimal places:
+
+.. code-block:: vyper
+
+    price: decimal = 1.5
+    quantity: decimal = 2.0
+    total: decimal = price * quantity  # 3.0
```
Sporarum: I think it would be useful for the example to instead use quantities that lead to issues with non-base-10 representations, for example `0.2`

docs/solidity-differences.rst
```diff
@@ -0,0 +1,502 @@
...
+
+    addr: address = 0x1234...
+    num: uint160 = convert(addr, uint160)
+
+Conversions between signed/unsigned integers or addresses and integers are visible in the code. No implicit casts that might hide bugs.
+
+Decimal Type
+============
+
+Native fixed-point arithmetic with 10 decimal places:
```
Sporarum: Should be more explicit about them really being "decimal" places, as in base-10, as "decimal place" is sometimes used colloquially to refer to "binary place"

docs/solidity-differences.rst
```diff
@@ -0,0 +1,502 @@
...
+    price: decimal = 1.5
+    quantity: decimal = 2.0
+    total: decimal = price * quantity  # 3.0
+
+Solidity lacks a native fixed-point type, requiring manual integer scaling.
+
+Bounds Checking
+===============
+
+Array accesses and arithmetic are bounds-checked. Out-of-bounds access reverts. Integer overflow reverts.
```
Sporarum:
```suggestion
Array accesses and arithmetic are bounds-checked at runtime. Out-of-bounds access reverts. Integer overflow reverts.
```

docs/solidity-differences.rst
```diff
@@ -0,0 +1,502 @@
...
+
+Solidity 0.8+ provides similar overflow protection, but it can be disabled with ``unchecked`` blocks. Vyper keeps checks on by default; opt-out requires explicit ``unsafe_*`` builtins.
```
Sporarum: Solidity also keeps check on by default, so this sentence structure is weird, maybe instead:
```suggestion
Solidity 0.8+ provides similar overflow protection, which are disabled in ``unchecked`` blocks. Vyper there is no way to disable the checks. For cases where it would be useful to do so, there are explicit ``unsafe_*`` builtins.
```

docs/solidity-differences.rst
```diff
@@ -0,0 +1,502 @@
...
+    def withdraw():
+        # Cannot be re-entered
+
+The compiler generates the mutex. No manual reentrancy guard implementation required.
+
+.. note::
+
+   The 2016 DAO hack exploited reentrancy to drain ~$60M in ETH (worth ~$150M at the time). This led to the Ethereum hard fork that created Ethereum Classic.
+
+The ``@nonreentrant`` decorator, combined with the ``extcall`` keyword, makes reentrancy points visible during review. Note that ``@nonreentrant`` is opt-in and does not prevent all reentrancy patterns (e.g., cross-contract reentrancy).
```
Sporarum: > The ``@nonreentrant`` decorator, combined with the ``extcall`` keyword, makes reentrancy points visible during review.

Not sure what that means ?

> Note that ``@nonreentrant`` is opt-in and does not prevent all reentrancy patterns (e.g., cross-contract reentrancy).

In case where things sometimes don't behave as expected, we should be very precise about the when
It does not have to be written here, a link to another page would also be acceptable

docs/solidity-differences.rst
```diff
@@ -0,0 +1,502 @@
...
+
+Vyper:
+
+.. code-block:: vyper
+
+    @deploy
+    def __init__(_owner: address):
+        self.owner = _owner
+
+The ``@deploy`` decorator marks the constructor (added in 0.4.0). By convention, constructor parameters use underscore prefixes to distinguish them from state variables.
```
Sporarum: > (added in 0.4.0)

I'm hesitant on having this here: On one hand, it doesn't cost much, on the other it feels like it distracts from comparing vyper to Solidity, since now we are also thinking about different vyper versions
Anyways, either is fine

> By convention, constructor parameters use underscore prefixes to distinguish them from state variables.

Is this a real convention ?
State variables are already distinguished by `self.`

docs/solidity-differences.rst
```diff
@@ -0,0 +1,502 @@
...
+    }
+
+Vyper:
+
+.. code-block:: vyper
+
+    interface IERC20:
+        def transfer(to: address, amount: uint256) -> bool: nonpayable
+
+Interface files use ``.vyi`` extension. Vyper ships with built-in interfaces for ERC20, ERC721, etc. via ``from ethereum.ercs import IERC20``.
```
Sporarum: Should make it clearer that the above is inline code, and interface files have a different syntax

docs/solidity-differences.rst
```diff
@@ -0,0 +1,502 @@
...
+
+Why Vyper?
+==========
+
+Use Vyper if:
+
+- **You have Python experience.** The syntax is familiar.
+- **You want compiler-enforced constraints.** The compiler rejects unbounded loops, implicit conversions, and recursive calls.
+- **You prefer explicit code.** One way to do most things. No modifiers, no inheritance, no operator overloading.
+- **You want safety checks on by default.** Overflow protection and bounds checking require explicit opt-out via ``unsafe_*`` builtins.
```
Sporarum: As per this document, Solidity also has safety checks by default

docs/solidity-differences.rst
```diff
@@ -0,0 +1,502 @@
...
+
+
+Further Reading
+===============
+
+- :ref:`Types <types>` : Type system reference
+- :ref:`Control Structures <control-structures>` : Loops, conditionals, functions
+- :ref:`Interfaces <interfaces>` : Working with external contracts
```
Sporarum: This is good to have, but I think it would be better to include links directly in each section

Sporarum (commented):
Overall seems like a good change, thanks !

Here is what I reviewed and how:
I read the .rst files from the github UI, I have therefore not checked how the site is built
I read the docs pages, but not the example pages, as it was a lot of text, and I'm not super familiar anyways

I found a couple of things weird or confusing, so I commented on them.
When I propose changes, I don't necessarily expect them to be committed like so, it's just a pointer to what I think would make sense

Here are a couple questions I had that cannot be located to specific lines:
(And are also addressed to other reviewers)

Should removed pages have some redirecting logic so we don't lead people to 404s ?

Is it possible to separate sentences by newlines without changing the output ?
(like in markdown)
If yes, I think we should do it, as it makes the underlying files easier to read

docs/scoping-and-declarations.rst
lufa23: Removed in b44c84db 👍

docs/solidity-differences.rst
lufa23: now fixed in ca2f76cc

docs/solidity-differences.rst
lufa23: good catch, thanks👍 (06aeae91)

docs/solidity-differences.rst
lufa23: good point (f30b5fda)

docs/solidity-differences.rst
lufa23: yep, makes sense (377ebb80)

docs/solidity-differences.rst
lufa23: i guess what i meant is: 
"This keeps gas costs predictable and prevents denial-of-service attacks where an attacker grows an array until iteration exceeds the block gas limit."

Would this be correct?

docs/solidity-differences.rst
lufa23: applied your suggestion with a minor syntax fix(acea1bd5)

docs/solidity-differences.rst
lufa23: cc @charles-cooper @benber86

docs/solidity-differences.rst
lufa23: 3c1d3664

docs/solidity-differences.rst
lufa23: 9e11a87a

docs/solidity-differences.rst
lufa23: a839637b

docs/solidity-differences.rst
Sporarum: That sounds a lot clearer, nice !

I'm not an expert, it seems correct, but also weirdly specific ?
IIRC there are many ways to increase gas cost, growing an array is just one of them
So maybe we should stay vague, or say explicitly that this is an example

lufa23: 81cc5cf6

docs/solidity-differences.rst
lufa23: 1d817ab1

docs/compiling-a-contract.rst
```diff
@@ -135,9 +135,9 @@ In codesize optimized mode, the compiler will try hard to minimize codesize by
 * using more loops for data copies.
 
 Enabling Experimental Code Generation
-===========================
+=====================================
 
-When compiling, you can use the CLI flag ``--experimental-codegen`` or its alias ``--venom`` to activate the new `Venom IR <https://github.com/vyperlang/vyper/blob/master/vyper/venom/README.md>`_.
+When compiling, you can use the CLI flag ``--experimental-codegen`` or its alias ``--venom-experimental`` to activate the new `Venom IR <https://github.com/vyperlang/vyper/blob/master/vyper/venom/README.md>`_.
```
charles-cooper: this seems wrong

docs/installing-vyper.rst
```diff
@@ -60,8 +65,8 @@ It is **strongly recommended** to install Vyper in **a virtual Python
 environment**, so that new packages installed and dependencies built are
 strictly contained in your Vyper project and will not alter or affect your
 other development environment set-up.
-For easy virtualenv management, we recommend either `pyenv <https://github.com/pyenv/pyenv>`_
-or `Poetry <https://github.com/python-poetry/poetry>`_.
+For easy virtualenv management, we recommend `uv <https://github.com/astral-sh/uv>`_,
+`pyenv <https://github.com/pyenv/pyenv>`_, or `Poetry <https://github.com/python-poetry/poetry>`_.
```
charles-cooper: let's get rid of poetry. recommend uv tool install or pipx install for standalone installs, otherwise pip install or uv pip install for inside a venv

docs/compiling-a-contract.rst
lufa23: 36ff8cf3

docs/installing-vyper.rst
lufa23: 2c10a2e0

charles-cooper: what about standalone installs?

lufa23: 95765194

Sporarum: In 2c10a2e We lose the clarity that `uv` does not need a `venv`, whereas `pip` does
(assuming that is how it does work)

docs/index.rst
```diff
@@ -6,35 +6,68 @@
...
+
+* Bounds and overflow checking on array accesses and arithmetic
+* Reentrancy protection via the ``@nonreentrant`` decorator (see :ref:`control-structures`)
+* Strong typing with explicit :ref:`type conversions <type_conversions>`
+
+**Predictable execution**
+
+* Decidable gas consumption: every function call has a calculable upper bound
+* Bounded loops only (compile-time maximum iterations)
+* No recursion: execution flow is always linear
```
charles-cooper: not sure what "linear" means. we could say "structurally decreasing"

docs/installing-vyper.rst
charles-cooper: uv pip does need a venv

charles-cooper: but uv can manage its own venv sometimes.. it's unclear to me exactly when actually

docs/event-logging.rst
```diff
@@ -76,7 +77,7 @@ For instance, ``LOG1`` includes one topic, ``LOG2`` includes two topics, and so
 Event declarations look similar to struct declarations, containing one or more arguments that are passed to the event. Typical events will contain two kinds of arguments:
 
     * **Indexed** arguments (topics), which can be searched for by listeners. Each indexed argument is identified by the ``indexed`` keyword.  Here, each indexed argument is an address. You can have up to four indexed arguments (``LOG4``), but indexed arguments are not passed directly to listeners, although some of this information (such as the sender) may be available in the listener's `results` object.
-    * **Value** arguments (data), which are passed through to listeners. You can have any number of value arguments and they can have arbitrary names, but each is limited by the EVM to be no more than 32 bytes.
```
charles-cooper, line 79: why this change?

docs/solidity-differences.rst
```diff
@@ -0,0 +1,496 @@
...
+    initializes: ownable
+    exports: ownable.transfer_ownership
+
+    @deploy
+    def __init__():
+        ownable.__init__()
+
+Three declarations manage module relationships: ``initializes`` (this contract manages the module's storage), ``uses`` (this contract reads module state without initializing), and ``exports`` (expose module functions in the ABI). See :doc:`using-modules` for details.
+
+A contract can be understood by reading one file and its direct imports, and dependencies are explicit.
```
charles-cooper: i would say 
```suggestion
A contract can be understood by reading one file and its direct imports; dependencies and what is exposed in the external function table are explicit.
```

docs/event-logging.rst
lufa23: My understanding was that the 32-byte limit applies to topics (indexed arguments) but not to value/data arguments.

docs/index.rst
lufa23: b156a9b6

docs/event-logging.rst
charles-cooper: Ah right ok

docs/contributing.rst
```diff
@@ -66,6 +66,19 @@ If you need to pull in any changes from ``master`` after making your fork (for
 example, to resolve potential merge conflicts), please avoid using ``git merge``
 and instead, ``git rebase`` your branch.
 
+Commit Messages
```
charles-cooper, line 69: i will want to add AGENTS.md / CLAUDE.md and potentially skills too, but will do in another PR

docs/solidity-differences.rst
```diff
@@ -0,0 +1,496 @@
...
+    def withdraw():
+        # Cannot be re-entered
+
+The compiler generates the mutex. No manual reentrancy guard implementation required.
+
+.. note::
+
+   The 2016 DAO hack exploited reentrancy to drain ~$60M in ETH. This led to the Ethereum hard fork that created Ethereum Classic.
+
+The ``extcall`` keyword makes external call sites explicit and easy to spot during code review. Note that ``@nonreentrant`` is opt-in and uses a global lock that protects against same-contract reentrancy: if any ``@nonreentrant`` function is executing, no other ``@nonreentrant`` function in the same contract can be entered. It does not prevent cross-contract reentrancy (i.e., contract A calling contract B which calls back into contract A). See :ref:`control-structures` for details on the lock behavior.
```
charles-cooper: pragma nonreentrancy turns on nonreentrancy on by default for all functions

docs/solidity-differences.rst
```diff
@@ -0,0 +1,496 @@
...
+        # ...
+
+Inline checks keep the control flow visible from top to bottom.
+
+No Class Inheritance
+====================
+
+Solidity supports multiple inheritance, which introduces the diamond problem and C3 linearization complexity. Vyper excludes inheritance entirely.
+
+For code reuse, Vyper 0.4.0 introduced a module system:
```
charles-cooper:
```suggestion
In Vyper 0.4.0, introduced a module system for powerful code reuse:
```

docs/testing-contracts-titanoboa.rst
charles-cooper, line 1: why is this file being deleted?

docs/solidity-differences.rst
Sporarum: To me that sounds wrong, shouldn't it be instead:
```suggestion
In Vyper 0.4.0, a module system was introduced for powerful code reuse:
```
or
```suggestion
Vyper 0.4.0 introduced a module system for powerful code reuse:
```

charles-cooper: ah yes, let's go with your first suggestion

docs/solidity-differences.rst
lufa23: Are you suggesting to reframe the paragraph to present the pragma as the recommended default, with ``@nonreentrant`` mentioned as the opt-in alternative?

docs/testing-contracts-titanoboa.rst
lufa23: The content of that file was very short and only pointing to external docs. I've merged its content into` testing-contracts.rst`, and let `conf.py `keep the old URLs working. Happy to restore if you prefer it as a separate file

docs/solidity-differences.rst
charles-cooper: not necessarily, that's up to you. but it should definitely be mentioned, since the 2nd sentence says "nonreentrant is opt-in"

docs/testing-contracts-titanoboa.rst
charles-cooper: No, this reasoning is fine

docs/solidity-differences.rst
lufa23: 23c355aa

docs/installing-vyper.rst
lufa23: Addressed in 9576519, poetry removed, standalone (uv tool / pipx) and venv (pip / uv pip) installs are now in separate sections.

docs/solidity-differences.rst
lufa23: da455c41

charles-cooper (approved):

PR #4831, opened 2026-01-31
feat[ci]: add PR commenting to bytecode-size workflow

charles-cooper:
### What I did

Post bytecode size reports as PR comments (currently only shows in step summary).

### How I did it

Follow-up to cf880fd50 which added the bytecode size workflow. Couldn't add PR comments in that PR because `pull_request_target` runs from the base branch, and the workflow file didn't exist on master yet.

The workflow now has two triggers:
- `pull_request` for outside contributors — just writes to step summary
- `pull_request_target` for org members/collaborators — also posts a PR comment

These are mutually exclusive (keyed on `author_association`) so the job doesn't run twice.

`pull_request_target` is needed because `pull_request` from forks gets a read-only token regardless of declared permissions. The contributor gate limits who can trigger `pip install .` of PR code. Token is scoped to `pull-requests: write` so the worst that can happen is comment spam.

### How to verify it

Merge this, then open a PR that changes bytecode output. Should get a comment with the size report.

### Commit message

```
follow-up to cf880fd50 — couldn't add comments there because
pull_request_target runs from the base branch and the workflow didn't
exist on master yet.

uses both pull_request and pull_request_target, mutually exclusive via
author_association so the job doesn't run twice. outside contributors
get step summary only, org members also get a PR comment.

pull_request_target is needed because fork PRs always get a read-only
GITHUB_TOKEN. contributor gate limits who can trigger pip install of PR
code. token scoped to pull-requests write only so the worst case is
comment spam.
```

N/A (CI only)

### Cute Animal Picture

capybara

### DIFF
--- a/.github/workflows/bytecode-size.yml
+++ b/.github/workflows/bytecode-size.yml
@@ -82,7 +82,7 @@ jobs:
       - name: Generate report
         id: report
         run: |
-          python3 head/.github/scripts/compare_bytecode.py base-sizes.json head-sizes.json > report.md
+          python3 base/.github/scripts/compare_bytecode.py base-sizes.json head-sizes.json > report.md
           cat report.md >> "$GITHUB_STEP_SUMMARY"
           {
             echo 'REPORT<<EOF'
--- a/.github/workflows/bytecode-size.yml
+++ b/.github/workflows/bytecode-size.yml
@@ -19,7 +19,7 @@ jobs:
         with:
           script: |
             const marker = '<!-- bytecode-size-report -->';
-            const { data: comments } = await github.rest.issues.listComments({
+            const comments = await github.paginate(github.rest.issues.listComments, {
               owner: context.repo.owner,
               repo: context.repo.repo,
               issue_number: context.payload.pull_request.number
@@ -90,7 +90,7 @@ jobs:
           script: |
             const marker = '<!-- bytecode-size-report -->';
             const body = marker + '\n\n' + process.env.REPORT;
-            const { data: comments } = await github.rest.issues.listComments({
+            const comments = await github.paginate(github.rest.issues.listComments, {
               owner: context.repo.owner,
               repo: context.repo.repo,
               issue_number: context.payload.pull_request.number
--- a/.github/workflows/bytecode-size.yml
+++ b/.github/workflows/bytecode-size.yml
@@ -2,9 +2,9 @@ name: Bytecode Size Report
 
 on:
   pull_request:
-    branches: [master, main]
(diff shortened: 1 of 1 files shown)


.github/workflows/bytecode-size.yml
```diff
@@ -52,7 +80,45 @@ jobs:
         run: python .github/scripts/measure_bytecode.py > ../head-sizes.json
 
       - name: Generate report
-        run: python3 head/.github/scripts/compare_bytecode.py base-sizes.json head-sizes.json >> "$GITHUB_STEP_SUMMARY"
+        id: report
+        run: |
+          python3 head/.github/scripts/compare_bytecode.py base-sizes.json head-sizes.json > report.md
```
pcaversaccio, line 85: hmm, I know you're only allowing it to be run for `"OWNER","MEMBER","COLLABORATOR"`, but still, let's not take `head` here but use directly `master` instead (like `measure_bytecode` above):
```yaml
python3 .github/scripts/compare_bytecode.py base-sizes.json head-sizes.json > report.md
```
so we don't execute PR-controlled scripts in a `pull_request_target` job.

pcaversaccio, line 85:
```suggestion
          python3 .github/scripts/compare_bytecode.py base-sizes.json head-sizes.json > report.md
```

pcaversaccio (changes requested):

.github/workflows/bytecode-size.yml
charles-cooper: 0077fbb4af15eb51b554e89df5ccdf6072d65570

pcaversaccio: yeah, I mean in theory you could still fuck it up like this:

```yaml
  with:
    ref: ${{ github.event.pull_request.head.sha }}
    path: base
```

but I hope nobody ever merges such code into `master` 😄.

.github/workflows/bytecode-size.yml
```diff
@@ -52,7 +80,45 @@ jobs:
         run: python .github/scripts/measure_bytecode.py > ../head-sizes.json
```
pcaversaccio, line 80:
```suggestion
        run: python3 base/.github/scripts/measure_bytecode.py > ../head-sizes.json
```

pcaversaccio: for consistency

pcaversaccio (changes requested):

pcaversaccio (approved):

PR #4835, opened 2026-02-03
fix[lang]: use correct DataLocation for internal function args

### LINKED ISSUE #4754: parse_args assumes calldata location for `type_from_annotation` call
### Version Information

* vyper Version (output of `vyper --version`): 34a9e08f77039b0d4274c551b0e87e07fc7bac47

### What's your issue about?

the following assumes call the calldata location
https://github.com/vyperlang/vyper/blob/4efae6dd659b221b514ccc5aaee82469c3d18d95/vyper/semantics/types/function.py#L917

this doesn't have to be the case for e.g. internal functions

as such, the following check can be imprecise:
https://github.com/vyperlang/vyper/blob/214a352f5e310a6a899b304a9aa7a5b83582eac5/vyper/semantics/types/utils.py#L90

frederikgramkortegaard:
_parse_args was hardcoding CALLDATA for all functions, but internal function arguments live in memory. Added visibility parameter to select the correct location.

Fixes #4754

### What I did

Fixed `_parse_args` to use the correct `DataLocation` when validating function argument types. Internal functions now correctly check against `MEMORY` instead of `CALLDATA`.

### How I did it

Added a `visibility` parameter to `_parse_args` (defaults to `EXTERNAL` to avoid changing parameter order). When visibility is `INTERNAL` we use `DataLocation.MEMORY` otherwise we uses `DataLocation.CALLDATA`.

### How to verify it

```vyper
@internal
def _foo(x: HashMap[uint256, uint256]) -> uint256:
    return x[0]
```

**Before:** `HashMap[uint256, uint256] is not instantiable in calldata`
**After:** `HashMap[uint256, uint256] is not instantiable in memory`

### Commit message

fix: use correct DataLocation for internal function args

Fixed incorrect error messages for internal function arguments - now correctly reports "memory" instead of "calldata"

### Cute Animal Picture

baby turtle

### DIFF
--- a/tests/unit/semantics/types/test_type_from_annotation.py
+++ b/tests/unit/semantics/types/test_type_from_annotation.py
@@ -104,3 +104,33 @@ def test_multidimensional_mapping(build_node, type_str, type_str2):
     v_t = types[type_str2]
 
     assert ann_t == HashMapT(k_t, HashMapT(k_t, v_t))
+
+
+def test_internal_function_args_use_memory_location():
+    # test that internal function arguments are validated against MEMORY location,
+    # while external function arguments are validated against CALLDATA (issue #4754)
+    #
+    # HashMap is invalid in both locations, but the error message should
+    # correctly report "memory" for internal and "calldata" for external
+    from vyper.compiler import compile_code
+    from vyper.exceptions import InstantiationException
+
+    internal_code = """
+@internal
+def _foo(x: HashMap[uint256, uint256]) -> uint256:
+    return 0
+    """
+
+    external_code = """
+@external
+def foo(x: HashMap[uint256, uint256]) -> uint256:
+    return 0
+    """
+
+    # internal function should report "memory" in error
+    with pytest.raises(InstantiationException, match="not instantiable in memory"):
+        compile_code(internal_code)
+
+    # external function should report "calldata" in error
+    with pytest.raises(InstantiationException, match="not instantiable in calldata"):
+        compile_code(external_code)
--- a/vyper/semantics/types/function.py
+++ b/vyper/semantics/types/function.py
@@ -346,7 +346,9 @@ class ContractFunctionT(VyperType):
                 "Default functions cannot appear in interfaces", funcdef
             )
 
-        positional_args, keyword_args = _parse_args(funcdef, is_interface=True)
+        positional_args, keyword_args = _parse_args(
+            funcdef, FunctionVisibility.EXTERNAL, is_interface=True
+        )
 
         return_type = _parse_return_type(funcdef)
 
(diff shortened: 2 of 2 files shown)


vyper/semantics/types/function.py
```diff
@@ -902,7 +902,9 @@ def _parse_decorators(funcdef: vy_ast.FunctionDef) -> _ParsedDecorators:
 
 
 def _parse_args(
-    funcdef: vy_ast.FunctionDef, is_interface: bool = False
+    funcdef: vy_ast.FunctionDef,
+    is_interface: bool = False,
+    visibility: FunctionVisibility = FunctionVisibility.EXTERNAL,
```
charles-cooper: it's ok to change the arg order here btw

charles-cooper (approved):
looks fine, please add a test for this

vyper/semantics/types/function.py
charles-cooper: @frederikgramkortegaard can you move visibility to be a positional arg, i'd like it to be required at the call site

charles-cooper:
sorry why did you force push? not sure it matches what i reviewed now

charles-cooper (changes requested):
unresolved comments

frederikgramkortegaard:
@charles-cooper    Force pushed to squash the merge commits into a single clean commit. The actual fix is unchanged - just cleaned up the history. Sorry for the confusion, I'll avoid it in any future contributions.

Sporarum (commented):
Looks good to me !

I quickly check, and as far as I can tell there is no type which is instantiable in calldata but not in memory, nor vice-versa

Sporarum:
> @charles-cooper Force pushed to squash the merge commits into a single clean commit. The actual fix is unchanged - just cleaned up the history. Sorry for the confusion, I'll avoid it in any future contributions.

Every PR is squashed before being put on master:
The commit name is the name of the PR.
The commit body is the "Commit message" section of the PR body, which should be a code-block correctly wrapped using `./fmt_commit_msg.py`

See for example https://github.com/vyperlang/vyper/pull/4895

Sporarum (approved):
LGTM!

tests/unit/semantics/types/test_type_from_annotation.py
```diff
@@ -104,3 +104,33 @@ def test_multidimensional_mapping(build_node, type_str, type_str2):
...
     assert ann_t == HashMapT(k_t, HashMapT(k_t, v_t))
+
+
+def test_internal_function_args_use_memory_location():
+    # test that internal function arguments are validated against MEMORY location,
+    # while external function arguments are validated against CALLDATA (issue #4754)
+    #
+    # HashMap is invalid in both locations, but the error message should
+    # correctly report "memory" for internal and "calldata" for external
+    from vyper.compiler import compile_code
```
charles-cooper, line 115: no reason to have inline imports here really

charles-cooper: oops wrote 'inline comments' originally

PR #4836, opened 2026-02-03
fix[venom]: stack param delayed initialization error

HodanPlodky:
### What I did
Fix for the error with calling convention where the param passed via stack would not be stored into the memory at the start of the internal function.

This would occur when the palloca was used for the first time further down the function. The error manifested when the palloca occured for the first time at the loop which reset the param value each iteration.

```py
@internal
def _helper(x: uint256) -> uint256:
    for i: uint256 in range(2):
        # used x for the first time so here is
        # is the first occurence of palloca
        x %= x
    return 0
```

this resulted 

```
function "internal 0 _helper(uint256)_runtime" {
  "internal 0 _helper(uint256)_runtime":  ; OUT=[7_condition]
      %2 = param                     ; x
      %3 = param                     ; return_pc
      %20 = alloca 32                ; free var 0
      %21 = alloca 32                ; free var 32
      %1 = alloca 0x100000001, 32    ; return buffer
      %10 = alloca 2, 32             ; i

      ; !!!!!!!!
      ; palloca it self was floated to entry
      %11 = palloca 1, 32            ; x (memory) 
      %4 = 0
      %6 = iszero 0
      nop                            ; assert %6
      jmp @7_condition

  7_condition:  ; OUT=[10_exit, 8_body]
      %4:1 = phi @"internal 0 _helper(uint256)_runtime", %4, @8_body, %4:2
      %8 = xor 2, %4:1
      %9 = iszero %8
      jnz %9, @10_exit, @8_body

  8_body:  ; OUT=[7_condition]
      mstore %10, %4:1

      ; !!!!!!!!
      ; setting of the param to allocated
      ; memory was not
      mstore %11, %2

      %12 = mload %11
      %14 = mload %11
      %22 = iszero %14
      %16 = iszero %22
      assert %16
      %17 = mod %12, %14
      mstore %11, %17
      %4:2 = add 1, %4:1
      jmp @7_condition

  10_exit:  ; OUT=[]
      mstore %1, 0
      %19 = mload %1
      ret %3, %19
}  ; close function "internal 0 _helper(uint256)_runtime"
```

After `Mem2Var` pass, the value was moved to stack variable but as a result of this reset in the loop the assert was checked just for the passed argument and not for the result of the modulus (the mod instruction was later even removed since the result was not used here)

Fix is done by explicitly marking this as a parameter initialization and floating it with the `palloca`. I added new instruction `paraminit` which is used for that purpose and after that it is transformed into `mstore`

### Commit message

```
This commit fixes the problem with the calling convention when the parameters are passed via stack, when this is the case the parameter is stored in memory at the first occurrence of corresponding palloca instruction. If the first occurrence is in the loop it could create error in which the value would be overwritten every iteration and not reflect changes that should have happened 

@internal
def _helper(x: uint256) -> uint256:
    for i: uint256 in range(2):
        # used x for the first time so here is
        # is the first occurence of palloca
        x %= x
    return 0

Fix is done by explicitly marking this as a parameter initialization and floating it with the `palloca`. I added new instruction `paraminit` which is used for that purpose and after that it is transformed into `mstore`
```

### Cute Animal Picture

### DIFF
--- a/tests/functional/codegen/calling_convention/test_internal_call.py
+++ b/tests/functional/codegen/calling_convention/test_internal_call.py
@@ -448,6 +448,32 @@ def test_selfcall_call_violation(failing_contract_code, assert_compile_failed):
         _ = compile_code(failing_contract_code)
 
 
+def test_mod_by_zero_reverts(get_contract, tx_failed):
+    """
+    Reproduce: mod-by-zero inside a loop in an internal function with parameters
+    should revert, but venom codegen silently returns 0.
+
+    The problem arose when the value was passed via stack and 
+    palloca was used for the first time int the loop
+    """
+
+    src = """
+@internal
+def _helper(x: uint256) -> uint256:
+    for i: uint256 in range(2):
+        x %= x  # iter1: 1%1=0, iter2: 0%0 should revert
+    return 0
+
+@external
+def foo() -> uint256:
+    self._helper(1)
+    return 42
+    """
+    c = get_contract(src)
+    with tx_failed():
+        c.foo()
+
+
 FAILING_CONTRACTS_ARGUMENT_EXCEPTION = [
     """
 # expected no args, args given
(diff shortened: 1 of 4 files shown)


harkal (commented):
no longer relevant

PR #4838, opened 2026-02-03
fix[lang]: detect duplicate imports via different relative paths

### LINKED ISSUE #4770: Modules can be imported more than once
### Version Information

* vyper Version (output of `vyper --version`): a68eb5736eacb63c98db85682c1d2c123604732b
* OS: linux
* Python Version (output of `python --version`): 3.12.11

### What's your issue about?

```
# playground/test.vy
from . import x as x1
from ..playground import x as x2
```
```
# playground/x.vy

# empty file
```

`vyper playground/test.vy` compiles without issue

### How can it be fixed?

It should raise a `vyper.exceptions.DuplicateImport`, like:
```
# playground/test.vy
from . import x as x1
from . import x as x2
```

frederikgramkortegaard:
### What I did

Fixed duplicate import detection to catch when the same module is imported via different relative paths (e.g., `from . import x` and `from ..pkg import x`).

Fixes #4770

### How I did it

The duplicate check was using the relative import path as the key, but different relative paths can resolve to the same file. Now it checks both:

1. **Fast check** (before disk I/O): Uses the relative path to catch obvious duplicates like `from . import x` twice
2. **Resolved path check** (after file load): Uses `file.resolved_path` to catch different relative paths that resolve to the same file

Added a `_check_duplicate_import` helper that uses the resolved path as the key.

### How to verify it

Create a test directory with:
```
test_duplicate/
  test.vy
  x.vy
```

`x.vy`:
```vyper
# empty module
```

`test.vy`:
```vyper
from . import x as x1
from ..test_duplicate import x as x2

@external
def foo() -> bool:
    return True
```

Before fix: compiles without error (silently allows duplicate import)
After fix: raises `DuplicateImport: x2 imported more than once!`

### Commit message

```
The duplicate import check used the relative import path as the key, but
different relative paths can resolve to the same file (e.g., `from .
import x` and `from ..playground import x`).

Use resolved path to detect duplicates. The check is performed at the
join point after `_load_import()`, which avoids duplicating the check
across the `.vy`, `.vyi`, and `.json` branches.
```

### Description for CHANGELOG

Fixed a bug where the same module could be imported multiple times using different relative paths.

### Cute Animal Picture

hammerhead

### DIFF
--- a/vyper/semantics/analysis/imports.py
+++ b/vyper/semantics/analysis/imports.py
@@ -187,7 +187,7 @@ class ImportAnalyzer:
     # raises FileNotFoundError
     def _check_duplicate_import(
         self, file: CompilerInput, node: vy_ast.VyperNode, alias: str
-    ) -> None:
+    ):
         # use resolved_path to detect duplicates, not the relative import path,
         # since different relative paths can resolve to the same file
         # (e.g., `from . import x` and `from ..pkg import x`)
--- a/vyper/semantics/analysis/imports.py
+++ b/vyper/semantics/analysis/imports.py
@@ -185,6 +185,18 @@ class ImportAnalyzer:
 
     # load an InterfaceT or ModuleInfo from an import.
     # raises FileNotFoundError
+    def _check_duplicate_import(
+        self, file: CompilerInput, node: vy_ast.VyperNode, alias: str
+    ) -> None:
+        # use resolved_path to detect duplicates, not the relative import path,
+        # since different relative paths can resolve to the same file
+        # (e.g., `from . import x` and `from ..pkg import x`)
+        resolved = file.resolved_path
+        if resolved in self.graph.imported_modules:
+            previous_import_stmt = self.graph.imported_modules[resolved]
+            raise DuplicateImport(f"{alias} imported more than once!", previous_import_stmt, node)
+        self.graph.imported_modules[resolved] = node
+
     def _load_import(
         self, node: vy_ast.VyperNode, level: int, module_str: str, alias: str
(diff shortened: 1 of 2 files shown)


Sporarum (commented):
Thanks for your contribution !

Here are a couple things I think could be improved:
1. When adding new logic, we should always have corresponding tests, in this case that would be tests that check duplicate imports really do fail
2. Do we really need the fast check ? It creates duplicate logic but I'm not sure the performance gain is that big, and it might actually hurt, since it makes the happy path slower
3. (probably out of scope for this PR) Maybe we should normalize paths in some unique way, and then use that for all path comparisons

Sporarum (commented):
Looks good to me !

vyper/semantics/analysis/imports.py
```diff
@@ -232,6 +244,9 @@ def _load_import(
             if isinstance(file, FileInput):
                 file = try_parse_abi(file)
             assert isinstance(file, JSONInput)  # mypy hint
+
+            self._check_duplicate_import(file, node, alias)
```
charles-cooper: i don't think abi files can have imports in them

frederikgramkortegaard: This code does not check for duplicate imports inside abi files, it checks if an abi file is being imported twice using different paths.

So this is a correct check

charles-cooper: hmm in this case, i think the check should be performed at the join point (right after the call to load_import) so we minimize duplicated logic

charles-cooper (approved):

PR #4841, opened 2026-02-04
fix[venom]: uninitialized mem after mem2var

HodanPlodky:
### What I did
Fixed problem with `Mem2Var` pass where if the value is read before written for the first time the semantics with memory would be to return 0. This was not reflected after the pass. To achieve this I added the initial assign to variable.

Before mem2var:

```
main:
    %x = source
    %ptr = alloca 1, 32
    jmp @loop_header
loop_header:
    %i = phi @main, %x, @loop_body, %nexti
    %cond = iszero %i
    jnz %i, @loop_body, @exit
loop_body:
    %inv = sload 0
    %y = add %i, %inv
    mstore %ptr, %y
    %nexti = add %i, 1
    jmp @loop_header
exit:
    ; if the loop body is not run
    ; then the result should be 0
    %res = mload %ptr 
    sink %res
```

After mem2var and makessa:

```
main:
    %x = source
    %ptr = alloca 1, 32

    ; by default set to zero to mimic 
    ; uninitialized memory
    %alloca_ptr_0 = 0
    jmp @loop_header
loop_header:
    %alloca_ptr_0:1 = phi @main, %alloca_ptr_0, @loop_body, %alloca_ptr_0:2
    %i:1 = phi @main, %x, @loop_body, %nexti
    %i = %i:1
    %cond = iszero %i:1
    jnz %i:1, @loop_body, @exit
loop_body:
    %inv = sload 0
    %y = add %i:1, %inv
    %alloca_ptr_0:2 = %y
    %nexti = add %i:1, 1
    jmp @loop_header
exit:
    %res = %alloca_ptr_0:1
    sink %res
```

### Commit message

Added initial assign to the `Mem2Var` pass for cases where it would be necessary to read from uninitialized memory.

### Cute Animal Picture

### DIFF
--- a/tests/unit/compiler/venom/test_mem2var.py
+++ b/tests/unit/compiler/venom/test_mem2var.py
@@ -1,7 +1,6 @@
 from tests.venom_utils import PrePostChecker
-from vyper.venom.passes import MakeSSA, Mem2Var
-
-_check_pre_post = PrePostChecker([Mem2Var, MakeSSA])
+from vyper.venom.passes import Mem2Var
+from vyper.utils import CompilerPanic
 
 
 def test_mem2var_alloca_without_initalization():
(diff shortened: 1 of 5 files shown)


charles-cooper:
is this generated by vyper code? or is the example handcrafted venom

i think that for unallocated memory, the assumption that the first read returns 0 is not true, if the allocator has given it a memory space which was previously written to already

charles-cooper:
like it should be an error to try to read alloca space before it has been written to, no?

banteg:
While reviewing `Mem2Var`, I found a separate correctness issue that this PR does not cover: #5070.

The short version is that `Mem2Var` currently checks use opcodes but not operand positions. An alloca used as the **value** being stored by `mstore` can be promoted as if it were only a local memory slot:

```venom
function main {
main:
    %a = alloca 32
    %b = alloca 32
    mstore %b, %a
    %x = mload %a
    return %b, 32
}
```

Current `MakeSSA` + `Mem2Var` turns the stored pointer value into a stack assignment:

```text
%alloca_a_0 = %a
%x = %alloca_a_0
```

and removes the original `mstore %b, %a`.

If this PR continues touching `Mem2Var`, it would be worth making the promotability check operand-position aware: `mstore` should count as a promotable use of an alloca only when the alloca is the destination/address operand, not when it is the stored value.

charles-cooper:
@banteg is this superseded by #5070 or should we resolve merge conflicts and get this in?

PR #4842, opened 2026-02-09
feat[venom]: add tail-merge pass for equivalent terminal blocks

harkal:
### What I did

I added a conservative `TailMergePass` that merges structurally equivalent halting
basic blocks to reduce code size in Venom IR, along with tests. Also refactored relabeling
to pass base to be used by SimplifyCFG and TailMerge.

### Commit message

```
Adds a conservative `TailMergePass` that merges structurally equivalent halting
basic blocks to reduce code size in Venom IR.

- implements `TailMergePass` for reachable, non-entry, halting blocks
- wires TailMergePass into O3 only
- runs SimplifyCFGPass after TailMergePass in O3 for CFG cleanup
- [important] centralize label/data-segment relabeling in IRPass and reuse from passes
- add/expand unit tests for merge and non-merge scenarios
```

### Cute Animal Picture

### DIFF
--- a/tests/unit/compiler/venom/test_tail_merge.py
+++ b/tests/unit/compiler/venom/test_tail_merge.py
@@ -94,3 +94,126 @@ def test_no_merge_for_non_halting_blocks():
     """
 
     _check_no_change(pre)
+
+
+def test_merge_identical_stop_blocks():
+    pre = """
+    main:
+        %cond = source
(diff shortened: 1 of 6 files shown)


charles-cooper (approved):
lgtm. are there plans to generalize this to other blocks? in which case i think we may want to rename it to blockmerge or something

PR #4844, opened 2026-02-11
fix[venom]: fix alloca overlap in `ConcretizeMemLoc`

harkal:
### Commit message

```
ConcretizeMemLocPass and DSE could disagree on memory liveness around invoke:
DSE may retain a store (due to unknown invoke aliasing), while MemLiveness
could treat the same alloca as dead and allow overlapping allocation.

This could overlap FREE_VAR_SPACE-backed scratch with active DynArray memory,
clobbering length words and causing runtime reverts.

Fix by conservatively marking destinations of surviving memory stores as live
in MemLiveness before allocation concretization. This aligns concretization
with the actual IR after DSE.

Add regression test for the internal-call + keccak scratch + DynArray case that
previously reverted under --experimental-codegen.
```

### Cute Animal Picture


vyper/venom/passes/concretize_mem_loc.py
HodanPlodky: This condition can be made to assert with the new change

charles-cooper: only if remove unused variables is run, but yes

charles-cooper (approved):

PR #4851, opened 2026-02-18
feat[venom]: elide full-allocation memory copies

HodanPlodky:
_co-authored by claude fable 5 (PR description update)_

### What I did
Improved memory copy elision to achieve detection of full copies between different allocations. This can at the end cause total elimination of that allocation if it is only used temporarily.

Example:

before
```
main:
        %dst = alloca 1, 64
        %ret_buf = alloca 2, 64
        invoke @fn, %ret_buf
        mcopy %dst, %ret_buf, 64
        %res = mload %dst
        sink %res
```

after
```
main:
        ; this alloca is later removed by
        ; RemoveUnusedVariablesPass
        %dst = alloca 1, 64
        %ret_buf = alloca 2, 64
        invoke @fn, %ret_buf
        nop
        %res = mload %ret_buf
        sink %res
```

### How I did it

Added a `TranslateAnalysis` dataflow analysis to `memory_copy_elision.py`. When an `mcopy` copies an entire allocation into another allocation of the same size, it records a "translation": accesses to the destination allocation may be redirected to the source, valid as long as the two buffers stay in sync.

- Translation facts propagate across basic blocks (intersection at joins) and are invalidated by intervening writes to either allocation.
- A translation is only accepted when every memory-effect use of the destination sees a live, consistent mapping and all pointers derived from the destination have statically known offsets.
- `MemoryCopyElisionPass` alternates copy elision with translation rewriting until fixpoint: destination reads/writes are redirected to the source, the establishing `mcopy` becomes a redundant self-copy and is nopped, and the dead destination alloca is later removed by `RemoveUnusedVariablesPass`.

Supporting changes:
- Volatile memory writers (`call`/`invoke`/...) now invalidate only the copy facts their write location can clobber, instead of clearing all copy state.
- Self-copies within the same allocation at equal fixed offsets are elided.
- `update_write_location`/`update_read_location` in `memory_location.py` are replaced by `write_location_idx`/`read_location_idx`, which return the operand index so rewrites go through `InstUpdater`.
- `BasePtrAnalysis` gains a reverse allocation→variables index (`vars_in_allocations`) and `new`/`new_gep` helpers to register pointers created during rewriting.

### How to verify it

- New unit tests in `tests/unit/compiler/venom/test_memory_copy_elision.py` (translation of invoke return buffers, chained copies, and negative cases where rewriting must not happen), checked for equivalence with hevm.
- Bytecode size report shows size reductions across all benchmark contracts.

### Commit message

```
memory copy elision could only remove copies that were provably
redundant at the copy site. a full copy from one allocation into another
-- e.g. copying an invoke's return buffer into a local buffer -- always
survived, keeping both the mcopy and the destination allocation in the
bytecode even when the destination was just a temporary alias of the
source.

introduce a dataflow "translation" analysis (TranslateAnalysis): when an
mcopy copies an entire allocation into another allocation of the same
size, record that accesses to the destination may be redirected to the
source, valid as long as the two buffers stay in sync. facts propagate
across basic blocks with intersection semantics at joins and are
invalidated by intervening writes to either allocation. a translation is
only accepted when every memory-effect use of the destination sees a
live, consistent mapping and all pointers derived from the destination
have statically known offsets.

the pass then alternates copy elision with translation rewriting until
fixpoint: reads and writes of the destination are redirected to the
source, the establishing mcopy becomes redundant and is nopped, and the
now-unused destination allocation can be removed by
RemoveUnusedVariablesPass.

supporting changes:

- volatile memory writers (call/invoke etc.) now invalidate only the
  copy facts their write location can clobber, instead of clearing all
  copy state
- self-copies within the same allocation at equal fixed offsets are
  elided
- update_write_location/update_read_location are replaced by
  write_location_idx/read_location_idx, which return the operand index
  so rewrites go through InstUpdater
- BasePtrAnalysis gains a reverse allocation->variables index and
  helpers to register pointers created during rewriting
```

Improve the Venom memory copy elision pass to detect full copies between allocations and redirect uses of the destination to the source, allowing temporary buffers (e.g. internal call return buffers) to be eliminated entirely.

### Cute Animal Picture


vyper/venom/passes/memory_copy_elision.py
```diff
@@ -22,14 +27,17 @@
 
 # Type alias for copy tracking: maps memory location to the copy instruction
 CopyMap = dict[MemoryLocation, IRInstruction]
+TranslateMap = dict[Allocation, tuple[Allocation, bool]]
```
charles-cooper: comment explaining what it is

vyper/venom/passes/memory_copy_elision.py
```diff
@@ -22,14 +27,17 @@
...
 
 
 class MemoryCopyElisionPass(IRPass):
     base_ptr: BasePtrAnalysis
     copies: CopyMap
+    total_translation: TranslateMap
```
charles-cooper: what is a total_translation?

HodanPlodky: Added explanation comment

charles-cooper:
@HodanPlodky please check conflicts from @banteg's PR storm (guessing https://github.com/vyperlang/vyper/issues/5041)

charles-cooper:
@HodanPlodky looks like quite substantial gas regressions?

charles-cooper (changes requested):
please investigate gas regressions

harkal:
Fable review at 3e44eedc8. The approach looks good, but I think there are two miscompiles and a crash here, plus some cleanups.

### Correctness

**1. The new redundant-copy elision can nop a real copy when offsets are unknown** (`memory_copy_elision.py:292`)

```python
if not write_loc.is_concrete and write_loc == read_loc:
    self.updater.nop(inst)
```

`MemoryLocation` equality is structural, and distinct dynamic pointers into the same alloca all resolve to `MemoryLocation(offset=None, size=N, alloca=buf)`:

```
%buf = alloca ...
%p1 = add %buf, %i   ; dynamic i
%p2 = add %buf, %j   ; dynamic j
mcopy %p2, %p1, 64   ; write_loc == read_loc -> nop'd, but it moves bytes
```

Needs an `offset is not None` guard.

**2. Calls no longer clear the storage/transient load caches** (`memory_copy_elision.py:188`)

`call` has MEMORY in its write effects, so it now takes the new `elif _volatile_memory(inst)` branch, which only invalidates `loads[Effects.MEMORY]`. Previously it fell through to the `else` branch, which also cleared `loads[STORAGE]` / `loads[TRANSIENT]`. So in

```
%v = sload %slot
call ...             ; may reenter and write %slot
sstore %slot, %v
```

the store-back gets elided. (Side effect of the same restructure: the `if Effects.MEMORY in ...` block in the `else` branch is now unreachable, since every MEMORY-writer is caught by the elif first.)

Related but weaker: for `dload` the write location is the placeholder `MemoryLocation(offset=0, size=32)` (the `TODO: use FreeVarSpace` one), so the targeted `_invalidate` only kills copies overlapping `[0, 32)` where the old code cleared everything.

**3. `alloca_size` assertion crash on dynamic allocations** (`memory_copy_elision.py:548,551`)

`_try_create_translate` guards on `is_concrete` / offset / literal size but not `is_dynamic`, and `BasePtrAnalysis` tracks `dalloca`/`bump` allocations too. A full-size literal `mcopy` between dynamic buffers reaches `read_loc.alloca.alloca_size`, which asserts `opcode == "alloca"`. Other consumers guard this (`copy_forwarding.py`, `mem_liveness.py` check `is_dynamic` first). The pass runs both before and after `FmpLoweringPass`, so both `dalloca` and `bump` can reach it.

**4. Stale DFG across outer loop iterations** (`memory_copy_elision.py:353,385`)

`update_read_location` / `update_write_location` mutate `inst.operands` directly without updating DFG use-lists, and `DFGAnalysis` is only invalidated after the `while True` loop exits. So `force_analysis(TranslateAnalysis)` on the next iteration validates "all uses agree" against use-lists missing the instructions rewritten in the previous iteration. A missing use on a src-derived var could make the check wrongly pass if that source allocation later becomes a translate dst itself. Routing the rewrite through `self.updater` (as the inserted `add` already is) would close this.

### Cleanups

- `_use_dominates` (`:315`) has no callers; `self.dom` and the `DominatorTreeAnalysis` import exist only to feed it, so every `run_pass` computes a dominator tree for nothing.
- `checked` (`:417`) is never added to, so `if dst in checked: continue` is inert and the per-dst validation reruns once per instruction map containing that dst. Also `if not all_ok: break` (`:439`) exits the whole map — did you mean `continue` (skip just this dst)?
- `_try_update_from_translates_read/_write` + `_update_base_allocation_read/_write` are read/write mirrors differing only in the location accessor pair, and the inner pair re-checks `alloca not in translates` which the caller already guarantees. One helper parameterized by the accessors would halve it.
- `_merge_translates` (`:499`) duplicates the shape of `_merge_copies` (only the per-key equality differs) — could share a merge helper.
- In `base_ptr_analysis.py`, `new` is only reachable via `new_gep`, and both re-inline the `vars_in_allocations` insert idiom that `analyze()` also has (`setdefault` would collapse all three).


---
*🤖 Review by [Claude Code](https://claude.com/claude-code) (Claude Fable 5)*

vyper/venom/passes/memory_copy_elision.py
```diff
@@ -256,3 +311,251 @@ def _try_elide_load_store(self, inst: IRInstruction, write_loc: MemoryLocation,
...
+        return self._update_base_allocation_read(inst, read_loc, translates)
+
+    def _update_base_allocation_read(
+        self, inst: IRInstruction, read_loc: MemoryLocation, translates: TranslateMap
+    ):
+        if read_loc.alloca not in translates:
+            return False
+
+        if read_loc.offset is None:
+            return False
```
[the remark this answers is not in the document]
HodanPlodky: both of the `dst` and `src` must have backing buffer with same size so this would be error any way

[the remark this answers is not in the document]
HodanPlodky: but I added sanity check for that

harkal:
Thanks for the quick fixes — re-reviewed at 3225e42d7. Items 1–4 from my previous comment all look correctly resolved (the `is_fixed` guard, the restored `else` invalidation, the `is_dynamic` check, and routing rewrites through the updater). A few leftovers:

**New bug from the `01fa52a45` refactor:** `update_read_location` (`memory_location.py:347`) calls `write_location_idx` instead of `read_location_idx` — copy-paste slip. Both `update_read_location` and `update_write_location` now have no callers (the pass uses the `*_idx` functions directly), so this is currently dead — but as-is, calling it on an `mcopy` would silently overwrite the dst operand. Suggest just deleting both wrappers, or fixing the call if you want to keep them.

**Leftovers from the previous round:**
- `self.dom = self.analyses_cache.request_analysis(DominatorTreeAnalysis)` (`memory_copy_elision.py:61`) has no readers now that `_use_dominates` is gone — every `run_pass` still computes a dominator tree for nothing. The import can go too.
- `checked` (`:407`) is still never added to, so the `if dst in checked: continue` guard is inert and per-dst validation reruns once per instruction map containing that dst.
- The `dload` point still stands: its write location is the placeholder `MemoryLocation(0, 32)`, which is `is_fixed`, so the targeted `_invalidate` only kills copies overlapping `[0, 32)` where the pre-PR code cleared everything. Until the `TODO: use FreeVarSpace` is resolved, it may be safer to treat `dload`/`dloadbytes` writes as non-fixed (full clear).


---
*🤖 Review by [Claude Code](https://claude.com/claude-code) (Claude Fable 5)*

harkal:
should we bail if pointer escapes to memory or something?

HodanPlodky:
> should we bail if pointer escapes to memory or something?

I dont think that it should bail it that would happen it would not be correctly concretize. If anything it should assert, though I dont know it should be done in this pass may be it would be good idea to do it in cocretization

harkal (approved):

charles-cooper (approved):

PR #4859, opened 2026-02-26
fix[venom]: venom to asm stack order target

HodanPlodky:
### What I did
Problem is created when the liveness order for the input into basicblock is different from two different predecessors. In this case it occurs in the `6_for_cond`. The liveness is used in the `3_match_0dff1a19` and `7_for_body` as a stack target.

This is only problem for the phi instruction because of which there is a possibility to get different liveness for variable which should be at the end in the same stack index.

```
function runtime {

  ...

  3_match_0dff1a19:  ; OUT=[6_for_cond]
      ...
      %9 = sload 1
      ...
      %11 = 0
      jmp @6_for_cond
    ; (runtime)


  6_for_cond:  ; OUT=[7_for_body, 9_for_exit]
      %30 = phi @3_match_0dff1a19, %9, @7_for_body, %21
      %11:1:1 = phi @3_match_0dff1a19, %11, @7_for_body, %11:2
      %29 = %9
      %35 = xor 1, %11:1:1
      jnz %35, @7_for_body, @9_for_exit
    ; (runtime)


  7_for_body:  ; OUT=[6_for_cond]
      ...
      %21 = %19
      ...
      %23 = add 1, %11:1:1
      %11:2 = %23
      jmp @6_for_cond
    ; (runtime)

  ...

}  ; close function runtime
```

First stack input:
```6_for_cond <StackModel: [%9, %11]>```

Second stack input:
```6_for_cond <StackModel: [%11:2, %21]>```

To fix that I forced to add the input vars for `phi` instruction to the end of the order.

### How I did it

### How to verify it

### Commit message

```
input_vars_from() computed incoming liveness across phi nodes by                                                                                                                                                                                                                                             
iterating phis sequentially and mutating the set — adding the
source-matching operand and removing non-matching ones. when two                                                                                                                                                                                                                                             
phis shared operands or when the liveness set contained a variable                                                                                                                                                                                                                                           
at a "canonical" position that should have been substituted, the                                                                                                                                                                                                                                             
removal order could drop live variables or leave stale ones,                                                                                                                                                                                                                                                 
producing an incorrect stack layout during assembly lowering.                                                                                                                                                                                                                                                
                                                                                                                                                                                                                                                                                                               
replace the incremental mutation with a positional substitution:                                                                                                                                                                                                                                             
build a map from every phi operand to its phi index, then walk                                                                                                                                                                                                                                               
the liveness set once, replacing phi-related entries with the                                                                                                                                                                                                                                                
source-matching operand (deduplicated by phi index) and passing                                                                                                                                                                                                                                              
non-phi entries through unchanged.
```

### Cute Animal Picture

![Put a link to a cute animal picture inside the parenthesis-->]()

### CODE REVIEW

File: tests/functional/codegen/calling_convention/test_external_contract_calls.py [hunk 1]
```diff
@@ -2629,3 +2629,26 @@ def boo() -> uint256:
 
     assert c.foo() == [1, 2, 3, 4]
     assert c.bar() == [1, 2, 3, 4]
+
+
+def test_stack_order_liveness_dependence_error(get_contract):
```
harkal: Why was this put into this file? Should we move it to `tests/unit/compiler/venom/test_stack_order.py` for example?

vyper/venom/analysis/liveness.py
```diff
@@ -113,6 +113,9 @@ def input_vars_from(self, source: IRBasicBlock, target: IRBasicBlock) -> Ordered
 
                 for label, var in inst.phi_operands:
                     if label == source.label:
+                        # to force order
```
harkal:
```suggestion
                        # Force phi operand to end of the OrderedSet. If `var` is already                                                                                      
                        # in liveness (from a non-phi use), add() is a no-op and it stays                                                                                      
                        # at its original position. This can cause predecessors of a join
                        # point to disagree on stack layout, since each predecessor                                                                                            
                        # substitutes different phi operands but expects non-phi vars at                                                                                       
                        # the same depths. Moving phi operands to the end ensures a                                                                                            
                        # consistent layout across all predecessors.
```

charles-cooper (approved):

PR #4860, opened 2026-02-26
refactor[venom]: normalize bytestring memory handling and dedupe hash/slice paths

harkal:
### What I did

Reduced duplicated normalization logic and make direct-to-venom behavior easier to reason about and maintain, while preserving semantics. Also fixes a latent `CompilerPanic` when `DataLocation` is compared with `None` via `==`/`in`.

### How I did it

- `ensure_bytestring_in_memory()`: centralizes the "copy from STORAGE/TRANSIENT/CODE into a memory buffer" pattern that was duplicated 4× across hashing.py and expr.py. Uses `is` identity checks to avoid `StringEnum.__eq__` panic when `location` is `None`.
- `_prepare_hash_input()`: shared input normalization for keccak256/sha256, handling both bytestring and fixed-size (bytesM/bytes32) cases.
- `_with_byte_offset()`: replaces ~8 copies of the 6-line `if offset == 0 / elif isinstance(base, IRLiteral) / else add` pattern in copy_memory, code_to_memory, load_immutable, load_immutable_ctor, store_immutable. Also elides dead `add 0` instructions.
- `_assert_slice_bounds()`: extracts the 4× duplicated overflow-safe bounds check from slice lowering paths.

### How to verify it

Existing test suite — pure refactor with no semantic changes (except the `is` vs `==` bugfix).

### Commit message

```
refactor[venom]: normalize bytestring memory handling and dedupe
hash/slice paths

the direct-to-venom codegen had four independent copies of the pattern
"if bytestring is in STORAGE/TRANSIENT/CODE, copy it to a memory buffer
first". keccak256 and sha256 each duplicated this in hashing.py, and
expr.py had two more copies in `_lower_keccak256_key` and
`_get_bytestring_hash` (used for mapping key hashing and bytestring
comparison). the slice builtin similarly had four identical copies of
the bounds-check sequence `start + length <= src_len` with overflow
protection.

separately, the offset arithmetic in `copy_memory`, `code_to_memory`,
`load_immutable`, `load_immutable_ctor`, and `store_immutable` each
manually special-cased `offset == 0` and `isinstance(base, IRLiteral)`
with 6-line if/elif/else blocks, repeated ~8 times.

add three helpers to eliminate the duplication:
- `ensure_bytestring_in_memory()`: centralizes STORAGE/TRANSIENT/CODE →
  memory copy; returns the value unchanged if already in memory
- `_prepare_hash_input()`: shared input normalization for keccak256 and
  sha256, handling both bytestring and fixed-size (bytesM/bytes32) cases
- `_with_byte_offset()`: compile-time offset addition that preserves
  IRLiteral folding and elides dead `add 0` instructions
- `_assert_slice_bounds()`: the overflow-safe bounds check extracted
  from slice lowering

also fixes a latent bug: expr.py used `vv.location in (DataLocation.X,
...)` which invokes `StringEnum.__eq__` — this panics with
`CompilerPanic("bad comparison")` when `location` is `None` (the
representation for memory-resident values). the new code uses identity
checks (`is`) throughout, which is safe for enum singletons.
```

refactor bytestring memory normalization and deduplicate hash/slice codegen paths

### Cute Animal Picture

### DIFF
--- a/vyper/codegen_venom/context.py
+++ b/vyper/codegen_venom/context.py
@@ -194,7 +194,7 @@ class VenomCodegenContext:
             DataLocation.TRANSIENT,
         ), f"bytes_data_ptr expects MEMORY, STORAGE, or TRANSIENT, got {loc}"
         word_scale = 32 if loc is None or loc == DataLocation.MEMORY else 1
-        return self.builder.add(vv.operand, IRLiteral(word_scale))
+        return self._with_byte_offset(vv.operand, word_scale)
 
     def bytestring_length(self, vv: VyperValue) -> IROperand:
         """Get length of bytestring from its pointer.
@@ -212,6 +212,24 @@ class VenomCodegenContext:
         if loc is None:
             return self.builder.mload(vv.operand)
         return self.builder.load(vv.operand, loc)
+
+    def ensure_bytestring_in_memory(self, vv: VyperValue, typ: _BytestringT) -> VyperValue:
+        """Return a bytestring value guaranteed to be in memory.
+
+        Hashing and certain builtins require memory-backed bytestring data.
+        STORAGE/TRANSIENT/CODE values are copied into a temporary memory buffer.
+        """
+        if vv.location is DataLocation.STORAGE or vv.location is DataLocation.TRANSIENT:
+            buf_val = self.new_temporary_value(typ)
+            self.slot_to_memory(vv.operand, buf_val.operand, typ.storage_size_in_words, vv.location)
+            return buf_val
+
+        if vv.location is DataLocation.CODE:
+            buf_val = self.new_temporary_value(typ)
+            self.code_to_memory(vv.operand, buf_val.operand, typ.storage_size_in_words)
+            return buf_val
+
+        return vv
 
     def is_constant(self) -> bool:
         """Check if in constant (view) context or range expression."""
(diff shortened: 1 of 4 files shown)


charles-cooper (approved):

charles-cooper (commented):
clean refactor, lgtm. one minor suggestion inline.

vyper/codegen_venom/context.py
```diff
@@ -213,6 +213,24 @@ def bytestring_length(self, vv: VyperValue) -> IROperand:
...
+            buf_val = self.new_temporary_value(typ)
+            self.slot_to_memory(vv.operand, buf_val.operand, typ.storage_size_in_words, vv.location)
+            return buf_val
+
+        if vv.location is DataLocation.CODE:
+            buf_val = self.new_temporary_value(typ)
+            self.code_to_memory(vv.operand, buf_val.operand, typ.storage_size_in_words)
+            return buf_val
+
+        return vv
```
charles-cooper, line 233: CALLDATA-located bytestrings silently pass through here unchanged, then hit the assert in `bytes_data_ptr` downstream. In practice unreachable (calldata bytestrings are copied to memory during param loading), but worth making the invariant explicit:

```suggestion
        assert vv.location is None or vv.location is DataLocation.MEMORY
        return vv
```

charles-cooper (commented):
clean refactor, lgtm. one minor suggestion inline.

PR #4862, opened 2026-02-27
feat[tool]: add `--disable-static-exceptions` flag

cyberthirst:
### Commit message

```
Add a CLI flag to suppress compile-time `StaticAssertionException` for
provably failing assertions, letting them revert at runtime instead.
This is useful for fuzzer-based contract generation, where the fuzzer
frequently produces contracts with statically failing assertions, which
dwarfs throughput on actually interesting contracts.
```

### Cute Animal Picture


charles-cooper:
if we go with this, we should just call it like `--no-static-assert`

HodanPlodky (approved):

PR #4866, opened 2026-03-01
fix[venom]: proper array subscript lowering

harkal:
### What I did

Fixes miscompile: 
```
arr: immutable(DynArray[uint256, 1])

@deploy
def __init__(arg: uint256):
    x: uint256 = arr[0]
    arr = []
```

Expected behavior:
- arr is uninitialized at that read site (length should be `0`)
- arr[0] should always fail bounds check and revert, independent of constructor arg value and codegen pipeline

### Commit message

```
Three bugs caused immutable dynarray reads to miscompile in the venom
codegen pipeline:

1. `is_word_type` treated single-element arrays (`uint256[1]`) as word-
   sized, causing them to be passed on the stack instead of through
   memory. Fix by checking the type is actually a primitive word type,
   not just that its size fits in a word.

2. `DataLocation.CODE` was overloaded in constructor context to mean
   both "constructor args in the data section" (needs `dload`) and
   "immutables staging area" (needs `iload`). When reading an immutable
   dynarray during the constructor, `builder.load(addr, CODE)` emitted
   `dload` — reading from constructor arg bytes instead of the
   immutables buffer. Fix by adding `DataLocation.IMMUTABLES` as a
   distinct location. Immutable variables now use IMMUTABLES everywhere;
   `load_word` emits `iload` in constructor context and `dload` at
   runtime. `CODE` is reserved for constructor args only.

3. Ternary expressions (`x if cond else y`) with complex types lost
   their location tracking, producing `VyperValue`s with `location=None`
   that downstream code couldn't load correctly. Fix by having
   `lower_IfExp` return `from_ptr(MEMORY)` for complex types so the
   location propagates.

Also clean up duplicated copy routines: collapse
`load_immutable`/`load_immutable_ctor` into `load_immutable_to_memory`,
replace `code_to_memory` with generic `copy_to_memory`, rename
`load_storage`/`load_transient` to `load_*_to_memory` for consistency,
and dedup the iter-loop element copy logic.
```

### Cute Animal Picture

### DIFF
--- a/vyper/codegen_venom/context.py
+++ b/vyper/codegen_venom/context.py
@@ -454,8 +454,6 @@ class VenomCodegenContext:
         success = b.staticcall(b.gas(), IRLiteral(IDENTITY_PRECOMPILE), src, length, dst, length)
         b.assert_(success)
 
-
-
     def load_calldata(self, offset: IROperand, typ: VyperType) -> IROperand:
         """Load from calldata.
 
@@ -569,9 +567,15 @@ class VenomCodegenContext:
         """
         # TODO: refactor — DataLocation should have word_scale/word_addressable
         # properties (like AddrSpace does) instead of hardcoding this.
-        _byte_addressed = (DataLocation.MEMORY, DataLocation.CODE, DataLocation.IMMUTABLES, DataLocation.CALLDATA)
-        assert location in _byte_addressed, \
-            f"copy_to_memory: expected byte-addressed location, got {location}"
+        _byte_addressed = (
+            DataLocation.MEMORY,
+            DataLocation.CODE,
+            DataLocation.IMMUTABLES,
+            DataLocation.CALLDATA,
+        )
+        assert (
+            location in _byte_addressed
+        ), f"copy_to_memory: expected byte-addressed location, got {location}"
         for i in range(0, size, 32):
             src_ptr = self._with_byte_offset(src, i)
             dst_ptr = self._with_byte_offset(dst, i)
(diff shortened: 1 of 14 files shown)


vyper/codegen_venom/context.py
```diff
@@ -208,6 +208,44 @@ def bytestring_length(self, vv: VyperValue) -> IROperand:
             return self.builder.mload(vv.operand)
         return self.builder.load(vv.operand, loc)
 
+    def value_location(
+        self, vv: VyperValue, fallback: Optional[DataLocation] = None
+    ) -> DataLocation:
+        """Resolve the effective location for a VyperValue.
+
+        UNSET/None locations are normalized to MEMORY.
```
charles-cooper: this looks sketchy, why?

vyper/codegen_venom/context.py
```diff
@@ -447,6 +470,35 @@ def copy_memory_dynamic(self, dst: IROperand, src: IROperand, length: IROperand)
...
+        if src_location == DataLocation.MEMORY:
+            self.copy_memory(dst, src, size)
+            return
+
+        if src_location == DataLocation.CALLDATA:
+            self.builder.calldatacopy(dst, src, IRLiteral(size))
+            return
+
+        if src_location == DataLocation.CODE:
+            self.code_to_memory(src, dst, size // 32)
```
charles-cooper: what?

charles-cooper (approved):

PR #4868, opened 2026-03-01
fix[venom]: layout-aware memory copy for type-compatible assignments

harkal:
### What I did

Fixed a Venom codegen bug where copies between type-compatible values with different
in-memory layouts were treated as flat blobs.

The original failure mode was copying a value like `DynArray[Bytes[540], 9]` into
`DynArray[Bytes[704], 13]`. Venom used a plain memory copy based on the destination
layout, so every element after the first was read from the wrong offset. That caused
silent memory corruption in nested dynamic data.

This branch also fixes runtime iteration over immutable dynamic arrays with multi-word
elements. Immutables are byte-addressed code reads at runtime, so those loop values now
get materialized correctly before any layout-aware repacking.

### Commit message

```
fix venom typed copies for layout-mismatched compatible types
when copying between compatible values with different in-memory layouts
(for example `DynArray[Bytes[540], 9]` into
`DynArray[Bytes[704], 13]`), venom treated the source as a flat blob
and copied using the destination layout. that misaligned elements after
the first and corrupted nested dynamic data.

fix this by threading source type information through venom memory
stores. when source and destination layouts differ, dispatch to a
recursive typed copy that walks the value by shape
(dynarray, sarray, tuple, struct, bytestring, prim word) with
independent src/dst offsets. dynarrays with matching element layouts
keep a bulk-copy fast path; mismatched layouts use an IR loop.

also fix runtime iteration over immutable dynarrays with multi-word
elements by materializing byte-addressed non-memory loop values before
typed repacking, add targeted regression coverage, and clean up the
common unwrap/store plumbing with a small helper.
```

### Cute Animal Picture

### DIFF
--- a/vyper/codegen_venom/stmt.py
+++ b/vyper/codegen_venom/stmt.py
@@ -131,17 +131,18 @@ class Stmt:
         no aliasing is possible and the staging copy is skipped.
         """
         src_loc = src_vv.location  # None for stack values, else DataLocation
+        src_typ = src_vv.typ
         src = self.ctx.unwrap(src_vv)  # always a memory ptr for complex types
 
         if src_loc is DataLocation.MEMORY and dst_ptr.location is DataLocation.MEMORY:
             # Both in memory — stage through temp for overlap safety.
-            tmp_val = self.ctx.new_temporary_value(typ)
-            self.ctx.copy_memory(tmp_val.operand, src, typ.memory_bytes_required)
+            tmp_val = self.ctx.new_temporary_value(src_typ)
+            self.ctx.copy_memory(tmp_val.operand, src, src_typ.memory_bytes_required)
             src = tmp_val.operand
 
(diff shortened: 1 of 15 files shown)


vyper/codegen_venom/stmt.py
```diff
@@ -108,21 +108,47 @@ def lower_Assign(self) -> None:
...
+    ) -> None:
         """Assign a VyperValue to a destination pointer.
 
         Handles both primitive word types (direct store) and complex types
         (with overlap-safe copying when source and dest are in the same
         address space).
         """
+        if isinstance(typ, _BytestringT) and self._is_empty_value(src_node):
+            # Empty bytes/string assignment only needs a zero length word.
+            if dst_ptr.location == DataLocation.STORAGE:
```
charles-cooper: there are more high level ways of doing this, like just ptr_store (as in the last branch?)

charles-cooper (approved):

PR #4869, opened 2026-03-02
refactor[venom]: move more opt to venom

harkal:
### What I did

Move frontend AST-level optimizations into Venom IR passes, simplifying the codegen and making optimizations more composable.

### How I did it

- selfbalance/codesize: Remove self.balance and self.codesize special cases from expr.py. The codegen now emits balance(address()) and extcodesize(address()) uniformly. AlgebraicOptimizationPass folds these to selfbalance() and codesize().
- Alias analysis: Remove _get_root_variable and _potential_overlap_ast from the frontend. For dynarray append, always stage non-primitive elements through a temp buffer. For assignments, stage whenever both src and dst are in memory. MemoryCopyElisionPass elides redundant copies.
- Pass ordering: Add MemoryCopyElisionPass + LoadElimination to O2/Os (new) and O3 (LoadElimination only, after existing MemoryCopyElisionPass) to clean up the more conservative staging.

### Commit message

```
the venom codegen frontend had several AST-level optimizations that are
better expressed as IR passes: `self.balance` and `self.codesize` were
special-cased to emit `selfbalance`/`codesize` directly, and
`_potential_overlap_ast`/`_get_root_variable` performed fragile AST-
level alias analysis to decide when to stage complex copies through
temporaries.

move the `self.balance`/`self.codesize` folding into
`AlgebraicOptimizationPass` as `balance(address()) -> selfbalance()` and
`extcodesize(address()) -> codesize()`. for dynarray append, always
stage non-primitive elements through a temp buffer instead of guessing
aliasing from AST root variables -- `MemoryCopyElisionPass` elides the
copy when safe. for assignments, simplify the staging condition to
always stage when both src and dst are in memory, removing the AST-level
`_get_root_variable` check.

add `MemoryCopyElisionPass` + `LoadElimination` to O2/Os/O3 pass lists
(after `MemMergePass`, before `LowerDloadPass`) to compensate for the
more conservative staging.
```

### Cute Animal Picture

### DIFF
--- a/vyper/venom/passes/algebraic_optimization.py
+++ b/vyper/venom/passes/algebraic_optimization.py
@@ -153,15 +153,19 @@ class AlgebraicOptimizationPass(IRPass):
             operands = inst.operands
 
         # balance(address()) → selfbalance()
-        # extcodesize(address()) → codesize()
-        if inst.opcode in ("balance", "extcodesize"):
+        # note: extcodesize(address()) → codesize() is NOT safe because
+        # during initcode EXTCODESIZE(ADDRESS()) returns 0 while CODESIZE
+        # returns the initcode length.
+        if inst.opcode == "balance":
             op = operands[0]
             if isinstance(op, IRVariable):
                 producer = self.dfg.get_producing_instruction(op)
                 if producer is not None and producer.opcode == "address":
-                    new_op = "selfbalance" if inst.opcode == "balance" else "codesize"
-                    self.updater.update(inst, new_op, [])
+                    self.updater.update(inst, "selfbalance", [])
                     return
+            return
+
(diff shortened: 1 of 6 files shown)


vyper/venom/passes/algebraic_optimization.py
```diff
@@ -152,6 +152,18 @@ def _handle_inst_peephole(self, inst: IRInstruction):
             inst.flip()
             operands = inst.operands
 
+        # balance(address()) → selfbalance()
```
charles-cooper, line 155: these don't feel like algebraic optimizations, maybe we should have a new pass like evm_peepholes

charles-cooper (approved):

PR #4878, opened 2026-03-09
chore[docs]: fix module/contract/interface terminology

lufa23:
### What I did
Fixed inconsistent use of "module", "contract", and "interface" throughout the docs, based on feedback from @Sporarum. Changes are consistent with the direction of #4877.

### How I did it
Replaced incorrect uses of "contract" with "module" where the text referred to source files. Fixed the inversion pattern that described "any contract is a valid module" — the correct relationship is that the entry-point module becomes a contract upon deployment.

### How to verify it
Read the changed files and check terminology is consistent with the module/contract/interface distinctions.
### Commit message

```
chore[docs]: fix module/contract/interface terminology
update docs to consistently distinguish between modules (source files),
contracts (deployed artifacts), and interfaces. fixes incorrect
inversion pattern which described any contract as a valid module. the
correct relationship is that the entry-point module becomes a contract
upon deployment.
```

### DIFF
--- a/docs/compiler-exceptions.rst
+++ b/docs/compiler-exceptions.rst
@@ -5,7 +5,7 @@ Compiler Exceptions
 
 .. _exceptions-common:
 
-Vyper raises one or more of the following exceptions when an issue is encountered while compiling a contract.
+Vyper raises one or more of the following exceptions when an issue is encountered while compiling a module.
 
 Whenever possible, exceptions include a source highlight displaying the location
 of the error within the code:
@@ -36,7 +36,7 @@ of the error within the code:
 
 .. py:exception:: EvmVersionException
 
-    Raises when a contract contains an action that cannot be performed with the active EVM ruleset.
+    Raises when a module contains an action that cannot be performed with the active EVM ruleset.
 
 .. py:exception:: FunctionDeclarationException
 
@@ -120,7 +120,7 @@ of the error within the code:
 
 .. py:exception:: NatSpecSyntaxException
 
-    Raises when a contract contains an invalid :ref:`NatSpec<natspec>` docstring.
+    Raises when a module contains an invalid :ref:`NatSpec<natspec>` docstring.
 
     .. code-block:: python
 
@@ -208,7 +208,7 @@ of the error within the code:
 
 .. py:exception:: VersionException
 
-    Raises when a contract version string is malformed or incompatible with the current compiler version.
+    Raises when a module version string is malformed or incompatible with the current compiler version.
 
 .. py:exception:: ZeroDivisionException
 
--- a/docs/compiler-exceptions.rst
+++ b/docs/compiler-exceptions.rst
@@ -5,7 +5,7 @@ Compiler Exceptions
 
 .. _exceptions-common:
 
-Vyper raises one or more of the following exceptions when an issue is encountered while compiling a module.
+Vyper raises one or more of the following exceptions when an issue is encountered while compiling a contract.
 
 Whenever possible, exceptions include a source highlight displaying the location
 of the error within the code:
@@ -36,7 +36,7 @@ of the error within the code:
 
 .. py:exception:: EvmVersionException
 
-    Raises when a module contains an action that cannot be performed with the active EVM ruleset.
+    Raises when a contract contains an action that cannot be performed with the active EVM ruleset.
 
 .. py:exception:: FunctionDeclarationException
 
@@ -120,7 +120,7 @@ of the error within the code:
 
 .. py:exception:: NatSpecSyntaxException
 
-    Raises when a module contains an invalid :ref:`NatSpec<natspec>` docstring.
+    Raises when a contract contains an invalid :ref:`NatSpec<natspec>` docstring.
 
     .. code-block:: python
 
@@ -208,7 +208,7 @@ of the error within the code:
 
 .. py:exception:: VersionException
 
-    Raises when a module version string is malformed or incompatible with the current compiler version.
+    Raises when a contract version string is malformed or incompatible with the current compiler version.
 
 .. py:exception:: ZeroDivisionException
 
--- a/docs/interfaces.rst
+++ b/docs/interfaces.rst
@@ -8,7 +8,7 @@ An interface is a set of function definitions used to enable communication betwe
 Declaring and using Interfaces
 ==============================
 
-Interfaces can be added to contracts either through inline definition, or by importing them from a separate file.
+Interfaces can be added to a module either through inline definition, or by importing them from a separate file.
 
 The ``interface`` keyword is used to define an inline external interface:
 
--- a/docs/interfaces.rst
+++ b/docs/interfaces.rst
@@ -110,7 +110,7 @@ You can define an interface for your module with the ``implements`` statement:
     implements: FooBarInterface
 
 
-This imports the defined interface from the vyper file at ``an_interface.vyi`` (or ``an_interface.json`` if using ABI json interface type) and ensures your current contract implements all the necessary external functions. If any interface functions are not included in the contract, it will fail to compile. This is especially useful when developing contracts around well-defined standards such as ERC20.
+This imports the defined interface from the vyper file at ``an_interface.vyi`` (or ``an_interface.json`` if using ABI json interface type) and ensures your current module implements all the necessary external functions. If any interface functions are not included in the module, it will fail to compile. This is especially useful when developing contracts around well-defined standards such as ERC20.
 
 Multiple ``implements`` statements can be grouped into one:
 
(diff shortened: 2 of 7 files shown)


docs/structure-of-a-contract.rst
```diff
@@ -3,9 +3,9 @@
 Structure of a Contract
 #######################
 
-Vyper contracts are contained within files. Each file contains exactly one contract.
+Each Vyper source file contains exactly one module.
```
Sporarum, line 6: Missing explanation of what a contract is

The one from using-modules.rst is very good, I think its fine to reuse it
(might or might not require being adapted)

pcaversaccio (commented):
hmmmmmm, I'm not sure replacing "contract" with "module" everywhere improves the clarity of the docs tbh (reading the docs confused me here since what is a module really?; is it compileable? what about non-fully-to-bytecode-compileable module contracts? etc). While the terminology might be technically correct (up for debate), most devs still reason about deployed contracts rather than compilation modules. It's just a fact. Introducing "module" throughout the docs also raises the question of _how_ it differs from a contract, and I don't see that concept clearly defined before it's used. This could make the docs harder to follow for users coming from Solidity or the broader Ethereum ecosystem tbh where "contract" is the standard term. It might be better to either keep "contract" in user-facing explanations or introduce a clear explanation of the module vs contract distinction before adopting the new terminology throughout the docs.

docs/compiler-exceptions.rst
```diff
@@ -5,7 +5,7 @@ Compiler Exceptions
 
 .. _exceptions-common:
 
-Vyper raises one or more of the following exceptions when an issue is encountered while compiling a contract.
+Vyper raises one or more of the following exceptions when an issue is encountered while compiling a module.
```
Sporarum: I think all these still make more sense with contract

There's not really a meaning to compiling a single module
(other than it becoming the contract)

docs/interfaces.rst
```diff
@@ -110,7 +110,7 @@ You can define an interface for your contract with the ``implements`` statement:
     implements: FooBarInterface
 
 
-This imports the defined interface from the vyper file at ``an_interface.vyi`` (or ``an_interface.json`` if using ABI json interface type) and ensures your current contract implements all the necessary external functions. If any interface functions are not included in the contract, it will fail to compile. This is especially useful when developing contracts around well-defined standards such as ERC20.
+This imports the defined interface from the vyper file at ``an_interface.vyi`` (or ``an_interface.json`` if using ABI json interface type) and ensures your current module implements all the necessary external functions. If any interface functions are not included in the module, it will fail to compile. This is especially useful when developing contracts around well-defined standards such as ERC20.
```
Sporarum, line 113: We should make it clear that `implements` is only ever useful when used from a root-module, since other modules can chose whether to `exports` external methods

docs/structure-of-a-contract.rst
```diff
@@ -36,12 +36,12 @@ The following declaration is equivalent, and, prior to 0.3.10, was the only supp
...
 
 
-In the above examples, the contract will only compile with Vyper versions ``0.4.x``.
+In the above examples, the module will only compile with Vyper versions ``0.4.x``.
 
 Optimization Mode
 -----------------
 
-The optimization mode can be one of ``"none"``, ``"codesize"``, or ``"gas"`` (default). For example, adding the following line to a contract will cause it to try to optimize for codesize:
+The optimization mode can be one of ``"none"``, ``"codesize"``, or ``"gas"`` (default). For example, adding the following line to a module will cause it to try to optimize for codesize:
```
Sporarum, line 44: What happens when pragmas are used in non-root modules ?

docs/structure-of-a-contract.rst
```diff
@@ -109,7 +109,7 @@ Further higher directories can be accessed with ``...``, ``....`` etc., as in Py
 Searching For Imports
 -----------------------------
 
-When looking for a file to import, Vyper will first search relative to the same folder as the contract being compiled. It then checks for the file in the provided search paths, in the precedence provided. Vyper checks for the file name with a ``.vy`` suffix first, then ``.vyi``, then ``.json``.
+When looking for a file to import, Vyper will first search relative to the same folder as the module being compiled. It then checks for the file in the provided search paths, in the precedence provided. Vyper checks for the file name with a ``.vy`` suffix first, then ``.vyi``, then ``.json``.
```
Sporarum, line 112: I think the wording should be clearer, are paths relative to the root-module, or the current module ?

(I'm pretty sure it's the current module, but it should be clear from this paragraph)

docs/using-modules.rst
```diff
@@ -8,7 +8,7 @@ A module is a set of function definitions and variable declarations which enable
 Declaring and using modules
 ===========================
 
-The simplest way to define a module is to write a contract. In Vyper, any contract is a valid module! For example, the following contract is also a valid module.
+Any ``.vy`` file is a module. When a module is the entry point of compilation, it becomes a contract upon deployment. For example, the following module can be both deployed directly and imported by other modules.
```
Sporarum, line 11: Perfect explanation !

Sporarum (commented):
A couple comments on the changes

Note: I used "root module" in my comments to mean "the module we point the compiler at, which will become the contract"
I see you have used the concept of "entry point module", which also works for me

docs/compiler-exceptions.rst
lufa23: you're right, reverted with d7184ae7

docs/interfaces.rst
```diff
@@ -101,7 +101,7 @@ You can see all the available built-in interfaces in the `Vyper GitHub <https://
 Implementing an Interface
 =========================
 
-You can define an interface for your contract with the ``implements`` statement:
+You can define an interface for your module with the ``implements`` statement:
```
fubuloubu, line 104:
```suggestion
You can define an interface that your module must satisfy with the ``implements`` statement:
```

docs/interfaces.rst
```diff
@@ -158,7 +158,7 @@ Standalone interfaces are written using a variant of standard Vyper syntax. The
 Extracting Interfaces
 =====================
 
-Vyper has a built-in format option to allow you to easily export a Vyper interface from a pre-existing contract.
+Vyper has a built-in format option to allow you to easily export a Vyper interface from a pre-existing module.
```
fubuloubu, line 161: Should this be "top-level" module, artifact, etc. (Stay "contract"?)

docs/structure-of-a-contract.rst
```diff
@@ -36,12 +36,12 @@ The following declaration is equivalent, and, prior to 0.3.10, was the only supp
     # @version ^0.4.0
 
 
-In the above examples, the contract will only compile with Vyper versions ``0.4.x``.
+In the above examples, the module will only compile with Vyper versions ``0.4.x``.
```
fubuloubu, line 39: Wonder if `file` would actually be clearer here (or `module file`)

docs/structure-of-a-contract.rst
```diff
@@ -185,7 +185,7 @@ Therefore, a module encapsulates
 * functionality (types and functions), and
 * state (variables), which may be tightly coupled with that functionality 
 
-Modules can be added to contracts by importing them from a ``.vy`` file. Any ``.vy`` file is a valid module which can be imported into another contract! This is a very powerful feature which allows you to assemble contracts via other contracts as building blocks.
+Modules can be imported from any ``.vy`` file. Any ``.vy`` file is a valid module which can be imported into another module. When the entry-point module is compiled and deployed, it becomes a contract.
```
fubuloubu, line 188: This seems very important, maybe consider listing higher? 

Also I haven't see "entry-point module" defined yet, I see it below I'm "Using Modules"

fubuloubu (approved):
Some minor suggestions

docs/interfaces.rst
```diff
@@ -8,7 +8,7 @@ An interface is a set of function definitions used to enable communication betwe
 Declaring and using Interfaces
 ==============================
 
-Interfaces can be added to contracts either through inline definition, or by importing them from a separate file.
+Interfaces can be added to a module either through inline definition, or by importing them from a separate file.
```
rafael-abuawad, line 11: `modules` would be better

```suggestion
Interfaces can be added to modules either through inline definition, or by importing them from a separate file.
```

docs/structure-of-a-contract.rst
rafael-abuawad, line 6: I think since this is really early in the docs, we should start using the word modules as a synonym, keeping the original and adding the word module

```suggestion
Vyper contracts (or modules) are contained within files. Each file contains exactly one contract.
```

Something in that line

rafael-abuawad (commented):
Interesting approach, but I don't think it is going to be useful, especially because the word '**contract**' is engraved so deeply in EVM development. Maybe adding a new section on modules would be better

docs/structure-of-a-contract.rst
Sporarum: The goal of this PR is specifically to make the two concepts more distinct !

And in particular "Each file contains exactly one contract." is incorrect.
That is especially true with the upcoming abstract modules, which are not valid contracts

charles-cooper:
@lufa23 @pcaversaccio @Sporarum are we still doing this or is it superseded by https://github.com/vyperlang/vyper/pull/4946?

pcaversaccio:
> @lufa23 @pcaversaccio @Sporarum are we still doing this or is it superseded by #4946?

well the scope is a bit different here, so i wouldn't close it, but i have still a lot of open questions and my PoV is still https://github.com/vyperlang/vyper/pull/4878#pullrequestreview-3915949286.

Sporarum:
> @lufa23 @pcaversaccio @Sporarum are we still doing this or is it superseded by #4946?

They are orthogonal.

I used module throughout the abstract module docs because it made more sense
Notably given abstract modules are not valid top-level modules

PR #4881, opened 2026-03-16
chore[build]: streamline devenv using uv and pyproject.toml

Sporarum:
### What I did

Make it so `uv run` automatically downloads the correct dependencies

Note!:
This changes how dev versions are computed:
* old:  `<next patch version>`/`<next patch version>+commit.<commit hash>` depending on location (custom vyper format)
* new: `<next patch version>.dev<number of commit since last release>.g0<commit hash>` everywhere ([default setuptools_scm](https://setuptools-scm.readthedocs.io/en/latest/extending/#available-implementations))

Notably the new format correctly precedes the next patch, whereas the old succeeded it.

### Commit message

### Cute Animal Picture


uv.lock
charles-cooper, line 1: do we really want to start committing this

Sporarum: I really think we should:

It only changes when we update the `pyproject.toml`, not when new package versions are released (see [relevant docs](https://docs.astral.sh/uv/concepts/projects/sync/#checking-the-lockfile))

It guarantees that "it works on my machine" means it works on all machines, as the package versions are exactly the same

Sporarum: See also [these docs](https://docs.astral.sh/uv/concepts/projects/sync/#upgrading-locked-package-versions) for how to update package versions, namely:
```
uv lock --upgrade
uv lock --upgrade-package ...
```

vyper/cli/vyper_ir.py
```diff
@@ -15,9 +15,7 @@ def _parse_cli_args():
 def _parse_args(argv):
     parser = argparse.ArgumentParser(description="Vyper IR IR compiler")
     parser.add_argument("input_file", help="Vyper sourcecode to compile")
-    parser.add_argument(
-        "--version", action="version", version=f"{vyper.__version__}+commit{vyper.__commit__}"
-    )
+    parser.add_argument("--version", action="version", version=vyper.__long_version__)
```
Sporarum, line 18: TODO: Add a test for this line

Sporarum:
Scope reduced as per feedback:
https://github.com/vyperlang/vyper/pull/4886

PR #4886, opened 2026-03-20
chore[build]: migrate mostly to `pyproject.toml`

Sporarum:
### Commit message

```
this commit moves most configuration to pyproject.toml to enable
downloading all dependencies with `uv sync`. this commit makes the
contributing section of the readme clearer and adds installation steps
for both pip and uv. there should not be any semantic differences in the
metadata, and the pip workflow should remain identical.
py_modules=["vyper"] is removed, it did nothing as there is no
`vyper.py` module file (packages=["vyper"] remains). the version scheme
remains identical, but is refactored for simplicity. the docs
dependencies are moved to pyproject.toml, and the readthedocs settings
are updated accordingly.
```

### Cute Animal Picture


setup.py
```diff
@@ -54,67 +29,10 @@ def _global_version(version):
...
-    pass
-
-
 setup(
-    name="vyper",
     use_scm_version={
-        "local_scheme": _local_version,
+        "local_scheme": _local_version_suffix,
         "version_scheme": _global_version,
         "write_to": "vyper/version.py",
```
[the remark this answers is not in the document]
Sporarum: Following the setup steps in the readme for both uv and pip leads to the correct version
(checked on a `mktemp --directory`)

A version of 0.0.0 is most likely due to forgetting to fetch the tags

README.md
```diff
@@ -45,40 +45,64 @@ the language and compile to ``bytecode`` and/or ``IR``.
...
+3. Run `python -m venv .venv` to create a virtual environment
+4. Run `source .venv/bin/activate` to activate it
+5. Run `make init` to install the main dependencies
+6. Run `make dev-init` to install the dev dependencies
+7. Run `vyper --version` to verify the setup
+
+### Commands
+
+* `vyper path/to/file.vy` to run your local copy of the compiler
+* `pytest -m "not fuzzing"` to test your changes (if you forget the `-m "not fuzzing"` it will take very long)
```
charles-cooper, line 83: we should just make `not fuzzing` the default (may require some changes to workflow files)

.gitignore
```diff
@@ -44,8 +47,6 @@ docs/vyper.*.rst
 
 # vyper
 vyper/version.py
-vyper/vyper_git_version.txt
```
charles-cooper, line 47: why are these removed?

.readthedocs.yaml
```diff
@@ -6,6 +6,15 @@ build:
   os: ubuntu-22.04
   tools:
     python: "3.11"
+  jobs:
+    pre_create_environment:
+      - asdf plugin add uv
```
charles-cooper: readthedocs env has asdf?

charles-cooper:
looks like readthedocs build is failing

.readthedocs.yaml
Sporarum: It appears to do, Claude did that and it worked (the current readthedocs failure was caused by the outdated main, not these changes)

charles-cooper: i just don't understand why we need it, is it so that readthedocs runs uv? but why do we need that

Sporarum: It's to be able to remove the requirements-docs.txt file
(as suggested by its own TODO)

There might be a way to make it work with pip, but IIRC it also involved using `asdf`

Sporarum: It also makes it easier to test the docs locally

Sporarum: Moved to pip in the end

pyproject.toml
```diff
@@ -22,3 +49,126 @@ exclude = '''
...
+[tool.coverage.run]
+branch = true
+source = ["vyper"]
+# this is not available on the CI step that performs `coverage combine`
+omit = ["vyper/version.py"]
+# allow `coverage combine` to combine reports from heterogeneous OSes.
+# (mainly important for consolidating coverage reports in the CI).
+relative_files = true
+
+# TODO: Remove once fully transitioned to uv
```
charles-cooper: i don't think this is ever going to happen

charles-cooper: like pip is still the standard even if we provide uv compatibility for convenience

README.md
```diff
@@ -45,40 +45,69 @@ the language and compile to ``bytecode`` and/or ``IR``.
...
+* Add PRs if you discover a solution to an existing issue
+* For further discussions and questions, post in [Discussions](https://github.com/vyperlang/vyper/discussions) or talk to us on [Discord](https://discord.gg/6tw7PTM7C2)
+* For more information, see [Contributing](http://docs.vyperlang.org/en/latest/contributing.html)
 
-(Complete [installation steps](https://docs.vyperlang.org/en/latest/installing-vyper.html) first.)
+## Contributing Code
 
-```bash
-make dev-init
-./quicktest.sh -m "not fuzzing"
```
charles-cooper, line 54: why did you remove this?

README.md
```diff
@@ -45,40 +45,69 @@ the language and compile to ``bytecode`` and/or ``IR``.
...
+#### uv (recommended)
 
-Install hevm by downloading it from the releases page (https://github.com/ethereum/hevm/releases/latest) and making sure it is in your PATH. hevm tests can be enabled with `--hevm` flag, and hevm tests can be selected with the `-m hevm` marker. For instance, `./quicktest.sh -m "hevm" --hevm`.
+Make sure you have `uv` installed
 
-## Developing (working on the compiler)
+1. Clone this repo and `cd` into it
+2. Run `git fetch --tags git@github.com:vyperlang/vyper.git` to get the tags necessary for version inference
+3. Run `uv sync`
+4. Run `source .venv/bin/activate` to activate the virtual environment
```
charles-cooper, line 65: does `uv venv` or something do this

README.md
```diff
@@ -45,40 +45,69 @@ the language and compile to ``bytecode`` and/or ``IR``.
...
 
-## Testing (with hevm)
+#### uv (recommended)
 
-Install hevm by downloading it from the releases page (https://github.com/ethereum/hevm/releases/latest) and making sure it is in your PATH. hevm tests can be enabled with `--hevm` flag, and hevm tests can be selected with the `-m hevm` marker. For instance, `./quicktest.sh -m "hevm" --hevm`.
+Make sure you have `uv` installed
 
-## Developing (working on the compiler)
+1. Clone this repo and `cd` into it
+2. Run `git fetch --tags git@github.com:vyperlang/vyper.git` to get the tags necessary for version inference
```
charles-cooper, line 63: does the compiler not work without this step?

setup.cfg
```diff
@@ -1,11 +1,16 @@
+
 [metadata]
 license_files = LICENSE
 
+# TODO: Does anyone use `python setup.py test` ?
```
charles-cooper: no, that should probably be removed

setup.py
charles-cooper, line 1: why do we now need setup.py in addition to pyproject?

README.md
Sporarum: It does run, but fails if the file has a version pragma, and as a result the tests do not pass

README.md
Sporarum: I moved it to the installation instructions, since it only concerns pip

README.md
Sporarum: `uv venv` create a venv, but does not activate it (probably for safety reasons ?)
And it's performed as part of `uv sync`

For automatically activating the venv there are things like `direnv`, which I highly recommend, but was not sure if it should be added to the readme, since it's not required

Sporarum: Seems like a good tutorial ?
https://www.ashleyconnor.co.uk/til/2025-03-01-how-to-automatically-activate-venv-with-direnv

I use a more convoluted way with nix shells, so I've not tried it tho

pyproject.toml
Sporarum: Rmoved

setup.cfg
Sporarum: Done

setup.py
Sporarum: As requested, I moved the metadata back to setup.py, while the rest (dependencies and their settings) are in pyproject.toml

It's possible to also have almost everything in pyproject.toml
(the only thing I was not able to move was the git-tag-based version logic, since we do something non-standard)

And it's also not possible to not have pyproject.toml
* black requires it, which is why it existed before this PR
* IIRC uv requires it

Sporarum: As an additional argument to having metadata in pyproject.toml, the PyPA recommends it:
> On the other hand, configuring setuptools with a setup.py file is still fully supported, although it is recommended to use the modern [[project] table in pyproject.toml](https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#writing-pyproject-toml) (or setup.cfg) whenever possible

Source: https://packaging.python.org/en/latest/guides/tool-recommendations/#build-backends

charles-cooper (approved):

PR #4896, opened 2026-03-25
refactor[venom]: remove `fix_memlocs` pass

HodanPlodky:

### What I did

Remove the `FixMemLocationsPass` and replace it with a compile-time assertion that no concrete literal memory addresses exist in the IR before optimization.

Tighten the `VenomBuilder` API so memory-destination operands require `IRVariable` (alloca-derived pointers) instead of the looser `Operand` type.

### How I did it

- Delete `vyper/venom/passes/fix_mem_locations.py` and remove it from all optimization pipelines (O2, O3, Os).
- Add `check_mem_ops()` in `check_venom.py` that asserts no `IRLiteral` memory read/write targets exist before passes run.
- Change `VenomBuilder` signatures for `mload`, `mstore`, `mcopy`, `dloadbytes`, `istore`, `call`/`staticcall`/`delegatecall` pointer args, and `create`/`create2` offset from `Operand` → `IRVariable`.
- Update all codegen callsites in `codegen_venom/` (context, expr, stmt, module, builtins) to produce `IRVariable` pointers where literals were previously used.
- Remove the now-unused `Allocation.pinned` field.

### How to verify it

Full test suite: `pytest --experimental-codegen -m "not fuzzing"`

### Commit message

```
the `FixMemLocationsPass` was a post-hoc fixup that replaced hardcoded
memory offsets (FREE_VAR_SPACE, FREE_VAR_SPACE2) with alloca-relative
pointers. this was necessary because the legacy IR-to-venom translation
emitted raw literal addresses for scratch memory, which the allocator
could not track or relocate.

with the new direct AST-to-venom codegen (`codegen_venom/`), all memory
operations go through VenomBuilder, which already emits alloca-based
pointers. the fixup pass has no remaining work to do.

remove the pass entirely and replace it with an assertion in
`check_venom.py` (`check_mem_ops`) that verifies no concrete literal
memory reads or writes exist before the optimization pipeline runs. this
turns a silent correctness risk into a hard failure at compile time.

additionally, tighten the VenomBuilder API so that memory-destination
operands (`mload`, `mstore`, `mcopy`, `dloadbytes`, `istore`,
`call`/`staticcall`/`delegatecall` pointer args, `create`/`create2`
offset) require `IRVariable` instead of the looser `Operand` type. this
makes it a type error to pass a literal address where an alloca-derived
pointer is expected, catching the class of bugs that
`FixMemLocationsPass` was designed to paper over.
```

Remove `FixMemLocationsPass` — all memory operations now use alloca-derived pointers by construction. Added compile-time assertion and tightened builder type signatures to prevent regressions.

### Cute Animal Picture


vyper/codegen_venom/module.py
```diff
@@ -848,7 +863,10 @@ def _generate_selector_section_dense(
         _generate_fallback_body(builder, module_t, func_t, default_function)
     else:
         # No fallback - revert
-        builder.revert(IRLiteral(0), IRLiteral(0))
+        revert_buffer = fallback_codegen_ctx.allocate_pinned_buffer(
```
charles-cooper: i think this one is fine actually? or i guess we can allocate a buffer with size 0 but idk what it means really (or how it will interact in the allocator)

vyper/venom/memory_allocator.py
```diff
@@ -111,3 +111,12 @@ def reserve(self, alloca: Allocation):
...
             self.reserve(mem)
+
+    def clone_alloca(self, orig: IRInstruction, clone: IRInstruction):
+        assert orig.opcode == "alloca"
+        assert clone.opcode == "alloca"
+
+        orig_allocation = Allocation(orig)
+        if self.is_allocated(orig_allocation):
+            clone_allocation = Allocation(clone)
+            self.set_position(clone_allocation, self.allocated[orig_allocation])
```
charles-cooper, line 127: comment explaining what this does and why

vyper/venom/builder.py
```diff
@@ -146,13 +160,14 @@ def iszero(self, a: Operand) -> IRVariable:
 
     # === Memory ===
     def mload(self, ptr: Operand) -> IRVariable:
```
harkal: `IRVariable` maybe?

harkal (changes requested):

harkal (approved):

PR #4898, opened 2026-03-27
fix[ux]: improve error message for exprs in type position

rafael-abuawad:
### What I did
Fixed a compiler panic when a name in a type position resolves to a `VarInfo`. The compiler now raises `InvalidType` with the existing `'…' is not a type!` message for that case, and keeps `CompilerPanic` for other unexpected non-VyperType namespace values.

### How I did it
In `_type_from_annotation` in `vyper/semantics/types/utils.py`, after resolving a `Name` through the namespace and unwrapping `ModuleInfo`, the code now checks `isinstance(typ_, VarInfo)` before the fallthrough `CompilerPanic`. If it is a `VarInfo`, it raises `InvalidType(err_msg, node)` using the existing error message; otherwise the `CompilerPanic` is preserved for truly unexpected values.

### How to verify it
1. **Minimal repro** using GH 4865
2. **Local compiler**

### Commit message
```
when a name used in a type position resolves to `VarInfo` (e.g.
`empty(MAX_MESSAGES)` where `MAX_MESSAGES` is a constant, or using a
constant as a type annotation like `x: N`), the compiler raised
`CompilerPanic`. the namespace can legitimately contain `VarInfo`
entries — constants, immutables, state variables, and function
parameters all resolve to `VarInfo` — so hitting this branch is a
user error, not a compiler bug.

raise `InvalidType` with the existing "not a type" message for
`VarInfo`, and keep `CompilerPanic` for other unexpected non-type
namespace values.
```

Semantics: treat VarInfo in type position as InvalidType, not CompilerPanic (GH 4865).

### Cute Animal Picture


vyper/semantics/types/utils.py
```diff
@@ -171,6 +171,10 @@ def _type_from_annotation(node: vy_ast.VyperNode) -> VyperType:
         typ_ = typ_.module_t
 
     if not isinstance(typ_, VyperType):
+        from vyper.semantics.analysis.base import VarInfo
+
+        if isinstance(typ_, VarInfo):
+            raise InvalidType(err_msg, node)
```
charles-cooper, line 176: we should probably just unconditionally raise this rather than ever falling thru to compiler panic

charles-cooper: then the inline import above can be removed too

tests/unit/semantics/types/test_type_from_annotation.py
```diff
@@ -104,3 +105,10 @@ def test_multidimensional_mapping(build_node, type_str, type_str2):
     v_t = types[type_str2]
 
     assert ann_t == HashMapT(k_t, HashMapT(k_t, v_t))
+
+
+def test_name_resolving_to_non_vyper_non_varinfo_raises_compiler_panic(build_node, namespace):
```
charles-cooper: kind of a useless test, i would rather just have `# pragma: nocover` on the compiler panic in the source

charles-cooper (changes requested):
thanks for the submission, looks fine overall, requested style changes

Sporarum:
I'm not sure this is the right fix, it seems to me like a better error for
```python
MAX_MESSAGES: constant(uint256) = 64

@external
def parse_blob(blob: Bytes[4096]):
    starts: DynArray[uint16, MAX_MESSAGES] = empty(MAX_MESSAGES)
```
would be something like:
"Only literal constants and expressions (`10`, `1 + 1`) are allowed as lengths for `DynArray`"

Because `2` is not a type either, but `DynArray[uint16, 2]` is valid !

There's also the question of whether we should just allow the above, since `MAX_MESSAGES` is a compile-time constant

rafael-abuawad:
@Sporarum The error is not that they are using a constant for the length of the `DynArray`. The issue is that they are calling `empty(<CONSTANT>)` and the compiler panics

tests/unit/semantics/types/test_type_from_annotation.py
rafael-abuawad: Added the teset just for the codecov bot to stop yelling, but I dont understand how to have it 100% coverage, without the test

vyper/semantics/types/utils.py
rafael-abuawad: I added the conditional raise because I don't know if this is going to handle all the cases

tests/unit/semantics/types/test_type_from_annotation.py
charles-cooper: you add `# pragma: nocover` on the CompilerPanic -- but like i said i think we should remove the panic line

vyper/semantics/types/utils.py
charles-cooper: well i don't know either :P like if it's not a type or a varinfo yea maybe we should panic...

Sporarum:
@rafael-abuawad oh, of course

Then I think it would make sense to add a couple more tests (if they don't already exist) of similar cases, for example:
```
convert(something, MAX_MESSAGES)
```
(and every other spot where a type can be)

rafael-abuawad:
@Sporarum im going to add a few more tests maybe 2 or 3 to see how it handles that. And going to remove the compile panic line

vyper/semantics/types/utils.py
rafael-abuawad: Im going to move the import to the top, should I add more test cases?

tests/functional/builtins/codegen/test_empty.py
```diff
@@ -264,6 +269,19 @@ def test_clear_literals(contract, assert_compile_failed, get_contract):
     assert_compile_failed(lambda: get_contract(contract), Exception)
 
 
+def test_empty_constant_name_not_a_type_regression_4865():
+    code = """
+# pragma: nocover
```
Sporarum: I don't think this does anything

I'm pretty sure Charles meant you should put this on the line that raises, in the python code, see other usage in the codebase

rafael-abuawad: @Sporarum updated, let me know if it looks right

tests/unit/semantics/types/test_type_from_annotation.py
rafael-abuawad: @charles-cooper im going to keep the compiler panic because I'm not 100% sure that this is going to cover all the edgecases

charles-cooper: ok but we definitely don't need a test for it

vyper/semantics/types/utils.py
```diff
@@ -171,6 +172,8 @@ def _type_from_annotation(node: vy_ast.VyperNode) -> VyperType:
         typ_ = typ_.module_t
 
     if not isinstance(typ_, VyperType):
+        if isinstance(typ_, VarInfo): # pragma: no cover
```
charles-cooper: uhh i think THIS one should not have pragma nocover! the panic on line 176 should though

vyper/semantics/types/utils.py
```diff
@@ -171,6 +172,8 @@ def _type_from_annotation(node: vy_ast.VyperNode) -> VyperType:
...
+            raise InvalidType(err_msg, node)
         raise CompilerPanic(f"Not a type: {typ_}", node)
```
charles-cooper:
```suggestion
        if isinstance(typ_, VarInfo):
            raise InvalidType(err_msg, node)
raise CompilerPanic(f"Not a type: {typ_}", node)  # pragma: no cover
```

vyper/semantics/types/utils.py
rafael-abuawad: Ohh sorry, understood that the branch should be nocover

vyper/semantics/types/utils.py
rafael-abuawad: @charles-cooper updated, was the indent removal intentional?

charles-cooper: no!

Sporarum:
@rafael-abuawad did you get around to add more tests ?

rafael-abuawad:
> @rafael-abuawad did you get around to add more tests ?

Hey @Sporarum , no, not yet I'm going to do so this weekend hopefully

rafael-abuawad:
@Sporarum I've added more tests, but I don't know if that's enough

Sporarum:
@rafael-abuawad
My idea was more to test more breadth of errors than more cases of specifically `empty(<something which is not a type>)`

For example:
```
x: <not a type>
convert(x, <not a type>)
x: <not a type>[10]
def foo(x: <not a type>)
def foo() -> <not a type>
```

But it's also something I can do myself if you prefer

rafael-abuawad:
> My idea was more to test more breadth of errors than more cases of specifically `empty(<something which is not a type>)`

@Sporarum understood, I’ve added more tests. Sorry, I became too short-sighted. I didn’t want to go overboard with the tests, but I think this list covers enough

Sporarum:
Very nice, thanks !

rafael-abuawad:
@Sporarum @charles-cooper do you need anything else from my side to have this merged?

charles-cooper (approved):

charles-cooper:
Looks good, will merge shortly

charles-cooper:
@rafael-abuawad looks like there is something not resolving correctly on initialization, can you fix? probably if you get lint passing it should be good

rafael-abuawad:
@charles-cooper I think that was it, I was able to run the lint properly, but I don't know how to re-run the checks

charles-cooper (approved):

charles-cooper:
thanks for your patience here!

charles-cooper:
oh yea this is weird -- https://github.com/vyperlang/vyper/actions/runs/25700504504/job/76139242908?pr=4898

rafael-abuawad:
> https://github.com/vyperlang/vyper/actions/runs/25700504504/job/76139242908?pr=4898

@charles-cooper Seems that commit d52ce769 ("Remove redundant import of VarInfo in utils.py") mistakenly moved that import to module level (line 13), which reintroduced the cycle. I'm going to fix that in this PR, but I tink should be separate

Work on this PR at:
https://github.com/vyperlang/vyper/pull/4898/commits/9437906dfd6700acbc77846fac199e6634945eb8

rafael-abuawad:
@charles-cooper should pass tests/builds now

charles-cooper:
hmm still failing lint @rafael-abuawad -- did you run `make lint`?

rafael-abuawad:
@charles-cooper ran `make link` and it worked as expected, the issue seemed to have come from a import order

Sporarum:
@charles-cooper Let's merge this ?

rafael-abuawad:
@Sporarum any updates?

Sporarum:
Only @charles-cooper can merge, I'll ping him in private

PR #4900, opened 2026-03-30
refactor[venom]: faster load analysis and memalias

HodanPlodky:
### What I did

Replaced the plain `dict` lattice in `LoadAnalysis` with a `Lattice` class that maintains a reverse index from `MemoryLocation` to lattice keys, and added precomputed alias set queries to `MemoryAliasAnalysis`. This eliminates the O(lattice size) scan on every store instruction during load analysis.

### How I did it

- `load_analysis.py` — new `Lattice` class wrapping a dict with a `_by_memloc` reverse index. `remove_aliases` uses the reverse index to invalidate only entries that actually alias the written location, replacing the previous loop over all keys with `may_alias` checks. The `_merge` path now uses `Lattice` instances instead of raw dicts.
- `mem_alias.py` — split location tracking into `concrete_locs` and `abstract_locs` (per-allocation, bisect-sorted). Added `ensure_analyzed` (lazy analysis guard), `get_alias_set` (returns precomputed alias set), and `insert_memloc` (bisect-based insertion for abstract locations sharing an allocation).

### How to verify it

```bash
uv run pytest tests/unit/compiler/venom/ -q
uv run pytest -m "not fuzzing" --optimize gas --evm-version prague --evm-backend revm --experimental-codegen tests/ -q
```

### Commit message

```
perf[venom]: speed up load analysis store invalidation

the store-invalidation path in `LoadAnalysis._analyze_bb` iterated every
lattice key on each store instruction, calling `may_alias` per key. for
functions with many memory operations this is O(n*m) where n is the
number of stores and m is the lattice size — it dominated compile time
on large contracts.

replace the plain `dict` lattice with a `Lattice` class that maintains a
reverse index from `MemoryLocation` to lattice keys. on a store, look up
the precomputed alias set and remove only the matching entries via
`remove_aliases`, making invalidation proportional to the alias set
rather than the full lattice.

on the `MemoryAliasAnalysis` side, split concrete and abstract location
tracking, add bisect-based insertion for abstract locations sharing an
allocation, and expose `ensure_analyzed` / `get_alias_set` so callers
can query aliases without redundant re-analysis.
```

Load analysis store invalidation is now proportional to the alias set size instead of the full lattice, reducing compile time on memory-heavy contracts.

### Cute Animal Picture


vyper/venom/analysis/mem_alias.py
```diff
@@ -74,6 +114,15 @@ def may_alias(self, loc1: MemoryLocation, loc2: MemoryLocation) -> bool:
 
         return result
 
+    def get_alias_set(self, loc: MemoryLocation) -> OrderedSet[MemoryLocation] | None:
+        if loc not in self.alias_sets:
+            self._analyze_mem_location(loc)
+        return self.alias_sets.get(loc, None)
```
harkal: the set will not contain loc so load analysis might not invalidate it properly?

harkal (approved):

Issue #2098, opened 2020-07-05
VIP: add `method_id` member to interface functions


iamdefinitelyahuman:
## Simple Summary
Add a `method_id` member to interface functions:

```python
from vyper.interfaces import ERC20

m: bytes[4] = ERC20.transfer.method_id
```

## Motivation
Currently, when using `raw_call` we have to generate the method ID from a string of the function selector, using the `method_id` builtin.  Adding `method_id` as a member to interfaces is less error-prone and arguably more readable.

## Specification

```python
<INTERFACE>.<FUNCTION>.method_id
```

The member is accessible **only** from the primitive type:

```python
# valid, accessing via the primitive
m: bytes[4] = ERC20.transfer.method_id

# not valid, trying to access from a defined type
m: bytes[4] = ERC20(_addr).transfer.method_id
```

## Backwards Compatibility
Not a breaking change.

## Copyright
Copyright and related rights waived via [CC0](https://creativecommons.org/publicdomain/zero/1.0/)

Sporarum:
I know this was already approved, but I hope its not too late to change things:

In the following situation, `method_id` is under-specified:
```python
def foo(x: uint256 = 0) -> uint256:
    return x
```
Since this method has two abi signatures: `"foo()"` and `"foo(uint256)"`, and it's not clear which `method_id` encodes
(And of course this gets worse the more default parameters there are)


I'm also not sure what is the motivation for this, if we have access to the interface, what upside is there of going through `raw_call` ?
And if it is only to get the return raw, then something like `raw extcall ERC20(_addr).transfer(...)` seems like a cleaner solution


Nitpick: It would feel more pythonic for the name to be `__method_id__`, and it makes it more apparent something is going on.
(`a.b.method_id` can refer to method `method_id` of module `a.b`)

Sporarum:
> Re default args, current implementation uses the full signature (all args). Happy to change if a different behavior is preferred
>
> Re naming (method_id vs __method_id__), FWIW, the reasoning in @Sporarum's https://github.com/vyperlang/vyper/issues/2098#issuecomment-4207082245 makes sense to me
- @johnnyonline, https://github.com/vyperlang/vyper/pull/4903#issuecomment-4207215178

charles-cooper:
maybe it should be a builtin then, since method_id needs to know the number of kwargs

johnnyonline:
[...]

PR #4903, opened 2026-04-06
feat: add method_id member to interface functions

johnnyonline:
### What I did

Implemented [VIP #2098](https://github.com/vyperlang/vyper/issues/2098) - added a `method_id_of` builtin function, allowing direct access to the 4-byte function selector via `method_id_of(Interface.function)`.

### How I did it

- `InterfaceT.get_member_in_expr` exposes functions in expression context (gated from annotation path to prevent use as type)
- Added `method_id_of` builtin in `builtins/functions.py` with build_IR for legacy codegen
- Added venom handler in `codegen_venom/builtins/misc.py`
- Optional `n_optional_args` kwarg to specify how many default args to include in the signature (defaults to `0`, positional-only)

### How to verify it

```bash
pytest tests/functional/codegen/test_interface_method_id.py tests/functional/syntax/test_interface_method_id.py -x -v
```

### Commit message

```
feat: add method_id_of builtin function
```

Added `method_id_of` builtin function, enabling `method_id_of(ERC20.transfer)` syntax to retrieve the 4-byte function selector as bytes4.

### Cute Animal Picture

image

### DIFF
--- a/vyper/builtins/functions.py
+++ b/vyper/builtins/functions.py
@@ -752,12 +752,12 @@ class MethodID(FoldedFunctionT):
 class MethodIDOf(BuiltinFunctionT):
     _id = "method_id_of"
     _inputs = [("func_ref", TYPE_T.any())]
-    _kwargs = {"n_args": KwargSettings(UINT256_T, 0, require_literal=True)}
+    _kwargs = {"n_optional_args": KwargSettings(UINT256_T, 0, require_literal=True)}
     _return_type = BYTES4_T
     _modifiability = Modifiability.CONSTANT
 
     def fetch_call_return(self, node):
-        validate_call_args(node, 1, ["n_args"])
+        validate_call_args(node, 1, ["n_optional_args"])
         self._validate_func_ref(node)
         return BYTES4_T
 
@@ -779,35 +779,32 @@ class MethodIDOf(BuiltinFunctionT):
         return arg_type.typedef
 
     def _validate_func_ref(self, node):
-        from vyper.semantics.types.function import ContractFunctionT
-
         func_t = self._get_func_t(node)
 
-        n_args = self._get_n_args(node, func_t)
-        if not (func_t.n_positional_args <= n_args <= func_t.n_total_args):
+        n_optional = self._get_n_optional_args(node, func_t)
+        if not (0 <= n_optional <= func_t.n_keyword_args):
             raise ArgumentException(
-                f"n_args must be between {func_t.n_positional_args} and "
-                f"{func_t.n_total_args}, got {n_args}",
+                f"n_optional_args must be between 0 and "
+                f"{func_t.n_keyword_args}, got {n_optional}",
                 node,
             )
 
-    def _get_n_args(self, node, func_t):
+    def _get_n_optional_args(self, node, func_t):
         for kw in node.keywords:
-            if kw.arg == "n_args":
+            if kw.arg == "n_optional_args":
                 val = kw.value.get_folded_value()
                 if not isinstance(val, vy_ast.Int):
-                    raise InvalidType("n_args must be an integer literal", kw.value)
+                    raise InvalidType("n_optional_args must be an integer literal", kw.value)
                 return val.value
-        # default: all args (positional + keyword)
-        return func_t.n_total_args
+        # default: 0 optional args (positional only)
+        return 0
 
     def _compute_method_id(self, node):
-        from vyper.semantics.types.function import ContractFunctionT
-
         func_t = self._get_func_t(node)
-        n_args = self._get_n_args(node, func_t)
+        n_optional = self._get_n_optional_args(node, func_t)
+        n_total = func_t.n_positional_args + n_optional
 
-        arg_types = [i.canonical_abi_type for i in func_t.argument_types[:n_args]]
+        arg_types = [i.canonical_abi_type for i in func_t.argument_types[:n_total]]
         function_sig = f"{func_t.name}({','.join(arg_types)})"
         selector = method_id(function_sig)
         return selector
--- a/vyper/builtins/functions.py
+++ b/vyper/builtins/functions.py
@@ -749,6 +749,75 @@ class MethodID(FoldedFunctionT):
         return {"output_type": TYPE_T(output_type)}
 
 
+class MethodIDOf(BuiltinFunctionT):
+    _id = "method_id_of"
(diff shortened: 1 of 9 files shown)


charles-cooper (commented):
i think maybe it makes sense for the evaluation to happen during constant folding? cc @Sporarum

vyper/codegen_venom/expr.py
```diff
@@ -637,8 +637,19 @@ def lower_Attribute(self) -> VyperValue:
                 value = 2**flag_id  # 0 => 1, 1 => 2, 2 => 4, etc.
                 return VyperValue.from_stack_op(IRLiteral(value), typ)
 
-        # Case 2: Address properties
+        # Case 1b: Interface.function.method_id (e.g. ERC20.transfer.method_id)
```
Sporarum: Case 1 should be modified to Case 1a then

Sporarum:
As for constant folding, I guess it technically is a compile-time constant, but I would not expect constant folding to be doing that job

We could instead move the computation to something like `ContractFunctionT.encoded_method_id`
(`encoded` because we apparently need to modify the value we got from `method_ids`: `value = method_id << 224`)

Also see my comments on the VIP: https://github.com/vyperlang/vyper/issues/2098#issuecomment-4207082245

vyper/codegen_venom/expr.py
johnnyonline: Fixed here https://github.com/vyperlang/vyper/pull/4903/changes/532e685475ed3794976880648ba1214bfb5046d4

johnnyonline:
> We could instead move the computation to something like ContractFunctionT.encoded_method_id
(encoded because we apparently need to modify the value we got from method_ids: value = method_id << 224)

Done here https://github.com/vyperlang/vyper/pull/4903/changes/9f51422ff3560ffd95ab52490b282c02ed3f260d

> Also see my comments on the VIP: https://github.com/vyperlang/vyper/issues/2098#issuecomment-4207082245

Re default args, current implementation uses the full signature (all args). Happy to change if a different behavior is preferred

Re naming (`method_id` vs `__method_id__`), FWIW,  the reasoning in @Sporarum's [comment](https://github.com/vyperlang/vyper/issues/2098#issuecomment-4207082245) makes sense to me

charles-cooper:
cf. https://github.com/vyperlang/vyper/issues/2098#issuecomment-4232025678 on functions with default args

johnnyonline:
First commit defaulted `n_args` to all args (full signature) but realized that probably not alligns with the [spec](https://github.com/vyperlang/vyper/issues/2098#issuecomment-4232025678) so second commit renames to `n_optional_args` and defaults to `0`

This way `method_id_of(ERC20.transfer)` just works for the common case and for functions with defaults like `def foo(x: uint256 = 0)`, `method_id_of(IFoo.foo)` --> `foo()` and `method_id_of(IFoo.foo, n_optional_args=1)` --> `foo(uint256)`

If `n_optional_args` is too verbose i will happily shorten it back to `n_args`. Mostly went with this so we're fully clear on the intent (tbh personally i also prefer more explicit too)

charles-cooper:
do we want to handle things like `n_args=-2`?

charles-cooper:
> do we want to handle things like `n_args=-2`?

ok i think we can actually handle this as a follow up. the PR looks good to me, @Sporarum do you have any more comments?

charles-cooper (approved):

vyper/codegen_venom/builtins/misc.py
```diff
@@ -646,6 +646,20 @@ def lower_breakpoint(node: vy_ast.Call, ctx: VenomCodegenContext) -> IROperand:
...
+def lower_method_id_of(node: vy_ast.Call, ctx: VenomCodegenContext) -> IROperand:
+    """
+    method_id_of(IFoo.bar) -> bytes4
+
+    Returns the 4-byte function selector, evaluated at compile time.
+    """
+    from vyper.builtins.functions import MethodIDOf
+
+    builtin = MethodIDOf()
+    selector = builtin._compute_method_id(node)
```
Sporarum, line 658:
```suggestion
    selector = MethodIDOf()._compute_method_id(node)
```

vyper/semantics/types/base.py
```diff
@@ -443,6 +443,10 @@ def infer_kwarg_types(self, node):
     # dispatch into get_type_member if it's dereferenced, ex.
     # MyFlag.FOO
     def get_member(self, key, node):
```
Sporarum, line 445: Why is this change required ?

vyper/semantics/types/module.py
```diff
@@ -70,6 +70,13 @@ def get_type_member(self, attr, node):
         # get an event, struct or flag from this interface
         return TYPE_T(self._helper.get_member(attr, node))
 
+    def get_member_in_expr(self, attr, node):
```
Sporarum, line 73: I think this function name is confusing, I don't see what it semantically means

vyper/builtins/functions.py
```diff
@@ -749,6 +749,72 @@ def infer_kwarg_types(self, node):
...
+        if not (0 <= n_optional <= func_t.n_keyword_args):
+            raise ArgumentException(
+                f"n_optional_args must be between 0 and "
+                f"{func_t.n_keyword_args}, got {n_optional}",
+                node,
+            )
+
+    def _get_n_optional_args(self, node, func_t):
+        for kw in node.keywords:
+            if kw.arg == "n_optional_args":
```
Sporarum, line 794: This seems a bit convoluted, wouldn't `node.keywords.get("n_optional_args", 0)` or something like it work ?

How do the other built-ins deal with kwargs ?

vyper/builtins/functions.py
```diff
@@ -749,6 +749,72 @@ def infer_kwarg_types(self, node):
...
+                val = kw.value.get_folded_value()
+                if not isinstance(val, vy_ast.Int):
+                    raise InvalidType("n_optional_args must be an integer literal", kw.value)
+                return val.value
+        # default: 0 optional args (positional only)
+        return 0
+
+    def _compute_method_id(self, node):
```
Sporarum, line 802: Is this logic a duplicate ?
It seems like we would already have something to compute that

Sporarum (changes requested):
@charles-cooper I think we should not default to 0, and instead force the user to specify the count if there are kwargs: https://github.com/vyperlang/vyper/issues/2098#issuecomment-4234675803

And a couple small comments:

PR #4904, opened 2026-04-06
fix[ux]: handle missing inputs/outputs keys in ABI JSON interface

### LINKED ISSUE #2625: Handle missing "inputs" or "outputs" in abi json file
There is an unhandled exception `KerError` when you are working with json interface files. Since these can sometimes be missing one of these two items, either it should raise a better error or simple skip if that key is not in the array (preferred)

yasumorishima:
### What I did

`ContractFunctionT.from_abi` and `EventT.from_abi` used direct dict key access (`abi["inputs"]`, `abi["outputs"]`) to read ABI entries. The Ethereum ABI spec allows these keys to be omitted when empty — for example, fallback functions have neither `inputs` nor `outputs — so passing such entries raised `KeyError`.

### How I did it

Changed `abi["inputs"]` / `abi["outputs"]` to `abi.get("inputs", [])` / `abi.get("outputs", [])` in both `from_abi` methods, so missing keys default to empty lists instead of raising. Added roundtrip tests that compile Vyper contracts to ABI JSON, parse them back via `from_abi` / `from_json_abi`, re-serialize, and verify the output matches the original.

### How to verify it

```bash
pytest tests/unit/semantics/types/test_from_abi_missing_keys.py tests/unit/semantics/types/test_abi_roundtrip.py -v
```

### Commit message

```
the ABI spec permits "inputs" and "outputs" keys to be absent from ABI
entries — fallback and constructor functions typically omit them.
ContractFunctionT.from_abi and EventT.from_abi accessed these keys
directly, raising KeyError on valid ABI JSON that left them out.

use dict.get() with empty-list defaults so missing keys are treated as
empty argument/return lists, matching the spec's semantics.

GH 2625
```

Fix `KeyError` when parsing ABI JSON entries that omit `"inputs"` or `"outputs"` keys (e.g. fallback functions).

### Cute Animal Picture

Cute red panda


charles-cooper (changes requested):
sorry, this changes correctly failing inputs into silently succeeding inputs. is there tooling which produces malformed inputs like this?

yasumorishima (commented):
Thank you for the review, @charles-cooper.

To answer your question — **yes, Vyper's own compiler produces ABI entries without these keys.**

A minimal Vyper contract with `__default__` (fallback):
```python
@external
@payable
def __default__():
    pass
```
compiles to:
```json
[{"stateMutability": "payable", "type": "fallback"}]
```
— no `inputs`, no `outputs` key.

I verified this in a Codespace by:
1. Checking out the parent commit (before this fix) and calling `ContractFunctionT.from_abi()` with a function ABI missing `inputs`/`outputs` → **KeyError** on all 3 patterns (missing inputs, missing outputs, missing both)
2. With the fix applied → all resolve correctly
3. Running the existing 22 ABI unit tests → **all pass** (no regression)

Beyond Vyper's own output, the [Solidity ABI spec](https://docs.soliditylang.org/en/latest/abi-spec.html) explicitly states:
- Constructor, receive, and fallback **never have `outputs`**
- Receive and fallback **never have `inputs`**
- Event and error types have **no `outputs` defined**

The `solc` compiler follows this — its [test fixtures](https://github.com/ethereum/solidity/blob/develop/test/libsolidity/ABIJson/include_fallback_function.sol) confirm fallback entries contain only `type` and `stateMutability`. The `eth-typing` library (used by web3.py) also marks both fields as `NotRequired`.

I understand the concern about silently accepting malformed inputs. Currently `from_json_abi` filters to `type == "function"` before calling `from_abi`, so this doesn't trigger in the normal import path. But the `from_abi` method itself is a public API, and using `.get(key, [])` is the standard defensive pattern for ABI parsing — consistent with how web3.py and ethers.js handle it.

Let me know if you'd like any changes to the approach.

charles-cooper:
interesting ok in that case i would like some roundtrip tests too -- that we can correctly parse abi jsons produced by vyper itself

yasumorishima:
Thanks for the feedback! I've added roundtrip tests in `e1fe0fb9`.

The new test file (`tests/unit/semantics/types/test_abi_roundtrip.py`) compiles various Vyper contracts via `CompilerData` + `build_abi_output`, then parses the resulting ABI JSON back through `ContractFunctionT.from_abi` / `InterfaceT.from_json_abi` / `EventT.from_abi` to verify Vyper can consume its own output.

**Covered contract patterns (20 tests):**
- **Fallback** (`__default__`) — confirms ABI has no `inputs`/`outputs` keys and `from_json_abi` handles it gracefully
- **Constructor** (with and without args)
- **Simple function** — name, arguments, return type all survive the roundtrip
- **No-args / no-return function**
- **Multiple functions + events** — full interface reconstruction
- **Default arguments** — 3 expanded ABI entries each parse individually via `from_abi`
- **Public state variables** — getter functions roundtrip correctly
- **Tuple return values**
- **Mixed contract** (fallback + constructor + functions + events + public var) — verifies all ABI types coexist and roundtrip

All 20 new tests pass, along with the existing 898 `semantics/types` tests. Lint (flake8/isort/black) clean.

tests/unit/semantics/types/test_abi_roundtrip.py
```diff
@@ -0,0 +1,335 @@
+"""
+Roundtrip tests: compile Vyper contracts to ABI JSON, then parse back
+via from_abi / from_json_abi to verify Vyper can consume its own output.
+
+Addresses reviewer request in vyperlang/vyper#4904.
```
Sporarum: This doesn't seem useful once this PR is merged

tests/unit/semantics/types/test_abi_roundtrip.py
```diff
@@ -0,0 +1,335 @@
...
+"""
+
+from vyper.compiler.output import build_abi_output
+from vyper.compiler.phases import CompilerData
+from vyper.semantics.types.function import ContractFunctionT
+from vyper.semantics.types.module import InterfaceT
+from vyper.semantics.types.user import EventT
+
+# ---------------------------------------------------------------------------
+# Helper
```
Sporarum: This is not the typical format for these things in this codebase

I think our typical way is to not have these section separators at all, but please check

tests/unit/semantics/types/test_abi_roundtrip.py
```diff
@@ -0,0 +1,335 @@
...
+    self.owner = initial_owner
+"""
+
+
+# ---------------------------------------------------------------------------
+# Tests
+# ---------------------------------------------------------------------------
+
+
+class TestFallbackRoundtrip:
```
Sporarum, line 120: We usually don't use test-classes in vyper, but maybe that is something we should change
@charles-cooper opinions ?

tests/unit/semantics/types/test_abi_roundtrip.py
```diff
@@ -0,0 +1,335 @@
...
+    def test_fallback_compiles_without_inputs_outputs(self):
+        abi = _compile_abi(FALLBACK_ONLY)
+        assert len(abi) == 1
+        entry = abi[0]
+        assert entry["type"] == "fallback"
+        assert "inputs" not in entry
+        assert "outputs" not in entry
```
Sporarum: I think it would be good to instead compare with the expected JSON/dict output

(Not only for this test)

tests/unit/semantics/types/test_abi_roundtrip.py
```diff
@@ -0,0 +1,335 @@
...
+            fn = ContractFunctionT.from_abi(entry)
+            assert fn.name == "foo"
+            assert fn.return_type is not None
+
+    def test_deduplicated_entry_via_from_json_abi(self):
+        abi = _compile_abi(DEFAULT_ARGS)
+        # Default-arg expansion produces multiple entries with the same name.
+        # from_json_abi rejects duplicate names, so we keep only the fullest
+        # overload (most inputs) — mirroring how external tools would
+        # deduplicate before feeding ABI JSON back into Vyper.
```
Sporarum: This seems fishy

To me it suggests either:
1. We don't produce a spec-compliant ABI
2. `from_json_abi` cannot accept spec-compliant ABI

Both of which seem wrong, but maybe this is out of scope for this PR
If it seems too daunting, let me know and I'll open a different PR

Sporarum (changes requested):
Hello, thanks for your contribution! 

I found a couple issues which I think should be fixed before this is merged (mostly refactoring)

tests/unit/semantics/types/test_abi_roundtrip.py
charles-cooper: we don't, but i don't think it's a big deal, it only affects grouping. i care more that the things we are testing are sane

tests/unit/semantics/types/test_abi_roundtrip.py
charles-cooper: yes that is weird - i would check the historical issues on how default functions work to see how they were originally intended to work with abi json

yasumorishima:
Addressed in `058da0a0`:
- Removed PR number reference from docstring
- Removed section separator comments
- Replaced individual field asserts with expected dict comparisons

---

Regarding the default-args duplicate name issue (L225), I checked the historical issues and commit history:

- **#903** introduced default parameters as Vyper's alternative to function overloading (motivated by ERC-721's `safeTransferFrom`). The compiler intentionally produces multiple ABI entries with the same name but different input counts.

- **`d1859cd8`** (2020, Ben Hauser) added duplicate-name resolution to the **AST-based** interface builder (`_get_module_functions`) — it keeps only the overload with the most inputs.

- However, `from_json_abi` (module.py:198) passes all entries straight to `_from_lists`, which raises `NamespaceCollision` on duplicate names. The AST-side dedup logic was never ported to the ABI JSON path.

- **#3170** documents related inconsistencies between `implements` and default parameters (still open).

So the compiler output is intentional, but `from_json_abi` lacks the equivalent dedup logic — it's a separate bug. I think this is out of scope for this PR. Happy to open a follow-up issue if you'd like.

Sporarum:
> Happy to open a follow-up issue if you'd like.

Thanks, please do it !
(and remove the test from this PR which relates to it)

yasumorishima:
Thanks @Sporarum.

Removed the `TestDefaultArgsRoundtrip` class in `cfa36dcb` — all 18 remaining roundtrip tests pass, lint is clean.

Follow-ups as requested:
- Issue: #4907 — root cause analysis, reproduction, and references to #903 / d1859cd8
- Fix PR: #4908 — ports the full d1859cd8 semantics (input-type prefix check + keep-longest-overload) to `from_json_abi`, with 6 regression tests. 1097 `tests/unit/semantics/` tests pass.

---

**Correction on my earlier comment**: I wrote that #3170 is "still open", but it was actually closed on 2026-02-11 as completed. Apologies for the inaccuracy.

tests/unit/semantics/types/test_abi_roundtrip.py
```diff
@@ -0,0 +1,291 @@
...
+        assert set(interface.events.keys()) == {"Transfer", "Approval"}
+
+    def test_individual_function_roundtrip(self):
+        abi = _compile_abi(MULTIPLE_FUNCTIONS_AND_EVENTS)
+        func_entries = [e for e in abi if e.get("type") == "function"]
+
+        for entry in func_entries:
+            fn = ContractFunctionT.from_abi(entry)
+            assert fn.name == entry["name"]
+            assert len(fn.arguments) == len(entry["inputs"])
```
Sporarum: This (and other like it) does not quite check the round-trip, since it compares the abi output to the reconstructed vyper.

To properly test the round-trip we should compare the straight compiler output with the one reconstructed from the abi.
Note however that some information is lost in the process (for example vyper `Bytes[10]` to abi `bytes`).
Therefore it can't just be object deep equality.
(If it gets too messy, we'll do something simpler)

Sporarum (changes requested):
Thanks for the changes, only one thing remaining !

tests/unit/semantics/types/test_abi_roundtrip.py
yasumorishima: Addressed in `bfcb72ba`: roundtrip tests now compile -> ABI -> `from_abi()` -> `to_toplevel_abi_dict()` -> compare with the original compiler ABI output.

Since the comparison operates entirely at the ABI level (`dict == dict`), the information loss you mentioned (`Bytes[10]` -> `bytes`) doesn't affect the tests — that lossy step happens before the roundtrip starts.

Sporarum: Nice !

tests/unit/semantics/types/test_abi_roundtrip.py
```diff
@@ -0,0 +1,243 @@
...
+        assert "inputs" not in fallbacks[0]
+        assert "outputs" not in fallbacks[0]
+
+    def test_fallback_skipped_by_from_json_abi(self):
+        abi = _compile_abi(FALLBACK_ONLY)
+        interface = InterfaceT.from_json_abi("FallbackContract", abi)
+        assert len(interface.functions) == 0
+
+
+class TestConstructorRoundtrip:
```
Sporarum: It doesn't look like any of these tests is a round-trip, no ?

Sporarum (changes requested):

Sporarum (approved):
LGTM

charles-cooper:
as a review request, please do not force push to a branch under review, it makes it very hard to keep track of how the latest code relates to the code that has been reviewed

charles-cooper (approved):

PR #4905, opened 2026-04-08
refactor[codegen]: clean up venom lowering pipeline

HodanPlodky:
### What I did

general follow on to https://github.com/vyperlang/vyper/pull/4811

Clean up the `codegen_venom` lowering pipeline: remove dead code, eliminate unnecessary zero-offset special cases, move `pragma: nocover` from function-level to line-level on unreachable raises, and make the `typ` parameter mandatory in `store_vyper_value`.

### How I did it

- Remove unused `LocalVariable.typ` and `LocalVariable.buf` property accessors, unused `child_for_function` method, and unused `load_calldata`, `storage_to_memory`, `memory_to_storage` functions.
- Make the `typ` parameter mandatory in `store_vyper_value` — callers always had the type available, the optional default was unnecessary indirection.
- Remove zero-offset special cases throughout (`if offset == 0: dst = ptr; else: dst = add(ptr, offset)` → just `dst = add(ptr, offset)`). Venom's constant folding eliminates the overhead, so the runtime cost is zero and the code is simpler.
- Tighten `_get_element_ptr` in the ABI encoder to accept only `IRLiteral` keys — dynamic indexing was never exercised and is not supported by the type system.
- Remove the dead primitive-word fast path in `_abi_encode_to_buf` — the general path handles it correctly.
- Replace blanket function-level `pragma: nocover` with line-level annotations on unreachable `CompilerPanic`/`TypeCheckFailure`/`TypeMismatch` raises, so real logic stays covered.
- Remove the zero-length optimization in `lower_create_from_blueprint` (`total_len = codesize` when `args_len == 0` → always `add(codesize, args_len)`), consistent with removing zero-offset shortcuts elsewhere.

### How to verify it

Existing test suite passes. No functional changes — all modifications are dead-code removal, simplification of equivalent code paths, and coverage annotation adjustments.

### Commit message

```
the venom lowering pipeline had accumulated several patterns that added
complexity without benefit: zero-offset special cases (if offset == 0:
use ptr directly, else add offset) were sprinkled across abi
encode/decode, create, stmt, and module — but venom's constant folding
makes the add-of-zero path free at runtime. remove all of them.

several functions and properties were unused: `LocalVariable.typ`,
`LocalVariable.buf`, `child_for_function`, `load_calldata`,
`storage_to_memory`, `memory_to_storage`. `store_vyper_value` had an
optional `typ` parameter that every caller already supplied — make it
mandatory.

replace blanket function-level `pragma: nocover` with line-level
annotations on unreachable raises (CompilerPanic, TypeCheckFailure,
TypeMismatch) so real logic stays covered. narrow
`_get_element_ptr` to accept only IRLiteral keys since dynamic
indexing was unreachable.
```

Cleanup: remove dead code, zero-offset special cases, and narrow nocover annotations to unreachable raises in `codegen_venom`.

### Cute Animal Picture


charles-cooper (commented):
seems fine to me, i would put pragma: nocover rather on the if stmt than the `raise CompilerPanic` stmt if one of the branches is not taken

vyper/codegen_venom/buffer.py
```diff
@@ -49,6 +49,6 @@ class Ptr:
 
     def __post_init__(self):
         if self.buf is not None and self.location != DataLocation.MEMORY:
```
charles-cooper, line 51: move the pragma nocover to this line

vyper/codegen_venom/context.py
```diff
@@ -46,21 +46,9 @@ class LocalVariable:
 
     def __post_init__(self):
         if self.value.is_stack_value:
-            raise CompilerPanic("LocalVariable.value must be located")
+            raise CompilerPanic("LocalVariable.value must be located") # pragma: nocover
         if self.value.location != DataLocation.MEMORY:
-            raise CompilerPanic("LocalVariable must be in MEMORY")
-
-    @property
-    def typ(self) -> VyperType:
```
charles-cooper, line 54: i feel like we should be using this though :P

vyper/codegen_venom/context.py
```diff
@@ -46,21 +46,9 @@ class LocalVariable:
...
-        return self.value.typ
-
-    @property
-    def buf(self) -> Buffer:
```
charles-cooper, line 58: ditto -- like i guess we are peering into var.value.ptr().buf everywhere?

charles-cooper (approved):

PR #4908, opened 2026-04-10
fix[ux]: dedup default-argument expanded ABI entries in `from_json_abi`

### LINKED ISSUE #4907: `InterfaceT.from_json_abi` raises `NamespaceCollision` on default-argument expanded ABI entries
## Summary

`InterfaceT.from_json_abi` cannot consume ABI JSON produced by Vyper itself when the contract contains functions with default arguments. The compiler expands such functions into multiple same-name ABI entries (one per arity), and `from_json_abi` passes them directly to `_from_lists`, which raises `NamespaceCollision` on the duplicate names.

## Reproduction

```python
from vyper.compiler import CompilerData
from vyper.compiler.output import build_abi_output
from vyper.semantics.types.module import InterfaceT

source = """
@external
@view
def foo(a: uint256, b: uint256 = 10) -> uint256:
    return a + b
"""

abi = build_abi_output(CompilerData(source))
# abi contains two entries both named "foo" (arity 1 and arity 2)
InterfaceT.from_json_abi("T", abi)
# vyper.exceptions.NamespaceCollision: multiple functions or events named 'foo'!
```

## Root cause

Default-argument expansion was introduced in #903 as Vyper's alternative to function overloading. The compiler emits N ABI entries with identical names but increasing input counts.

Commit d1859cd8 ("fix: function overloading in interfaces", 2020-07-01) added dedup logic to the AST-side interface builder `_get_module_functions`: it checks input-type prefix compatibility between same-name entries and keeps only the overload with the most inputs. That handling was never ported to the ABI JSON path. `from_json_abi` at `vyper/semantics/types/module.py:198` passes all same-name entries straight to `_from_lists`, which raises `NamespaceCollision` via `_mark_seen`.

Related: #903 (default args motivation), #3170 (implements + default params, closed 2026-02-11 as completed).

## Impact

- Vyper cannot consume its own ABI JSON output for contracts using default arguments.
- External tooling that feeds ABI JSON into Vyper via `InterfaceT.from_json_abi` must manually deduplicate before calling.

## Proposed fix

Port the full d1859cd8 semantics to `from_json_abi`: when multiple function entries share a name, check input-type prefix compatibility (raising `NamespaceCollision` on genuine collisions), and keep the overload with the most inputs.

I've opened a draft PR with the fix and 6 regression tests covering default-arg roundtrip (2-entry and 3-entry), no-default-args unchanged behavior, order-independence, and two incompatible-overload rejection cases.

Discovered while addressing review feedback on #4904.

yasumorishima:
### What I did

`InterfaceT.from_json_abi` previously raised `NamespaceCollision` when parsing ABI JSON produced by Vyper itself for functions with default arguments. The compiler expands such functions into multiple same-name ABI entries (one per arity, see #903), which the AST-side interface builder already deduplicates in `_get_module_functions` (d1859cd8) — but the ABI JSON path did not.

This PR ports the full d1859cd8 semantics to `from_json_abi`:
- When multiple entries share a name, check input-type prefix compatibility.
- If any prefix element type mismatches, raise `NamespaceCollision` (genuine collision).
- Otherwise, keep the overload with the most inputs.

Fixes #4907

### How I did it

`vyper/semantics/types/module.py` — pre-deduplicate function entries by name before handing them to `_from_lists`, comparing input-type prefixes and keeping the longest compatible overload.

### How to verify it

```python
from vyper.compiler import CompilerData
from vyper.compiler.output import build_abi_output
from vyper.semantics.types.module import InterfaceT

abi = build_abi_output(CompilerData("""
@external
@view
def foo(a: uint256, b: uint256 = 10) -> uint256:
    return a + b
"""))
iface = InterfaceT.from_json_abi("T", abi)
assert len(iface.functions["foo"].arguments) == 2  # used to raise NamespaceCollision
```

6 regression tests added at `tests/unit/semantics/types/test_from_json_abi_default_args.py`:

- 2-entry default-args roundtrip
- 3-entry default-args roundtrip
- no-default-args unchanged
- ordering-independent (longest overload wins regardless of order)
- incompatible overload at position 0 → raises `NamespaceCollision`
- incompatible overload at position 1 → raises `NamespaceCollision`

All 1097 `tests/unit/semantics/` tests pass.

### Commit message

```
fix: deduplicate default-argument expanded ABI entries in from_json_abi

InterfaceT.from_json_abi previously raised NamespaceCollision when
parsing ABI JSON produced by Vyper itself for functions with default
arguments. The compiler expands such functions into multiple same-name
ABI entries (one per arity, see #903), which the AST-side interface
builder deduplicates in _get_module_functions (d1859cd8) but the ABI
JSON path did not.

This change ports the full d1859cd8 semantics to from_json_abi: when
multiple entries share a name, check input-type prefix compatibility
(raising NamespaceCollision on genuine collisions), and keep the
overload with the most inputs.
```

Fix `InterfaceT.from_json_abi` raising `NamespaceCollision` when parsing ABI JSON for functions with default arguments.

### Cute Animal Picture

cat


vyper/semantics/types/module.py
```diff
@@ -211,13 +211,38 @@ def from_json_abi(cls, name: str, abi: dict) -> "InterfaceT":
...
+            fn_name = item["name"]
+            prev = fn_by_name.get(fn_name)
+            if prev is None:
+                fn_by_name[fn_name] = item
+                continue
+            prev_inputs = prev.get("inputs", [])
+            curr_inputs = item.get("inputs", [])
+            overlap = min(len(prev_inputs), len(curr_inputs))
+            for a, b in zip(prev_inputs[:overlap], curr_inputs[:overlap]):
+                if a.get("type") != b.get("type"):
```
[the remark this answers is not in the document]
yasumorishima: Thanks for the catch @chatgpt-codex-connector — confirmed and fixed in `d53a2c7a`.

I verified the bug concretely: Vyper emits both `decimal` and `int168` with `"type": "int168"` in the raw ABI, distinguished only by `internalType: "decimal"`. The old implementation compared raw `"type"` strings and silently accepted these as compatible default-arg expansions, discarding one overload.

The fix parses each entry via `ContractFunctionT.from_abi` first, then compares the parsed argument objects' `.typ` attribute. That correctly treats `decimal` (`DecimalT`) and `int168` (`IntegerT`) as incompatible and raises `NamespaceCollision`.

New regression test `test_decimal_vs_int168_rejected_despite_same_raw_type` added. Additionally verified with edge cases: `uint128` vs `uint256`, `int128` vs `int256`, `bool` vs `uint256`, `address` vs `uint160`, `bytes32` vs `bytes16` all correctly raise `NamespaceCollision` (Vyper's type `__eq__` handles bit-width correctly). 7 tests in `test_from_json_abi_default_args.py` pass, and 1098 tests in `tests/unit/semantics/` plus 1938 in `tests/unit/compiler/` pass. 👍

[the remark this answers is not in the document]
Sporarum: This is might also be the case for flags (which are uint256), make sure that is tested too

tests/unit/semantics/types/test_from_json_abi_default_args.py
```diff
@@ -0,0 +1,155 @@
...
+def foo(a: uint256, b: uint256 = 10) -> uint256:
+    return a + b
+"""
+    abi = _compile_abi(source)
+    fn_entries = [e for e in abi if e.get("type") == "function"]
+    assert len(fn_entries) == 2
+
+    interface = InterfaceT.from_json_abi("T", abi)
+    assert "foo" in interface.functions
+    assert len(interface.functions["foo"].arguments) == 2
```
Sporarum: This doesn't test that the defaults are actually carried over

vyper/semantics/types/module.py
```diff
@@ -211,13 +211,40 @@ def from_json_abi(cls, name: str, abi: dict) -> "InterfaceT":
         InterfaceT
             primitive interface type
         """
-        functions: list = []
-        events: list = []
-
+        # Functions with default arguments are expanded by the Vyper compiler
+        # into multiple ABI entries sharing the same name but with different
+        # input counts (see #903). Deduplicate by keeping the overload with
```
Sporarum: Please don't refer to github issues from the code

This policy means we can safely move to a different git host in the future

Sporarum: (Referring to git commits is fine)

Sporarum (dismissed):
Nice work!
A couple comments

Sporarum (commented):
Looks good to me !

vyper/semantics/types/module.py
```diff
@@ -205,13 +205,40 @@ def from_json_abi(cls, name: str, abi: dict) -> "InterfaceT":
...
+                if a.typ != b.typ:
+                    raise NamespaceCollision(
+                        f"ABI contains multiple functions named '{fn_name}' "
+                        f"with incompatible input types"
+                    )
+            if len(curr_args) > len(prev_args):
+                parsed_fns[fn_name] = curr_fn
+
+        functions: list = [(fn_name, fn) for fn_name, fn in parsed_fns.items()]
+        events: list = [
```
charles-cooper: the style change for events to use a list comprehension is unnecessary

vyper/semantics/types/module.py
```diff
@@ -205,13 +205,40 @@ def from_json_abi(cls, name: str, abi: dict) -> "InterfaceT":
...
+        # into multiple ABI entries sharing the same name but with different
+        # input counts (one entry per arity). Deduplicate by keeping the
+        # overload with the most inputs, mirroring the AST-side handling in
+        # _get_module_functions (d1859cd8). Entries with the same name but
+        # incompatible input-type prefixes represent a genuine collision and
+        # are rejected. Types are compared on the parsed argument objects
+        # rather than the raw ABI "type" string, so e.g. `int168` and
+        # `decimal` (which share `"type": "int168"` but differ in
+        # `internalType`) are correctly treated as incompatible.
+        parsed_fns: dict = {}
```
charles-cooper, line 234: since this logic is somewhat nontrivial i'd like to see it broken out into another function

vyper/semantics/types/module.py
```diff
@@ -205,13 +205,40 @@ def from_json_abi(cls, name: str, abi: dict) -> "InterfaceT":
...
+            prev_args = list(prev.arguments)
+            curr_args = list(curr_fn.arguments)
+            overlap = min(len(prev_args), len(curr_args))
+            for a, b in zip(prev_args[:overlap], curr_args[:overlap]):
+                if a.typ != b.typ:
+                    raise NamespaceCollision(
+                        f"ABI contains multiple functions named '{fn_name}' "
+                        f"with incompatible input types"
+                    )
+            if len(curr_args) > len(prev_args):
```
charles-cooper, line 251: hmm, this line is key. like the algorithm used here looks correct, but it doesn't pass "immediately obvious by inspection". i would like to see a test for example that mixing prefix lengths works, e.g. `foo(uint256, address, decimal)`, `foo(uint256, address)`, `foo(uint256, address, int128)` is correctly rejected.

charles-cooper (approved):
lgtm. thanks!

charles-cooper:
need to run `make lint` @yasumorishima

PR #4909, opened 2026-04-10
refactor[venom]: sunset msize instruction

harkal:
### What I did

Removed the `msize` instruction and its `MSIZE` effect flag from Venom IR. The flag was load-bearing for five passes (RemoveUnusedVariables fence, AvailableExpressionAnalysis `ignore_msize`, CSE skip-list, DFT mask, memmerging volatile check) and forced a constructor `mload` touch in `codegen_venom/module.py` — all of it existing only so the optimizer could see `msize`.

Replaced with `memtop`, an instruction that lowers to a single `MSIZE` byte at assembly time and carries only a `MEMORY` read effect (enough to prevent CSE merging across memory writes and DFT reordering past them, while still letting two consecutive memtops merge). The three call sites that need runtime-sized scratch (`raw_call(msg.data, ...)`, `create_copy_of`, `create_from_blueprint`) now go through a new `ctx.allocate_dyn()` helper, mirroring the existing `allocate_buffer` API for static buffers.

Drive-bys:
- Fixed an aliasing bug in `raw_call(msg.data)`: the calldatacopy ran before kwarg evaluation, so any nested expression that wrote to scratch could clobber the calldata buffer. Deferred the calldatacopy to immediately before the call.
- Guarded `Mem2Var` against non-literal alloca sizes (one-line `IRLiteral` check) so a future dynamic alloca cannot be silently mis-promoted to SSA.
- Added the first direct unit tests for `MemLivenessAnalysis` + `ConcretizeMemLocPass` over loop CFGs (existing coverage was incidental functional tests), plus a `test_memtop.py` covering `memtop` lowering, CSE merge/no-merge, and DFT scheduling.

### Commit message

```
The msize Venom instruction carried a dedicated MSIZE effect flag that
effects.py propagated to every memory-touching opcode, so the optimizer
could honor the implicit "any memory touch advances msize" semantics.
That single flag was load-bearing for five separate code paths — a fence
worklist in RemoveUnusedVariables, the ignore_msize shortcut in
AvailableExpressionAnalysis, a CSE skip-list entry, the `& ~MSIZE` mask
in DFT scheduling, and a volatile-memory check in memmerging — plus a
constructor mload touch in codegen_venom/module.py to advance msize past
the immutables region. All of that machinery existed solely so the
optimizer could see msize at all.

Replace msize with memtop, which lowers to a single MSIZE byte at
assembly emission time and carries only a MEMORY read effect. That
effect is the minimal constraint needed to prevent CSE from merging two
memtops across a memory write and to give DFT an edge that keeps memtop
ordered after prior writes; consecutive memtops with no intervening
write remain mergeable. The three callers that need runtime-sized
scratch (raw_call(msg.data), create_copy_of, create_from_blueprint) now
go through a new codegen_venom/context.py::allocate_dyn() helper,
mirroring the existing allocate_buffer API for static buffers.
```

### Cute Animal Picture

### DIFF
--- a/tests/unit/compiler/venom/test_alloca_in_loop.py
+++ b/tests/unit/compiler/venom/test_alloca_in_loop.py
@@ -1,0 +1,337 @@
+"""
+Unit tests for alloca behavior in loop CFGs.
+
+These tests pin down the current (static) alloca-in-loop invariants before
+dynamic-alloca support is added. They exercise the three cooperating parts
+of the pipeline:
+
+1. MemLivenessAnalysis — extends an alloca's live range across loop
+   back-edges via its fixpoint propagation, so allocas used in a loop body
+   are live for the entire loop.
+2. ConcretizeMemLocPass + MemoryAllocator — must NOT reuse a slot between
+   two allocas both live in the same loop, but MAY reuse it between
+   disjoint branches within a loop, and between allocas whose lifetimes
+   are disjoint with respect to the loop.
+3. Mem2Var + MakeSSA — a loop-body mload/mstore pair on a 32-byte alloca
+   is promoted to SSA, and MakeSSA then inserts the expected loop-header
+   phi for the loop-carried value.
+
+The direct test coverage for these cases was previously limited to
+incidental functional tests; see tests/functional/codegen/features/
+test_alloca_loop_param_init.py (PR #4840 regression). These unit-level
+tests catch regressions in the analysis layer itself.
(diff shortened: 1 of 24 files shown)


charles-cooper (commented):
i think there is a problem that there is not really an allocator. if the live range of two `alloca_top`s overlap then they will trample each others memory

charles-cooper (approved):

PR #4902, opened 2026-04-05
fix[docs]: add blank line after code block

hwhsu1231:
### What I did

Fix the build warnings "maximum 1 argument(s) allowed, X supplied" raised by the `.. code-block:: vyper` directive.

### How I did it

Add a blank line after the `.. code-block:: vyper` directive.

### How to verify it

```bash
sphinx-build -b html docs docs/_build
```

### Commit message

```
fix[docs]: add blank line after code block

Add a blank line after the `.. code-block:: vyper` directives.

Otherwise, Sphinx parsed the first code line as extra directive
arguments and raised "maximum 1 argument(s) allowed, X supplied".

This prevents the parser from misinterpreting the code block and
fixes the documentation build warnings.
```

PR #4910, opened 2026-04-13
fix[docs]: fix docs CI build

charles-cooper:

### What I did

Fix the docs CI build which is currently failing with `AttributeError: 'int' object has no attribute 'replace'`. Supersedes #4902 (includes those changes plus additional fixes).

### How I did it

- **Pin `Pygments==2.18.0`**: `pygments>=2.19` changed `linespans` to require strings, which crashes with `shibuya==2024.1.17`. This is the root cause of the CI failure.
- **Add blank lines after `code-block` directives** (from #4902): missing blank lines caused sphinx to treat code as directive arguments.
- **Use anonymous refs (`` >`__ ``)** for audit links in `release-notes.rst`: repeated named targets like `ChainSecurity` caused duplicate target warnings.
- **Fix inline literal**: `` ``phi``s `` → `` ``phi``\s `` to avoid "start-string without end-string" warning.
- **Fix title underline length** in `structure-of-a-contract.rst` for "Experimental Code Generation" section.
- **Rename duplicate label**: `.. _interfaces:` in `structure-of-a-contract.rst` → `.. _structure-interfaces:` to avoid conflict with `interfaces.rst`.
- **Update skills/contributing.md**: note that `fmt_commit_msg.py` prints to stdout, and that commit messages should not use issue-tracker-specific syntax.

### How to verify it

CI docs job should pass.

### Commit message

```
fix[docs]: fix docs CI build

the docs CI build crashes with `AttributeError: 'int' object has no
attribute 'replace'` due to pygments>=2.19 passing an int for
`linespans` which shibuya==2024.1.17 doesn't handle. pin
Pygments==2.18.0 as a stopgap.

also fix several rst warnings that were already present: duplicate named
targets for audit links (use anonymous refs), an inline literal without
end-string, a title underline that was too short, and a duplicate `..
_interfaces:` label.

also update skills/contributing.md to note that fmt_commit_msg.py prints
to stdout, and that commit messages should not use issue-tracker-
specific syntax (use `GH 1234` not `#1234`).
```

n/a

### Cute Animal Picture

capybara

### DIFF
--- a/requirements-docs.txt
+++ b/requirements-docs.txt
@@ -1,3 +1,4 @@
+Pygments<2.19
 shibuya==2024.1.17
 sphinx==7.2.6
 sphinx-copybutton==0.5.2
--- a/requirements-docs.txt
+++ b/requirements-docs.txt
@@ -1,4 +1,4 @@
-Pygments<2.19
+Pygments==2.18.0
 shibuya==2024.1.17
 sphinx==7.2.6
 sphinx-copybutton==0.5.2
--- a/docs/control-structures.rst
+++ b/docs/control-structures.rst
@@ -96,6 +96,7 @@ Internal functions (optionally marked with the ``@internal`` decorator) are only
 Or for internal functions which are defined in :ref:`imported modules <modules>`, they are invoked by prefixing the name of the module to the function name:
 
 .. code-block:: vyper
+
     import calculator_library
 
     @external
--- a/docs/release-notes.rst
+++ b/docs/release-notes.rst
@@ -26,7 +26,7 @@ Date released: 2025-06-19
 
 v0.4.3 introduces the ``@raw_return`` decorator which allows contracts to return bytes directly without ABI-encoding, which enables new proxy contract use cases. The default EVM version has been updated to ``prague``, and several improvements have been made to the Venom optimizer pipeline.
 
-Audits: `ChainSecurity <https://github.com/vyperlang/audits/blob/master/audits/ChainSecurity_Vyper_June_2025_limited_review.pdf>`_,  `Anatomist <https://github.com/vyperlang/audits/blob/master/audits/Anatomist_Vyper_April_2025.pdf>`_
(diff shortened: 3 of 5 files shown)


requirements-docs.txt
```diff
@@ -1,3 +1,4 @@
+Pygments<2.19
```
pcaversaccio:
```suggestion
Pygments==2.18.0
```

pcaversaccio: let's pin an exact version, there is always a possibility that the package is compromised and someone pushes a malicious `2.18.x` version.

pcaversaccio (approved):

Sporarum (approved):
I don't really know how the internals of the docs work, but this seems like a clean fix

Having to add blank lines at the beginning of code blocks is a bit confusing, is there an option to have it work as it was ?

hwhsu1231:
Hello @charles-cooper,

> **Pin `Pygments==2.18.0`:** `pygments>=2.19` changed linespans to require strings, which crashes with `shibuya==2024.1.17`. This is the root cause of the CI failure.

I found that the docs build works even though we pin `pygments>=2.19.0,<2.20.0`:

```bash
BRANCH=master
COMMIT=6a53488e99aa9f0396747570e747734a83cd1ed0
git clone --branch=${BRANCH} https://github.com/vyperlang/vyper.git vyper
cd vyper
git checkout ${COMMIT}
conda create --prefix ./.conda --yes
conda activate ./.conda
conda install python=3.11 -c conda-forge -c nodefaults --yes
export PYTHONNOUSERSITE=1 LANG=en_US.UTF-8 LANGUAGE=en_US
cat <<EOF > requirements-docs.txt
pygments>=2.19.0,<2.20.0
shibuya==2024.1.17
sphinx==7.2.6
sphinx-copybutton==0.5.2
sphinx-reredirects==0.1.5
EOF
pip install -r requirements-docs.txt
sphinx-build -b html docs docs/_build
```

The docs build failed only if we installed `pygments==2.20.0`.

PR #4914, opened 2026-04-13
fix[venom]: proper handling of effects in the memory copy elision

HodanPlodky:

### What I did

Fix the memory copy elision pass to properly invalidate cached storage and transient load results when those locations are written.

### How I did it

The `_volatile_memory` helper only checked `Effects.MEMORY` — when an `sstore` or `tstore` occurred, cached `sload`/`tload` results were never cleared, so subsequent reads could be replaced with stale values.

Inlined the helper into the else branch and extended it to check all three effect types independently:
- `Effects.MEMORY` → clear `copies` and memory load cache (existing behavior)
- `Effects.STORAGE` → clear storage load cache (new)
- `Effects.TRANSIENT` → clear transient load cache (new)

### How to verify it

`pytest --experimental-codegen -m "not fuzzing"`

### Commit message

```
the memory copy elision pass only tracked and invalidated MEMORY
effects. when an instruction wrote to STORAGE or TRANSIENT, cached
sload/tload results were not invalidated, so later reads could be
replaced with stale values.

inline the _volatile_memory helper and extend the else branch to check
all three effect types independently: clear memory copy/load caches on
MEMORY writes, clear storage load caches on STORAGE writes, and clear
transient load caches on TRANSIENT writes.
```

Fix memory copy elision pass to invalidate cached storage/transient loads on write effects.

### Cute Animal Picture


charles-cooper (approved):

charles-cooper (approved):

charles-cooper:
check lint please

PR #4916, opened 2026-04-15
fix[codegen]: fix zero pad implementation in venom abi encoder

HodanPlodky:
### What I did
The new codegen directly into venom handled padding in abi encode for `Bytes[N]` incorrectly. That could cause the write outside of allocated memory. 

### How I did it
This problem was discovered by running venom pipeline without the `Mem2VarPass` if you disable this pass the tests/functional/builtins/codegen/test_abi_encode.py:test_abi_encode_private test should fail on master but should be ok with this fix

### Commit message

```
the venom codegen ABI encoder padded `Bytes[N]` values by writing a
single 32-byte zero word at the first byte past the data. this is
incorrect when the padding needed to reach the next 32-byte boundary is
less than 32 bytes: the write would extend past the allocated ABI
buffer, clobbering adjacent data.

the correct approach is to write only the number of zero bytes actually
required: `padded_len - (32 + length)`, where `padded_len = ceil32(32 +
length)`. the fix computes that count before calling `_zero_pad` and
uses `calldatacopy` from `calldatasize` (a guaranteed out-of-bounds
calldata source of zeros) to write exactly that many zero bytes.

the fix also removes the mload of `length` from inside `_zero_pad`
itself, instead requiring callers to pass both `length` and `count`
explicitly, which makes the control flow easier to follow and avoids the
now-unnecessary assertion on the pointer type.
```

### Cute Animal Picture


charles-cooper:
i'm guessing the codesize regression is from the larger literals right?

charles-cooper (commented):
this works. for the case where count < 32 i think we can write a 0 to `dst & ~5` (floor32) and it will be more efficient than this whole zero pad gymnastics, but i'll do it in a follow up

charles-cooper (approved):

PR #4922, opened 2026-04-16
feat[venom]: add dynamic allocation lowering

harkal:
### What I did

Added first-class Venom support for runtime-sized memory allocation and
internal dynamic returns, lowered through an explicit **FMP virtual
register** so the free-memory pointer is ordinary, analyzable IR instead of
a hidden side-channel.

The main pieces:

- One-output `dalloca` as the producer-facing runtime allocation
  instruction; no explicit free — the compiler reclaims conservatively, and
  leaving the FMP advanced until return is always sound.
- `dret` for internal functions that return dynamically-sized buffers
  (pack-by-copy; the caller adopts the callee's FMP).
- The old `memtop`/`dfree`/MSIZE allocation model removed from the IR
  surface; `raw_call(msg.data)`, `create_copy_of`, and
  `create_from_blueprint` migrated to `dalloca`.
- The FMP register opcodes (`getfmp`, `setfmp`, `retfmp`) with a dedicated
  effects row, dedicated hidden-param opcodes (`fmp_param`, `retpc_param`),
  and a function-header annotation for lowered IR — the calling convention
  is carried by syntax, not inferred by heuristics.
- A reclaim engine with a written soundness argument (dataflow over
  allocation-mark stacks), fail-closed escape analysis, and post-lowering
  invariant checks that run in every build.

### How I did it

The IR has three vocabulary levels; each pass boundary can be checked
mechanically because no instruction from an earlier level may survive it:

```text
producer:   dalloca, dret
              │  DretDesugarPass    (pre-inline, purely local)
register:   getfmp, setfmp, retfmp
              │  FunctionInlinerPass (a plain cloner — no FMP knowledge)
              │  ... optimization pipeline ...
              │  FmpLoweringPass    (post-SSA, single owner) + FmpPrunePass
lowered:    bump, fmp_param, retpc_param, ret with hidden adopted-FMP value
```

**`dret` desugars locally.** `DretDesugarPass` rewrites a `dret` inside the
callee into `getfmp`-rooted pack copies, a `setfmp` advance, and a `retfmp`
terminator. It touches no callers, no invokes, and no params — so inlining
needs no special FMP handling, and no pass can ever observe a half-threaded
call edge.

**The publish fact lives in the terminator.** `retfmp` means "return these
values and the caller adopts my FMP"; plain `ret` is callee-save (the FMP
implicitly reverts). After inlining, the host's own terminators alone
determine whether the host publishes — nothing is inferred from body
contents.

**Exactly one pass owns the calling convention.** `FmpLoweringPass`
materializes the hidden `fmp_param`, seeds the entry function with
`initial_fmp` (which resolves to an assembly-time constant — leaf functions
get zero plumbing), threads one SSA FMP runner, lowers `dalloca` to
`ceil32` + `bump`, and writes every hidden invoke operand from the value in
its hand. It *asserts* no one else wrote one (half-lowered input is rejected
by the validator up front). `FmpPrunePass` then deletes dead FMP plumbing
and seals each function's `fmp_signature` callee-first, so callers only ever
lower against final callee shapes. `check_post_lowering` verifies the
result in every build: no raw ops survive, physical shape matches the
sealed signature, invoke operand/output counts match the callee, and every
allocation base is FMP-rooted.

**Reclaim is proved, never guessed.** The engine runs a forward dataflow
over stacks of allocation marks: the meet at CFG joins keeps the longest
common *top* segment (anything below a divergence is simply never
reclaimed), one shared interpreter drives both the fixpoint and the
rewrite, and a rewind is emitted only when the popped allocations are
provably dead — where "dead" includes derived pointers (`BasePtrAnalysis`),
escaped pointers pin their allocation permanently (fail-closed), and a live
`getfmp` capture vetoes all rewinds (its derived region is unbounded
above). Each emitted restore asserts that the mark's definition dominates
the restore point.

For internal dynamic returns the callee lowers

```text
dret dyn_count, <ordinary returns...>, src0, size0, ..., return_pc
```

into a physical `ret` returning

```text
<ordinary returns...>, dst0, ..., dstN, adopted_fmp, return_pc
```

with the buffers packed at the callee's entry FMP. The caller adopts the
hidden FMP output at the invoke edge, so its later allocations never reuse
returned memory. The producer contract is unchanged: pack sources must be
ordered so an earlier destination does not clobber a later source.

Hand-written Venom text round-trips at every level: raw IR needs nothing,
lowered IR carries `function f [fmp_lowered, fmp_publishes] { ... }`, and
mixed-level IR is rejected with structured errors.

## Commit message

```
Add one-output `dalloca` for runtime-sized Venom memory and `dret` for
internal functions returning dynamically-sized buffers, lowered through
an explicit free-memory-pointer virtual register (`getfmp`/`setfmp` and
the publishing `retfmp` terminator) so the FMP is ordinary, analyzable
IR instead of a hidden side-channel.

`DretDesugarPass` rewrites `dret` locally (pre-inline) into pack copies
plus a register advance; the inliner needs no FMP knowledge, and whether
a function publishes its FMP is determined by its terminator alone
(plain `ret` is callee-save). `FmpLoweringPass` (post-SSA) is the single
owner of the calling convention: it materializes the hidden `fmp_param`
and `retpc_param` (dedicated opcodes -- the convention is carried by
syntax, not inferred), seeds the entry function via `initial_fmp`,
lowers `dalloca` to `ceil32` + `bump`, and threads one FMP runner
through invokes. `FmpPrunePass` deletes dead FMP plumbing and seals each
function's frozen `fmp_signature` callee-first; post-lowering checks
validate shape against signature in every build.

Reclaim is conservative and proved: a forward dataflow over allocation
marks (top-segment meet at joins) rewinds the FMP only when liveness,
base-pointer facts, and a fail-closed escape analysis show the popped
allocations and all derived pointers are dead; live `getfmp` captures
veto rewinds. Anything unprovable leaves the FMP advanced until return,
which is always sound.

Replace `memtop` scratch allocation in the create/raw_call paths with
`dalloca`, remove `memtop`/`dfree`/MSIZE from the IR surface, and update
validation, analyses, copy-forwarding (including transitive callee-frame
cost accounting that keeps the migration gas-neutral), pipeline
ordering, parser round-tripping (function-header annotations for lowered
IR), docs, and tests for the new model. Generated bytecode and gas are
unchanged across the example corpus.
```


charles-cooper:
I think we should track the fmp explicitly as a stack variable and not use msize

tests/unit/compiler/venom/test_dalloca.py
```diff
@@ -5,24 +5,44 @@
...
 
+def _apply_lowering(fn):
+    """Apply the post-dalloca-lowering pass pipeline to a single function.
+
+    Run in callee-first order when there are multiple functions.
+    """
+    ConcretizeMemLocPass(IRAnalysesCache(fn), fn).run_pass()
+    DallocaLoweringPass(IRAnalysesCache(fn), fn).run_pass()
+    ac = IRAnalysesCache(fn)
+    MakeSSA(ac, fn).run_pass()
```
HodanPlodky, line 32: Why is this `MakeSSA` necessary in the test but not in the optimizations order?

vyper/venom/passes/dalloca_lowering.py
```diff
@@ -0,0 +1,158 @@
...
+    `bump`. A subsequent MakeSSA run renames the reassignments into
+    fresh SSA versions and places phis at merge points.
+
+    Each invoke to a needs-fmp callee takes the caller's current fmp
+    as the first (deepest-on-stack) argument. The callee does not
+    propagate FMP back to the caller; the caller's FMP SSA variable
+    survives across `invoke` via standard liveness (the stack
+    scheduler DUPs it before the call when needed). This implicitly
+    reclaims the callee's dynamically allocated memory when the
+    callee returns.
```
HodanPlodky, line 37: would this mean that the `FMP` is reset to the value before call if it is needed?

If that is the case that would mean we could not return the dynamically (with dynalloca) allocated value

vyper/venom/passes/dalloca_lowering.py
```diff
@@ -0,0 +1,158 @@
...
+        # whole function in pre-SSA form. All re-definitions share
+        # this name; MakeSSA will version them and insert phis at
+        # merge points.
+        fmp_var = fn.get_next_variable()
+
+        # add leading `param` at entry, before any other instructions.
+        entry = fn.entry
+        param_inst = IRInstruction("param", [], [fmp_var])
+        param_inst.parent = entry
+        entry.instructions.insert(0, param_inst)
```
HodanPlodky, line 88: Why not using `insert_instruction` method

vyper/venom/passes/dalloca_lowering.py
```diff
@@ -0,0 +1,158 @@
...
+    def _rewrite_bb(self, bb, fn, fmp_var: IRVariable) -> None:
+        new_instructions: list[IRInstruction] = []
+        for inst in bb.instructions:
+            if inst.opcode == "dalloca":
+                new_instructions.extend(self._lower_dalloca(inst, fn, fmp_var, bb))
+                continue
+            if inst.opcode == "invoke":
+                target = inst.operands[0]
+                assert isinstance(target, IRLabel)
+                callee = fn.ctx.functions.get(target)
```
HodanPlodky, line 106: You can use `get_function` method on ctx

vyper/venom/passes/dalloca_lowering.py
```diff
@@ -0,0 +1,158 @@
...
+                if callee is not None and getattr(callee, "_needs_fmp", False):
+                    calls_needs_fmp = True
+                    break
+            if calls_needs_fmp:
+                break
+
+        needs_fmp = has_dalloca or calls_needs_fmp
+        fn._needs_fmp = needs_fmp  # type: ignore[attr-defined]
```
HodanPlodky, line 73: Do we want to add it to `IRFunction` properly or do you think that there is no need?

vyper/venom/passes/dalloca_lowering.py
harkal: We wanted to free the dynamic memory a function allocates

tests/unit/compiler/venom/test_dalloca.py
harkal: The pass actually reuses the fmp_var with each bump so it needs makessa, but it gets aways with it in the pipeline :/ Proper is to have to make ssa. Adding it.

vyper/venom/venom_to_assembly.py
```diff
@@ -200,6 +214,19 @@ def generate_evm_assembly(self, no_optimize: bool = False) -> list[AssemblyInstr
...
+        # AND above every function's spill region. Spill slots start
+        # at fn_eom[fn] and grow upward; `peak_spill_end` tracks the
+        # maximum `fn_eom + N*32` reached across all functions during
+        # codegen. Placing the initial FMP at or above that value
+        # guarantees that the dynamic allocation region (bumped from
+        # FMP by each `bump` instruction) never aliases with any
+        # function's spill area.
+        eoms = [self.ctx.mem_allocator.fn_eom.get(fn, 0) for fn in self.ctx.functions.values()]
+        max_eom = max(eoms, default=0)
+        return ceil32(max(max_eom, self.spiller.peak_spill_end, MemoryPositions.RESERVED_MEMORY))
```
HodanPlodky: I dont think `MemoryPositions.RESERVED_MEMORY` is relevant here since the venom can allocate form 0 and the reserved mem should be used only in old pipe line

vyper/venom/function.py
```diff
@@ -26,6 +26,14 @@ class IRFunction:
     _invoke_param_count: Optional[int]
     _has_memory_return_buffer_param: Optional[bool]
 
+    # Set by DallocaLoweringPass: True if this function threads the
+    # free-memory pointer (i.e. contains a dalloca or invokes a callee
+    # that does). When present, the hidden FMP param is placed at the
+    # tail of the param list, immediately before return_pc (if any).
+    # Functions on the `initial_fmp` fast path keep this False.
+    _needs_fmp: bool
+    _has_fmp_param: bool
```
HodanPlodky: I have seen them, only set to same thing. Is there a difference between these two.

vyper/venom/passes/dalloca_lowering.py
```diff
@@ -0,0 +1,419 @@
...
+            return False
+
+        liveness = self.analyses_cache.request_analysis(LivenessAnalysis)
+        for bb in fn.get_basic_blocks():
+            open_allocs: list[tuple[IRVariable, IRVariable]] = []
+            ptr_aliases: set[IRVariable] = set()
+            closed_ptrs: set[IRVariable] = set()
+            for inst in bb.instructions:
+                if inst.opcode == "dalloca":
+                    if any(op in closed_ptrs for op in inst.operands):
```
HodanPlodky: operands of `dalloca` should be only size no? So this condition should be always False

vyper/venom/passes/dalloca_lowering.py
```diff
@@ -0,0 +1,419 @@
...
+        new_instructions: list[IRInstruction],
+    ) -> None:
+        mark = inst.operands[0]
+        top = bump_stack[-1] if bump_stack else None
+
+        if top is None or mark != top["mark"]:
+            # Unstructured restore. This is still valid low-level IR; it just
+            # does not qualify for the local scratch rewrite.
+            bump_stack.clear()
+            new_instructions.append(self._restore_fmp_inst(mark, fmp_var, bb, inst))
```
HodanPlodky: Doesn't this create possible use after free?

vyper/venom/passes/dalloca_lowering.py
```diff
@@ -0,0 +1,419 @@
...
+        for inst in dead_chain:
+            inst.parent.remove_instruction(inst)
+        fn.entry.remove_instruction(param_inst)
+        return True
+
+    def _collect_dead_fmp_chain(self, root: IRVariable) -> list[IRInstruction] | None:
+        dfg = self.analyses_cache.request_analysis(DFGAnalysis)
+        dead_insts: list[IRInstruction] = []
+        seen_insts: set[IRInstruction] = set()
+        seen_vars: set[IRVariable] = set()
```
HodanPlodky: Why is there two mechanism for checking loops? The short circuit condition is unreachable (the vars I think)

vyper/venom/passes/dalloca_lowering.py
harkal: It can, but only if the source IR already performs a non-LIFO/invalid restore and then keeps using invalidated pointers. `dfree` is a low-level FMP cursor restore, not lifetime validation. I added comments for this.

vyper/utils.py
```diff
@@ -425,8 +425,6 @@ def evm_pow(x, y):
 
 # memory used for system purposes, not for variables
 class MemoryPositions:
-    # CMC 2025-11-28 these are actually bad names. they should be
```
HodanPlodky, line 422: Why was this removed

vyper/codegen_venom/builtins/create.py
```diff
@@ -537,7 +543,12 @@ def lower_create_from_blueprint(node: vy_ast.Call, ctx: VenomCodegenContext) ->
         args_len = IRLiteral(0)
         args_ptr = IRLiteral(0)
 
-    mem_ofst = ctx.allocate_dyn()
+    # Total length = codesize + args_len (if any).
+    if isinstance(args_len, IRLiteral) and args_len.value == 0:
```
HodanPlodky: We were removing this kind of pre optimiation from frontend, since if this is true it would be removed anyway but this is just to point it out not blocker

vyper/venom/__init__.py
```diff
@@ -129,16 +130,41 @@ def _run_global_passes(
...
+
+    def walk(fn: IRFunction) -> None:
+        if fn in visited:
+            return
+        visited.add(fn)
+
+        for callee in fcg.get_callees(fn):
+            walk(callee)
+
+        DallocaLoweringPass(ir_analyses[fn], fn).run_pass()
```
HodanPlodky: why is it called in global passes?

harkal: mostly to be called before inlining

vyper/utils.py
harkal: accident :)

vyper/venom/__init__.py
HodanPlodky: but if it is called here why would is it called in the optimization passes, since all of the dalloca should be already lowered there?

vyper/venom/function.py
```diff
@@ -26,6 +26,21 @@ class IRFunction:
...
+    # return_pc (if any). Functions on the `initial_fmp` fast path keep this
+    # False.
+    _needs_fmp: bool
+
+    # Internal dynamic-return metadata. A function with `dret` packs its
+    # dynamic return buffers into the caller's entry FMP and returns an
+    # extra hidden adopted FMP value. `_dret_shape` is
+    # (ordinary_return_count, dynamic_return_count).
+    _returns_adopted_fmp: bool
+    _dret_shape: Optional[tuple[int, int]]
```
charles-cooper: these should just be analyses no?

vyper/venom/call_layout.py
```diff
@@ -13,6 +13,68 @@
 # present, lives at the tail.
 
 
+def parse_dret_shape(inst: IRInstruction) -> tuple[int, int] | None:
```
HodanPlodky, line 34: Could it assert instead of returning None, if it is malformed?

vyper/venom/analysis/dynamic_memory.py
```diff
@@ -0,0 +1,199 @@
...
+                    continue
+                if inst.opcode == "dret":
+                    has_dret = True
+                    shape = parse_dret_shape(inst)
+                    if shape is not None:
+                        shapes.add(shape)
+
+        dret_shape = next(iter(shapes)) if len(shapes) == 1 else None
+
+        ret_arities = _ret_arities(fn)
```
HodanPlodky: This could be done in the loop with rest of the checks (with `dalloca` and `dret`)

vyper/venom/analysis/dynamic_memory.py
```diff
@@ -0,0 +1,199 @@
...
+                calls_need_dret_fmp=calls_need_dret_fmp,
+                needs_dret_fmp=needs_dret_fmp[fn],
+                has_physical_hidden_fmp=info.has_physical_hidden_fmp,
+                returns_adopted_fmp=info.returns_adopted_fmp,
+                dret_shape=info.dret_shape,
+                user_return_count=info.user_return_count,
+            )
+
+    def get_info(self, fn: IRFunction) -> DynamicMemoryInfo:
+        return self.infos.get(
```
HodanPlodky: When would we need default, should not info always be there?

vyper/venom/analysis/dynamic_memory.py
```diff
@@ -0,0 +1,199 @@
...
+    def function_returns_adopted_fmp(self, fn: IRFunction) -> bool:
+        return self.get_info(fn).returns_adopted_fmp
+
+    def _iter_callees(self, fn: IRFunction):
+        for bb in fn.get_basic_blocks():
+            for inst in bb.instructions:
+                if inst.opcode != "invoke" or len(inst.operands) == 0:
+                    continue
+                target = inst.operands[0]
+                if not isinstance(target, IRLabel):
```
HodanPlodky: This should be an assert

vyper/venom/analysis/dynamic_memory.py
```diff
@@ -0,0 +1,199 @@
...
+                    continue
+                callee = fn.ctx.functions.get(target)
```
HodanPlodky: this should also be assert

vyper/venom/__init__.py
```diff
@@ -23,7 +23,9 @@
     AlgebraicOptimizationPass,
     AssertEliminationPass,
     BranchOptimizationPass,
+    DallocaLoweringPass,
```
HodanPlodky: unused

harkal:
Redesign rationale: https://gist.github.com/harkal/cb1ae12ee8cc8e09953b83022a7d1cfc

tests/unit/compiler/venom/test_dalloca.py
```diff
@@ -0,0 +1,1901 @@
...
+    # callee desugared: no dret, no new params; getfmp right after the params
+    callee = ctx.get_function(IRLabel("callee"))
+    callee_params = [inst for inst in callee.entry.instructions if inst.opcode == "param"]
+    assert [p.output for p in callee_params] == [IRVariable("%retpc")]
+    assert callee.entry.instructions[len(callee_params)].opcode == "getfmp"
+
+    callee_insts = [inst for bb in callee.get_basic_blocks() for inst in bb.instructions]
+    callee_opcodes = [inst.opcode for inst in callee_insts]
+    assert "dret" not in callee_opcodes
+    assert "mcopy" in callee_opcodes
```
HodanPlodky: this will fail for anything before cancun

vyper/codegen_venom/builtins/system.py
```diff
@@ -146,11 +146,14 @@ def lower_raw_call(node: vy_ast.Call, ctx: VenomCodegenContext) -> Union[IROpera
...
-    # else overwrites the memtop scratch region before the call.
+    # Copy msg.data into the scratch buffer AFTER all kwarg evaluation:
+    # evaluating a kwarg may itself make calls that change msg.data-derived
+    # state or clobber calldata-staging assumptions, so the copy must be the
+    # last thing before the call. (The buffer itself is a tracked `dalloca`
+    # allocation -- owned until reclaim, not borrowed.)
     if use_msg_data:
-        data_ptr = ctx.allocate_dyn()
         data_len = b.calldatasize()
+        data_ptr = ctx.allocate_scratch(data_len)
```
HodanPlodky, line 158: is there a reason to switch the order of getting buffer and `calldatasize` or not important?

vyper/venom/analysis/dfg.py
```diff
@@ -42,8 +42,8 @@ def get_transitive_uses(self, inst: IRInstruction) -> OrderedSet[IRInstruction]:
             if curr in result:
                 continue
             result.add(curr)
-            if curr.has_outputs:
-                worklist.extend(self.get_uses(curr.output))
+            for output in curr.get_outputs():
```
HodanPlodky, line 45: was this latent bug?

vyper/venom/analysis/dynamic_memory.py
```diff
@@ -0,0 +1,117 @@
...
+                # gains a hidden FMP param via the closure nor loses one
+                needs_fmp[fn] = sig.has_fmp_param or sig.publishes
+                publishes[fn] = sig.publishes
+            else:
+                needs_fmp[fn] = self._touches_fmp(fn)
+                publishes[fn] = self._publishes(fn)
+
+        # order-insensitive transitive fixpoint over invokes
+        changed = True
+        while changed:
```
HodanPlodky, line 69: Dont we have order since there should not be recursion (what I mean why is the worklist necessary)?

vyper/venom/analysis/liveness.py
```diff
@@ -113,24 +113,24 @@ def input_vars_from(self, source: IRBasicBlock, target: IRBasicBlock) -> Ordered
         operand_to_phi_idx: dict[IROperand, int] = {}
         phi_matching: dict[int, IRVariable] = {}
```
HodanPlodky, line 114: Are the changes here relevant I only see renames and assert move and not something relevant to PR

vyper/venom/optimization_levels/O3.py
```diff
@@ -79,6 +81,11 @@
     AssignElimination,
     RemoveUnusedVariablesPass,
     ConcretizeMemLocPass,
+    MakeSSA,
```
HodanPlodky: Why is the `MakeSSA` necessary `ConcretizeMemLocPass`

HodanPlodky: same question for the rest of the O levels

vyper/venom/passes/internal_return_copy_forwarding.py
```diff
@@ -157,7 +164,16 @@ def _invoke_may_clobber_src(self, invoke_inst: IRInstruction, src_loc: MemoryLoc
...
         return False
 
-    def _invoke_operand_may_write(self, invoke_inst: IRInstruction, operand_pos: int) -> bool:
-        if operand_pos == 1 and self._invoke_has_return_buffer(invoke_inst):
+    def _invoke_operand_may_write(
+        self, layout: InvokeLayout, operand_pos: int, retbuf_pos: int | None
+    ) -> bool:
+        if operand_pos == retbuf_pos:
+            return True
+        arg_idx = layout.user_arg_index(operand_pos)
```
HodanPlodky, line 175: why is it not using `not self._is_readonly_invoke_operand(invoke_inst, operand_pos)` like before what is the difference?

banteg:
FYI: I opened #5071 for a current-master `memtop`/DFT ordering bug. DFT can move `memtop` before a prior `mload`:

```text
before: source, mload, memtop, sink
after:  source, memtop, mload, sink
```

Since this PR removes the old `memtop`/`MSIZE` dynamic allocation model, it may obsolete #5071 if it lands first. If any equivalent FMP/high-water read remains after this PR, it should have a dependency on prior memory-expanding accesses, not only prior memory writes.

vyper/codegen_venom/builtins/system.py
harkal: The reorder is forced by the API change rather than a choice: the old allocate_dyn() took no size (it lowered to memtop/MSIZE), so the two statements were independent and the order was arbitrary. The new allocate_scratch(data_len) is an exact-size allocation (lowers to dalloca data_len), so the size is now an operand of the allocation and calldatasize has to be emitted first as a plain data dependency.

vyper/venom/analysis/dfg.py
harkal: yes!

vyper/venom/optimization_levels/O3.py
harkal: defensive decision for passes up to this one. I think it's ok to remove it, since you also think so.

vyper/venom/passes/internal_return_copy_forwarding.py
harkal: No difference in behavior. It's the same logic inlined. `_is_readonly_invoke_operand` builds a fresh `InvokeLayout` per call, so calling it per operand rebuilt the layout in a loop; this reuses the one layout already in hand. The unknown-callee case is also unchanged: old not is_readonly -> may-write, new `True` directly.

vyper/venom/analysis/dynamic_memory.py
harkal: future proofing, for supporting recursion (even at venom level)

charles-cooper (approved):

PR #4923, opened 2026-04-20
feat[stdlib]!: move `isqrt` from builtin to stdlib math module

charles-cooper:

### What I did

Moved `isqrt` from a builtin function to the `math` stdlib module, matching the existing pattern for `sqrt`. The bare `isqrt()` builtin now raises a deprecation error directing users to `import math` and use `math.isqrt()`.

### How I did it

- Added `isqrt(x: uint256) -> uint256` to `vyper/builtins/stdlib/math.vy` using the same Babylonian method as the legacy IR implementation
- Converted the `ISqrt` builtin class into a deprecation stub that raises `UnimplementedException` with a helpful message (same pattern as `Sqrt`)
- Removed `lower_isqrt` from venom codegen (`codegen_venom/builtins/misc.py`)
- Updated tests and CurveFi examples to use `import math` + `math.isqrt()`

### How to verify it

```bash
./quicktest.sh tests/functional/codegen/types/numbers/test_isqrt.py -n0
./quicktest.sh tests/functional/codegen/modules/ -n32 -m "not fuzzing"
```

Confirm bare `isqrt()` fails:
```bash
cat > /tmp/test.vy << 'EOF'
@external
def test(a: uint256) -> uint256:
    return isqrt(a)
EOF
vyc /tmp/test.vy  # should error with deprecation message
```

### Commit message

```
`isqrt` was the last integer math function implemented as a builtin
with hand-written IR. `sqrt` was already moved to the stdlib `math`
module in a prior release. Move `isqrt` alongside it so `math` is the
single entry point for all math functions.

The bare `isqrt()` builtin now raises a deprecation error directing
users to `import math` and use `math.isqrt()`, matching the `sqrt`
deprecation pattern. Removed the venom `lower_isqrt` codegen path
since the stdlib implementation is compiled like any other Vyper
function.
```

Move `isqrt` from builtin to `math` stdlib module. Use `import math; math.isqrt(x)` instead of bare `isqrt(x)`.

### Cute Animal Picture

Red panda


vyper/builtins/stdlib/math.vy
```diff
@@ -1,3 +1,37 @@
...
+        y = y >> 32
+        z = z << 16
+    if y >= 2 ** 24:
+        y = y >> 16
+        z = z << 8
+
+    z = z * (y + 65536) // 4 ** 9
+
+    for i: uint256 in range(7):
+        z = (x // z + z) // 2
```
Sporarum: Nit: I'd prefer it not depend on order of operations, something like:

```suggestion
    z = z * (y + 65536) // (4 ** 9)

    for i: uint256 in range(7):
        z = ((x // z) + z) // 2
```

vyper/builtins/stdlib/math.vy
```diff
@@ -1,3 +1,37 @@
...
+        z = z << 64
+    if y >= 2 ** 72:
+        y = y >> 64
+        z = z << 32
+    if y >= 2 ** 40:
+        y = y >> 32
+        z = z << 16
+    if y >= 2 ** 24:
+        y = y >> 16
+        z = z << 8
```
Sporarum: Nit: I preferred the previous parenthesization 
```suggestion
    if y >= 2 ** (128 + 8):
        y = y >> 128
        z = z << 64
    if y >= 2 ** (64 + 8):
        y = y >> 64
        z = z << 32
    if y >= 2 ** (32 + 8):
        y = y >> 32
        z = z << 16
    if y >= 2 ** (16 + 8):
        y = y >> 16
        z = z << 8
```

vyper/builtins/stdlib/math.vy
```diff
@@ -1,3 +1,37 @@
...
+
+    z = z * (y + 65536) // 4 ** 9
+
+    for i: uint256 in range(7):
+        z = (x // z + z) // 2
+
+    t: uint256 = x // z
+    return min(z, t)
```
Sporarum, line 35: Add back comment:

```suggestion
    # note: If ``x+1`` is a perfect square, then the Babylonian
    # algorithm oscillates between floor(sqrt(x)) and ceil(sqrt(x)) in
    # consecutive iterations. return the floor value always.
    return min(z, t)
```

Sporarum (commented):
I think the docs should also be updated in this PR, so they stay in sync

I can't guarantee the square root algorithm is correct, but as far as I can tell, it is a correct port of the previous logic

I would like a test for the deprecation (one is even mentioned in the PR body)

Couple of nitpicks, see below

charles-cooper:
> I can't guarantee the square root algorithm is correct, but as far as I can tell, it is a correct port of the previous logic

i would say this is out of scope, eventually we will formally verify the stdlib

pcaversaccio (commented):
we should also update the docs here:
- https://docs.vyperlang.org/en/stable/built-in-functions.html#isqrt

to

```vy
import math

@external
@view
def foo(x: uint256) -> uint256:
    return math.isqrt(x)
```


Also, "For instance, `math.isqrt(101) == 10`".

pcaversaccio:
Hmm, now thinking about the docs; we have this section here: https://docs.vyperlang.org/en/stable/built-in-functions.html#math

we might wanna have a different section on the `math` stdlib and put the `isqrt` there. wdyt?

vyper/builtins/functions.py
```diff
@@ -2091,49 +2090,10 @@ class ISqrt(BuiltinFunctionT):
...
-            # note: If ``x+1`` is a perfect square, then the Babylonian
-            # algorithm oscillates between floor(sqrt(x)) and ceil(sqrt(x)) in
-            # consecutive iterations. return the floor value always.
-
-            ret.append(["with", "t", ["div", x, z], ["select", ["lt", z, "t"], z, "t"]])
-
-            ret = ["with", y, x, ["with", z, 181, ret]]
-            return b1.resolve(IRnode.from_list(ret, typ=UINT256_T))
+    def fetch_call_return(self, node):
+        message = "The `isqrt` builtin was removed. Instead import module "
```
pcaversaccio, line 2094: can we test this pls? otherwise PR lgtm (pending discussion around the proper structure for the docs)

pcaversaccio:
oh another nit: i think this PR title deserves a `!`, i.e. "feat[stdlib]!: move isqrt from builtin to math module" since it's breaking technically

docs/built-in-functions.rst
```diff
@@ -665,12 +665,18 @@ Math
 
     Return the square root of the provided decimal number, using the Babylonian square root algorithm. The rounding mode is to round down to the nearest epsilon. For instance, ``sqrt(0.9999999998) == 0.9999999998``.
 
+    .. note::
+
+        ``sqrt`` has been moved to the ``math`` stdlib module. See :ref:`stdlib-math`. Import it with ``import math`` and call ``math.sqrt(d)``.
```
pcaversaccio, line 670:
```suggestion
        ``sqrt`` has been moved to the ``math`` stdlib module as part of the ``0.4.2`` release (see `PR #4520 <https://github.com/vyperlang/vyper/pull/4520>`_). See :ref:`stdlib-math`. Import it with ``import math`` and call ``math.sqrt(d)``.
```

docs/built-in-functions.rst
```diff
@@ -681,12 +687,18 @@ Math
 
     Return the (integer) square root of the provided integer number, using the Babylonian square root algorithm. The rounding mode is to round down to the nearest integer. For instance, ``isqrt(101) == 10``.
 
+    .. note::
+
+        ``isqrt`` has been moved to the ``math`` stdlib module. See :ref:`stdlib-math`. Import it with ``import math`` and call ``math.isqrt(x)``.
```
pcaversaccio, line 692:
```suggestion
        ``isqrt`` has been moved to the ``math`` stdlib module as part of the ``0.5.0`` release (see `PR #4923 <https://github.com/vyperlang/vyper/pull/4923>`_). See :ref:`stdlib-math`. Import it with ``import math`` and call ``math.isqrt(x)``.
```

docs/built-in-functions.rst
Sporarum: Is this necessary ?
Since the docs are versioned

pcaversaccio: i never read outdated docs, always the `stable` version and it's good to know when a certain feature was moved.

vyper/builtins/functions.py
Sporarum: Done

pcaversaccio (approved):

PR #4932, opened 2026-04-24
fix[build]!: remove dev dependencies from published package

### LINKED ISSUE #4929: Remove dev dependencies from the published package
It is currently possible to download vyper's dev dependencies with things like `pip install vyper[dev]`.
This means the dev dependencies are actually packaged with the rest of vyper, offering little to no benefit.

[PEP-735](https://peps.python.org/pep-0735/) introduced a way to specify dev dependencies in a way which does not get packaged and published, by moving dependencies to `[dependency-groups]` in `pyproject.toml`.

Both `pip>=25.1` and `uv` are compatible with this PEP by using the `--group <group name>` flag.
(Additionnally, `uv sync` automatically includes the `dev` group, i.e. `uv sync` and `uv sync --group dev` are equivalent.)

Given the above, we should transition from `optional-dependencies` to `dependency-groups`.

Sporarum:
### What I did

Fix #4929 

### How I did it

Moved optional dependencies to dependency groups, which do not get packaged

This is therefore a breaking change, `pip install vyper[test]` (or `dev` or `lint`) will no longer work!
(We expect the effect to be negligible to null, as we don't document this anywhere, and do not know of anyone using it)

Updated all consumers (CI workflows, Dockerfile, Makefile, readthedocs config) to use `pip install . --group <name>` instead of `pip install .[name]`.

### Commit message

```
development dependencies (test, lint, docs, dev) were declared as
[project.optional-dependencies] in pyproject.toml, which means they were
bundled into the published wheel on PyPI. anyone installing vyper got
these dev-only packages transitively, inflating the package size and
pulling in irrelevant dependencies like pytest, pre-commit, and
pyinstaller.

move dev dependencies to [dependency-groups], which pip (>=25.1)
installs via `pip install . --group <name>` rather than the extras
syntax `pip install .[name]`. this is a breaking change for anyone using
`pip install vyper[test]` (or dev/lint/docs), but the impact should be
negligible since these extras were never documented.

update all consumers: CI workflows, Dockerfile, Makefile, and
readthedocs config to use the new `--group` syntax. the dev group now
uses `{include-group = ...}` cross-references instead of the self-
referential `vyper[test,lint,docs]` hack that existed only to bridge the
gap between uv's dependency-groups and pip's optional-dependencies.

closes GH 4929
```

shrink published package by not bundling development dependencies

### Cute Animal Picture


Sporarum:
Appears to be working well, needs more checks

(Reference for me only:
`nix-shell -p python311`
in a `maketmp --directory`
by running `pip install --upgrade pip` inside the `venv`
)

charles-cooper (approved):

PR #4875, opened 2026-03-05
feat[lang]: add abstract methods

Sporarum:
### What I did

Implement "initializes-based proposal v2" of GH 4730 — add `@abstract` and `@override` decorators for module methods.

Abstract methods have no body (written as `def foo() -> uint256: ...`) and must be overridden in another module that `initializes` the abstract module. Calling an abstract method dispatches to its concrete override at compile time. Calling an abstract function requires `uses`-ing its module, which forces the abstract module to be `initializes`-d (and hence overridden) somewhere in the import tree.

**Key changes:**
- `@abstract` / `@override` decorators on `FunctionDef` nodes
- `ContractFunctionT.is_abstract`, `overridden_by`, `get_concrete_override()` 
- Decoupled call graph from import graph: `_build_call_graph_edges` / `_analyze_call_graph` replaced the merged `_analyze_call_graph`
- Override validation: signature matching via `structurally_equivalent()`, single-override rule, duty-cycle checks
- `analyze_module` → `analyze_modules` (takes `ImportAnalyzer`, processes all modules in import order with separate phases: type computation → override checking → call graph → function bodies)
- Type comparison: added `is_supertype_of`, `is_subtype_of`, `is_equivalent_to` to `VyperType`; disambiguated `compare_type` usages
- Event collection moved from AST traversal to `FunctionAnalyzer.visit_Log` → `ContractFunctionT._emitted_events`; `used_events` now computed from reachable functions
- `ModuleT.reachable_functions` factored out; ABI output and metadata now use it instead of duplicating logic
- `ExprInfo._writes`/`_reads` changed from post-init mutation to `field(default_factory=OrderedSet)` for proper dataclass semantics

Also updates `skills/contributing.md` to clarify that the PR template's "Commit message" section should contain only the body (the subject line comes from the PR title on squash-merge).

No docs yet — will be added in a follow-up.

### How I did it

The core design: abstract functions are resolved to their concrete override at analysis time, not runtime. When building the call graph, calls to abstract methods are rewritten to reference `fn.get_concrete_override()`. This means the call graph, reachable set, and used-modules set all reflect the override's effects — no virtual dispatch needed.

Analysis is structured in phases to handle cross-module dependencies:
1. `_compute_module_type_r` — collect members, build `ModuleT` (no cross-module info needed)
2. `_check_overrides` — validate `@override` signatures and duty-cycle rules (needs `ModuleT` of both abstract and override module)
[...]

PR #4935, opened 2026-04-24
chore[docs]: document @abstract and @override decorators

lufa23:
### What I did

Documented the `@abstract` and `@override` decorators added in [GH 4875](https://github.com/vyperlang/vyper/pull/4875). The feature landed without user-facing docs; this adds a new "Abstract methods and overrides" section in `using-modules.rst` and cross-references it from three adjacent pages.

### How I did it

- New section in `docs/using-modules.rst` placed between *Initializing a module with dependencies* and *Exporting functions*. Opens with a worked two-module example (`hook.vy` + `token.vy`), then four subsections: Rules, Signature compatibility, Dispatch and call paths, Abstract methods versus `.vyi` interfaces.
- Two rows added to the Decorators Reference table in `docs/control-structures.rst`.
- One sentence appended to the Modules subsection of `docs/structure-of-a-contract.rst` pointing to the new section.
- Two-sentence entry added under *No Class Inheritance* in `docs/solidity-differences.rst` contrasting with Solidity's `abstract contract` / `virtual` / `override`.

### How to verify it

```bash
# Sphinx build produces no new warnings on the edited files
uv run --with-requirements requirements-docs.txt \
  sphinx-build -b html docs /tmp/vyper-docs

# The worked example compiles as-is
mkdir -p /tmp/docs-ex
# (copy hook.vy and token.vy from the new section into /tmp/docs-ex/)
vyper /tmp/docs-ex/token.vy -p /tmp/docs-ex
```

### Commit message

```
GH 4875 merged the @abstract and @override decorators without user-
facing documentation. Add a new "Abstract methods and overrides" section
in using-modules.rst covering syntax, the override-coverage and
initialization rules, signature compatibility (parameter names and
sizes, return type, mutability ordering, @nonreentrant, default values),
compile-time dispatch with the shortest-reachable-path rule, and the
distinction from .vyi external interfaces. The worked example compiles
as-is.

Add cross-references from the Decorators Reference table in control-
structures.rst, the Modules subsection of structure-of-a-contract.rst,
and the No Class Inheritance section of solidity-differences.rst so
readers arriving from those entry points land in the new section.
```

Documented the `@abstract` and `@override` decorators.

### DIFF
--- a/docs/solidity-differences.rst
+++ b/docs/solidity-differences.rst
@@ -98,6 +98,8 @@ Three declarations manage module relationships: ``initializes`` (this contract m
 
 A contract can be understood by reading one file and its direct imports; dependencies and what is exposed in the external function table are explicit.
 
+Solidity's ``abstract contract`` and ``virtual`` / ``override`` keywords belong to its inheritance model. Vyper's ``@abstract`` and ``@override`` decorators apply only to internal module functions and are resolved at compile time against a single initializing module, so there is no virtual dispatch and no method-resolution order. See :ref:`abstract-methods` for details.
+
 No Inline Assembly
 ==================
 
--- a/docs/structure-of-a-contract.rst
+++ b/docs/structure-of-a-contract.rst
@@ -208,7 +208,7 @@ Modules can be added to contracts by importing them from a ``.vy`` file. Any ``.
     def foo() -> uint256:
         return my_module.perform_some_computation()
 
-Modules are opt-in by design. That is, any operations involving state or exposing external functions must be explicitly opted into using the ``exports``, ``uses`` or ``initializes`` keywords. See the :ref:`Modules <modules>` documentation for more information.
+Modules are opt-in by design. That is, any operations involving state or exposing external functions must be explicitly opted into using the ``exports``, ``uses`` or ``initializes`` keywords. A module can also declare internal functions whose bodies another module must supply, using the :ref:`@abstract and @override decorators <abstract-methods>`. See the :ref:`Modules <modules>` documentation for more information.
 
 Events
 ======
--- a/docs/using-modules.rst
+++ b/docs/using-modules.rst
@@ -189,6 +189,92 @@ Sometimes, you may encounter a module which itself ``uses`` other modules. Vyper
 .. warning::
     In normal usage, you should make sure that ``__init__()`` functions are called in dependency order. In the above example, you can get unexpected behavior if ``ownable_2step.__init__()`` is called before ``ownable.__init__()``! The compiler may enforce this behavior in the future.
 
+.. _abstract-methods:
+
+Abstract methods and overrides
+==============================
+
+A module can declare an **internal** function whose body another module must supply. The declaration uses the ``@abstract`` decorator and leaves the body as ``...``. A concrete function in the module that initializes the abstract's module supplies the body, marked with ``@override``. Every call to an abstract function is resolved to its concrete override at compile time; there is no runtime dispatch table.
+
+The following module declares a transfer hook and calls it from its own code:
+
+.. code-block:: vyper
+
+    # hook.vy
+
+    event Transfer:
+        sender: indexed(address)
+        recipient: indexed(address)
+        amount: uint256
+
+    @external
+    def transfer(to: address, amount: uint256):
+        # ... transfer logic
+        self._on_transfer(to, amount)
+
+    @abstract
+    def _on_transfer(to: address, amount: uint256): ...
+
+A module that initializes ``hook`` must provide ``_on_transfer``:
+
+.. code-block:: vyper
+
+    # token.vy
+
+    import hook
+
+    initializes: hook
+
+    exports: hook.transfer
+
+    @override(hook)
+    def _on_transfer(to: address, amount: uint256):
+        log hook.Transfer(sender=msg.sender, recipient=to, amount=amount)
+
+The call to ``self._on_transfer`` inside ``hook.transfer`` resolves at compile time to the concrete body in ``token``. Compiling ``token.vy`` produces a contract whose ``transfer`` entry runs the override's body.
+
+Rules
+-----
+
+* ``@abstract`` and ``@override`` apply only to internal functions. They are rejected on ``@external``, on ``@deploy``, and in ``.vyi`` interface files.
+* An ``@abstract`` function's body must be ``...``. A leading docstring is allowed.
+* A module that declares ``@override(M)`` must also declare ``initializes: M``.
+* A module that initializes another module which contains abstract functions must provide exactly one override for each of them. Zero overrides raise ``Abstract function was not overridden``. A duplicate override raises ``... was already overridden in ...``.
+* A single function may carry several ``@override(...)`` decorators to override abstracts from multiple modules, as long as the containing module initializes each one. The overriding function's name must match the abstract's name in every listed module; a mismatch is reported as ``Tried to override M.name, but it does not exist``.
+* Abstract functions can only live in imported modules. A top-level contract that declares ``@abstract`` has no module above it to supply the override, so it fails the override-coverage rule with ``Abstract function was not overridden`` pointing at the abstract's own line.
+
+Signature compatibility
+-----------------------
+
+The compiler checks each override's signature against the abstract's:
+
+* Parameter names must match, in order.
+* For parameters whose type takes a size (``String[N]``, ``Bytes[N]``, ``DynArray[_, N]``), the override's ``N`` must be greater than or equal to the abstract's. For other parameter types the two must be the same.
+* The return type must match, or for sized types the override's ``N`` must be less than or equal to the abstract's.
+* Mutability on the override may be the same or stricter, in the order ``payable > nonpayable > view > pure``.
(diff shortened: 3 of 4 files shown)


docs/using-modules.rst
```diff
@@ -189,6 +189,92 @@ Sometimes, you may encounter a module which itself ``uses`` other modules. Vyper
 .. warning::
     In normal usage, you should make sure that ``__init__()`` functions are called in dependency order. In the above example, you can get unexpected behavior if ``ownable_2step.__init__()`` is called before ``ownable.__init__()``! The compiler may enforce this behavior in the future.
 
+.. _abstract-methods:
+
+Abstract methods and overrides
+==============================
+
+A module can declare an **internal** function whose body another module must supply. The declaration uses the ``@abstract`` decorator and leaves the body as ``...``. A concrete function in the module that initializes the abstract's module supplies the body, marked with ``@override``. Every call to an abstract function is resolved to its concrete override at compile time; there is no runtime dispatch table.
```
Sporarum, line 197: Since methods can be both abstract and override, this sentence is misleading

```suggestion
A module can declare an **internal** function whose body another module must supply. The declaration uses the ``@abstract`` decorator and leaves the body as ``...``. A function in the module that initializes the abstract's module supplies the body, marked with ``@override``. Every call to an abstract function is resolved to its concrete override at compile time; there is no runtime dispatch table.
```

docs/structure-of-a-contract.rst
```diff
@@ -208,7 +208,7 @@ Modules can be added to contracts by importing them from a ``.vy`` file. Any ``.
     def foo() -> uint256:
         return my_module.perform_some_computation()
 
-Modules are opt-in by design. That is, any operations involving state or exposing external functions must be explicitly opted into using the ``exports``, ``uses`` or ``initializes`` keywords. See the :ref:`Modules <modules>` documentation for more information.
+Modules are opt-in by design. That is, any operations involving state or exposing external functions must be explicitly opted into using the ``exports``, ``uses`` or ``initializes`` keywords. A module can also declare internal functions whose bodies another module must supply, using the :ref:`@abstract and @override decorators <abstract-methods>`. See the :ref:`Modules <modules>` documentation for more information.
```
Sporarum, line 211: Weird flow

docs/solidity-differences.rst
```diff
@@ -98,6 +98,8 @@ Three declarations manage module relationships: ``initializes`` (this contract m
 
 A contract can be understood by reading one file and its direct imports; dependencies and what is exposed in the external function table are explicit.
 
+Solidity's ``abstract contract`` and ``virtual`` / ``override`` keywords belong to its inheritance model. Vyper's ``@abstract`` and ``@override`` decorators apply only to internal module functions and are resolved at compile time against a single initializing module, so there is no virtual dispatch and no method-resolution order. See :ref:`abstract-methods` for details.
```
Sporarum, line 101: The contrast between "inheritance model" and "apply only to internal module ..." is a bit confusing

It kinda implies Vyper does not have an inheritance model
(which one could argue is wrong, it just has a simple, compile-time only one)

docs/using-modules.rst
```diff
@@ -189,6 +189,92 @@ Sometimes, you may encounter a module which itself ``uses`` other modules. Vyper
...
+
+    initializes: hook
+
+    exports: hook.transfer
+
+    @override(hook)
+    def _on_transfer(to: address, amount: uint256):
+        log hook.Transfer(sender=msg.sender, recipient=to, amount=amount)
+
+The call to ``self._on_transfer`` inside ``hook.transfer`` resolves at compile time to the concrete body in ``token``. Compiling ``token.vy`` produces a contract whose ``transfer`` entry runs the override's body.
```
Sporarum, line 234: Very clearly explained !

docs/using-modules.rst
```diff
@@ -189,6 +189,92 @@ Sometimes, you may encounter a module which itself ``uses`` other modules. Vyper
...
+
+Rules
+-----
+
+* ``@abstract`` and ``@override`` apply only to internal functions. They are rejected on ``@external``, on ``@deploy``, and in ``.vyi`` interface files.
+* An ``@abstract`` function's body must be ``...``. A leading docstring is allowed.
+* A module that declares ``@override(M)`` must also declare ``initializes: M``.
+* A module that initializes another module which contains abstract functions must provide exactly one override for each of them. Zero overrides raise ``Abstract function was not overridden``. A duplicate override raises ``... was already overridden in ...``.
```
Sporarum, line 242: I think the part about error messages is not really necessary:

```suggestion
* A module that initializes another module which contains abstract functions must provide exactly one override for each of them.
```

docs/using-modules.rst
```diff
@@ -189,6 +189,92 @@ Sometimes, you may encounter a module which itself ``uses`` other modules. Vyper
...
+* A single function may carry several ``@override(...)`` decorators to override abstracts from multiple modules, as long as the containing module initializes each one. The overriding function's name must match the abstract's name in every listed module; a mismatch is reported as ``Tried to override M.name, but it does not exist``.
+* Abstract functions can only live in imported modules. A top-level contract that declares ``@abstract`` has no module above it to supply the override, so it fails the override-coverage rule with ``Abstract function was not overridden`` pointing at the abstract's own line.
```
Sporarum, line 244: "override-coverage rule" seems very anchored in the compiler implementation

docs/using-modules.rst
```diff
@@ -189,6 +189,92 @@ Sometimes, you may encounter a module which itself ``uses`` other modules. Vyper
...
+Signature compatibility
+-----------------------
+
+The compiler checks each override's signature against the abstract's:
+
+* Parameter names must match, in order.
+* For parameters whose type takes a size (``String[N]``, ``Bytes[N]``, ``DynArray[_, N]``), the override's ``N`` must be greater than or equal to the abstract's. For other parameter types the two must be the same.
+* The return type must match, or for sized types the override's ``N`` must be less than or equal to the abstract's.
+* Mutability on the override may be the same or stricter, in the order ``payable > nonpayable > view > pure``.
+* ``@nonreentrant`` must be present on both or on neither.
```
Sporarum, line 255: Technically more subtle, reentrancy must match, but this might come from `@reentrant` if we're in "non-reentrant by default" mode

docs/using-modules.rst
Sporarum: Also applies more broadly to this whole section

docs/using-modules.rst
```diff
@@ -189,6 +189,92 @@ Sometimes, you may encounter a module which itself ``uses`` other modules. Vyper
...
+When a signature fails these rules, the compiler reports ``Override parameter mismatch: Got X, but expected Y (or more general)``.
+
+Dispatch and call paths
+-----------------------
+
+Calls to an abstract function resolve at compile time to the concrete override. There is no virtual dispatch.
+
+To call an abstract function from a module other than the one that declared it, add ``uses:`` for that module at the top of the calling file. Omitting this raises ``Cannot access abstract methods of 'X'`` with the hint ``add 'uses: X' as a top-level statement to your contract``. If a shorter path to the override is already reachable from the caller (through ``self``, or through a module the caller has initialized), the compiler requires that shorter path instead and reports ``Abstract method `X` is overridden by `Y`, call that instead.``
+
+Call sites see the **abstract's** signature, not the override's. An abstract declared without ``@view`` cannot be called from a ``@view`` caller, even when the concrete override happens to be ``@view``. Mark the abstract with the mutability its call sites require.
```
Sporarum, line 267: Example would be clearer with a concrete decorator instead of "without view"

docs/using-modules.rst
```diff
@@ -189,6 +189,92 @@ Sometimes, you may encounter a module which itself ``uses`` other modules. Vyper
...
+
+The compiler tracks an override's side effects. Events declared in a module that contains abstract functions and emitted only inside an override still appear in the compiled contract's ABI. Reads and writes performed by an override count as accesses on the initializing module.
```
Sporarum, line 269: Seems very implementation-y, for example I would expect users not to associate side-effects with the ABI

I think users would just expect that any event which is used just shows up in the ABI, it maybe doesn't even bear mentionning wrt abstract methods

docs/using-modules.rst
```diff
@@ -189,6 +189,92 @@ Sometimes, you may encounter a module which itself ``uses`` other modules. Vyper
...
+
+An overriding function may itself be marked ``@abstract``, stacking ``@abstract`` with ``@override(parent)``. The concrete body then lives in a further module down the import tree. This is a niche pattern; typical code provides a concrete override directly.
```
Sporarum, line 271: Hard to tell for the "niche pattern", I think a simpler way of explaining it is to say it delegates the overriding of the abstract method

docs/using-modules.rst
```diff
@@ -189,6 +189,92 @@ Sometimes, you may encounter a module which itself ``uses`` other modules. Vyper
...
+
+Abstract methods versus ``.vyi`` interfaces
```
Sporarum, line 273: Maybe would better fit closer to the top of this section ?

Sporarum (changes requested):
Nice work, see comments below

Sporarum:
Superseded by #4946

PR #4940, opened 2026-04-25
Problem: function copy don't work

yihuang:
### What I did

Fix `IRFunction` copy method to preserve necessary fields.

### Commit message

### Cute Animal Picture

### DIFF
--- a/vyper/venom/function.py
+++ b/vyper/venom/function.py
@@ -128,8 +128,11 @@ class IRFunction:
 
     def copy(self):
         new = IRFunction(self.name)
+        new.clear_basic_blocks()
+        new.last_variable = self.last_variable
         for bb in self.get_basic_blocks():
             new_bb = bb.copy()
+            new_bb.parent = new
             new.append_basic_block(new_bb)
 
         return new


harkal (commented):
This method is essentially dead code. We should probably remove it all-together.

yihuang:
> This method is essentially dead code. We should probably remove it all-together.

we are experimenting with the API to build EVM bytecodes programmatically, we are not sure if it's the right approach yet, so no strong opinions on this.

banteg:
Audit note: this PR addresses the immediate `IRFunction.copy()` assertion by clearing the auto-created entry block before appending copied blocks, preserving `last_variable`, and fixing the copied block parent pointers.

There are two remaining details if we keep this helper instead of deleting it:

- `ctx` is not preserved. A copied function whose blocks/instructions refer back through function context can still behave differently from the original.
- Calling-convention metadata is not preserved: `_invoke_param_count` and `_has_memory_return_buffer_param`. If `copy()` is later used by inlining/copy-forwarding experiments, the clone can lose invoke/return-buffer shape information even though the block copy succeeded.

So I think the decision is either:

1. remove `IRFunction.copy()` as dead API, as suggested above; or
2. make the clone complete enough to preserve `ctx`, `last_variable`, block parents, and the invoke/memory-return metadata, with a small regression that calls `copy()` on a function with at least one real block and non-default metadata.

The current patch is enough for the duplicate-label assert, but not quite a complete semantic clone yet.

charles-cooper:
@harkal should we go with the approach in this PR or remove the code entirely?

charles-cooper:
superseded by https://github.com/vyperlang/vyper/pull/5225

PR #4942, opened 2026-04-27
feat[venom]: memmerge abstract

HodanPlodky:
### What I did
Added support for memory allocated through `alloca` to `MemMergePass`

### Commit message

### Cute Animal Picture

### DIFF
--- a/vyper/venom/passes/concretize_mem_loc.py
+++ b/vyper/venom/passes/concretize_mem_loc.py
@@ -66,7 +66,7 @@ class ConcretizeMemLocPass(IRPass):
                 #   %3 = add 16, %1
                 #   %cond = gt %2, %3
                 #   assert %cond
-                # this can happends when other uses removed but the 
+                # this can happends when other uses removed but the
                 # bounds check remain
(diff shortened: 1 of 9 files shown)


vyper/venom/passes/concretize_mem_loc.py
```diff
@@ -56,6 +56,12 @@ def _handle_bb(self, bb: IRBasicBlock):
             if inst.opcode == "alloca":
                 base_ptr = self.base_ptrs.ptr_from_op(inst.output)
                 assert base_ptr is not None, f"alloca without base ptr: {inst}"
+
+                # there can be the case where the allocation is only used
+                # in assert which may be can be removed but for now
+                # lets just allocate it
```
harkal, line 62: Can we give a better explanation here?

harkal (changes requested):
The frontend will currently not mix abstract / concrete, but should we just make it clear at the begining of the pass by asserting it? Or make it accept a literal but make it a hard barrier so we are future proof?

harkal (approved):

PR #4944, opened 2026-04-29
fix[venom]: fix venom invoke copy-forwarding gas regressions

harkal:
### What I did

- Unified common copy-forwarding logic behind a shared `CopyForwardingPolicy`.
- Replaced readonly invoke-argument fixed thresholds with a profitability-aware memory-expansion guard.
- Fixed large Timelock gas regressions by avoiding false-positive liveness-extension blocking.
- Added a retained-copy cost comparison so the guard does not preserve huge staged copies without proof they are cheaper.
- Kept conservative handling for source-write early-exit cases, preserving the ERC1155 early-revert gas improvements.

### How I did it

`MemoryCopyElisionPass`, `InternalReturnCopyForwardingPass`, and
`ReadonlyInvokeArgCopyForwardingPass` now share policy helpers for copy
size/source handling, copy equivalence, destination alloca lookup, and invoke
forwarding cost decisions.

The readonly invoke-arg guard now only blocks forwarding when it can justify
keeping the staged copy. It checks whether forwarding would actually extend the
source allocation across the invoke, estimates the source memory-expansion
penalty across the callee frame, and compares that against the retained-copy
cost.

For read-only source accesses, the retained-copy cost includes the removed
`MCOPY` runtime/deploy savings plus the destination buffer's own callee-frame
memory expansion. This prevents unknown user contracts from keeping a very large
staged copy unless the model has evidence that forwarding is worse.

Source-write cases remain conservative because those correspond to decoder/setup
work on paths that may skip the invoke. That preserves the ERC1155 early-exit
shape where forwarding can otherwise pull expensive callee-frame memory into
revert paths.

The path analysis was also tightened: an access only counts as cross-path
evidence if it can both reach the copy and reach an exit path without executing
that copy. Invoke operands are treated as source accesses, and invalid Venom
shapes now assert instead of silently falling back.

### Commit message

```
Venom has several copy-forwarding paths that can remove temporary
memory staging copies, including readonly invoke-argument forwarding.
The old invoke-arg guard was too one-sided: it estimated the potential
memory-expansion cost of forwarding a source allocation across a callee
frame, but it did not account for the cost of keeping the staged
destination copy.

That caused large gas regressions in Timelock batch paths. The pass
kept huge staged `MCOPY`s before `_hash_operation_batch` even though the
source payload buffer was already live after the invoke. In that case,
forwarding did not actually extend source liveness, so preserving the
copy only added work.

Teach `CopyForwardingPolicy` to make a more complete decision:

- detect whether forwarding would actually extend source liveness across
  the invoke;
- treat invoke operands as source accesses for the skip-path analysis;
- require a source access to both reach the copy and have a path that
  skips the copy before it counts as cross-path evidence;
- estimate the source memory-expansion penalty across the callee frame;
- compare that penalty with the retained staged-copy cost, including the
  removed `MCOPY` runtime/deploy savings and, for read-only cases, the
  destination buffer's own callee-frame expansion;
- keep source-write cases conservative so decoder/setup writes on
  early-exit paths do not pull expensive callee-frame memory into revert
  paths.

This restores the Timelock batch gas numbers to the base level while
preserving the ERC1155 early-revert improvements.

Also assert invalid Venom shapes in the policy helper where invoke
targets, callee lookup, block termination, and alloca sizes are expected
to be well-formed IR rather than silently falling back to conservative
behavior.
```

Improve Venom copy forwarding by sharing common forwarding policy logic and
using a profitability-aware guard for readonly invoke-argument forwarding.

### Cute Animal Picture


vyper/venom/passes/copy_forwarding.py
```diff
@@ -0,0 +1,235 @@
...
+            return size.value
+        return None
+
+    def copy_source(self, copy_inst: IRInstruction) -> IROperand:
+        _, src, _ = copy_inst.operands
+        if isinstance(src, IRVariable):
+            return self.dfg._traverse_assign_chain(src)
+        return src
+
+    def copy_destination_alloca(self, copy_inst: IRInstruction) -> Allocation | None:
```
HodanPlodky: I think this is unused

vyper/venom/passes/copy_forwarding.py
```diff
@@ -0,0 +1,226 @@
...
+                if not isinstance(size, IRLiteral):
+                    continue
+                intervals.append((ptr, size.value))
+                ptr += size.value
+
+        return intervals
+
+    def _get_invoke_callee(self, invoke_inst: IRInstruction) -> IRFunction | None:
+        target = invoke_inst.operands[0]
+        if not isinstance(target, IRLabel):
```
HodanPlodky: I think this should be assert

vyper/venom/passes/copy_forwarding.py
```diff
@@ -0,0 +1,226 @@
...
+            ]
+
+        intervals: list[tuple[int, int]] = []
+        ptr = 0
+        for bb in callee.get_basic_blocks():
+            for inst in bb.instructions:
+                if inst.opcode != "alloca":
+                    continue
+                size = inst.operands[0]
+                if not isinstance(size, IRLiteral):
```
HodanPlodky: Also should be assert

vyper/venom/passes/copy_forwarding.py
```diff
@@ -0,0 +1,226 @@
...
+            successors = list(self._successors(bb))
+            if len(successors) == 0:
+                return True
+
+            worklist.extend(successors)
+
+        return False
+
+    def _successors(self, bb: IRBasicBlock) -> list[IRBasicBlock]:
+        if not bb.is_terminated:
```
HodanPlodky: Also assert

vyper/venom/passes/copy_forwarding.py
```diff
@@ -0,0 +1,226 @@
...
+            return []
+        return bb.out_bbs
+
+    def _lowest_position_across_callee_frame(
+        self, invoke_inst: IRInstruction, size: int
+    ) -> int | None:
+        callee = self._get_invoke_callee(invoke_inst)
+        if callee is None:
```
HodanPlodky: This check could disappear it the `_get_invoke_callee` has an assert

HodanPlodky (commented):
There were few places where condition that returns `None` should be assert since otherwise it should not be valid venom. There were few more similar cases, but they would have to assume variable at the place of the `ptr` (ie. in the copy) so I did not pointed them explicitly, since you could have done that as future proof

HodanPlodky (approved):

charles-cooper (approved):
amazing

PR #4946, opened 2026-05-01
chore[docs]: document abstract methods

### LINKED ISSUE #4927: Document abstract methods
Currently no documentation, add some

Sporarum:
### What I did

Fix #4927

### How I did it

Add a new page, and modify existing ones

### Commit message

```
this commit adds docs/abstract-modules.rst which documents the new
abstract methods. other pages are also adapted accordingly.
```

Add abstract modules (https://docs.vyperlang.org/en/latest/abstract-modules.html)

### Cute Animal Picture


pcaversaccio (commented):
all the `tests/` and `vyper/` changes don't belong into this PR? maybe time for a rebase

Sporarum:
> all the `tests/` and `vyper/` changes don't belong into this PR? maybe time for a rebase

Fixed

docs/abstract-modules.rst
```diff
@@ -0,0 +1,496 @@
...
+This takes the form of **abstract methods**, methods decorated with ``@abstract``.
+These methods can then be overridden to supply the custom logic.
+Here is an example:
+
+.. code-block:: vyper
+
+    # base_token.vy
+    # This is the abstract module
+
+    balances: HashMap[address, uint256]
```
pcaversaccio: this is super nit and has no impact but _usually_ any kind of balances are declared as `public` in contracts (so the compiler auto-generates an external getter).
```suggestion
    balances: public(HashMap[address, uint256])
```

docs/abstract-modules.rst
```diff
@@ -0,0 +1,496 @@
...
+
+    # pausable_token.vy
+
+    import base_token
+
+    initializes: base_token
+
+    exports: base_token.transfer
+
+    paused: bool
```
pcaversaccio: same as above; pause states are _usually_ declares as `public`.
```suggestion
    paused: public(bool)
```

docs/abstract-modules.rst
```diff
@@ -0,0 +1,496 @@
+.. _abstract-modules:
+
+Abstract Modules
+################
+
+An abstract module is a special kind of :ref:`module <modules>` which offers points at which its logic can be customized.
+This takes the form of **abstract methods**, methods decorated with ``@abstract``.
```
pcaversaccio: i know in the later section you specify it but it's already helpful for the reader to know that it only applies to `internal` functions.
```suggestion
This takes the form of **abstract methods**, :ref:`internal <structure-functions-internal>` methods decorated with ``@abstract``.
```

docs/abstract-modules.rst
```diff
@@ -0,0 +1,496 @@
...
+Its body must consist of an Ellipsis literal (``...``) potentially preceded by a docstring and comments.
+
+.. code-block:: vyper
+
+    @abstract
+    def an_abstract_method(x: uint256) -> uint256:
+        """This is a docstring"""
+        ...
+
+There are no other restrictions on abstract methods, they can have any signature, can take any decorator, and be called like concrete methods:
```
pcaversaccio: thought of an edge case when reading this (haven't tested): can i declare an `internal` function in a `.vyi` function and also decorate it as `abstract`? if yes, we should disallow this as it doesn't make sense :)

pcaversaccio: "can take any decorator" -> i think this is not true as you cannot declare an `external` function as `abstract`? same for example for `@deploy`.

docs/abstract-modules.rst
pcaversaccio, line 1: another nit that i personally don't like; i hate not adding explicitly the `@internal` decorator to the functions; i know it's default but i like explicitness. you can ignore this comment for now and i know this is the case for other parts of the docs; i personally prefer pushing for explicitness in docs.

docs/abstract-modules.rst
```diff
@@ -0,0 +1,496 @@
...
+    @override(abstract_m)
+    def compute(y: uint256) -> uint256:
+        return y + 1
+
+.. code-block:: vyper
+
+    import abstract_m
+    initializes: abstract_m
+
+    # Invalid: Has looser mutability: nonpayable vs the original which is view
```
pcaversaccio: the `payable` one is interesting (not directly related to this example); as a library writer you might want to allow a function to be "generic" enough but the implementer wants to prevent any sent `mgs.value` for example, so the abstract method is declared as `payable` but the implementation is `nonpayable`. is this something that would be currently rejected?

docs/abstract-modules.rst
```diff
@@ -0,0 +1,496 @@
...
+
+    @abstract
+    def check_cap(amount: uint256) -> bool:
+        ...
+
+    @abstract
+    def check_address(addr: address) -> bool:
+        ...
+
+Here ``checked_token`` provides a concrete ``_before_transfer`` for ``base_token``, but it is itself an abstract module because it introduces ``check_cap`` and ``check_address``. A final module completes the chain by overriding those:
```
pcaversaccio:
```suggestion
Here ``checked_token.vy`` provides a concrete ``_before_transfer`` for ``base_token``, but it is itself an abstract module because it introduces ``check_cap`` and ``check_address``. A final module completes the chain by overriding those:
```

docs/abstract-modules.rst
```diff
@@ -0,0 +1,496 @@
...
+This is how you compose unrelated concerns, for instance combining transfer validation from one module with fee configuration from another.
+
+Using ``base_token`` from the :ref:`earlier example <abstract-modules>`, consider a second abstract module that manages access control:
+
+.. code-block:: vyper
+
+    # access_control.vy
+
+    @abstract
+    def _is_allowed(user: address) -> bool: ...
```
pcaversaccio:
```suggestion
    def _is_allowed(user: address) -> bool:
        ...
```

docs/abstract-modules.rst
```diff
@@ -0,0 +1,496 @@
...
+    initializes: minter
+    initializes: authentication_provider
+
+    roles: HashMap[address, my_roles.ROLE]
+
+    # By having `default` as an optional parameter, it's a valid override of both
+    # `minter.get_role` which does not have that parameter, and
+    # `authentication_provider.get_role` which has it as a mandatory parameter
+
+    @override(minter)
```
pcaversaccio: i have to admit that this can create very fast complex patters; not sure we should encourage this. here it would be "cleaner" to have `authentication_provider` import  `minter` and call the `minter.get_role` function for `get_role` if you ask me (think if it like a `super` call).

docs/abstract-modules.rst
```diff
@@ -0,0 +1,496 @@
...
+--------------------
+
+In some cases a module might not want to override every method of another, if such is the case, ``@abstract`` and ``@override`` can be combined to defer providing an implementation:
+
+.. code-block:: vyper
+
+    # base.vy
+
+    @abstract
+    def method_a() -> uint256: ...
```
pcaversaccio:
```suggestion
    def method_a() -> uint256:
        ...
```

docs/abstract-modules.rst
```diff
@@ -0,0 +1,496 @@
...
+
+    @abstract
+    def method_b() -> uint256: ...
```
pcaversaccio:
```suggestion
    def method_b() -> uint256:
        ...
```

docs/abstract-modules.rst
```diff
@@ -0,0 +1,496 @@
...
+    
+    # Concrete override
+    @override(base)
+    def method_a() -> uint256:
+        return 0
+
+    # Delegates
+    @abstract
+    @override(base)
+    def method_b() -> uint256: ...
```
pcaversaccio:
```suggestion
    def method_b() -> uint256:
        ...
```

docs/abstract-modules.rst
```diff
@@ -0,0 +1,496 @@
...
+
+.. code-block:: vyper
+
+    # pausable_token.vy
+
+    import base_token
+
+    initializes: base_token
+
+    paused: bool
```
pcaversaccio:
```suggestion
    paused: public(bool)
```

docs/solidity-differences.rst
```diff
@@ -76,10 +76,13 @@ In Vyper, checks are written inline:
...
-No Class Inheritance
-====================
+.. Preserve old anchor for external links
+.. _no-class-inheritance:
 
-Solidity supports multiple inheritance, which introduces the diamond problem and C3 linearization complexity. Vyper excludes inheritance entirely.
+Modules Instead of Inheritance
+==============================
+
+Solidity supports multiple inheritance, which introduces the diamond problem and C3 linearization complexity. Vyper replaces inheritance with a module system that achieves the same goals of code reuse and behavioral customization through different, more explicit mechanisms.
```
pcaversaccio, line 85:
```suggestion
Solidity supports multiple inheritance, which introduces the diamond problem and C3 linearization complexity. Vyper replaces inheritance with a module system that favors composition rather than inheritance and achieves the same goals of code reuse and behavioral customization through different, more explicit mechanisms.
```

docs/abstract-modules.rst
Sporarum: That's why I said "other restrictions", I also omitted that they can't take decorators that are only valid on external methods, since it seems pretty intuitive

If you think I should be more precise given the above, let me know

docs/abstract-modules.rst
Sporarum: Yeah I think this should be a separate issue

Also for this in particular, I think it benefits from omitting `@internal`, since that makes things easier to parse

docs/abstract-modules.rst
pcaversaccio: personally i think it should be written more precisely, wdyt @charles-cooper?

docs/abstract-modules.rst
Sporarum: The mutability check is only one way: the override cannot have looser mutability, but it can have stricter mutability
Also note that abstract methods can only be internal, so there is no "sent msg.value", nonpayable only affects reading it

docs/abstract-modules.rst
Sporarum: Yeah I agree that this is very circumstantial, but I wanted to document it since it is allowed
I'm not sure how to handle it better than that

docs/solidity-differences.rst
Sporarum: Hmm I disagree, the abstract module system is just inheritance

docs/abstract-modules.rst
Sporarum: Can .vyi files even have internal functions ?

pcaversaccio: yes, i just tested it:

```vyi
# test.vyi
# pragma version ~=0.4.3

@internal
def _foo():
    ...
```

```console
vyper -f ast test.vyi 
```

which will successfully compile. I think we should disallow this.

pcaversaccio: Created an issue here https://github.com/vyperlang/vyper/issues/4947.

docs/solidity-differences.rst
pcaversaccio: well the entire module system is based on composition and also described as this, example: https://docs.vyperlang.org/en/stable/using-modules.html

image

docs/abstract-modules.rst
pcaversaccio: sure, i understand the circumstances but i have to admit the complexity patterns that this allows concern me; @charles-cooper thoughts? i mean for the docs it's fine, this is more of a general discussion.

docs/abstract-modules.rst
pcaversaccio: "the override cannot have looser mutability, but it can have stricter mutability"; that's what i wanted to know and i also saw that you described it later in the docs, which is good!

docs/abstract-modules.rst
pcaversaccio: Created an issue here: https://github.com/vyperlang/vyper/issues/4948.

docs/solidity-differences.rst
Sporarum: I know I know, but in this section we're specifically speaking about inheritance, so it feels weird to say "oh and by the way, we allow it, but don't recommend it"

pcaversaccio: hmm, well, i think it's still important to highlight "favors composition rather than inheritance" in the "Differences from Solidity" section somehow, specifically because that's what people will google and then will land here: https://en.wikipedia.org/wiki/Composition_over_inheritance. And folks should understand this.

charles-cooper: should we just link to the wikipedia article in that case? lol

docs/abstract-modules.rst
charles-cooper: uhmm i think it's fine as is, we could also say any non-visibility decorator?

docs/abstract-modules.rst
charles-cooper: it's a bit odd yea. let's document it as is for now and can discuss whether we want to ban it in the future separately

Sporarum: Opened separate issue for discussion:
https://github.com/vyperlang/vyper/issues/4963

docs/abstract-modules.rst
Sporarum: The issue being not all non-visibility decorators are valid on internal methods, so this might actually make it more confusing

docs/abstract-modules.rst
```diff
@@ -0,0 +1,500 @@
...
+----------------------
+
+It's sometimes desirable to provide a default implementation for an abstract method, this can be done by defining a separate method which holds this logic:
+
+.. code-block:: vyper
+
+    # base_token.vy
+
+    @abstract
+    def _before_transfer(sender: address, recipient: address, amount: uint256): ...
```
pcaversaccio:
```suggestion
    def _before_transfer(sender: address, recipient: address, amount: uint256):
        ...
```

docs/abstract-modules.rst
```diff
@@ -0,0 +1,500 @@
...
+.. code-block:: vyper
+
+    # base_token.vy
+    # This is the abstract module
+
+    balances: public(HashMap[address, uint256])
+
+    # Abstract method is declared here
+    @abstract
+    def _before_transfer(sender: address, recipient: address, amount: uint256): ...
```
pcaversaccio:
```suggestion
    def _before_transfer(sender: address, recipient: address, amount: uint256):
        ...
```

docs/abstract-modules.rst
```diff
@@ -0,0 +1,500 @@
...
+        assert not self.paused, "transfers are paused"
+
+Since any call to ``base_token._before_transfer`` will be replaced at compile-time to a call to the overriding ``_before_transfer``, it is necessary that any call to the former is a valid call to the latter:
+
+.. code-block:: vyper
+
+    # abstract_m.vy
+    @pure
+    @abstract
+    def compute(x: uint256) -> uint256: ...
```
pcaversaccio:
```suggestion
    def compute(x: uint256) -> uint256:
        ...
```

docs/abstract-modules.rst
```diff
@@ -0,0 +1,500 @@
...
+
+Abstract methods are called in the same way concrete methods are.
+Additionally, calling the abstract methods of another module requires :ref:`using <uses-statement>` it.
+
+.. code-block:: vyper
+
+    # base_token.vy
+
+    @abstract
+    def _before_transfer(sender: address, recipient: address, amount: uint256): ...
```
pcaversaccio:
```suggestion
    def _before_transfer(sender: address, recipient: address, amount: uint256):
        ...
```

docs/abstract-modules.rst
```diff
@@ -0,0 +1,500 @@
...
+Note however that the method must be a :ref:`valid override <overriding-abstract-methods>` for every abstract method:
+
+.. code-block:: vyper
+
+    # minter.vy
+
+    import my_roles
+
+    @abstract
+    def get_role(user: address) -> my_roles.ROLE: ...
```
pcaversaccio:
```suggestion
    def get_role(user: address) -> my_roles.ROLE:
        ...
```

docs/abstract-modules.rst
```diff
@@ -0,0 +1,500 @@
...
+
+.. code-block:: vyper
+
+    # authentication_provider.vy
+
+    import my_roles
+
+    @abstract
+    def get_role(user: address, default: my_roles.ROLE) -> my_roles.ROLE: ...
```
pcaversaccio:
```suggestion
    def get_role(user: address, default: my_roles.ROLE) -> my_roles.ROLE:
        ...
```

pcaversaccio (commented):
due to consistency

docs/abstract-modules.rst
pcaversaccio: what about:
> There are no other restrictions on abstract methods; they can have any signature, take _most_ decorators (for example, `@external` and `@deploy` are not allowed), and be called like concrete methods:

docs/solidity-differences.rst
pcaversaccio: why not :)

docs/abstract-modules.rst
Sporarum: I'd prefer something like this:
> There are no other restrictions on abstract methods, they can have any signature, can take any decorator (provided they are valid on internal methods), and be called like concrete methods:

Because it's not really a restriction on the abstract methods, just a consequence of them being internal methods

pcaversaccio: sgtm

Sporarum:
```suggestion
There are no other restrictions on abstract methods, they can have any signature, can take any decorator (provided they are valid on internal methods), and be called like concrete methods:
```

docs/abstract-modules.rst
```diff
@@ -0,0 +1,506 @@
...
+        ...
+
+    @external
+    def transfer(recipient: address, amount: uint256):
+        # and used here
+        self._before_transfer(msg.sender, recipient, amount)
+        self.balances[msg.sender] -= amount
+        self.balances[recipient] += amount
+
+The ``base_token`` module defines a transfer hook, ``_before_transfer()``, as an abstract method. It is called during every transfer, but has no implementation — that is left to whoever initializes this module. This lets this module focus on *where* custom logic runs, while the overriding module decide *what* it does.
```
lufa23:
```suggestion
The ``base_token`` module defines a transfer hook, ``_before_transfer()``, as an abstract method. It is called during every transfer, but has no implementation — that is left to whoever initializes this module. This lets this module focus on *where* custom logic runs, while the overriding module decides *what* it does.
```

docs/abstract-modules.rst
```diff
@@ -0,0 +1,506 @@
...
+Following the rules above will always result in a valid override.
+However these might prove too restrictive in your use-case, for this reason the actual rules are more flexible:
+
+- Each parameter of the abstract method must appear in the override at the same position, with the same name.
+- If the abstract method defines a default argument for a parameter, the value of the default argument must match in the override. Any value matches an ellipsis (``...``).
+- The override may add extra parameters to the right, as long as they have default values.
+- The override may add a default value to a parameter that was mandatory in the abstract method.
+- Each parameter type in the override must be a super-type of the corresponding parameter type in the abstract method.
+- The return type of the override must be a sub-type of the abstract method's return type.
+- The override's :ref:`mutability <function-mutability>` must the same (or stricter) than the abstract's.
```
lufa23:
```suggestion
- The override's :ref:`mutability <function-mutability>` must be the same (or stricter) than the abstract's.
```

docs/abstract-modules.rst
```diff
@@ -0,0 +1,506 @@
...
+As abstract modules are by essence incomplete, it is necessary for another module to complete them, by providing implementations for its abstract methods:
+
+.. code-block:: vyper
+
+    import base_token
+
+    initializes: base_token
+
+    @override(base_token)
+    def before_transfer(sender: address, recipient: address, amount: uint256):
```
lufa23: I got a compiler error with `Tried to override 'base_token.before_transfer', but it does not exist`, as `base_token` defines `_before_transfer` but this snippet uses `before_transfer`.


```suggestion
    def _before_transfer(sender: address, recipient: address, amount: uint256):
```

docs/abstract-modules.rst
```diff
@@ -0,0 +1,506 @@
...
+.. code-block:: vyper
+
+    # checked_token.vy
+
+    import base_token
+
+    initializes: base_token
+
+    @override(base_token)
+    def before_transfer(sender: address, recipient: address, amount: uint256):
```
lufa23: Same here, as  `base_token` defines `_before_transfer`, not `before_transfer`.

```suggestion
    def _before_transfer(sender: address, recipient: address, amount: uint256):
```

docs/abstract-modules.rst
```diff
@@ -0,0 +1,506 @@
...
+    import checked_token
+
+    initializes: checked_token
+
+    MAX_AMOUNT: constant(uint256) = 10**24
+    total_supply: uint256
+
+    @override(checked_token)
+    def check_cap(amount: uint256) -> bool:
+        return 0 < amount <= MAX_AMOUNT
```
lufa23: i got: `SyntaxException: Cannot have a comparison with more than two elements`

```suggestion
        return 0 < amount and amount <= MAX_AMOUNT
```

docs/abstract-modules.rst
```diff
@@ -0,0 +1,506 @@
...
+    @override(abstract_m)
+    def compute(y: uint256) -> uint256:
+        return y + 1
+
+.. code-block:: vyper
+
+    import abstract_m
+    initializes: abstract_m
+
+    # Invalid: Has looser mutability: nonpayable vs the original which is view
```
lufa23: The abstract `compute` is `@pure`, not `@view`.

```suggestion
    # Invalid: Has looser mutability: nonpayable vs the original which is pure
```

docs/abstract-modules.rst
```diff
@@ -0,0 +1,506 @@
...
+
+    @abstract
+    def get_role(user: address, default: my_roles.ROLE) -> my_roles.ROLE:
+        ...
+
+.. code-block:: vyper
+
+    import minter
+    import authentication_provider
+
```
lufa23: The snippet uses `my_roles.ROLE` but doesn't import `my_roles` (the prior two snippets do). Compiler errors with `InvalidType: 'my_roles.ROLE' is not a type!`.

```suggestion
    import minter
    import authentication_provider
    import my_roles
```

docs/abstract-modules.rst
```diff
@@ -0,0 +1,506 @@
...
+        ...
+
+.. code-block:: vyper
+
+    # authentication_provider.vy
+
+    import my_roles
+
+    @abstract
+    def get_role(user: address, default: my_roles.ROLE) -> my_roles.ROLE:
```
lufa23: `default` is a reserved keyword in Vyper. Rename also needed in the combined override below to keep the signature matched.

```suggestion
    def get_role(user: address, default_role: my_roles.ROLE) -> my_roles.ROLE:
```

charles-cooper (approved):
approved for merge once @pcaversaccio and @lufa23 comments have been addressed

docs/abstract-modules.rst
```diff
@@ -0,0 +1,506 @@
...
+
+    roles: HashMap[address, my_roles.ROLE]
+
+    # By having `default` as an optional parameter, it's a valid override of both
+    # `minter.get_role` which does not have that parameter, and
+    # `authentication_provider.get_role` which has it as a mandatory parameter
+
+    @override(minter)
+    @override(authentication_provider)
+    def get_role(user: address, default: my_roles.ROLE = empty(my_roles.ROLE)) -> my_roles.ROLE:
```
Sporarum:
```suggestion
    def get_role(user: address, default_role: my_roles.ROLE = empty(my_roles.ROLE)) -> my_roles.ROLE:
```

docs/abstract-modules.rst
```diff
@@ -0,0 +1,506 @@
...
+
+    import minter
+    import authentication_provider
+
+    initializes: minter
+    initializes: authentication_provider
+
+    roles: HashMap[address, my_roles.ROLE]
+
+    # By having `default` as an optional parameter, it's a valid override of both
```
Sporarum:
```suggestion
    # By having `default_role` as an optional parameter, it's a valid override of both
```

docs/abstract-modules.rst
```diff
@@ -0,0 +1,506 @@
...
+    # `minter.get_role` which does not have that parameter, and
+    # `authentication_provider.get_role` which has it as a mandatory parameter
+
+    @override(minter)
+    @override(authentication_provider)
+    def get_role(user: address, default: my_roles.ROLE = empty(my_roles.ROLE)) -> my_roles.ROLE:
+        role: my_roles.ROLE = self.roles[user]
+        if role == empty(my_roles.ROLE):
+            return default
```
Sporarum:
```suggestion
            return default_role
```

docs/solidity-differences.rst
pcaversaccio: i think this is the last pending discussion before we can merge - i won't die on this hill, but i really think it would be helpful/insightful for users/devs to highlight "favors composition rather than inheritance" in some form.

PR #4949, opened 2026-05-05
fix[codegen]: gate decimals by usage

harkal:
### Commit message

### Cute Animal Picture

### DIFF
--- a/vyper/semantics/analysis/local.py
+++ b/vyper/semantics/analysis/local.py
@@ -75,14 +75,25 @@ from vyper.semantics.types.utils import type_from_annotation
 def analyze_functions(vy_module: vy_ast.Module) -> None:
     """Analyzes a vyper ast and validates the function bodies"""
     err_list = ExceptionList()
+    module_t = vy_module._metadata["type"]
 
-    for node in vy_module.get_children(vy_ast.FunctionDef):
-        _analyze_function_r(node, err_list)
+    function_defs = vy_module.get_children(vy_ast.FunctionDef)
 
     for node in vy_module.get_children(vy_ast.VariableDecl):
         if not node.is_public:
             continue
-        _analyze_function_r(node._expanded_getter, err_list)
+        function_defs.append(node._expanded_getter)
+
+    for fn_t in module_t.exposed_functions:
+        if isinstance(fn_t.ast_def, vy_ast.FunctionDef):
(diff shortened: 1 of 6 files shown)


vyper/semantics/analysis/local.py
```diff
@@ -75,14 +75,25 @@
...
+            function_defs.append(fn_t.ast_def)
+
+    seen = set()
+    for node in function_defs:
+        if id(node) in seen:
+            continue
+        seen.add(id(node))
+        _analyze_function_r(node, err_list)
 
     err_list.raise_if_not_empty()
```
Sporarum, line 98: Logic seems a bit weird, notably doing both vy_module.get_children(vy_ast.FunctionDef) and module_t.exposed_functions

More broadly, there should* be a method on ModuleT to get all external methods, including getters from public variables

*meaning I think it exists, but if not, it should be created

vyper/semantics/analysis/module.py
```diff
@@ -130,6 +131,8 @@ def _analyze_module_bodies(module_ast: vy_ast.Module) -> None:
     # interfaces don't have function bodies to validate
     if module_ast.is_interface:
         return
+    if not module_ast._metadata.get("is_compilation_target"):
+        return
```
Sporarum, line 135: This means we skip module analysis for all but the root module, which seems incorrect
(and surprising it doesn't brake anything)

vyper/semantics/analysis/module.py
```diff
@@ -61,6 +61,7 @@ def analyze_modules(imports: ImportAnalyzer) -> ModuleT:
     type-check/validate semantics and annotate with type and analysis info.
     """
     root_module_ast = imports.toplevel_module
+    root_module_ast._metadata["is_compilation_target"] = True
```
Sporarum, line 64: This seems flawed, the root module is not the only contract-like module
Consider the following:
```vyper
# contract1.vy

<contract code which gets deployed, was a root-module when deployed>

# contract2.vy

import contract1

other: address

def foo():
  extcall contract1.__interface__(other).bar(...)
```

Sporarum (changes requested):
See the comments for specifics, but I don't think this is the correct way forward ^^'

Sporarum:
Superceeded by #4950

PR #4950, opened 2026-05-05
fix[lang]: allow unused decimals without --allow-decimals

### LINKED ISSUE #4580: Make `decimals` optional when `math` is imported generically
Only require `decimals` for stdlib `math` imports if it's reachable in the code. I.e. a simple empty contract like this should not throw:

```vyper
import math
```

Sporarum:
### What I did

Fix #4580

Fix tests failing

Always allow decimals if they are unreachable

### How I did it

Add new pass in module analysis which checks no reachable function uses `decimal`

### Commit message

```
move the decimal check from type resolution time to `analyze_modules`,
where it inspects only `reachable_functions`. done through helper
`ContractFunctionT.uses_decimal()` which walks param types, return type,
and body annotations to detect actual decimal usage in a function. 
```
### Cute Animal Picture

### DIFF
--- a/vyper/semantics/types/function.py
+++ b/vyper/semantics/types/function.py
@@ -869,6 +869,12 @@ class ContractFunctionT(VyperType):
                 if _contains_decimal(typ):
                     return True
 
+        # body variable accesses
+        for var_access in self.get_variable_accesses():
+            if _contains_decimal(var_access.variable.typ):
+                return True
+
+
         # emitted events
         for event_t in self._emitted_events:
             for member_t in event_t.arguments.values():
--- a/vyper/semantics/types/function.py
+++ b/vyper/semantics/types/function.py
@@ -869,6 +869,12 @@ class ContractFunctionT(VyperType):
                 if _contains_decimal(typ):
                     return True
 
+        # emitted events
+        for event_t in self._emitted_events:
+            for member_t in event_t.arguments.values():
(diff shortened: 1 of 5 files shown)


harkal:
this seems to compile without `--enable-decimals` now:

```vyper
@external
def foo(x: uint256) -> uint256:
    return convert(convert(x, decimal) + 1.0, uint256)
```

Sporarum:
> this seems to compile without `--enable-decimals` now:
> 
> ```vyper
> @external
> def foo(x: uint256) -> uint256:
>     return convert(convert(x, decimal) + 1.0, uint256)
> ```

Fixed

vyper/semantics/types/function.py
```diff
@@ -852,6 +852,55 @@ def abi_signature_for_kwargs(self, kwargs: list[KeywordArg]) -> str:
         args = self.positional_args + kwargs  # type: ignore
         return self.name + "(" + ",".join([arg.typ.abi_type.selector_name() for arg in args]) + ")"
 
+    def uses_decimal(self) -> bool:
```
charles-cooper, line 855: i would prefer to see this just be in type_from_annotation; otherwise trying to find out all the types from walking the ast is probably a losing game

charles-cooper:
superseded for now by https://github.com/vyperlang/vyper/pull/4978

Issue #4856, opened 2026-02-19
VIP: Unbounded Sequence Types


Sporarum:
# Unbounded Sequence Types

## Introduction

This section introduces relevant terminology

### Parameters and arguments

Parameters are the inputs as defined by the function, arguments are the values or expressions effectively passed to the input:
```python
#       v parameter
def foo(x: uint256)...

#   v argument
foo(4)
```

### Bounded sequence types

We define bounded sequence types to be the types taking as type parameter their maximum length:

1. `Bytes[n]`
2. `String[n]`
3. `DynArray[T, n]`

Notably this list does not include fixed-size lists (`T[n]` where `T` is any type except `Bytes[n]`, `String[n]`, flag types), since there the `n` is not an upper bound, but the actual length of the sequence.

For bounded sequence types, variables with a smaller bound can be assigned to variables with a bigger bound:
```python
def foo() -> None:
	s1: String[10] = "Hi" # valid since "Hi" has 2 characters, so fits into "at most 10"
	s2: String[20] = s1 # valid since "at most 10" fits into "at most 20"
	s3: String[2]  = s1 # invalid since "at most 10" does not fit into "at most 2"
	return
```
In other words, for all `n1 <= n2`, `String[n1]` is a subtype of `String[n2]`: `String[n1] <: String[n2]`.

### Data locations

The EVM contains essentially 4 locations where data can be located:

1. Calldata
2. Memory
3. Transient storage
4. Storage

(There are also others like the stack and the contract bytecode itself, but those are not relevant for this discussion.)

Calldata contains the parameters to an external function, and is immutable.
Memory contains parameters to internal functions, as well as temporary variables (ones defined inside functions).
Transient storage and storage store persistent data, for example contract variables (which can be accessed with `self`).

Memory, while cheaper than storage, is not free.
The total cost depends on the highest index allocated, and is quadratic.
This means allocating more memory than needed can be _extremely_ expensive.

## Motivation

Currently, before a function's arguments are passed, memory is allocated to fit the expected type into memory.
If a function has a `String[1000]` as parameter, even if the argument is `String[10]`, we allocate the full 1000 chars worth of memory !

And of course you cannot pass a `String[11]` as an argument for a `String[10]` parameter, as that could break assumptions on the other end.

This leads to a dilemma: How big do I make that sequence type ?
1. If you make it too small, this heavily contrains users in what they can do
2. If you make it too big, the cost will get astronomical, even if the user calls the method with a small instance

[...]

Issue #4685, opened 2025-06-08
`slice()` with String[N] fails if the sliced region exceeds the underlying string’s actual length


benber86:
### Version Information

* vyper Version (output of `vyper --version`): 0.4.1
* OS: linux
* Python Version (output of `python --version`): 3.12

### What's your issue about?

Using `slice` with a `String[N]` variable uses the string's actual length rather than the specified `N`.
For instance:

```
@view
@external
def foo() -> String[5]:
    id_label: String[78] = uint2str(123)
    short_id: String[5] = slice(id_label, 0, 5)
    return short_id
```
Will revert, but:

```
@view
@external
def foo() -> String[5]:
    id_label: String[78] = uint2str(123456)
    short_id: String[5] = slice(id_label, 0, 5)
    return short_id
```

will work fine.

Both use `String[78]` for id_label. The only difference is the runtime length: "123" vs "123456".
I'd expect in both cases `slice(id_label, 0, 5)` should succeed. 

You could check the length and use `convert` but it's not great UX

cyberthirst:
related https://github.com/vyperlang/vyper/issues/4345#issuecomment-2454933113

PR #4967, opened 2026-05-12
feat[lang]!: add unbounded types

Sporarum:
### What I did

Implement #4856

Fix interface implementation working backwards for the return type, this is a breaking change !
(text below mentions `Bytes`, but the logic is the same for `String` and `DynArray`)
In interfaces, replace `Bytes[1]` in return types and variables by `Bytes[...]` to keep them behaving as previously.
There is no replacement for  `Bytes[<n>]` in return types meaning "can be implemented by `Bytes[<k>]` where k >= n"
In return types, `Bytes[n]` now means "can be implemented by `Bytes[<k>]` where k <= n".
Note that this is required for safety: if a function is guaranteed by its interface to return at most 2 bytes, we shouldn't be able to implement it with one which returns 3 bytes.

### Commit message

```
interface return types were checked contravariantly instead of
covariantly: a function declared as returning `Bytes[1]` in an interface
could be implemented by one returning `Bytes[100]`. this violates the
Liskov substitution principle — if an interface promises "at most 1
byte", the implementation must not return 100. the check was
intentionally inverted (with a TODO) to let IERC20's `String[0]` return
type be implemented by `String[..]`, because there was no other way to
express "any length".

fix the variance direction and add two new length kinds so the
intentional hack is no longer needed:

- `T[...]` (wildcard): only valid in interface declarations. means "any
  length" and is checked on assignment. allows interfaces to declare
  `Bytes[...]` where they previously used `Bytes[1]` to work around the
  contravariance bug. breaking: existing interfaces with bounded return
  types now enforce covariance correctly; migrate `Bytes[n]` ->
  `Bytes[...]` to preserve old behavior.

- `T[INF]` (infinity): only valid on local variables, parameters, and
  return types (no storage). represents an unbounded type whose actual
  size is determined at runtime. `Bytes[n] <: Bytes[INF]` for any n, so
  a bounded value can flow into an unbounded slot. enables patterns like
  forwarding, multicall, and virtual machines where the parameter size
  is not known at compile time.

internally, `set_length` / `set_min_length` on bytestring types are
replaced with direct constructor invocation — the two-phase construction
was a workaround for the old inference system.
```

add wildcard length for interfaces, for example `Bytes[...]`
BREAKING fix interface return types, update your interfaces to use a wildcard length to fix it, example: `Bytes[1]` -> `Bytes[...]`
add unbounded types, for example`Bytes[INF]`

### Cute Animal Picture


charles-cooper (commented):
nice cleanups here. looks generally good at first pass

vyper/semantics/types/subscriptable.py
```diff
@@ -278,8 +286,23 @@ def compare_type(self, other):
...
             return False
+
+        # Wildcard matches any length (bidirectional)
+        if self.length is WILDCARD or other.length is WILDCARD:
+            return self.value_type.compare_type(other.value_type)
+
+        if self.length is INF:
+            # INF >= INF
+            # INF >= n
+            return True
```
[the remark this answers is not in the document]
Sporarum: Fixed by 9e7475106fdf6dffc0fea2876e7321fe0760fbb3

vyper/semantics/analysis/utils.py
```diff
@@ -620,8 +620,23 @@ def validate_expected_type(node, expected_type):
...
+        hint = None
+        failed_bytes_subtyping = isinstance(given, BytesT) and any(
+            isinstance(expected, BytesT) for expected in expected_type
+        )
+        failed_string_subtyping = isinstance(given, StringT) and any(
+            isinstance(expected, StringT) for expected in expected_type
+        )
+
+        if failed_bytes_subtyping or failed_string_subtyping:
+            hint = "To reduce the maximum size, either: Use `slice` (with a constant length)"
```
charles-cooper: does slice work like this now

vyper/semantics/analysis/utils.py
```diff
@@ -620,8 +620,23 @@ def validate_expected_type(node, expected_type):
...
+            hint += "to get a subset. Or use `convert` which will revert if the runtime length "
```
charles-cooper: i would just say "try use slice or convert"

vyper/abi_types.py
```diff
@@ -185,7 +186,7 @@ def selector_name(self):
 
 class ABI_DynamicArray(ABIType):
     def __init__(self, subtyp, elems_bound):
-        if not elems_bound >= 0:
+        if elems_bound is not INF and elems_bound is not WILDCARD and not elems_bound >= 0:
```
charles-cooper: maybe we should have some helpers for this

vyper/semantics/types/bytestrings.py
```diff
@@ -46,13 +38,11 @@ def _addl_dict_fields(self):
         return {"length": self.length}
 
     @property
-    def length(self):
+    def length(self) -> int | Inf | Wildcard:
```
charles-cooper: maybe we should have a type alias like ValidLength

vyper/semantics/analysis/utils.py
Sporarum: It predates this PR, but note that it reverts if the input is smaller than the length:

For example:
```
x: String[5] = "Hi"

x2: String[1] = slice(x, 0, 1) # compiles and evaluates to "H"

x2: String[2] = slice(x, 0, 2) # compiles and evaluates to "Hi"

x3: String[3] = slice(x, 0, 3) # compiles, but reverts
```

Sporarum: In other words, for all `x` and `n` :
```
sliced: String[n] = slice(x, 0, n)
assert len(sliced) == n
```
The assert will never revert (but the slice might)

I would expect the opposite, given the length on the type as the meaning "n or less"

charles-cooper: yea, i kind of hate the semantics of slice. i think there is an issue in the issue tracker for it

charles-cooper: https://github.com/vyperlang/vyper/issues/2701
https://github.com/vyperlang/vyper/issues/4685
https://github.com/vyperlang/vyper/issues/4345#issuecomment-2454933113

charles-cooper: https://github.com/vyperlang/vyper/issues/3562

Sporarum: Yeah I think we should definitely change it

vyper/semantics/types/infinity.py
```diff
@@ -0,0 +1,39 @@
...
+    def __str__(self):
+        return "..."
+
+
+WILDCARD = Wildcard.WILDCARD
+
+LengthUpperBound: TypeAlias = int | Inf | Wildcard
+
+
+def is_bounded(val: LengthUpperBound) -> TypeGuard[int]:
```
charles-cooper:
```suggestion
def is_bounded(_lengthval: LengthUpperBound) -> TypeGuard[int]:
```

charles-cooper: nifty find on the type guard thing

Sporarum: Yeah claude mentioned it in passing once so I looked into it

vyper/semantics/analysis/local.py
```diff
@@ -988,12 +970,27 @@ def visit_Call(self, node: vy_ast.Call, typ: VyperType) -> None:
...
+            for arg, arg_typ in zip(node.args, func_type.argument_types):
+                self.visit(arg, arg_typ)
             for kwarg in node.keywords:
                 # We should only see special kwargs
-                typ = func_type.call_site_kwargs[kwarg.arg].typ
-                self.visit(kwarg.value, typ)
+                kwarg_typ = func_type.call_site_kwargs[kwarg.arg].typ
+                self.visit(kwarg.value, kwarg_typ)
+
+            if func_type.is_external:
```
charles-cooper, line 983: this code is tricky and i want to refactor it, extracting to a new issue

charles-cooper (approved):

PR #4968, opened 2026-05-12
perf[venom]: mem passes

HodanPlodky:
### What I did

Sped up the Venom memory passes (mostly dead store elimination).

### Commit message

```
Speed up the memory passes, mostly dead store elimination.

`_is_memory_def_live` used to walk the CFG for every store. Add fast
paths so we skip that when the answer is obvious: empty writes, stores
that are never read, and locations that are only read (where liveness is
just a reachability check). The slow worklist is only used when another
write might clobber.

To support this, track which instructions touch each memory location in
the alias analysis, and compute a full reachability closure (the
recursive version wasn't transitive across loops). Also stop deep-
copying the `_by_memloc` index in the load analysis lattice on every
copy -- sharing it is fine since it's only used to speed up removals.
```

### Cute Animal Picture


vyper/venom/analysis/mem_alias.py
```diff
@@ -121,6 +129,15 @@ def get_alias_set(self, loc: MemoryLocation) -> OrderedSet[MemoryLocation] | Non
         res.add(loc)
         return res
 
+    def get_all_insts(self, loc: MemoryLocation) -> set[IRInstruction]:
+        alias_set = self.get_alias_set(loc)
+        assert alias_set is not None
+        res = set()
+        for alias_loc in alias_set:
+            tmp = self.mem_loc_insts[alias_loc]
```
harkal:
```suggestion
            tmp = self.mem_loc_insts.get(alias_loc, set())
```

harkal (changes requested):

harkal:
lgtm

harkal (approved):

PR #4974, opened 2026-05-17
fix[venom]: guard copy-forwarding against deep-callee gas bombs

harkal:
### What I did

Hardened `CopyForwardingPolicy` against quadratic memory-expansion gas bombs when the mcopy source operand can't be resolved to an alloca (typically a function `param`). On master alone, no contract in the snekmate corpus currently triggers this shape, but it materializes on branches whose inlining decisions are less aggressive  (observed at ~135M gas per call on snekmate ERC1155 batch revert paths).

### How I did it

Two structural changes in `vyper/venom/passes/copy_forwarding.py`:

- `_callee_reserved_intervals` now walks transitive callees in its pre-allocator fallback. The previous local-only walk silently missed deeper allocas, which is a precision bug independent of the bomb scenario.
- `should_block_forwarding` now has an explicit branch for unresolved sources. It estimates the per-invoke memory-expansion penalty using the copy size as a proxy and bails when the penalty exceeds an absolute 1M-gas threshold. Small-frame forwardings still win (hundreds of gas); ABI-encode-buffer-scale frames (megabytes -> millions of gas) are blocked.

An alternative defer all unresolved-source forwarding decisions to the post-inline per-function pass (was evaluated and rejected). It eliminates the magic number but loses ~67 forwarding wins across the snekmate corpus, because non-inlined functions with `param` sources keep `src=None` in both passes and the post-inline pass can't tell "never-inlined" apart from "will-be-inlined."

### Commit message

```
`CopyForwardingPolicy.should_block_forwarding` previously returned False
unconditionally when the mcopy source couldn't be resolved to an alloca
(typically a function `param`). That's a soundness gap: if the enclosing
function gets inlined later, the param resolves to a caller alloca whose
liveness now extends across the deep invoke. The conflict graph can
force that caller alloca above the callee's transitive pinned set, and
on snekmate ERC1155 batch revert paths under inlining shapes that expose
this pattern, the resulting quadratic memory-expansion gas is ~135M per
call.

Two changes plug the gap. `_callee_reserved_intervals` now walks
transitive callees in its pre-allocator fallback -- the old local-only
walk silently missed deeper allocas, a precision bug in its own right.
And `should_block_forwarding` now estimates the per-invoke memory-
expansion penalty even when the source is unresolved, using the copy
size as a proxy, and bails when the penalty exceeds an absolute 1M-gas
threshold. Small-frame forwardings still win (hundreds of gas); ABI-
encode-buffer-scale frames (megabytes -> millions of gas) are caught.

A defer-to-post-inline-pass alternative was evaluated and rejected. It
removes the magic number but loses ~67 forwarding wins across the
snekmate corpus: non-inlined functions with `param` sources have
src=None in both passes, so the post-inline pass can't tell "never-
inlined" apart from "will-be-inlined" and conservatively keeps the
mcopy.
```

### DIFF
--- a/vyper/venom/passes/copy_forwarding.py
+++ b/vyper/venom/passes/copy_forwarding.py
@@ -340,8 +340,7 @@ class CopyForwardingPolicy:
                     if inst.opcode != "invoke":
                         continue
                     target = inst.operands[0]
-                    if not isinstance(target, IRLabel):
-                        continue
+                    assert isinstance(target, IRLabel)
                     sub_callee = self.function.ctx.functions.get(target)
                     if sub_callee is None:
                         continue
--- a/vyper/venom/passes/copy_forwarding.py
+++ b/vyper/venom/passes/copy_forwarding.py
@@ -25,6 +25,12 @@ class CopyForwardingPolicy:
     # the one-byte MCOPY opcode from deployed bytecode.
     CODE_DEPOSIT_GAS_PER_BYTE: int = 200
     MIN_ELIDED_MCOPY_BYTES: int = 1
+    # When the copy source can't be resolved to an alloca, only block
+    # forwarding for catastrophic memory-expansion penalties (millions of
+    # gas, e.g. ABI-encode buffers of dynamic-array args). Moderate
+    # penalties (a few KB of frame growth) can still be outweighed by the
+    # forwarding win across many invocations.
+    UNRESOLVED_SOURCE_PENALTY_THRESHOLD: int = 1_000_000
 
     function: IRFunction
     dfg: DFGAnalysis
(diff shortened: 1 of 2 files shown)


vyper/venom/passes/copy_forwarding.py
```diff
@@ -285,19 +308,46 @@ def _callee_reserved_intervals(self, callee: IRFunction) -> list[tuple[int, int]
...
+                for inst in bb.instructions:
+                    if inst.opcode == "alloca":
+                        size_op = inst.operands[0]
+                        assert isinstance(size_op, IRLiteral)
+                        sizes.append(size_op.value)
+                        continue
+                    if inst.opcode != "invoke":
+                        continue
+                    target = inst.operands[0]
+                    if not isinstance(target, IRLabel):
```
HodanPlodky: I think this should be assert

HodanPlodky (changes requested):
Just two small things otherwise lgtm

HodanPlodky (approved):

harkal:
Merged though other PR

PR #4979, opened 2026-05-19
fix[codegen]: fix `convert()` codegen in venom pipeline

HodanPlodky:
### What I did
Fix the venom pipeline `convert()` to match the legacy pipeline behavior for bounds checking, literal validation, and sign extension.

### How I did it
- Fixed decimal bounds clamping to use `round_towards_zero` instead of integer division (`//`), which truncated towards negative infinity and could allow out-of-range negative values to pass the pre-multiplication clamp.
- Refactored `_get_folded_value` to return the AST constant node instead of the raw Python value, so literal-kind information (hex vs decimal vs bytes) is preserved for legacy-compatible validation.
- Added sign extension for hex/bytes constants converted to signed integer types (e.g. `convert(0xff <bytes1>, int16)` now correctly yields -1, not 255).
- Added `_check_literal_decimal_bounds` — the venom pipeline had no compile-time overflow checks for decimal conversions at all.
- Made `_check_literal_int_bounds` handle hex, bytes, and bytestring constants (previously only int/bool), so `convert(0xffff, uint8)` now correctly raises `InvalidLiteral`.
- Used `arg_node.reduced()` to resolve constant variable references before type-checking bytestring inputs.

### How to verify it
- `uv run ./quicktest.sh tests/functional/builtins/codegen/test_convert.py`

### Commit message
```
the venom pipeline `convert()` had several mismatches with the legacy
pipeline:

- decimal bounds checking used integer division for pre-multiplication
  clamping, which truncated towards negative infinity instead of towards
  zero. for negative inputs close to `DECIMAL_MIN / DECIMAL_DIVISOR`,
  this could produce a pre-lo bound that was one unit too high, allowing
  out-of-range values to pass the clamp and wrap around after scaling.

- `_check_literal_int_bounds` only handled plain int/bool constants. hex
  literals (`0xff`), bytes constants, and bytestring constants were
  silently skipped, so `convert(0xffff, uint8)` compiled without error
  instead of raising InvalidLiteral. additionally, sign extension was
  missing for hex/bytes-to-signed-integer conversions (e.g.
  `convert(0xff <bytes1>, int16)` should yield -1, not 255).

- there was no `_check_literal_decimal_bounds` at all, so compile-time
  overflow checks for decimal conversions (e.g. converting a too-large
  int136 to decimal) were absent in the venom pipeline.

refactor `_get_folded_value` to return the AST node instead of the
python value, so literal-kind information (hex vs decimal vs bytes) is
available for legacy-compatible sign-extension and bounds checks. add
sign extension for hex/bytes constants converted to signed types. add a
separate `_check_literal_decimal_bounds` with correct round-towards-zero
division. use `arg_node.reduced()` to resolve constant variable
references before type-checking bytestring inputs.
```

Fix `convert()` in the venom codegen pipeline: correct decimal bounds clamping, add missing literal overflow checks for hex/bytes/decimal, and apply sign extension for hex/bytes-to-signed conversions.

### Cute Animal Picture

Cute red panda


vyper/codegen_venom/builtins/convert.py
```diff
@@ -35,13 +38,14 @@ def lower_convert(node: vy_ast.Call, ctx: VenomCodegenContext) -> IROperand:
...
 
-    arg_node = node.args[0]
+    arg_node = node.args[0].reduced()
     in_t = arg_node._metadata["type"]
     out_t = node.args[1]._metadata["type"].typedef
 
     # For bytestrings we need pointer, for primitives we need value
     if isinstance(in_t, _BytestringT):
         arg_vv = Expr(arg_node, ctx).lower()
+        in_t = arg_vv.typ
```
charles-cooper, line 60: why is that necessary?

charles-cooper (approved):
seems good.. i want to do a deeper audit of convert in legacy vs venom

charles-cooper:
@HodanPlodky please check CI failures

tests/functional/builtins/codegen/test_convert.py
```diff
@@ -662,6 +663,54 @@ def foo() -> {t_bytes}:
...
+    code = r"""
+FOO: constant(Bytes[2]) = b'\xff'
+
+@external
+def foo() -> int16:
+    return convert(FOO, int16)
+    """
+
+    c = get_contract(code)
+    assert c.foo() == -1
```
charles-cooper, line 742: this is super sketchy, i think `c.foo()` should return 255

PR #4984, opened 2026-05-20
fix[codegen]: remove silent fallback for bool kwargs in venom pipeline

banteg:
### What I did

Fixed Venom builtin lowering for boolean keyword arguments that semantic analysis already accepted as folded constants.

This covers `abi_decode(..., unwrap_tuple=CONST_FALSE)` and `print(..., hardhat_compat=CONST_TRUE)`.

### How I did it

Updated the Venom bool kwarg extraction helpers in the ABI and debug-print builtins to use `kw_node.reduced()` before checking for boolean or integer literals.

If a provided bool kwarg still cannot be recognized after reduction, the helper raises a compiler panic instead of silently falling back to the default. That keeps the absent-kwarg default path, but avoids masking unexpected accepted input.

### How to verify it

- `uv run --python 3.12 pytest tests/unit/compiler/venom/test_builtin_kwargs.py tests/functional/builtins/codegen/test_abi_decode.py::test_abi_decode_folded_unwrap_tuple_kwarg tests/functional/syntax/test_print.py::test_print_folded_hardhat_compat_kwarg -q`
- `uv run --python 3.12 ruff check vyper/codegen_venom/builtins/abi.py vyper/codegen_venom/builtins/misc.py tests/unit/compiler/venom/test_builtin_kwargs.py tests/functional/builtins/codegen/test_abi_decode.py tests/functional/syntax/test_print.py`

### Commit message
```
_get_bool_kwarg in abi.py and misc.py silently returned the default
value when the kwarg node was not a NameConstant or Int literal. this
masked two bugs at once: folded constants like UNWRAP=False (a Name, not
NameConstant) were ignored, and any unexpected node type was swallowed
without error.

call kw_node.reduced() before the type checks so folded constants
resolve to their actual literal form. replace the silent fallback
with a CompilerPanic to make future issues immediately visible.
```

Fix Venom builtin lowering for folded constant boolean keyword arguments.

### DIFF
--- a/vyper/codegen_venom/builtins/abi.py
+++ b/vyper/codegen_venom/builtins/abi.py
@@ -45,7 +45,7 @@ def _get_bool_kwarg(node: vy_ast.Call, kwarg_name: str, default: bool) -> bool:
     # Could also be an Int with constant value
     if isinstance(kw_node, vy_ast.Int):
         return bool(kw_node.value)
-    raise CompilerPanic(f"unfoldable boolean kwarg: {kwarg_name}")
+    raise CompilerPanic(f"unfoldable boolean kwarg: {kwarg_name}", kw_node)
 
 
 def _parse_method_id(method_id_node: vy_ast.VyperNode) -> Optional[int]:
--- a/vyper/codegen_venom/builtins/abi.py
+++ b/vyper/codegen_venom/builtins/abi.py
@@ -15,6 +15,7 @@ from vyper.codegen.core import calculate_type_for_external_return
 from vyper.codegen_venom.abi import abi_decode_to_buf, abi_encode_to_buf
 from vyper.codegen_venom.buffer import Buffer, Ptr
 from vyper.codegen_venom.value import VyperValue
+from vyper.exceptions import CompilerPanic
 from vyper.semantics.data_locations import DataLocation
 from vyper.semantics.types import BytesT, TupleT
 from vyper.utils import fourbytes_to_int
@@ -37,13 +38,14 @@ def _get_bool_kwarg(node: vy_ast.Call, kwarg_name: str, default: bool) -> bool:
     kw_node = _get_kwarg_value(node, kwarg_name)
     if kw_node is None:
         return default
+    kw_node = kw_node.reduced()
     # The value should be a NameConstant (True/False)
     if isinstance(kw_node, vy_ast.NameConstant):
         return kw_node.value
     # Could also be an Int with constant value
     if isinstance(kw_node, vy_ast.Int):
         return bool(kw_node.value)
-    return default
+    raise CompilerPanic(f"unfoldable boolean kwarg: {kwarg_name}")
 
 
 def _parse_method_id(method_id_node: vy_ast.VyperNode) -> Optional[int]:
--- a/vyper/codegen_venom/builtins/misc.py
+++ b/vyper/codegen_venom/builtins/misc.py
@@ -19,7 +19,7 @@ from vyper.builtins.functions import AsWeiValue
 from vyper.codegen_venom.abi.abi_encoder import abi_encode_to_buf
 from vyper.codegen_venom.constants import BLOCKHASH_LOOKBACK_LIMIT, ECRECOVER_PRECOMPILE
 from vyper.evm.opcodes import version_check
-from vyper.exceptions import EvmVersionException
+from vyper.exceptions import CompilerPanic, EvmVersionException
 from vyper.semantics.types import BytesT, DecimalT, StringT, TupleT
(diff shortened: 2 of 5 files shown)


charles-cooper:
i think it's worse that we are falling back to something rather than not unfolding. we should remove the `return default` fallback line in these builtins (and actually all builtins as applicable)

banteg:
Updated. The helpers still return the default when the kwarg is absent, but a provided bool kwarg that cannot be interpreted after folding now raises `CompilerPanic` instead of falling back to the default.

vyper/codegen_venom/builtins/misc.py
```diff
@@ -389,11 +389,13 @@ def _get_bool_kwarg(node: vy_ast.Call, kwarg_name: str, default: bool) -> bool:
     kw_node = _get_kwarg_value(node, kwarg_name)
     if kw_node is None:
         return default
+    if kw_node.has_folded_value:
+        kw_node = kw_node.get_folded_value()
```
harkal:
```suggestion
    kw_node = kw_node.reduced()
```

vyper/codegen_venom/builtins/abi.py
```diff
@@ -37,13 +38,15 @@ def _get_bool_kwarg(node: vy_ast.Call, kwarg_name: str, default: bool) -> bool:
     kw_node = _get_kwarg_value(node, kwarg_name)
     if kw_node is None:
         return default
+    if kw_node.has_folded_value:
+        kw_node = kw_node.get_folded_value()
```
harkal:
```suggestion
    kw_node = kw_node.reduced()
```

harkal (changes requested):

banteg:
Updated to use `kw_node.reduced()` in both bool-kwarg helpers. I also added direct unit coverage for both the folded path and the unreduced/error path, so the Codecov patch complaint should clear on the new run.

vyper/codegen_venom/builtins/abi.py
```diff
@@ -37,13 +38,14 @@ def _get_bool_kwarg(node: vy_ast.Call, kwarg_name: str, default: bool) -> bool:
...
         return default
+    kw_node = kw_node.reduced()
     # The value should be a NameConstant (True/False)
     if isinstance(kw_node, vy_ast.NameConstant):
         return kw_node.value
     # Could also be an Int with constant value
     if isinstance(kw_node, vy_ast.Int):
         return bool(kw_node.value)
-    return default
+    raise CompilerPanic(f"unfoldable boolean kwarg: {kwarg_name}")
```
charles-cooper:
```suggestion
    raise CompilerPanic(f"unfoldable boolean kwarg: {kwarg_name}", kw_node)
```

vyper/codegen_venom/builtins/misc.py
```diff
@@ -389,11 +389,12 @@ def _get_bool_kwarg(node: vy_ast.Call, kwarg_name: str, default: bool) -> bool:
     kw_node = _get_kwarg_value(node, kwarg_name)
     if kw_node is None:
         return default
+    kw_node = kw_node.reduced()
     if isinstance(kw_node, vy_ast.NameConstant):
         return kw_node.value
     if isinstance(kw_node, vy_ast.Int):
         return bool(kw_node.value)
-    return default
+    raise CompilerPanic(f"unfoldable boolean kwarg: {kwarg_name}")
```
charles-cooper:
```suggestion
    raise CompilerPanic(f"unfoldable boolean kwarg: {kwarg_name}", kw_node)
```

banteg:
Applied the suggestion in both bool-kwarg helpers: the CompilerPanic now carries the offending kwarg node. Validated with the focused bool-kwarg, abi_decode, print tests, and ruff check.

harkal (approved):

charles-cooper (approved):

banteg:
Correction: #4984 was merged while this follow-up was being pushed, so the branch update is not part of the merged PR.

I opened the audited follow-up as #5001: https://github.com/vyperlang/vyper/pull/5001

Summary of the follow-up:
- removed every Venom _get_kwarg_value copy across abi, create, system, and misc
- added shared reduced-kwarg helpers so builtin callers receive kw.value.reduced()
- made required literal kwargs fail closed with CompilerPanic instead of silently falling back to defaults
- kept runtime fallback only for create_from_blueprint(code_offset=...), where runtime values are allowed
- found and fixed an adjacent ABIEncode.fetch_call_return raw ensure_tuple read that ICEd on ensure_tuple=CONST before codegen

Validated with focused Venom helper, abi_encode, abi_decode, print, constants, create, raw_call, and send tests; ruff check is clean on all touched files.

PR #4985, opened 2026-05-20
fix[venom]: fix `raw_call` kwarg evaluation order

banteg:
### What I did

Fixed Venom `raw_call` lowering so side-effectful `value=` and `gas=` keyword arguments are evaluated in source order.

### How I did it

Changed Venom `raw_call` lowering to walk `node.keywords` and materialize `gas=` / `value=` in the order written by the user.

The patch keeps the existing behavior that copies `msg.data` only after kwarg evaluation, so scratch memory is not clobbered before the call.

Legacy codegen currently evaluates this class of call kwargs in operand order; the regression test documents that legacy behavior against #2863 while requiring source-order behavior for Venom.

### How to verify it

- `uv run --python 3.12 pytest tests/functional/builtins/codegen/test_raw_call.py::test_raw_call_msg_data_kwarg_evaluation_order -q`
- `uv run --python 3.12 pytest tests/functional/builtins/codegen/test_raw_call.py::test_raw_call_msg_data_kwarg_evaluation_order --experimental-codegen -q`
- `uv run --python 3.12 ruff check vyper/codegen_venom/builtins/system.py tests/functional/builtins/codegen/test_raw_call.py`

### Commit message

```
the Venom `raw_call` lowering evaluated `gas=` before `value=`
unconditionally, regardless of the order the user wrote them. When
either kwarg has side effects (e.g. `value=self.set_value()`,
`gas=self.set_gas()`), the observable evaluation order was wrong — Vyper
requires left-to-right evaluation.

walk `node.keywords` in source order and materialize `gas=` / `value=`
as encountered, deferring the gas default to after the loop. also guard
against unexpected kwargs with a `CompilerPanic`.

the regression test documents that legacy codegen still evaluates kwargs
in operand order (GH 2863 tracks fixing that) while requiring source-
order for Venom.
```

Fix Venom `raw_call` keyword-argument evaluation order.

### DIFF
--- a/tests/functional/builtins/codegen/test_raw_call.py
+++ b/tests/functional/builtins/codegen/test_raw_call.py
@@ -601,6 +601,40 @@ def bar(f: uint256) -> Bytes[100]:
     )
 
 
+def test_raw_call_msg_data_kwarg_evaluation_order(get_contract, experimental_codegen):
+    code = """
+identity: constant(address) = 0x0000000000000000000000000000000000000004
+
+marker: public(uint256)
+
+@internal
+def set_value() -> uint256:
+    self.marker = 1
+    return 0
+
+@internal
+def set_gas() -> uint256:
+    self.marker = 2
+    return msg.gas
+
+@external
+def value_then_gas() -> uint256:
+    raw_call(identity, msg.data, value=self.set_value(), gas=self.set_gas())
+    return self.marker
+
+@external
+def gas_then_value() -> uint256:
+    raw_call(identity, msg.data, gas=self.set_gas(), value=self.set_value())
+    return self.marker
+    """
+    c = get_contract(code)
+
+    assert c.value_then_gas() == 2
+    # Legacy codegen evaluates raw_call kwargs in operand order; #2863 tracks
+    # fixing it to source order.
+    assert c.gas_then_value() == (1 if experimental_codegen else 2)
+
+
 uncompilable_code = [
     (
         """
--- a/tests/functional/builtins/codegen/test_raw_call.py
+++ b/tests/functional/builtins/codegen/test_raw_call.py
@@ -632,7 +632,11 @@ def gas_then_value() -> uint256:
     assert c.value_then_gas() == 2
     # Legacy codegen evaluates raw_call kwargs in operand order; #2863 tracks
     # fixing it to source order.
-    assert c.gas_then_value() == (1 if experimental_codegen else 2)
+    gas_then_value = c.gas_then_value()
+    if experimental_codegen:
+        assert gas_then_value == 1
+    else:
+        assert gas_then_value == 2
 
 
 uncompilable_code = [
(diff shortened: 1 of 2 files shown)


charles-cooper:
we should just match source order actually

charles-cooper:
tracking issue: https://github.com/vyperlang/vyper/issues/2863

banteg:
Updated to preserve source order for `raw_call` `gas=` / `value=` evaluation. The test now records legacy operand-order behavior under #2863 while asserting source-order behavior for Venom.

vyper/codegen_venom/builtins/system.py
```diff
@@ -118,21 +118,16 @@ def lower_raw_call(node: vy_ast.Call, ctx: VenomCodegenContext) -> Union[IROpera
...
-        value = IRLiteral(0)
-    else:
-        value = Expr(value_node, ctx).lower_value()
+    gas: Optional[IROperand] = None
+    value = IRLiteral(0)
+    for kw in node.keywords:
+        if kw.arg == "gas":
+            gas = Expr(kw.value, ctx).lower_value()
+        elif kw.arg == "value":
+            value = Expr(kw.value, ctx).lower_value()
```
charles-cooper, line 128: might be clearer to evaluate unconditionally.. although maybe it's fine as is. do we have an invariant that "gas" and "value" are the only possible kwargs? like we should check

```suggestion
    for kw in node.keywords:
        if kw.arg == "gas":
            gas = Expr(kw.value, ctx).lower_value()
        elif kw.arg == "value":
            value = Expr(kw.value, ctx).lower_value()
        else:
            raise CompilerPanic("unexpected kwarg", kw)
```

charles-cooper (changes requested):
this looks good; pending minor style suggestion https://github.com/vyperlang/vyper/pull/4985#discussion_r3279737279

banteg:
Updated the raw_call kwarg loop to make the invariant explicit: gas and value still lower in source order, known literal kwargs are accepted, and unexpected kwargs now panic. Validated with the raw-call kwarg-order test and focused ruff check.

tests/functional/builtins/codegen/test_raw_call.py
```diff
@@ -601,6 +601,40 @@ def bar(f: uint256) -> Bytes[100]:
...
+def gas_then_value() -> uint256:
+    raw_call(identity, msg.data, gas=self.set_gas(), value=self.set_value())
+    return self.marker
+    """
+    c = get_contract(code)
+
+    assert c.value_then_gas() == 2
+    # Legacy codegen evaluates raw_call kwargs in operand order; #2863 tracks
+    # fixing it to source order.
+    assert c.gas_then_value() == (1 if experimental_codegen else 2)
```
charles-cooper: branch more explicitly please:

```suggestion
    gas_then_value = c.gas_then_value()
    if experimental_codegen:
        assert gas_then_value == 1
    else:
        assert gas_then_value == 2
```

or if you want to keep the single comparison:
```suggestion
    expected_gas_then_value = 1 if experimental_codegen else 2
    assert c.gas_then_value() == expected_gas_then_value
```

but in general i prefer explicit if statements which reduce expr nesting in any single statement.

banteg: Applied in 840a62b79. The test now stores gas_then_value once and branches explicitly on experimental_codegen. Validated both legacy and experimental-codegen runs for that test, plus ruff on the file.

charles-cooper (approved):

PR #4986, opened 2026-05-20
fix[venom]: fix `extract32` overflow and `bytesN` clamping

banteg:
### What I did

Fixed two Venom `extract32` validation gaps:

- check arithmetic overflow in `start + 32` before accepting the source slice
- apply the same `bytesM` canonicality clamp as legacy codegen for `output_type=bytesN`

### How I did it

Reused the existing slice-bound helper for the `start + 32` check so wraparound cannot make an out-of-bounds read look valid.

Also routed `bytesN` output through the same clamp behavior as legacy codegen, preventing dirty trailing bytes in the returned `bytesN` word.

### How to verify it

- `uv run --python 3.12 pytest tests/functional/builtins/codegen/test_extract32.py --experimental-codegen -q`
- `uv run --python 3.12 pytest tests/functional/builtins/codegen/test_extract32.py -q`

### Commit message

```
Venom `extract32` had two validation gaps versus legacy codegen. The
bounds check computed `start + 32` without overflow protection — a near-
max start would wrap around to a small value, bypassing the out-of-
bounds assertion. Reuse `_assert_slice_bounds` (already used by `slice`)
which handles this correctly.

The output clamping was also incomplete: `_clamp_extract32_result` had
ad-hoc logic that didn't canonicalize `bytesN` output (dirty trailing
bytes from the 32-byte load could leak through). Legacy codegen applies
the basetype clamp for all primitive types. Consolidate all output
clamping into `clamp_basetype`, extending it to handle `BytesM_T`,
`AddressT`, `BoolT`, and 256-bit integers alongside the existing
signed/unsigned integer paths.
```

Fix Venom `extract32` bounds checking and `bytesN` output canonicalization.

### DIFF
--- a/vyper/codegen_venom/arithmetic.py
+++ b/vyper/codegen_venom/arithmetic.py
@@ -16,6 +16,8 @@ from vyper.semantics.types import AddressT, BoolT, BytesM_T, DecimalT, IntegerT
 from vyper.venom.basicblock import IRLiteral, IROperand
 from vyper.venom.builder import VenomBuilder
 
+AnyPrimType = Union[AddressT, BoolT, BytesM_T, DecimalT, IntegerT]  # TODO: move to shared location
+
 
 def safe_add(
     b: VenomBuilder, x: IROperand, y: IROperand, typ: Union[IntegerT, DecimalT]
@@ -235,9 +237,7 @@ def safe_pow(
     return b.exp(x, y)
 
 
-def clamp_basetype(
-    b: VenomBuilder, val: IROperand, typ: Union[AddressT, BoolT, BytesM_T, DecimalT, IntegerT]
-) -> IROperand:
+def clamp_basetype(b: VenomBuilder, val: IROperand, typ: AnyPrimType) -> IROperand:
     """Clamp value to type bounds."""
     if isinstance(typ, BytesM_T):
         if typ.m < 32:
--- a/vyper/codegen_venom/arithmetic.py
+++ b/vyper/codegen_venom/arithmetic.py
@@ -12,7 +12,7 @@ from typing import Optional, Union
 from vyper import ast as vy_ast
 from vyper.codegen.arithmetic import calculate_largest_base, calculate_largest_power
 from vyper.exceptions import CompilerPanic, TypeCheckFailure
-from vyper.semantics.types import DecimalT, IntegerT
+from vyper.semantics.types import AddressT, BoolT, BytesM_T, DecimalT, IntegerT
 from vyper.venom.basicblock import IRLiteral, IROperand
 from vyper.venom.builder import VenomBuilder
 
@@ -235,8 +235,28 @@ def safe_pow(
     return b.exp(x, y)
 
 
-def clamp_basetype(b: VenomBuilder, val: IROperand, typ: Union[IntegerT, DecimalT]) -> IROperand:
+def clamp_basetype(
+    b: VenomBuilder, val: IROperand, typ: Union[AddressT, BoolT, BytesM_T, DecimalT, IntegerT]
+) -> IROperand:
     """Clamp value to type bounds."""
+    if isinstance(typ, BytesM_T):
+        if typ.m < 32:
+            b.assert_(b.iszero(b.shl(IRLiteral(typ.m * 8), val)))
+        return val
+
+    if isinstance(typ, AddressT):
+        ok = b.iszero(b.gt(val, IRLiteral((1 << 160) - 1)))
+        b.assert_(ok)
+        return val
+
+    if isinstance(typ, BoolT):
+        ok = b.iszero(b.gt(val, IRLiteral(1)))
+        b.assert_(ok)
+        return val
+
+    if isinstance(typ, IntegerT) and typ.bits == 256:
+        return val
+
     lo, hi = typ.int_bounds
 
     if typ.is_signed:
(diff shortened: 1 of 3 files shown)


charles-cooper (commented):
surprised we did not have tests for this case already (or maybe they are not active in the experimental-codegen path), lgtm

vyper/codegen_venom/builtins/bytes.py
```diff
@@ -278,9 +278,7 @@ def lower_extract32(node: vy_ast.Call, ctx: VenomCodegenContext) -> IROperand:
     start = Expr(start_node, ctx).lower_value()
 
     # Bounds check: start + 32 <= length
-    end = b.add(start, IRLiteral(32))
-    oob = b.gt(end, src_len)
-    b.assert_(b.iszero(oob))
+    _assert_slice_bounds(ctx, start, IRLiteral(32), src_len)
```
charles-cooper, line 282: legacy uses: `assert start > 0 and start < length - 32`

vyper/codegen_venom/builtins/bytes.py
```diff
@@ -314,8 +312,11 @@ def _clamp_extract32_result(val: IROperand, out_t, ctx: VenomCodegenContext) ->
         mask = (1 << 160) - 1
         too_big = b.gt(val, IRLiteral(mask))
         b.assert_(b.iszero(too_big))
+    elif isinstance(out_t, BytesM_T) and out_t.m < 32:
+        # BytesM values are left-aligned, so any bytes after M must be zero.
+        b.assert_(b.iszero(b.shl(IRLiteral(out_t.m * 8), val)))
```
charles-cooper: this is fine but this whole function is just not needed actually, it should just use the venom pipeline equivalent of clamp_basetype as legacy pipeline does

banteg:
Followed up on the clamp suggestion: extract32 now reuses the Venom basetype clamp helper, and that helper handles address, bool, and bytesM clamping. Validated with the focused extract32 regressions under experimental codegen and ruff check.

vyper/codegen_venom/arithmetic.py
```diff
@@ -235,8 +235,28 @@ def safe_pow(
     return b.exp(x, y)
 
 
-def clamp_basetype(b: VenomBuilder, val: IROperand, typ: Union[IntegerT, DecimalT]) -> IROperand:
+def clamp_basetype(
+    b: VenomBuilder, val: IROperand, typ: Union[AddressT, BoolT, BytesM_T, DecimalT, IntegerT]
```
charles-cooper: will be prettier if we can get `PrimTypes` uniont

vyper/codegen_venom/arithmetic.py
```diff
@@ -235,8 +235,28 @@ def safe_pow(
...
+) -> IROperand:
```
charles-cooper:
```suggestion
AnyPrimType = Union[AddressT, BoolT, BytesM_T, DecimalT, IntegerT]  # TODO: move to more shared location
def clamp_basetype(b: VenomBuilder, val: IROperand, typ: AnyPrimType) -> IROperand:
```

vyper/codegen_venom/arithmetic.py
```diff
@@ -235,8 +235,28 @@ def safe_pow(
...
     """Clamp value to type bounds."""
+    if isinstance(typ, BytesM_T):
+        if typ.m < 32:
+            b.assert_(b.iszero(b.shl(IRLiteral(typ.m * 8), val)))
+        return val
+
+    if isinstance(typ, AddressT):
+        ok = b.iszero(b.gt(val, IRLiteral((1 << 160) - 1)))
```
charles-cooper, line 248: wow, this lost optimizations from the original pipeline (out of scope for this PR, just commenting)

vyper/codegen_venom/builtins/bytes.py
banteg: The current branch keeps this in the shared slice-bounds helper: extract32 calls _assert_slice_bounds(ctx, start, 32, src_len), which asserts start + 32 <= src_len and separately catches arithmetic wraparound. The focused extract32 tests still pass under experimental codegen.

vyper/codegen_venom/builtins/bytes.py
banteg: Done on the current branch: the bespoke extract32 clamp helper is gone, and lower_extract32 now returns clamp_basetype(b, result, out_t). 8619f3b14 only adds the small AnyPrimType alias requested below.

vyper/codegen_venom/arithmetic.py
banteg: Applied in 8619f3b14. clamp_basetype now takes AnyPrimType, with the TODO kept local to arithmetic.py for now.

vyper/codegen_venom/arithmetic.py
banteg: Agreed; 8619f3b14 adds AnyPrimType and uses it in clamp_basetype.

vyper/codegen_venom/arithmetic.py
banteg: Acknowledged. I left the optimization parity issue out of this PR as suggested; this branch stays focused on the missing bounds/clamp behavior.

charles-cooper (approved):
lgtm; thanks

banteg:
Pushed 127b7ecc0 to fix the red CI from the missing `AddressT` import in `vyper/codegen_venom/builtins/bytes.py`.

Validation:
- `uv run --python 3.12 pytest --experimental-codegen tests/functional/builtins/codegen/test_slice.py tests/functional/syntax/test_address_code.py tests/functional/syntax/test_slice.py -q`

PR #4988, opened 2026-05-20
fix[venom]: fix `raw_call` kwarg evaluation order

banteg:
_co-authored by chatgpt_

### What I did

Fixed two Venom external-call lowering issues:

- evaluate side-effectful call kwargs in source order
- freeze `default_return_value` before the external call, including
  composite defaults that need a stable memory copy

### How I did it

Changed external call kwarg parsing to walk `node.keywords` directly,
materializing `value=`, `gas=`, and `skip_contract_check=` in the order
written by the user.

`default_return_value` follows the same source-order rule: it is
materialized when its keyword is encountered, before the external call.
Because the fallback is only consumed later, after empty returndata is
observed, primitive defaults stay on the stack and composite defaults are
copied to a fresh temporary so the call or later memory writes cannot
mutate the fallback.

The regression tests cover both source-order kwarg side effects and the
empty-returndata fallback timing. Legacy codegen still documents its
operand-order behavior pending GH 2863.

### How to verify it

- `uv run --python 3.12 pytest tests/functional/codegen/calling_convention/test_external_contract_calls.py::test_default_return_value_reads_storage_before_empty_returndata_call tests/functional/codegen/calling_convention/test_external_contract_calls.py::test_external_call_kwarg_source_order --experimental-codegen -q`
- `uv run --python 3.12 pytest tests/functional/codegen/calling_convention/test_external_contract_calls.py::test_default_return_value_reads_storage_before_empty_returndata_call tests/functional/codegen/calling_convention/test_external_contract_calls.py::test_external_call_kwarg_source_order -q`
- `uv run --python 3.12 ruff check vyper/codegen_venom/expr.py tests/functional/codegen/calling_convention/test_external_contract_calls.py`

### Commit message

```
Venom lowered external call keyword arguments in a way that did not
match Vyper's source-order evaluation rules. `gas=` and `value=` were
materialized by keyword kind rather than in the order written, so side
effects in one kwarg could change the value observed by another.

The same source-order rule applies to `default_return_value`: the
expression must be evaluated when its keyword is encountered, before the
external call. Because the fallback value is only consumed later, after
empty returndata is observed, the evaluated value must also be frozen so
the call or subsequent memory writes cannot change it.

Parse external call kwargs by walking the AST keyword list, materialize
`default_return_value` at that point, and copy composite defaults into a
fresh temporary so the empty-returndata block reads the source-order
value.
```

Fix Venom external-call keyword argument evaluation order and stabilize
`default_return_value` fallback values.

### DIFF
--- a/tests/functional/codegen/calling_convention/test_external_contract_calls.py
+++ b/tests/functional/codegen/calling_convention/test_external_contract_calls.py
@@ -2450,19 +2450,21 @@ def bar(foo: Foo):
     c.bar(bad_2.address)
 
 
-def test_default_return_value_reads_storage_after_call(get_contract, experimental_codegen):
+def test_default_return_value_reads_storage_before_empty_returndata_call(
+    get_contract, experimental_codegen
+):
     callee_code = """
 interface Caller:
     def set_x(v: uint256): nonpayable
 
 @external
-def poke(c: address):
+def set_x_without_return(c: address):
     extcall Caller(c).set_x(99)
     """
 
     caller_code = """
 interface Callee:
-    def poke(c: address) -> uint256: nonpayable
+    def set_x_without_return(c: address) -> uint256: nonpayable
 
 x: public(uint256)
 
@@ -2473,7 +2475,7 @@ def set_x(v: uint256):
 @external
 def run(callee: address) -> uint256:
     self.x = 7
-    return extcall Callee(callee).poke(self, default_return_value=self.x)
+    return extcall Callee(callee).set_x_without_return(self, default_return_value=self.x)
     """
 
     callee = get_contract(callee_code)
--- a/tests/functional/codegen/calling_convention/test_external_contract_calls.py
+++ b/tests/functional/codegen/calling_convention/test_external_contract_calls.py
@@ -2450,7 +2450,7 @@ def bar(foo: Foo):
     c.bar(bad_2.address)
 
 
-def test_default_return_value_reads_storage_after_call(get_contract):
+def test_default_return_value_reads_storage_after_call(get_contract, experimental_codegen):
     callee_code = """
 interface Caller:
     def set_x(v: uint256): nonpayable
@@ -2479,7 +2479,11 @@ def run(callee: address) -> uint256:
     callee = get_contract(callee_code)
     caller = get_contract(caller_code)
 
-    assert caller.run(callee.address) == 99
+    output = caller.run(callee.address)
+    if experimental_codegen:
+        assert output == 7
+    else:
+        assert output == 99
     assert caller.x() == 99
 
 
@@ -2510,9 +2514,13 @@ def run() -> uint256:
     c = get_contract(code)
     env.set_balance(env.deployer, 10**18)
 
-    # Legacy codegen evaluates call kwargs in operand order; #2863 tracks
-    # fixing it to source order.
-    assert c.run(value=10) == (20 if experimental_codegen else 10)
+    output = c.run(value=10)
+    if experimental_codegen:
+        assert output == 20
+    else:
+        # Legacy codegen evaluates call kwargs in operand order; #2863 tracks
+        # fixing it to source order.
+        assert output == 10
     assert c.counter() == 2
 
 
--- a/tests/functional/codegen/calling_convention/test_external_contract_calls.py
+++ b/tests/functional/codegen/calling_convention/test_external_contract_calls.py
@@ -2450,6 +2450,72 @@ def bar(foo: Foo):
     c.bar(bad_2.address)
 
 
+def test_default_return_value_reads_storage_after_call(get_contract):
+    callee_code = """
+interface Caller:
+    def set_x(v: uint256): nonpayable
+
+@external
+def poke(c: address):
+    extcall Caller(c).set_x(99)
+    """
+
+    caller_code = """
+interface Callee:
+    def poke(c: address) -> uint256: nonpayable
+
+x: public(uint256)
+
+@external
+def set_x(v: uint256):
+    self.x = v
+
+@external
+def run(callee: address) -> uint256:
+    self.x = 7
+    return extcall Callee(callee).poke(self, default_return_value=self.x)
+    """
+
+    callee = get_contract(callee_code)
+    caller = get_contract(caller_code)
+
+    assert caller.run(callee.address) == 99
+    assert caller.x() == 99
+
+
+def test_external_call_kwarg_source_order(env, get_contract, experimental_codegen):
+    code = """
+interface Foo:
+    def pay() -> uint256: payable
+
+counter: public(uint256)
+
+@internal
+def bump() -> uint256:
+    self.counter += 1
+    return self.counter
+
+@external
+@payable
+def pay() -> uint256:
+    return msg.value * 10
+
+@external
+@payable
+def run() -> uint256:
+    self.counter = 0
+    return extcall Foo(self).pay(gas=50000 + self.bump(), value=self.bump())
+    """
+
+    c = get_contract(code)
+    env.set_balance(env.deployer, 10**18)
+
+    # Legacy codegen evaluates call kwargs in operand order; #2863 tracks
+    # fixing it to source order.
+    assert c.run(value=10) == (20 if experimental_codegen else 10)
+    assert c.counter() == 2
+
+
 def test_contract_address_evaluation(get_contract):
     callee_code = """
 # implements: Foo
--- a/vyper/codegen_venom/expr.py
+++ b/vyper/codegen_venom/expr.py
@@ -61,7 +61,7 @@ class _CallKwargs:
     value: IROperand  # ETH value to send (CALL only)
     gas: IROperand  # Gas limit for the call
     skip_contract_check: bool  # Skip extcodesize check
-    default_return_value: Optional[vy_ast.VyperNode]  # Default if returndatasize==0
+    default_return_value: Optional[VyperValue]  # Default if returndatasize==0
 
 
 # Environment variable prefixes for attribute access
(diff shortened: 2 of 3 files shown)


charles-cooper:
i think we should preserve source-order; if legacy does not respect source order (known class of bug), we should not adhere to legacy semantics

banteg:
Updated to preserve source order for external-call kwargs. The `default_return_value` fix remains deferred to the empty-returndata branch, and the test now records legacy operand-order behavior under #2863 while asserting source-order behavior for Venom.

vyper/codegen_venom/expr.py
```diff
@@ -1653,40 +1653,30 @@ def _parse_external_call_kwargs(self, call_node) -> _CallKwargs:
...
-                gas = kw_val
+                gas = Expr(kw.value, self.ctx).lower_value()
             elif kw.arg == "skip_contract_check":
+                kw_val = Expr(kw.value, self.ctx).lower_value()
                 # Must be a literal True/False
                 if not isinstance(kw_val, IRLiteral):  # pragma: nocover
                     raise CompilerPanic(f"Expected IRLiteral for keyword, got {type(kw_val)}")
                 skip_contract_check = bool(kw_val.value)
+            elif kw.arg == "default_return_value":
+                default_return_value = kw.value
```
charles-cooper: this is weird, why does it not evaluate the keyword like the other cases? like 
```suggestion
                default_return_value = Expr(kw.value, self.ctx).lower_value()
```

tests/functional/codegen/calling_convention/test_external_contract_calls.py
```diff
@@ -2450,6 +2450,72 @@ def bar(foo: Foo):
...
+    self.counter = 0
+    return extcall Foo(self).pay(gas=50000 + self.bump(), value=self.bump())
+    """
+
+    c = get_contract(code)
+    env.set_balance(env.deployer, 10**18)
+
+    # Legacy codegen evaluates call kwargs in operand order; #2863 tracks
+    # fixing it to source order.
+    assert c.run(value=10) == (20 if experimental_codegen else 10)
```
charles-cooper: branch please -- 
```suggestion
    output = c.run(value=10)
    if experimental_codegen:
        assert output == 20
    else:
        assert output == 10
```

tests/functional/codegen/calling_convention/test_external_contract_calls.py
```diff
@@ -2450,6 +2450,72 @@ def bar(foo: Foo):
...
+x: public(uint256)
+
+@external
+def set_x(v: uint256):
+    self.x = v
+
+@external
+def run(callee: address) -> uint256:
+    self.x = 7
+    return extcall Callee(callee).poke(self, default_return_value=self.x)
```
charles-cooper: hang on, i would actually expect the default return value to just return the value of `self.x` before the call. like what does this test actually

vyper/codegen_venom/expr.py
```diff
@@ -1851,8 +1841,9 @@ def _lower_external_call(self) -> VyperValue:
             b.set_block(default_bb)
 
             # Store default value
-            default_vv = call_kwargs.default_return_value
-            assert default_vv is not None
+            default_node = call_kwargs.default_return_value
+            assert default_node is not None
+            default_vv = Expr(default_node, self.ctx).lower()
```
charles-cooper: this looks extremely wrong, i think this should be evaluated in source order. like the eval ordering should be `evaluate fn args in order, then evaluate fn`

charles-cooper (changes requested):
i think the source order changes look good, but i don't think the change to default_return_value is right. default_return_value should also be evaluated in source order.

charles-cooper:
> i think the source order changes look good, but i don't think the change to default_return_value is right. default_return_value should also be evaluated in source order.

i was thinking it might be more subtle than this, like maybe we want to elide the effects if the call succeeds? but yea, no, we should evaluate default_return_value unconditionally.

banteg:
Updated default_return_value handling so the fallback expression is lowered and materialized in keyword source order before the external call, then only stored in the empty-returndata branch. Also branched the legacy versus experimental assertions in the tests. Validated the two focused tests in both default and experimental-codegen modes plus ruff check.

tests/functional/codegen/calling_convention/test_external_contract_calls.py
charles-cooper: is `poke()` supposed to fail? like i don't see why returning default_return_value gets triggered here

vyper/codegen_venom/expr.py
```diff
@@ -1670,7 +1670,14 @@ def _parse_external_call_kwargs(self, call_node) -> _CallKwargs:
                     raise CompilerPanic(f"Expected IRLiteral for keyword, got {type(kw_val)}")
                 skip_contract_check = bool(kw_val.value)
             elif kw.arg == "default_return_value":
-                default_return_value = kw.value
+                default_vv = Expr(kw.value, self.ctx).lower()
+                if default_vv.location is not None and default_vv.location in (
+                    DataLocation.STORAGE,
+                    DataLocation.TRANSIENT,
+                ):
+                    materialized = self.ctx.unwrap(default_vv)
```
charles-cooper: i think this `materialized` case is overfitting from the agent that previously wrote this code. in theory we should just be able to use `Expr(kw.value, self.ctx).lower()` and the code that consumes it "just works"

vyper/codegen_venom/expr.py
banteg: Addressed on the current branch. default_return_value is now lowered while walking call_node.keywords in source order, before the external call is emitted; dcf6fe98a keeps that path as lower()+unwrap() so storage/transient and complex values are frozen at the same point.

tests/functional/codegen/calling_convention/test_external_contract_calls.py
banteg: This is already explicit on the current branch: output is stored once, then the assertion branches on experimental_codegen. I re-ran that test in both default and experimental-codegen modes after dcf6fe98a.

tests/functional/codegen/calling_convention/test_external_contract_calls.py
banteg: Clarified in dcf6fe98a by renaming the test and callee method. The callee is not supposed to fail; it mutates caller.x and returns zero returndata, while the interface expects uint256, so default_return_value is the fallback path being tested. The Venom assertion is that self.x was read before that call mutates storage.

banteg: No, it does not fail. The implementation returns empty returndata, and the caller-side interface return type is what makes the default_return_value branch relevant. I renamed the method to set_x_without_return in dcf6fe98a to make that setup explicit.

vyper/codegen_venom/expr.py
banteg: Agreed; this was fixed before the current head and remains so after dcf6fe98a. default_return_value is no longer lowered in the post-call default block; it is evaluated during the source-order keyword walk before the CALL/STATICCALL is emitted.

vyper/codegen_venom/expr.py
banteg: Simplified in dcf6fe98a. The explicit materialized storage/transient branch is gone; default_return_value now uses the same lower() path and then unwraps once immediately so the value is frozen in keyword source order before the external call.

vyper/codegen_venom/expr.py
```diff
@@ -1671,13 +1671,10 @@ def _parse_external_call_kwargs(self, call_node) -> _CallKwargs:
...
                 default_vv = Expr(kw.value, self.ctx).lower()
-                if default_vv.location is not None and default_vv.location in (
-                    DataLocation.STORAGE,
-                    DataLocation.TRANSIENT,
-                ):
-                    materialized = self.ctx.unwrap(default_vv)
-                    default_vv = VyperValue.from_stack_op(materialized, default_vv.typ)
-                default_return_value = default_vv
+                # Freeze the expression here; the default block runs after the external call.
+                default_return_value = VyperValue.from_stack_op(
```
charles-cooper: oh i see why this looks so wrong -- it *is* wrong for non-primitive types. i think we should just use the new `materialize_vyper_value` (iirc its name correctly)

charles-cooper: or maybe one of the `ensure_in_memory` helpers

banteg: Addressed in 7c373e5e2. Primitive defaults are still frozen as stack values, while composite default_return_value expressions are copied into a fresh memory temporary via materialize_value before the external call. Validation: uv run --python 3.12 pytest --experimental-codegen tests/functional/codegen/calling_convention/test_external_contract_calls.py -k "default_override or default_return_value_reads_storage_before_empty_returndata_call or external_call_kwarg_source_order" -q; uv run --python 3.12 ruff check vyper/codegen_venom/context.py vyper/codegen_venom/expr.py tests/functional/codegen/calling_convention/test_external_contract_calls.py

banteg: Covered by 7c373e5e2 as well; composite defaults now go through materialize_value before the call.

charles-cooper (approved):

vyper/codegen_venom/expr.py
```diff
@@ -1653,40 +1653,42 @@ def _parse_external_call_kwargs(self, call_node) -> _CallKwargs:
...
                 if not isinstance(kw_val, IRLiteral):  # pragma: nocover
                     raise CompilerPanic(f"Expected IRLiteral for keyword, got {type(kw_val)}")
                 skip_contract_check = bool(kw_val.value)
+            elif kw.arg == "default_return_value":
+                default_vv = Expr(kw.value, self.ctx).lower()
+                # Freeze the expression here; the default block runs after the
+                # external call. Primitive values can stay on the stack, but
+                # composite values need a fresh memory copy instead of a
+                # pointer to a source location that later code may mutate.
+                if default_vv.typ._is_prim_word:
```
charles-cooper, line 1681: we should refactor this into a function (and i'm surprised there isn't one in context.py already)

charles-cooper: extracting to new issue

charles-cooper: https://github.com/vyperlang/vyper/issues/5008

PR #4989, opened 2026-05-20
fix[venom]: short-circuit list membership

banteg:
### What I did

Changed Venom lowering for list-literal membership (`x in [a, b, c]` / `x not in [a, b, c]`) to short-circuit element evaluation.

### How I did it

Replaced the eager list-element evaluation path with conditional lowering that stops evaluating once the membership result is known.

The previous Venom path evaluated every list element first, then combined equality results. That made side effects in later list elements run even after the result had already been determined. Legacy codegen lowers this as nested conditionals, so `1 in [1, self.bump()]` does not call `self.bump()`.

### How to verify it

- `uv run --python 3.12 pytest tests/functional/codegen/features/test_comparison.py --experimental-codegen -q`
- `uv run --python 3.12 pytest tests/functional/codegen/features/test_comparison.py -q`
- `uv run --python 3.12 ruff check vyper/codegen_venom/expr.py tests/functional/codegen/features/test_comparison.py`

### Commit message

fix: short-circuit Venom list membership

Fix Venom list-literal membership lowering to short-circuit element evaluation.

### DIFF
--- a/tests/functional/codegen/features/test_comparison.py
+++ b/tests/functional/codegen/features/test_comparison.py
@@ -46,3 +46,26 @@ def test_in(addr: address) -> bool:
         assert c.test_in(t) is (True if not invert else False)
     for t in should_not_in:
         assert c.test_in(t) is (True if invert else False)
+
+
+@pytest.mark.parametrize("invert", (True, False))
+def test_in_list_short_circuits_side_effects(get_contract, invert):
+    INVERT = "not" if invert else ""
+    code = f"""
+counter: uint256
+
+@internal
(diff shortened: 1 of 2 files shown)


charles-cooper:
this is a known issue in the legacy pipeline https://github.com/vyperlang/vyper/issues/4465, i think the new version in the venom pipeline is what we want

charles-cooper:
if i'm understanding correctly, this is a regression which reintroduces the known bug in #4465; closing. (please reopen if i have misunderstood)

PR #4990, opened 2026-05-20
fix: treat venom gas as ordered

banteg:
### What I did

Treated Venom `gas` instructions as non-removable, non-substitutable, and ordered inside DFT.

### How I did it

Updated the Venom optimizer model so `gas` behaves like an ordered observation of the current execution point:

- marked `gas` volatile so remove-unused-vars does not delete it
- excluded `gas` from available-expression substitution
- made DFT preserve instruction order across `gas` reads

`GAS` observes remaining gas at its exact program point. Removing an unused `msg.gas` read or moving independent work such as `SLOAD` across it can change later gas observations.

### How to verify it

- `uv run --python 3.12 pytest tests/functional/codegen/features/test_gas.py --experimental-codegen -q`
- `uv run --python 3.12 pytest tests/functional/codegen/features/test_gas.py -q`
- `uv run --python 3.12 pytest tests/unit/compiler/venom/test_common_subexpression_elimination.py -q`
- `uv run --python 3.12 ruff check vyper/venom/basicblock.py vyper/venom/passes/dft.py vyper/venom/analysis/available_expression.py tests/functional/codegen/features/test_gas.py`

### Commit message

fix: treat Venom gas as ordered

Fix Venom optimizer handling of `msg.gas` so gas reads are preserved and remain ordered.

### DIFF
--- a/vyper/venom/passes/dft.py
+++ b/vyper/venom/passes/dft.py
@@ -8,6 +8,9 @@ from vyper.venom.function import IRFunction
 from vyper.venom.passes.base_pass import IRPass
 
 
+ORDERED_INSTRUCTIONS = frozenset(["gas"])
+
+
 class DFTPass(IRPass):
     function: IRFunction
     data_offspring: dict[IRInstruction, OrderedSet[IRInstruction]]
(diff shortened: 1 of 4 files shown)


charles-cooper:
i think gas should not be volatile, and we treat it as re-orderable.

vyper/venom/passes/dft.py
```diff
@@ -146,13 +149,23 @@ def _calculate_dependency_graphs(self, bb: IRBasicBlock) -> None:
         #
         last_write_effects: dict[effects.Effects, IRInstruction] = {}
         all_read_effects: dict[effects.Effects, list[IRInstruction]] = defaultdict(list)
+        last_ordered_instruction: IRInstruction | None = None
+        previous_instructions: OrderedSet[IRInstruction] = OrderedSet()
 
         for inst in non_phis:
+            if last_ordered_instruction is not None:
```
charles-cooper, line 156: hmm i why the agent wanted to make this change, @harkal any ideas?

banteg:
Closing per maintainer guidance: `gas` is intended to remain non-volatile and re-orderable in Venom, so this change should not land.

PR #4993, opened 2026-05-20
fix[venom]: handle duplicate phi inputs

banteg:
### What I did

Fixed Venom stack lowering for join blocks where multiple phi nodes use the same incoming value from one predecessor.

### How I did it

Preserved duplicate phi inputs for stack layout, allowed duplicate copies of the same phi operand, and kept the existing ambiguity check for different phi operands.

The old stack/liveness path used set semantics for predecessor inputs. When two phi nodes both needed the same incoming variable, one predecessor could enter the join with only one stack slot, causing stack underflow at O0 or wrong-code after optimization. A minimal join has `%x = phi @p1, %a, @p2, %b` and `%y = phi @p1, %a, @p2, %c`; the `p1` edge needs two copies of `%a` while the `p2` edge needs `%b` and `%c`.

### How to verify it

- `uv run --python 3.12 pytest tests/unit/compiler/venom/test_duplicate_operands.py -q`
- `uv run --python 3.12 pytest tests/unit/compiler/venom/test_stack_order.py tests/unit/compiler/venom/test_duplicate_operands.py -q`
- `uv run --python 3.12 ruff check vyper/venom/analysis/liveness.py vyper/venom/analysis/stack_order.py vyper/venom/stack_model.py vyper/venom/venom_to_assembly.py tests/unit/compiler/venom/test_duplicate_operands.py`

### Commit message

fix: handle duplicate phi inputs

Fix Venom stack lowering for duplicate phi inputs at join blocks.

### DIFF
--- a/vyper/venom/passes/single_use_expansion.py
+++ b/vyper/venom/passes/single_use_expansion.py
@@ -20,12 +20,41 @@ class SingleUseExpansion(IRPass):
     """
 
     def run_pass(self):
+        self._expand_phi_operands()
+        self.analyses_cache.invalidate_analysis(DFGAnalysis)
         self.dfg = self.analyses_cache.request_analysis(DFGAnalysis)
         for bb in self.function.get_basic_blocks():
             self._process_bb(bb)
 
         self.analyses_cache.invalidate_analysis(DFGAnalysis)
         self.analyses_cache.invalidate_analysis(LivenessAnalysis)
+
+    def _expand_phi_operands(self):
+        for target in self.function.get_basic_blocks():
+            phis = list(target.phi_instructions)
+            if len(phis) == 0:
+                continue
+
+            operands_by_source: dict[IRVariable, list[tuple[IRInstruction, int]]]
+            sources: dict[str, dict[IRVariable, list[tuple[IRInstruction, int]]]] = {}
+            for phi in phis:
+                for i in range(0, len(phi.operands), 2):
+                    label, var = phi.operands[i], phi.operands[i + 1]
+                    source = label.name
+                    operands_by_source = sources.setdefault(source, {})
+                    operands_by_source.setdefault(var, []).append((phi, i + 1))
+
+            for source, operands in sources.items():
+                source_bb = self.function.get_basic_block(source)
+                insert_idx = len(source_bb.instructions) - 1
+
+                for var, uses in operands.items():
+                    for phi, operand_idx in uses[1:]:
+                        new_var = self.function.get_next_variable()
+                        to_insert = IRInstruction("assign", [var], outputs=[new_var])
+                        source_bb.insert_instruction(to_insert, index=insert_idx)
+                        insert_idx += 1
+                        phi.operands[operand_idx] = new_var
 
     def _process_bb(self, bb):
         i = 0
--- a/vyper/venom/passes/single_use_expansion.py
+++ b/vyper/venom/passes/single_use_expansion.py
@@ -20,41 +20,12 @@ class SingleUseExpansion(IRPass):
     """
 
     def run_pass(self):
-        self._expand_phi_operands()
-        self.analyses_cache.invalidate_analysis(DFGAnalysis)
-        self.dfg = self.analyses_cache.request_analysis(DFGAnalysis)
+        self.dfg = self.analyses_cache.force_analysis(DFGAnalysis)
         for bb in self.function.get_basic_blocks():
             self._process_bb(bb)
 
         self.analyses_cache.invalidate_analysis(DFGAnalysis)
         self.analyses_cache.invalidate_analysis(LivenessAnalysis)
-
-    def _expand_phi_operands(self):
-        for target in self.function.get_basic_blocks():
-            phis = list(target.phi_instructions)
-            if len(phis) == 0:
-                continue
-
-            operands_by_source: dict[IRVariable, list[tuple[IRInstruction, int]]]
-            sources: dict[str, dict[IRVariable, list[tuple[IRInstruction, int]]]] = {}
-            for phi in phis:
-                for i in range(0, len(phi.operands), 2):
-                    label, var = phi.operands[i], phi.operands[i + 1]
-                    source = label.name
-                    operands_by_source = sources.setdefault(source, {})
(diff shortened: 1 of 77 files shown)


banteg:
CI follow-up: rebased this branch onto current master, applied Black formatting to the touched Venom backend files, and updated the PR title to include the required venom scope. Validated the duplicate-operands test, the tail-merge regression that failed in CI, Black, and ruff locally.

HodanPlodky:
This issue arises because we assume single use for variable and that is not handled correctly for `phi`s. I think the fix should be done in the `SingleUseExpansionPass` instead for creating more complicated `venom_to_assembly`

banteg:
Agreed, moved this into `SingleUseExpansion` in 04e141982. The pass now rewrites repeated incoming phi operands per predecessor by inserting assign copies before the predecessor terminator, so liveness/stack ordering/codegen can keep the existing unique stack operand assumptions. I also removed the duplicate-stack reorder path from the net PR diff.

Validated with:
- `uv run --python 3.12 pytest tests/unit/compiler/venom/test_duplicate_operands.py -q`

vyper/venom/passes/single_use_expansion.py
```diff
@@ -20,13 +20,42 @@ class SingleUseExpansion(IRPass):
     """
 
     def run_pass(self):
+        self._expand_phi_operands()
+        self.analyses_cache.invalidate_analysis(DFGAnalysis)
```
charles-cooper: or just force_analysis at this point

vyper/venom/passes/single_use_expansion.py
```diff
@@ -20,13 +20,42 @@ class SingleUseExpansion(IRPass):
...
         self.dfg = self.analyses_cache.request_analysis(DFGAnalysis)
         for bb in self.function.get_basic_blocks():
             self._process_bb(bb)
 
         self.analyses_cache.invalidate_analysis(DFGAnalysis)
         self.analyses_cache.invalidate_analysis(LivenessAnalysis)
 
+    def _expand_phi_operands(self):
```
charles-cooper: this... seems like a performance (bytecode) regression

vyper/venom/passes/single_use_expansion.py
banteg: Addressed in 09f390c1b: SingleUseExpansion now uses force_analysis(DFGAnalysis) instead of invalidate+request.

vyper/venom/passes/single_use_expansion.py
banteg: Reworked in 09f390c1b to avoid predecessor IR expansion entirely. The fix now keeps duplicate phi inputs in the predecessor stack layout and handles duplicate stack reorder/phi collapse in assembly generation, which also restores the internal-call copy-forwarding CI checks locally. Validation: uv run --python 3.12 pytest tests/unit/compiler/venom/test_duplicate_operands.py tests/unit/compiler/venom/test_internal_call_result_binding.py -q; uv run --python 3.12 pytest tests/unit/compiler/venom/test_stack_order.py tests/unit/compiler/venom/test_multi_entry_block.py -q; focused ruff check.

banteg:
Fixed the lint failure by applying Black to `vyper/venom/venom_to_assembly.py` and pushed `b1600c91e` without force-pushing.

Validation:
- `.venv/bin/black --check -C --force-exclude=vyper/version.py vyper/venom/venom_to_assembly.py`
- `.venv/bin/pytest tests/unit/compiler/venom/test_duplicate_operands.py -q` (2 passed)
- `git diff --check`

banteg:
CI follow-up: pushed ec10afbdb to fix the symbolic-tests failure. The branch liveness change no longer carries dead phi inputs into a halting block, so venom_to_assembly now skips phi stack lowering when the phi output is not live.

Validation:
- uv run --python 3.12 pytest tests/unit/compiler/venom/test_tail_merge.py::test_no_merge_with_phi_nodes -q
- uv run --python 3.12 python -c "from tests.hevm import _prep_hevm_venom; ..." for the CI-failing phi block
- uv run --python 3.12 pytest -m "not fuzzing" --optimize gas --evm-version shanghai --evm-backend py-evm tests/unit/compiler/venom/test_internal_call_result_binding.py::test_mutable_internal_memory_arg_is_not_forwarded tests/unit/compiler/venom/test_internal_call_result_binding.py::test_non_return_invoke_source_copy_is_not_forwarded -q
- uv run --python 3.12 pytest tests/unit/compiler/venom/test_duplicate_operands.py -q
- uv run --python 3.12 black --check -C --force-exclude=vyper/version.py vyper/venom/venom_to_assembly.py
- uv run --python 3.12 ruff check vyper/venom/venom_to_assembly.py
- git diff --check

charles-cooper:
i think @HodanPlodky is working on an alternative implementation in https://github.com/vyperlang/vyper/pull/5009 -- @HodanPlodky does #5009 supersede this?

banteg:
Followed up on the py311/shanghai CI failure in cc1bead80. The failing shard was order-dependent: the new duplicate-phi runtime test constructs the py-evm env with --evm-version shanghai, which mutates the process default EVM version; the later internal-call tests were implicitly relying on the default remaining prague so MCOPY exists. I pinned those MCOPY-specific tests to prague explicitly.

Validation:
- uv run --python 3.11 pytest -n 0 -m "not fuzzing" --optimize gas --evm-version shanghai --evm-backend py-evm tests/unit/compiler/venom/test_duplicate_operands.py::test_phi_duplicate_incoming_operands tests/unit/compiler/venom/test_internal_call_result_binding.py -q
- uv run --python 3.11 pytest -n 0 -m "not fuzzing" --optimize gas --evm-version shanghai --evm-backend py-evm tests/unit/compiler/venom -q
- uv run --python 3.11 ruff check tests/unit/compiler/venom/test_internal_call_result_binding.py

HodanPlodky:
> i think @HodanPlodky is working on an alternative implementation in #5009 -- @HodanPlodky does #5009 supersede this?

Yeah I think that PR is better way to fix this issue

banteg:
Closing this in favor of #5009 after @HodanPlodky confirmed that PR is the better way to fix the phi single-use issue. Leaving the branch intact for reference.

PR #4994, opened 2026-05-20
fix[venom]: align print selector storage

banteg:
### What I did

Fixed Venom lowering for the debug-only `print()` builtin so the console selector is stored at the offset used by the subsequent `staticcall`.

This applies to both default `log(string,bytes)` mode and `hardhat_compat=True` direct `log(...)` mode.

### How I did it

Stored the unshifted 4-byte method selector at `buf`, matching the legacy IR layout for calls that start calldata at `buf + 28`.

Before this patch, Venom stored `method_id << 224` at `buf`. That places the selector at `buf..buf+3`, while the call starts at `buf+28`, so the call sent zeroes as the selector. Storing the unshifted selector puts it at `buf+28..buf+31`, so calldata starts with the intended selector.

### How to verify it

- `uv run --python 3.12 pytest tests/functional/syntax/test_print.py -q`
- `uv run --python 3.12 ruff check vyper/codegen_venom/builtins/misc.py tests/functional/syntax/test_print.py`

### Commit message

```
venom stored `method_id << 224` (left-aligned in word) at `buf`, placing
the 4-byte selector at `buf..buf+3`. but the `staticcall` for the
console precompile starts calldata at `buf + 28`, so the call sent
zeroes as the method selector. store the unshifted `method_id` at `buf`
instead, which places the selector at `buf+28..buf+31` where the call
reads it.

applies to both default `log(string,bytes)` mode and
`hardhat_compat=True` direct `log(...)` mode.
```

Fix Venom debug `print()` console selector encoding.

### DIFF
--- a/tests/functional/syntax/test_print.py
+++ b/tests/functional/syntax/test_print.py
@@ -105,6 +105,9 @@ def foo(x: uint256):
     opcodes = out["opcodes_runtime"]
     selector = method_id_int(signature)
 
+    # This is intentionally a narrow opcode regression and is brittle to
+    # unrelated codegen changes. If it starts failing from harmless lowering
+    # churn, replace it with functional console-precompile instrumentation.
     assert f"PUSH4 0x{selector:08X}" in opcodes
     assert f"PUSH32 0x{selector:08X}{'0' * 56}" not in opcodes
 
--- a/tests/functional/syntax/test_print.py
+++ b/tests/functional/syntax/test_print.py
@@ -1,6 +1,8 @@
 import pytest
 
 from vyper import compiler
+from vyper.compiler.settings import Settings
+from vyper.utils import method_id_int
 
 valid_list = [
     """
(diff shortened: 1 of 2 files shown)


tests/functional/syntax/test_print.py
```diff
@@ -79,3 +81,29 @@ def foo():
...
+
+
+@pytest.mark.parametrize(
+    "print_call,signature",
+    [
+        ("print(x)", "log(string,bytes)"),
+        ("print(x, hardhat_compat=True)", "log(uint256)"),
+    ],
+)
+def test_venom_print_selector_stored_at_call_offset(print_call, signature):
```
charles-cooper, line 90: the fix looks good, but this test is... idk -- it tests for a specific form in the bytecode rather than be a functional test. i think the proper way to test this would be to instrument / have an actual precompile in the evm, not sure how hard this is to do though

banteg:
Agreed that a functional console precompile test would be stronger. I did not find an existing backend hook for that in the current test fixtures, so I left this as a narrow bytecode regression for the selector placement and only normalized the PR title scope for now.

charles-cooper:
there seems to be a merge conflict with master now

banteg:
Merged current master in 8fcef5973 with a normal merge commit; the only conflict was in tests/functional/syntax/test_print.py, where I kept both the folded hardhat_compat coverage from master and this PR selector-offset regression. Validated with `uv run --python 3.12 pytest tests/functional/syntax/test_print.py -q` and `uv run --python 3.12 ruff check tests/functional/syntax/test_print.py`.

tests/functional/syntax/test_print.py
banteg: Agreed a functional console-precompile test would be stronger. I looked for an existing EVM fixture hook to install or instrument the console address and did not find one in the current test setup, so this PR keeps the narrow selector-placement bytecode regression. I also merged current master in 8fcef5973 and kept the folded hardhat_compat coverage added there.

charles-cooper: i'm ok with this if there is a comment saying that the current test is a bit brittle to codegen changes and if that comes up we should try to the functional instrumenting test in

banteg: Done in 3b7752d52. I added a comment before the opcode assertions saying this is intentionally a narrow/brittle selector-placement regression and that harmless lowering churn should push us toward a functional console-precompile instrumentation test.

Validation:
- `uv run --python 3.12 pytest tests/functional/syntax/test_print.py -q`
- `uv run --python 3.12 ruff check tests/functional/syntax/test_print.py`

charles-cooper (approved):

charles-cooper:
looks like CI is failing on lint

banteg:
Fixed in 8900212e7 by applying the CI Black 26.3.1 formatting to `tests/functional/syntax/test_print.py`.

Validation:
- `uv run --python 3.12 pytest tests/functional/syntax/test_print.py -q`
- `uv run --python 3.12 black --check -C --force-exclude=vyper/version.py tests/functional/syntax/test_print.py`

PR #4996, opened 2026-05-21
fix[ux]: fix panic in pow folding

### LINKED ISSUE #4980: Negative-base literal power raises uncaught math domain error
### Version Information

* vyper Version: 0.4.4+commit.bdf00d31
* OS: Linux Ubuntu Jammy
* Python Version: Python 3.11

### What's your issue about?

Compiling a contract that contains a negative-base literal `**` expression — even when the result is a valid, in-range value — aborts the compiler with a bare, uncaught Python `ValueError: math domain error`. There is no Vyper exception class, no source location, and no hint; the failure surfaces as an internal Python traceback. This rejects *valid* code: `(-2) ** 2 == 4` and `(-1) ** 100 == 1` both fit comfortably in `int256`, yet neither compiles.

EVM version: cancun.

Code that caused the failure:

```vyper
@external
def t():
    x: int256 = (-2) ** 2   # == 4, fits int256, yet: ValueError: math domain error
```

Command run (compiling the above contract):

```
$ vyper contract.vy
```

Output (relevant part):

```
ValueError: math domain error
```

Control: `(2) ** 2` (positive base) compiles cleanly, isolating the negative base as the trigger.

Root cause: in the constant-fold path for `**` (`vyper/ast/nodes.py`, class `Pow._op`), a negative *exponent* is sign-checked at `nodes.py:1121` (clean `InvalidOperation: Cannot calculate a negative power`), but the subsequent "prevent a compiler hang" overflow heuristic at `nodes.py:1130` evaluates `math.log(decimal.Decimal(left))` on the *base* without guarding its sign:

```python
# vyper/ast/nodes.py:1118-1133  (class Pow)
def _op(self, left, right):
    if isinstance(left, decimal.Decimal):
        raise TypeMismatch("Cannot perform exponentiation on decimal values.", self._parent)
    if right < 0:
        raise InvalidOperation("Cannot calculate a negative power", self._parent)
    # ... "prevent a compiler hang" ...
    # r > ln(2 ** 256) / ln(l)
    if right > math.log(decimal.Decimal(2**257)) / math.log(decimal.Decimal(left)):  # left<0 -> ValueError
        raise InvalidLiteral("Out of bounds", self)
    return int(left**right)
```

For any negative base, `math.log(decimal.Decimal(left))` raises the bare `ValueError: math domain error`. This is the negative-base sibling of the same broken line that yields `0 ** 0` → math domain error and `1 ** N` → ZeroDivisionError; unlike those degenerate inputs, this one denies a legitimate, in-range computation, so a correct constant expression silently becomes uncompilable.

### How can it be fixed?

Guard the base before the `math.log` overflow heuristic at `nodes.py:1130`. A single early check such as `if left <= 0 or left == 1:` that skips the heuristic and falls through to the exact fold (`int(left**right)`) would fix all three related cases (negative base, `0 ** 0`, `1 ** N`) at once, since for those bases the logarithmic overflow estimate is either undefined or unnecessary and the exact result can be computed directly.

SAY-5:
### What I did

Fix #4980: a negative base in a literal `**` expression aborts the compiler with `ValueError: math domain error` (now `expected a positive input`) because `Pow._op` calls `math.log(decimal.Decimal(left))` in the overflow heuristic without guarding the base sign. Skipping the heuristic for `left <= 1` also fixes the related `0 ** 0` and `1 ** N` crashes since the exact fold handles those exactly.

### How I did it

Guarded the log-based overflow estimate with `left > 1` in `vyper/ast/nodes.py` so non-positive and unit bases fall through to `int(left**right)`.

### How to verify it

`uv run pytest tests/unit/ast/nodes/test_fold_binop_int.py::test_binop_pow_degenerate_base` covers `(-2)**2`, `(-2)**3`, `(-1)**100`, `(-1)**101`, `0**0`, `1**99`.

### Commit message

```
The constant-fold path for `**` in `Pow._op` calls
`math.log(decimal.Decimal(left))` to estimate whether `l**r` would
overflow `2**256` and result in a compiler hang or crash. The estimate
is only defined for `left > 1`; for any negative base it raises
`ValueError`, denying valid in-range expressions such as
`(-2) ** 2 == 4` and `(-1) ** 100 == 1`. The same math.log call also
crashes on `0 ** 0` and `1 ** N`.

Guard the heuristic with `left > 1` so degenerate bases skip the
estimate and fall through to the exact `int(left**right)` fold, which
already handles them correctly.
```

Fix uncaught `ValueError` when constant-folding `**` with a non-positive base.

### DIFF
--- a/vyper/ast/nodes.py
+++ b/vyper/ast/nodes.py
@@ -1136,7 +1136,11 @@ class Pow(Operator):
         # l**r > 2**256
         # r * ln(l) > ln(2 ** 256)
         # r > ln(2 ** 256) / ln(l)
-        if right > math.log(decimal.Decimal(2**257)) / math.log(decimal.Decimal(left)):
+        # math.log is undefined for left <= 0 and zero for left == 1; use the
+        # base magnitude so |left| > 1 still gets the early bound check.
+        if abs(left) > 1 and right > math.log(decimal.Decimal(2**257)) / math.log(
+            decimal.Decimal(abs(left))
+        ):
             raise InvalidLiteral("Out of bounds", self)
 
         return int(left**right)
--- a/vyper/ast/nodes.py
+++ b/vyper/ast/nodes.py
@@ -1133,14 +1133,16 @@ class Pow(Operator):
         # stage since we are just trying to filter out inputs which can cause
         # the compiler to hang. the others will get caught during constant
         # folding or codegen.
+        # |left| <= 1 can never overflow (result magnitude stays <= 1), so
+        # fast-path it before the log-based heuristic. math.log is undefined
+        # for left <= 0 and zero for left == 1, so the log check below would
+        # also be ill-defined for those cases.
+        if abs(left) <= 1:
+            return int(left**right)
         # l**r > 2**256
         # r * ln(l) > ln(2 ** 256)
         # r > ln(2 ** 256) / ln(l)
-        # math.log is undefined for left <= 0 and zero for left == 1; use the
-        # base magnitude so |left| > 1 still gets the early bound check.
-        if abs(left) > 1 and right > math.log(decimal.Decimal(2**257)) / math.log(
-            decimal.Decimal(abs(left))
-        ):
+        if right > math.log(decimal.Decimal(2**257)) / math.log(decimal.Decimal(abs(left))):
             raise InvalidLiteral("Out of bounds", self)
 
         return int(left**right)
--- a/tests/unit/ast/nodes/test_fold_binop_int.py
+++ b/tests/unit/ast/nodes/test_fold_binop_int.py
@@ -131,3 +131,32 @@ def foo({input_value}) -> int128:
     else:
         with tx_failed():
             contract.foo(*values)
+
+
+@pytest.mark.parametrize(
+    "expr,expected",
+    [
+        ("(-2) ** 2", 4),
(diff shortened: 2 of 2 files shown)


vyper/ast/nodes.py
```diff
@@ -1136,7 +1136,11 @@ def _op(self, left, right):
         # l**r > 2**256
         # r * ln(l) > ln(2 ** 256)
         # r > ln(2 ** 256) / ln(l)
-        if right > math.log(decimal.Decimal(2**257)) / math.log(decimal.Decimal(left)):
+        # math.log is undefined for left <= 0 and zero for left == 1; use the
+        # base magnitude so |left| > 1 still gets the early bound check.
+        if abs(left) > 1 and right > math.log(decimal.Decimal(2**257)) / math.log(
```
charles-cooper: i had to reference the original issue before i understood that this relies on short circuiting to avoid evaluation of `math.log` later in the condition. i think it would be good to fastpath on the previous line (and i think the short circuiting is also what led to the issue that codex pointed out)

charles-cooper:
```suggestion
        if left <= 1:
            raise InvalidLiteral("Out of bounds", self.left)
        if right > math.log(decimal.Decimal(2**257)) / math.log(
```

charles-cooper (approved):

PR #5001, opened 2026-05-21
refactor[venom]: prepare builtin call arguments

banteg:
### What I did

Centralize Venom builtin argument preparation so runtime inputs are evaluated once, in source order, before handler dispatch. This fixes delayed storage loads, runtime keyword order, and external-code slice source order.

For example, when `self.start()` changes the address returned by `self.source()`, `slice(self.source().code, self.start(), 1)` must capture the code address first. The regression reverts on current master and returns the expected byte on this branch. The existing CREATE salt-before-value regression also now passes under Venom. If a bounds expression deploys code at an initially empty address, the slice must still observe the source-time empty code and revert for a nonempty slice.

### How I did it

`BuiltinCall` holds the analyzed AST, codegen context, prepared values and explicit-keyword presence. It performs one suffix scan with master's effect analysis, eagerly loads primitive values, and uses the existing snapshot helper for memory composites that later expressions can mutate. Storage/transient/immutable composites are copied to memory at their evaluation point. Handlers consume the prepared values; they no longer call `Expr` themselves or duplicate keyword defaults.

Defaults come from the semantic signature. `ContextDefault.GAS` and integer defaults are backend-neutral; the legacy adapter creates IR nodes where needed. An absent salt remains distinguishable from an explicit zero salt. Type arguments and literal-only keywords remain compile-time inputs.

Lengths, raw-log topic lists, raw calldata and slice code views have explicit preparation rules. External-code slices capture both the address and its code length before evaluating the bounds. Existing code cannot change during this execution, so the captured length preserves source-time bounds without copying the entire contract code. Current master's unbounded bytes/string/array support is retained. The old parallel preparation-policy framework and extra value wrapper are removed; no new snapshot machinery is added to the context.

Move the `create_forwarder_to` deprecation warning to semantic analysis so both pipelines report it, and reject duplicate entries when assembling the handler registry.

### How to verify it

- Venom builtin and unit suites: 4,564 passed, 1 skipped, 15 expected failures.
- Focused preparation regressions: 28 passed under Venom; 26 passed and 2 expected failures under legacy codegen.
- The deployment-in-bounds regression fails on the previous PR head (the call unexpectedly succeeds) and passes after capturing the source-time code length.
- `make lint` passes (mypy, Black, Flake8, isort).

The two focused legacy expected failures cover external-code source ordering and code creation during bounds evaluation; this PR changes Venom preparation.

### Commit message

```text
Builtin handlers currently decide when to load inputs and evaluate
keywords, so argument order varies between handlers. Delayed storage
loads and mutable memory aliases can observe later argument effects;
CREATE keyword order and external-code slice address order are affected.

Prepare runtime inputs once in source order before dispatch. Reuse the
existing effect analysis and snapshot helpers, eagerly load words, and
copy composite values only when their location or later effects require
it. Resolve keyword defaults from the semantic signature, preserving
whether an optional salt was explicitly supplied.

Keep special preparation for lengths, log topics and raw data views, and
retain unbounded sequence support. Use backend-neutral defaults and
adapt them at the legacy IR boundary.
```

Fix builtin argument evaluation order in experimental codegen, including runtime CREATE/raw-call keywords and external-code slice sources.


charles-cooper:
hmm, i think the point is actually that `_get_kwarg_value` is a code smell that args are getting evaluated out of order. what i would like to see ideally is some shared infrastructure which makes all of these conform to a uniform spec. like we should probably have:
- `get_kwarg_values` (compiles kwargs in source order and returns a dict from kwarg name to ir value)
- `get_kwarg_constants` or `get_kwarg_ast_constants` (fold foldable kwargs and return a dict from kwarg name to constant ast values)
- each builtin should also have a list of allowed kwarg names, which makes the validation uniform rather than needing adhoc implementation for every builtin

the keys of `get_kwarg_values()` should always be a superset of `get_kwarg_constants()`. the only difference is when they run (during validation vs during codegen), and what they return -- get_kwarg_constants should return ast nodes which are guaranteed to be compile-time constants (and maybe even have valid `get_folded()`, not sure without reading all the code), vs get_kwarg_values should return things which live in the runtime (VyperValue or IROperand or the like)

banteg:
Pushed a normal follow-up commit for source order, no force push.

Runtime kwargs are now lowered by walking source-order node.keywords, and raw_create/create_from_blueprint materialize positional constructor args before keyword expressions. I also converted the existing raw_create source-order xfails into Venom-passing tests and added raw_call gas/value kwarg order coverage.

tests/functional/builtins/codegen/test_create_functions.py
```diff
@@ -1145,10 +1145,13 @@ def deploy_from_calldata(s: Bytes[1024], arg: uint256, salt: bytes32) -> address
     assert env.get_code(res) == runtime
 
 
-# evaluation of the value kwarg changes the value of the salt kwarg
-# value kwarg comes after the salt kwarg in the source code
-@pytest.mark.xfail(raises=AssertionError, reason="salt kwarg is evaluated after value kwarg")
-def test_raw_create_order_of_eval_of_kwargs(get_contract, env, create2_address_of, keccak):
+# value is evaluated after salt because that is the source order.
```
charles-cooper: oh these are good, looks like we are making progress

vyper/codegen_venom/builtins/create.py
```diff
@@ -61,6 +60,37 @@ def _check_create_result(
     return addr
 
 
+def _copy_bytes_value(ctx: VenomCodegenContext, b, src: IROperand, typ) -> IRVariable:
```
charles-cooper: i'm pretty sure there is already a utility for this. or like if there isn't, there should be (in context.py or maybe we need a core.py for venom)

banteg:
Addressed in 9e9070fa7.

Changes:
- replaced the reduced-kwarg lookup with shared get_kwarg_ast_constants() and get_kwarg_values() helpers;
- added explicit allowed kwarg lists for the Venom builtins that lower kwargs here: abi_encode/abi_decode, raw/create variants, raw_call, send, and print;
- kept runtime kwargs lowered by walking node.keywords in source order; literal-only kwargs go through folded AST constants and are not runtime-lowered unnecessarily;
- added unit coverage for folded constants, unexpected kwargs, and presence checks.

I also checked the Venom builtin tree after the refactor; the only direct node.keywords iteration left is inside _kwargs.py. extract32(output_type=...) is type-only in semantic analysis and has no Venom runtime lowering.

Validation:
- uv run --python 3.12 pytest --experimental-codegen tests/unit/compiler/venom/test_builtin_kwargs.py tests/functional/builtins/codegen/test_abi_encode.py tests/functional/builtins/codegen/test_create_functions.py tests/functional/builtins/codegen/test_raw_call.py -q
- uv run --python 3.12 pytest tests/functional/builtins/codegen/test_create_functions.py::test_raw_create_order_of_eval_of_kwargs tests/functional/builtins/codegen/test_create_functions.py::test_raw_create_eval_order tests/functional/builtins/codegen/test_raw_call.py::test_raw_call_runtime_kwargs_source_order -q
- uv run --python 3.12 ruff check vyper/codegen_venom/builtins/_kwargs.py vyper/codegen_venom/builtins/abi.py vyper/codegen_venom/builtins/create.py vyper/codegen_venom/builtins/system.py vyper/codegen_venom/builtins/misc.py tests/unit/compiler/venom/test_builtin_kwargs.py
- git diff --check

vyper/codegen_venom/builtins/create.py
```diff
@@ -61,6 +60,37 @@ def _check_create_result(
...
+    dst = ctx.new_temporary_value(typ)
+    assert isinstance(dst.operand, IRVariable)
+    assert isinstance(src, IRVariable)
+    src_len = b.mload(src)
+    copy_size = b.add(src_len, IRLiteral(32))
+    ctx.copy_memory_dynamic(dst.operand, src, copy_size)
+    return dst.operand
+
+
+def _materialize_ctor_args(
```
charles-cooper: i'm also suspicious that this is duplicated code; or like what does legacy do here? we should have uniform serialization / deserialization / abi encode/decode machinery

vyper/builtins/functions.py
```diff
@@ -2242,10 +2242,14 @@ def infer_kwarg_types(self, node):
...
-        ensure_tuple = next(
-            (arg.value.value for arg in node.keywords if arg.arg == "ensure_tuple"), True
-        )
-        assert isinstance(ensure_tuple, bool)
+        ensure_tuple = True
+        for arg in node.keywords:
+            if arg.arg == "ensure_tuple":
+                ensure_tuple_node = arg.value.reduced()
+                if not isinstance(ensure_tuple_node, vy_ast.NameConstant):  # pragma: nocover
+                    raise CompilerPanic("unfoldable boolean kwarg: ensure_tuple", ensure_tuple_node)
```
charles-cooper: i'd rather just not touch the legacy pipeline (currently it panics rather than emitting garbage code, right?)

vyper/codegen_venom/builtins/_kwargs.py
```diff
@@ -0,0 +1,118 @@
...
+_UNSET = object()
+
+
+def _validate_allowed_kwargs(
+    node: vy_ast.Call, allowed_kwarg_names: Iterable[str] | None
+) -> None:
+    if allowed_kwarg_names is None:
+        return
+
+    allowed_kwarg_names = set(allowed_kwarg_names)
```
charles-cooper: might as well also include a sanity check that there are no duplicates

vyper/codegen_venom/builtins/_kwargs.py
```diff
@@ -0,0 +1,118 @@
...
+
+def _literal_value(node: vy_ast.VyperNode) -> Any:
+    if isinstance(node, vy_ast.Int):
+        return node.value
+    if isinstance(node, vy_ast.NameConstant):
+        return node.value
+    return _UNSET
+
+
+def get_bool_kwarg(
```
charles-cooper: this looks kind of overgrown.. like it replaces the previous `get_bool_kwarg` right? but it doesn't actually improve the cleanliness of the code, just moves the messiness somewhere else

vyper/codegen_venom/builtins/_kwargs.py
```diff
@@ -0,0 +1,118 @@
...
+
+def get_literal_kwarg(
+    node: vy_ast.Call,
+    kwarg_name: str,
+    default,
+    allowed_kwarg_names: Iterable[str] | None = None,
+):
+    kw_node = get_kwarg_ast_constants(
+        node,
+        (kwarg_name,),
```
charles-cooper: oh honestly, like, there should be another rule which is that `get_kwarg_ast_constants` only gets called once per builtin. so like this should actually take the kwarg constants dict as input.

but also the kwargs dicts should just know what the defaults are?

vyper/builtins/functions.py
banteg: Removed in 29d696e6b. The legacy `ABIEncode.fetch_call_return()` change and its folded-kwarg functional test are gone, so this PR is back to Venom builtin lowering only.

vyper/codegen_venom/builtins/create.py
banteg: Agreed. In 29d696e6b I removed the local `_copy_bytes_value()` helper and added `ctx.materialize_value()`, which copies through the existing `store_vyper_value()` / `store_memory()` path instead of duplicating bytestring copy logic in `create.py`.

vyper/codegen_venom/builtins/create.py
banteg: I kept the actual constructor-argument serialization on the existing `abi_encode_to_buf()` path. The follow-up only snapshots values before later source-order kwargs run, and that snapshot now goes through `ctx.materialize_value()` rather than a create-specific bytes copy. A broader unification of ABI encode/decode machinery seems separate from this source-order PR.

vyper/codegen_venom/builtins/abi.py
```diff
@@ -70,7 +51,10 @@ def _parse_method_id(method_id_node: vy_ast.VyperNode) -> Optional[int]:
         return method_id_node.value
 
     # If it has a folded value (constant expression)
-    if hasattr(method_id_node, "_metadata") and "folded_value" in method_id_node._metadata:
+    if (
```
charles-cooper: this looks overly defensive; there are methods for getting reduced/folded values safetly out of an ast node

vyper/codegen_venom/builtins/_kwargs.py
banteg: Added in e2c3e4543. `_validate_allowed_kwargs()` now also checks for duplicate kwarg names, and the unit tests cover that sanity check.

vyper/codegen_venom/builtins/_kwargs.py
banteg: Simplified in e2c3e4543. `get_bool_kwarg()` and `get_literal_kwarg()` now just read from an already-collected constants dict, so the AST scanning and validation stays in `get_kwarg_ast_constants()` instead of being repeated inside each accessor.

vyper/codegen_venom/builtins/_kwargs.py
banteg: Done in e2c3e4543. Each builtin lowering now calls `get_kwarg_ast_constants()` once and passes the resulting constants dict into the bool/literal accessors. I left default ownership as explicit call-site data for now, since moving defaults into kwarg specs would be a larger API change than this PR needs.

vyper/codegen_venom/builtins/abi.py
```diff
@@ -201,7 +189,9 @@ def lower_abi_decode(node: vy_ast.Call, ctx: VenomCodegenContext) -> VyperValue:
     # Parse args
     data_node = node.args[0]
```
charles-cooper, line 179: like we should also have uniform source-order lowering for positional args! like i would expect to see either a more-or-less uniform set of calls at entry to each builtin, like 
```
constant_kwargs = get_kwarg_ast_constants(self.signature, ...)
arg_values = lower_pos_args(...)
kwarg_values = lower_kwargs(...)
```

OR just redesign the signature of lower_builtin entry point; my only concern here is that we will have a bit of a zoo (as in the legacy pipeline) to handle the various adhoc signatures that builtins have

vyper/codegen_venom/builtins/abi.py
banteg: Cleaned up in 85cf22443. `_parse_method_id()` now just calls `reduced()` on the node and no longer reaches into `_metadata["folded_value"]` directly.

vyper/codegen_venom/builtins/system.py
```diff
@@ -82,10 +73,15 @@ def lower_raw_call(node: vy_ast.Call, ctx: VenomCodegenContext) -> Union[IROpera
     to = Expr(node.args[0], ctx).lower_value()
 
     # Parse kwargs (need to know is_static before constancy check)
-    max_outsize = _get_literal_kwarg(node, "max_outsize", 0)
-    is_delegate = _get_literal_kwarg(node, "is_delegate_call", False)
-    is_static = _get_literal_kwarg(node, "is_static_call", False)
-    revert_on_failure = _get_literal_kwarg(node, "revert_on_failure", True)
+    kwarg_constants = get_kwarg_ast_constants(
+        node,
+        ("max_outsize", "is_delegate_call", "is_static_call", "revert_on_failure"),
```
charles-cooper: why does this not match RAW_CALL_KWARGS?

vyper/codegen_venom/context.py
```diff
@@ -155,6 +155,18 @@ def store_vyper_value(self, vv: VyperValue, ptr: IRVariable, typ: VyperType) ->
         """Store a VyperValue into memory, preserving its source layout."""
         self.store_memory(self.unwrap(vv), ptr, typ, src_typ=vv.typ)
 
+    def materialize_value(
```
charles-cooper, line 640: much better!

vyper/codegen_venom/builtins/_kwargs.py
```diff
@@ -0,0 +1,112 @@
...
+def _validate_allowed_kwargs(
+    node: vy_ast.Call, allowed_kwarg_names: Iterable[str] | None
+) -> None:
+    seen = set()
+    for kw in node.keywords:
+        if kw.arg in seen:  # pragma: nocover
+            raise CompilerPanic(f"duplicate kwarg: {kw.arg}", kw)
+        seen.add(kw.arg)
+
+    if allowed_kwarg_names is None:
```
charles-cooper: why would this happen

vyper/codegen_venom/builtins/_kwargs.py
```diff
@@ -0,0 +1,112 @@
...
+def kwarg_is_provided(
+    node: vy_ast.Call, kwarg_name: str, allowed_kwarg_names: Iterable[str] | None = None
+) -> bool:
+    _validate_allowed_kwargs(node, allowed_kwarg_names)
+    return any(kw.arg == kwarg_name for kw in node.keywords)
+
+
+def get_kwarg_ast_constants(
+    node: vy_ast.Call,
+    kwarg_names: Iterable[str],
```
charles-cooper: what is the point of having both kwarg_names and allowed_kwarg_names..?

vyper/codegen_venom/builtins/system.py
banteg: Adjusted in 51da445c3. `lower_raw_call()` now validates once against the full `_RAW_CALL_KWARGS` set, then separately asks for only the literal constants and runtime values it needs.

vyper/codegen_venom/builtins/_kwargs.py
banteg: This is now isolated to `validate_kwargs()` in 51da445c3. The duplicate case should already be rejected earlier, so this is just a centralized internal sanity check for malformed call nodes rather than something expected in normal user input.

vyper/codegen_venom/builtins/_kwargs.py
banteg: Simplified in 51da445c3. There is now a separate validate_kwargs(node, FULL_ALLOWED_KWARGS) step for the full builtin signature, while get_kwarg_ast_constants() and get_kwarg_values() only take the subset they actually return. That removes the dual kwarg_names/allowed_kwarg_names API.

vyper/codegen_venom/builtins/abi.py
banteg: Agreed on the direction. This PR now keeps the shared kwarg path uniform and source-order based, but I think source-order positional lowering / lower_builtin entrypoint redesign is a larger follow-up because it changes every builtin signature shape rather than just removing the reduced-kwarg lookup. I would keep that out of this PR unless you want this PR to absorb the full entrypoint redesign too.

charles-cooper: > unless you want this PR to absorb the full entrypoint redesign too.

yes, i do -- it will be easier to review the fully refactored version than go through and have to piecemeal keep track of all the outstanding things to fix

vyper/codegen_venom/builtins/create.py
```diff
@@ -195,46 +195,30 @@ def lower_raw_create(node: vy_ast.Call, ctx: VenomCodegenContext) -> IROperand:
     bytecode_node = node.args[0]
     ctor_arg_nodes = node.args[1:]
 
-    # Get bytecode - may be in memory, storage, or transient
+    # Snapshot bytecode before constructor args or kwargs can mutate its source
+    # or overlap its memory.
     bytecode_vv = Expr(bytecode_node, ctx).lower()
     bytecode_typ = bytecode_node._metadata["type"]
+    bytecode = ctx.materialize_value(bytecode_vv, bytecode_typ, "raw_create bytecode").operand
```
charles-cooper: whoa -- .operand loses location provenance, is that on purpose?

vyper/codegen_venom/builtins/create.py
```diff
@@ -309,15 +278,13 @@ def lower_create_minimal_proxy_to(node: vy_ast.Call, ctx: VenomCodegenContext) -
     # Parse args
     target = Expr(node.args[0], ctx).lower_value()
 
-    # Parse kwargs
-    value_node = _get_kwarg_value(node, "value")
-    salt_node = _get_kwarg_value(node, "salt")
-    revert_on_failure, _ = _get_literal_kwarg(node, "revert_on_failure", True)
+    validate_kwargs(node, _CREATE_KWARGS)
+    kwarg_constants = get_kwarg_ast_constants(node, ("revert_on_failure",))
```
charles-cooper: kwarg_constants should just have defaults filled in tbh

vyper/codegen_venom/builtins/create.py
```diff
@@ -309,15 +278,13 @@ def lower_create_minimal_proxy_to(node: vy_ast.Call, ctx: VenomCodegenContext) -
...
+    revert_on_failure = get_literal_kwarg(kwarg_constants, "revert_on_failure", True)
 
-    if value_node is not None:
-        value = Expr(value_node, ctx).lower_value()
-    else:
-        value = IRLiteral(0)
+    runtime_kwargs = get_kwarg_values(node, ctx, ("value", "salt"))
+    value = runtime_kwargs["value"] if "value" in runtime_kwargs else IRLiteral(0)
```
charles-cooper: ditto for runtime values altho it's harder since there might be (compile-time) side effects, like `b.gas()`. i recommend that get_kwarg_values takes a dict from names to functions which describe what to do on-missing. for example in this case, `get_kwarg_values(node, ctx, {"value": (lambda: IRLiteral(0)), "gas":  ctx.builder.gas})`. not sure if this is too magical, but it would be a uniform interface for builtins to inspect the kwargs dict

vyper/codegen_venom/builtins/create.py
```diff
@@ -441,7 +405,9 @@ def lower_create_copy_of(node: vy_ast.Call, ctx: VenomCodegenContext) -> IROpera
     return _check_create_result(ctx, b, addr, revert_on_failure)
 
 
-def lower_create_from_blueprint(node: vy_ast.Call, ctx: VenomCodegenContext) -> IROperand:
+def lower_create_from_blueprint(
+    node: vy_ast.Call, ctx: VenomCodegenContext
```
charles-cooper: this is going to fail lint; just use the same exact same lint cmd as `make lint`

vyper/codegen_venom/builtins/create.py
```diff
@@ -463,80 +429,65 @@ def lower_create_from_blueprint(node: vy_ast.Call, ctx: VenomCodegenContext) ->
...
-    has_code = b.sgt(codesize, IRLiteral(0))
-    b.assert_(has_code)
-
-    # Handle constructor arguments
-    args_len: IROperand
-    args_ptr: IROperand
-
+    validate_kwargs(node, _CREATE_FROM_BLUEPRINT_KWARGS)
+    kwarg_constants = get_kwarg_ast_constants(node, ("raw_args", "revert_on_failure"))
+    raw_args = get_literal_kwarg(kwarg_constants, "raw_args", False)
```
charles-cooper: why not just `bool(kwarg_constants.get("raw_args", False))`?

charles-cooper: or if this happens a lot, we should just have an adapter that is passed to get_kwarg_ast_constants so it knows what the output type is supposed to be, not sure if this can be designed so it's uniform though

vyper/codegen_venom/builtins/abi.py
banteg: Done in 2a92b0d6b. I added a shared BuiltinCall entry helper and moved the kwarg-aware builtins onto the same pattern: validate full kwarg set, collect default-filled constant kwargs, lower positional args through the call helper in source order, then collect default-filled runtime kwargs in source order. I kept the outer HANDLERS registry shape stable, but the actual abi/create/raw_call/send/print entry paths now use the same call object instead of each spelling the plumbing independently.

vyper/codegen_venom/builtins/create.py
banteg: Not on purpose; adjusted in 2a92b0d6b. The materialized bytecode/raw-args values now use `.ptr().operand`, so the code is explicit that these are memory-backed pointers after materialization rather than generic operands.

vyper/codegen_venom/builtins/create.py
banteg: Done in 2a92b0d6b. The kwarg helpers now accept default maps and return default-filled constant/runtime kwarg dicts, so create/raw_call/abi/print no longer need separate `if missing then default` handling at each use site.

vyper/codegen_venom/builtins/create.py
banteg: Implemented in 2a92b0d6b. `get_kwarg_values()` now takes either an iterable of names or a defaults mapping; defaults are filled after walking supplied kwargs, and callable defaults are invoked late. The raw_call path uses `{"gas": b.gas, "value": IRLiteral(0)}`, while create paths use default-filled value/salt/code_offset maps.

vyper/codegen_venom/builtins/create.py
banteg: Confirmed with the project lint command. `uv run --python 3.12 make lint` now passes after f61bc2ebf fixed the mypy issue and applied black/isort formatting; 2f6f1e4c0 also went through the same lint command.

vyper/codegen_venom/builtins/create.py
banteg: Adjusted in 2f6f1e4c0. Boolean constant kwargs now go through `get_bool_kwarg()` with defaults already filled by `get_kwarg_ast_constants()`, including `raw_args`, `revert_on_failure`, and the raw_call boolean switches. That avoids direct `bool(ast_node)` while still keeping the call sites uniform.

vyper/codegen_venom/builtins/abi.py
banteg: Done in 5075470b8.

`lower_builtin()` now constructs `BuiltinCall(node, ctx)` and every `HANDLERS` target takes that single entry object, so the dispatcher and per-builtin entry signatures match the full refactor. The kwarg-aware builtins still use the shared default-filled literal/runtime helpers, and non-kwarg builtins pull `node`/`ctx` from the same call object.

Validation:
- `uv run --python 3.12 pytest --experimental-codegen tests/unit/compiler/venom tests/functional/builtins/codegen tests/functional/syntax/test_print.py -q` (`4178 passed, 1 skipped, 15 xfailed`)

vyper/codegen_venom/builtins/_kwargs.py
```diff
@@ -0,0 +1,156 @@
...
+
+def validate_kwargs(node: vy_ast.Call, allowed_kwarg_names: Iterable[str]) -> None:
+    seen = set()
+    for kw in node.keywords:
+        if kw.arg in seen:  # pragma: nocover
+            raise CompilerPanic(f"duplicate kwarg: {kw.arg}", kw)
+        seen.add(kw.arg)
+
+    allowed_kwarg_names = set(allowed_kwarg_names)
+    for kw in node.keywords:
```
charles-cooper: double walk of `node.keywords`?

vyper/codegen_venom/builtins/_kwargs.py
charles-cooper: sorry, i wasn't clear -- sanity check there are not duplicates in `allowed_kwarg_names`

vyper/codegen_venom/builtins/_kwargs.py
```diff
@@ -0,0 +1,156 @@
...
+
+def get_kwarg_ast_constants(
+    node: vy_ast.Call,
+    kwarg_defaults: Mapping[str, Any] | Iterable[str],
+    error_prefix: str = "unfoldable constant kwarg",
+) -> dict[str, Any]:
+    kwarg_names, defaults = _kwarg_names_and_defaults(kwarg_defaults)
+    ret = {}
+    for kw in node.keywords:
+        if kw.arg not in kwarg_names:
```
charles-cooper: this seems like it should be a validation error

charles-cooper: oh, i see -- it's only the kwargs that are expected to be constant

vyper/codegen_venom/builtins/_kwargs.py
```diff
@@ -0,0 +1,156 @@
...
+            raise CompilerPanic(f"duplicate kwarg: {kw.arg}", kw)
+        seen.add(kw.arg)
+
+    allowed_kwarg_names = set(allowed_kwarg_names)
+    for kw in node.keywords:
+        if kw.arg not in allowed_kwarg_names:  # pragma: nocover
+            raise CompilerPanic(f"unexpected kwarg: {kw.arg}", kw)
+
+
+def kwarg_is_provided(node: vy_ast.Call, kwarg_name: str) -> bool:
```
charles-cooper: why doesn't this operate on a dict? like i think we should have an informal, hygiene rule that we only walk `node.keywords` once in builtins validation and once in codegen

vyper/codegen_venom/builtins/_kwargs.py
```diff
@@ -0,0 +1,156 @@
...
+    return any(kw.arg == kwarg_name for kw in node.keywords)
+
+
+def get_kwarg_ast_constants(
+    node: vy_ast.Call,
+    kwarg_defaults: Mapping[str, Any] | Iterable[str],
```
charles-cooper: what kind of type signature is this, just be `dict[str, VyperNode]` ..?

vyper/codegen_venom/builtins/_kwargs.py
```diff
@@ -0,0 +1,156 @@
...
+    kwarg_names, defaults = _kwarg_names_and_defaults(kwarg_defaults)
+    ret = {}
+    for kw in node.keywords:
+        if kw.arg not in kwarg_names:
+            continue
+
+        kw_node = kw.value.reduced()
+        if not isinstance(kw_node, vy_ast.Constant):  # pragma: nocover
+            raise CompilerPanic(f"{error_prefix}: {kw.arg}", kw_node)
+        ret[kw.arg] = kw_node
```
charles-cooper: since we already know it's a `Constant`, i'm pretty sure at this point we want to extract `kw_node.value`

vyper/codegen_venom/builtins/_kwargs.py
charles-cooper: or yea, if we are extracting vyper values, can be any of the extractable types...

vyper/codegen_venom/builtins/_kwargs.py
banteg: Done in 053367bb7. `validate_kwargs()` now returns the validated `dict[str, VyperNode]`; `BuiltinCall` caches it, and the constant/runtime helpers operate on that dict instead of walking `node.keywords` again.

vyper/codegen_venom/builtins/_kwargs.py
banteg: Done in 053367bb7. `validate_kwargs()` now also sanity-checks `allowed_kwarg_names` itself and raises on duplicate allowed names; the unit tests cover that case.

vyper/codegen_venom/builtins/_kwargs.py
banteg: Addressed in 053367bb7. `validate_kwargs()` now builds the validated kwarg dict in one pass over `node.keywords` instead of doing a duplicate pass and an allowed-name pass.

vyper/codegen_venom/builtins/_kwargs.py
banteg: Adjusted in 053367bb7. The helper boundary now uses a validated `dict[str, vy_ast.VyperNode]`; the broader defaults mapping type only remains where defaults can be IR literals, callables such as `b.gas`, or Python literal defaults.

vyper/codegen_venom/builtins/_kwargs.py
banteg: Done in 053367bb7. `get_literal_kwarg()` now returns `kw_node.value` directly for reduced `vy_ast.Constant` nodes instead of branching over individual constant node subclasses.

charles-cooper:
@banteg please fix merge conflicts

vyper/codegen_venom/builtins/_kwargs.py
```diff
@@ -0,0 +1,182 @@
...
+from typing import Any
+
+from vyper import ast as vy_ast
+from vyper.exceptions import CompilerPanic
+
+_UNSET = object()
+
+
+@dataclass(frozen=True)
+class BuiltinCall:
```
charles-cooper: represents a builtin callsite with additional metadata

vyper/codegen_venom/builtins/_kwargs.py
```diff
@@ -0,0 +1,182 @@
...
+        return get_kwarg_ast_constants(self._kwargs_dict(), kwarg_defaults, error_prefix)
+
+    def get_kwarg_values(self, kwarg_defaults: Mapping[str, Any] | Iterable[str]):
+        return get_kwarg_values(self._kwargs_dict(), self.ctx, kwarg_defaults)
+
+    def lower_pos_args(self, arg_nodes: Iterable[vy_ast.VyperNode] | None = None) -> list[Any]:
+        from vyper.codegen_venom.expr import Expr
+
+        arg_nodes = self.node.args if arg_nodes is None else arg_nodes
+        return [Expr(arg, self.ctx).lower() for arg in arg_nodes]
```
charles-cooper: lower() or lower_value() should be called exactly once for every single arg/kwarg, in order, and it's not clear by inspection that (or why) this invariant holds

vyper/codegen_venom/context.py
charles-cooper, line 640: i think this is merge resolution leftover, 620579717e2e2f4e2a4ad6806591258495850934 also adds `materialize_value`

banteg:
Resolved the merge conflicts by merging current `origin/master` into `fix/venom-reduced-builtin-kwargs` and pushing without force-pushing.

I kept the refactored `BuiltinCall.get_kwarg_values()` path for `raw_call` and added a follow-up docs commit (`3ea8f270b`) to make the builtin callsite metadata and source-order/one-lowering-helper invariant explicit in `_kwargs.py`.

Validation:
- `.venv/bin/black --check -C --force-exclude=vyper/version.py vyper/codegen_venom/builtins/_kwargs.py vyper/codegen_venom/builtins/system.py`
- `.venv/bin/pytest tests/unit/compiler/venom/test_builtin_kwargs.py tests/functional/builtins/codegen/test_raw_call.py -q` (41 passed, 1 xfailed)
- `.venv/bin/pytest tests/functional/builtins/codegen/test_create_functions.py tests/functional/builtins/codegen/test_extract32.py tests/functional/builtins/codegen/test_as_wei_value.py tests/functional/builtins/codegen/test_convert.py -q` (1583 passed, 2 xfailed)

vyper/codegen_venom/builtins/_kwargs.py
banteg: Added a `BuiltinCall` docstring describing it as the builtin callsite plus the codegen context used to lower it.

vyper/codegen_venom/builtins/_kwargs.py
banteg: Clarified the invariant at the helper boundary: positional args are lowered in AST/source order, and explicit runtime kwargs are lowered once by iterating the collected keyword nodes in the user's keyword order rather than by allowed-name order.

vyper/codegen_venom/context.py
banteg: Yes, this was a stale merge-resolution leftover. After merging current master, context.py is no longer in this PR diff; materialize_value now comes from 620579717e2e2f4e2a4ad6806591258495850934 on master.

charles-cooper:
> Resolved the merge conflicts by merging current `origin/master` into `fix/venom-reduced-builtin-kwargs` and pushing without force-pushing.
> 
> I kept the refactored `BuiltinCall.get_kwarg_values()` path for `raw_call` and added a follow-up docs commit (`3ea8f270b`) to make the builtin callsite metadata and source-order/one-lowering-helper invariant explicit in `_kwargs.py`.
> 
> Validation:
> 
> * `.venv/bin/black --check -C --force-exclude=vyper/version.py vyper/codegen_venom/builtins/_kwargs.py vyper/codegen_venom/builtins/system.py`
> * `.venv/bin/pytest tests/unit/compiler/venom/test_builtin_kwargs.py tests/functional/builtins/codegen/test_raw_call.py -q` (41 passed, 1 xfailed)
> * `.venv/bin/pytest tests/functional/builtins/codegen/test_create_functions.py tests/functional/builtins/codegen/test_extract32.py tests/functional/builtins/codegen/test_as_wei_value.py tests/functional/builtins/codegen/test_convert.py -q` (1583 passed, 2 xfailed)

re. the lowering invariant, i didn't really mean a docs clarification. i meant a code change so that the invariant is clear by inspection. like i think better would be if BuiltinCall calls `lower()/lower_value()` on all arguments as a pre-processing step before `lower_<builtin>` even gets called...

banteg:
Pushed a normal follow-up commit, `3e8fa3acf`, for the code-side invariant.

`BuiltinCall` now pre-lowers runtime positional args and runtime kwargs before dispatching to `lower_<builtin>`, then handlers consume the cached `VyperValue`s / unwrapped operands instead of calling `Expr(...)` again. I also wired create aliases into the runtime-kwarg map and updated the touched handlers so the source-order lowering path is visible at the call wrapper.

I left the explicit opt-outs (`len`, `slice`, `raw_log`) in the runtime-arg map because they have ad hoc lowering for `msg.data`/code or nested log topics; the opt-out is now centralized instead of implicit in each handler.

Validation:
- `.venv/bin/black --check -C --force-exclude=vyper/version.py vyper/codegen_venom/builtins`
- `.venv/bin/ruff check vyper/codegen_venom/builtins`
- `.venv/bin/pytest tests/unit/compiler/venom/test_builtin_kwargs.py tests/functional/builtins/codegen/test_raw_call.py -q` (41 passed, 1 xfailed)
- `.venv/bin/pytest tests/functional/builtins/codegen/test_create_functions.py tests/functional/builtins/codegen/test_extract32.py tests/functional/builtins/codegen/test_as_wei_value.py tests/functional/builtins/codegen/test_convert.py -q` (1583 passed, 2 xfailed)
- `.venv/bin/pytest tests/functional/builtins/codegen/test_slice.py tests/functional/builtins/codegen/test_extract32.py -q` (81 passed)
- `.venv/bin/pytest tests/functional/builtins/codegen/test_concat.py tests/functional/builtins/codegen/test_keccak256.py tests/functional/builtins/codegen/test_sha256.py tests/functional/builtins/codegen/test_unsafe_math.py tests/functional/builtins/codegen/test_floor.py tests/functional/builtins/codegen/test_ceil.py tests/functional/builtins/codegen/test_uint2str.py tests/functional/builtins/codegen/test_length.py tests/functional/builtins/codegen/test_blobhash.py tests/functional/builtins/codegen/test_ec.py tests/functional/builtins/codegen/test_ecrecover.py -q` (121 passed)
- `git diff --check`

banteg:
Pushed `30ddf6d22`, a consolidation pass that makes the invariant hold by construction rather than by convention.

`BuiltinCall.__init__` now does all the lowering before `lower_<builtin>` is dispatched: it validates kwargs in a single walk of `node.keywords`, folds constant kwargs to plain python values (defaults filled), then lowers every runtime expression exactly once in source order — positional args first, then keyword args. Handlers never call `Expr(...)` anymore; they read `call.arg_operand(i)` / `call.kwarg_constants` / `call.kwarg_values`.

Structural changes:

- The `RUNTIME_ARG_INDICES` / `RUNTIME_KWARGS` / `MATERIALIZE_COMPLEX_ARGS` string-keyed maps in `__init__.py` are gone. Each builtin declares its interface with a `@callsite(...)` decorator next to its handler: constant kwargs, runtime kwargs (with defaults — ints become `IRLiteral`s, callables like `gas` are invoked with the ctx when missing, `None` means "no default", which is how `salt` selects CREATE vs CREATE2), and type kwargs (`extract32(output_type=...)`).
- Positional args with no runtime value are skipped by two generic rules instead of index maps: type expressions (`empty(T)`, `convert(x, T)`, `abi_decode(..., T)`), and data views (`msg.data`/`self.code`/`<addr>.code` — for `<addr>.code` the address subexpression is lowered in the view's place, preserving source order). Only two builtins keep explicit handler-managed args, declared in their spec with the reason: `raw_log` (topics list literal, where data must follow topics) and `as_wei_value` (denomination literal).
- The per-builtin materialize opt-in is replaced by a generic freeze rule: after lowering, an arg is snapshotted (prim word → stack, composite → fresh temporary) iff a later arg/kwarg expression can have side effects, so the common case emits no extra copies.
- `get_bool_kwarg` / `get_literal_kwarg` / `_UNSET` are deleted; `kwarg_constants` holds extracted python values directly. `_parse_method_id` now operates on the folded constant and can no longer silently fall through to "no selector". Module renamed `_kwargs.py` → `_call.py` to match its scope.

Net: −197 lines vs the previous branch state.

Validation:
- `make lint` (mypy/black/flake8/isort clean)
- `pytest --experimental-codegen tests/functional/builtins/codegen tests/unit/compiler/venom tests/functional/syntax/test_print.py` (4184 passed)
- `pytest tests/functional/builtins/codegen` (legacy pipeline, 2526 passed)
- `pytest --experimental-codegen tests/functional/codegen/` (full sweep, 4176 passed)

banteg:
Pushed `05c17e04a` fixing a regression in `30ddf6d22`: `raw_call`'s payload check had switched from the msg.data-only predicate to the broader `is_data_view()`, so `self.code` / `<address>.code` payloads silently lowered to the calldata copy path.

There is no lowering for `.code` payloads in either pipeline (legacy panics with `invalid IRnode: ~selfcode`; semantic analysis lets them through), so the venom path now raises `CompilerPanic` for them instead of miscompiling, and `len()` is tightened to the msg.data predicate as well. Added a parametrized regression test (`test_raw_call_code_payload_panics`) that passes under both pipelines.

Validation: `pytest --experimental-codegen tests/functional/builtins/codegen tests/unit/compiler/venom` (4174 passed), legacy builtins suite (2528 passed), lint clean.

banteg:
Refreshed this branch in 65f277458 with a normal merge of current master (no force-push). The only content conflict was in codegen_venom/builtins/abi.py. The resolution keeps the PR’s BuiltinCall path and also preserves master’s zero-argument abi_encode validation via StructureException; without that guard, current master’s test_abi_encode_no_args regresses and abi_encode() compiles unexpectedly.

Validation on the merged tree:
- both default and experimental test_abi_encode_no_args cases pass
- uv run --python 3.12 -m pytest tests/functional/builtins/codegen --experimental-codegen -m "not fuzzing" -qq (2541 passed, 4 xfailed)
- focused Ruff and Black checks for the builtins and changed functional tests pass

banteg:
Architectural follow-up pushed as `6f6b50943` with a normal push (no history rewrite).

The main change is that builtin lowering now has one typed boundary derived from the analyzed `BuiltinFunctionT`: a linear evaluation plan owns all runtime `Expr` lowering, prepares primitive words eagerly and composites as stable memory values, then hands handlers emission-free typed accessors. Type arguments, folded literals, topic lists, lengths, and authorized calldata/code views are explicit policies rather than handler escape hatches.

This also closes the concrete correctness gaps found in the previous shape:

- storage/transient positional values are materialized before gas-sensitive kwargs; in the `raw_call(to, self.payload)` repro, `gas` moved from the entry block to the storage-copy exit after the `sload` loop;
- undeclared `self.code` / `<address>.code` consumers are rejected instead of treating an address as a bytes pointer;
- located handler results are checked against the semantic return type with `compare_type`, which caught and fixed `raw_call` returning `(uint256, Bytes[N])` instead of `(bool, Bytes[N])`;
- backend defaults are IR-neutral, ABI-layout calldata/code cannot be flat-copied accidentally, and `create_forwarder_to` now warns consistently in both codegen pipelines.

Validation on the final commit:

- `uv run --python 3.12 ./quicktest.sh -m 'not fuzzing'` — 12,296 passed, 24 skipped, 51 xfailed
- post-push Socket and Read the Docs checks pass

The PR title and description now document the final architecture and emitted before/after proof. It is mergeable and ready for another review.

banteg:
Reworked against current master in 4e2f06282. The preparation boundary is now a single `BuiltinCall`, reusing existing effect analysis and snapshot helpers rather than maintaining a second policy/value framework. All runtime expressions are lowered there in source order; handlers only read prepared inputs. Words and storage/transient composites are loaded before later gas-sensitive arguments, defaults come from semantic signatures, and explicit zero salt stays distinct from absence.

Current master’s unbounded data/code views are preserved. A new runtime regression also exposed a remaining master bug: `slice(self.source().code, self.start(), 1)` evaluated the bounds before the address. The refactor fixes that and the existing CREATE salt-before-value regression.

Validation: 2,611 Venom builtin tests, 4,450 Venom functional codegen tests, 1,995 Venom unit/warning tests, and 2,618 legacy builtin/warning tests pass, alongside expected skips/xfails. The 27 focused preparation regressions pass and `make lint` passes. Updated the description to match the smaller implementation and current master.

vyper/codegen_venom/builtins/_call.py
```diff
@@ -0,0 +1,167 @@
...
+                    value = Expr(arg, ctx).lower()
+                    assert value.location is not None
+                    self.length = ctx.load_word(value.operand, value.location)
+                continue
+            if self.func_t._id == "raw_log" and i == 0:
+                topics = arg.reduced()
+                assert isinstance(topics, vy_ast.List)
+                nodes.extend(topics.elements)
+                continue
+            if self.func_t._id == "slice" and i == 0 and is_slice_view(arg):
```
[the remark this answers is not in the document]
banteg: Confirmed with a reproducer: a bounds expression can deploy code at an initially empty address, and the previous head incorrectly lets the slice succeed. Fixed in 1ac93e1d6 by capturing EXTCODESIZE alongside the source address, before evaluating bounds. The handler checks against that captured length. Existing code cannot change during this execution, so this preserves the source-time view without copying the full code. The new regression passes; the Venom builtin and unit suites report 4,564 passed, and make lint passes.

PR #5003, opened 2026-05-21
fix[ci]: harden workflow trust boundaries

banteg:
### What I did

- Split bytecode-size and gas benchmark measurement into read-only PR jobs, and moved PR comment writes into separate `workflow_run` follow-up jobs.
- Restricted publish, coverage, and other token-bearing jobs to trusted events and job-scoped permissions.
- Pinned remaining workflow dependencies, updated stale GitHub Action pins to their latest released tag SHAs, and added Dependabot coverage for GitHub Actions.

### How I did it

- Measurement jobs upload Markdown reports as artifacts. Comment workflows download the triggering run's artifact and write only the PR comment body, without checkout or shell execution.
- Release, PyPI, GHCR, CodeQL, and coverage jobs keep write scopes on the narrow jobs that need them, with checkout credentials disabled.
- All `uses:` references remain full-SHA pinned, with comments identifying the resolved released tag.

### How to verify it

- `ruby -e 'require "yaml"; ARGV.each { |f| YAML.load_file(f) }' .github/workflows/*.yml .github/workflows/*.yaml .github/dependabot.yml`
- `zizmor --no-progress --persona regular .github/workflows .github/dependabot.yml`
- `zizmor --no-progress --persona auditor --min-severity high .github/workflows .github/dependabot.yml`
- `git diff --check origin/master..HEAD`
- A second CI security pass found no reportable findings after validating the `workflow_run` comment jobs and the `era-tester` cache path.

### Commit message

```text
Harden GitHub Actions workflows so untrusted pull request code does not
share execution paths with jobs that can write comments, publish
artifacts, or access repository secrets.

Benchmark and bytecode-size runs now produce artifacts from read-only
pull request jobs. Separate `workflow_run` follow-up jobs download those
reports and only write the comment body, without checking out or
executing pull request code. Release, PyPI, GHCR, coverage, and CodeQL
jobs now keep write permissions scoped to the jobs that need them and
avoid attacker-influenced cache state in publish paths.

All external actions remain pinned to full commit SHAs. The remaining
workflow dependencies are pinned, stale action pins are moved to their
latest released tag SHAs, and Dependabot is configured to keep GitHub
Actions pins visible for future updates.
```

No changelog entry required; CI-only hardening.

### Cute Animal Picture

sleepy kitten

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,8 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    cooldown:
+      default-days: 7
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -6,3 +6,7 @@ updates:
       interval: "weekly"
     cooldown:
       default-days: 7
+    groups:
+      github-actions:
+        patterns:
+          - "*"
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -5,6 +5,7 @@ on:
     inputs:
       tag:
         default: ''
+        type: string
   push:
     branches:
       - master
@@ -22,6 +23,7 @@ defaults:
 jobs:
   unix-build:
     runs-on: ${{ matrix.os }}
+    timeout-minutes: 45
     strategy:
       matrix:
         os:
@@ -68,6 +70,7 @@ jobs:
 
   windows-build:
     runs-on: windows-latest
+    timeout-minutes: 45
 
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
@@ -103,6 +106,7 @@ jobs:
     needs: [windows-build, unix-build]
     if: ${{ github.event_name == 'release' }}
     runs-on: ubuntu-latest
+    timeout-minutes: 45
     permissions:
       # Upload generated binaries to the GitHub release.
       contents: write
@@ -139,6 +143,7 @@ jobs:
   build-success:
     if: always()
     runs-on: ubuntu-latest
+    timeout-minutes: 45
     needs: [windows-build, unix-build]
     permissions: {}
     steps:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -12,6 +12,9 @@ on:
   release:
     types: [published]  # releases and pre-releases (release candidates)
 
+permissions:
+  contents: read
+
 defaults:
   run:
     shell: bash
@@ -29,22 +32,22 @@ jobs:
           - macos-latest
 
     steps:
-      - uses: actions/checkout@v6
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
         with:
             # grab the commit passed in via `tag`, if any
             ref: ${{ github.event.inputs.tag }}
             # need to fetch unshallow so that setuptools_scm can infer the version
             fetch-depth: 0
+            persist-credentials: false
 
       # debug
       - name: Git shorthash
         run: git rev-parse --short HEAD
 
       - name: Python
-        uses: actions/setup-python@v6
+        uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 # v6
         with:
           python-version: "3.11"
-          cache: "pip"
 
       - name: Generate Binary
         run: |
@@ -58,7 +61,7 @@ jobs:
 
 
       - name: Upload Artifact
-        uses: actions/upload-artifact@v4
+        uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4
         with:
           name: vyper-${{ runner.os }}
           path: dist/vyper.*
@@ -67,22 +70,22 @@ jobs:
     runs-on: windows-latest
 
     steps:
-      - uses: actions/checkout@v6
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
         with:
             # grab the commit passed in via `tag`, if any
             ref: ${{ github.event.inputs.tag }}
             # need to fetch unshallow so that setuptools_scm can infer the version
             fetch-depth: 0
+            persist-credentials: false
 
       # debug
       - name: Git shorthash
         run: git rev-parse --short HEAD
 
       - name: Python
-        uses: actions/setup-python@v6
+        uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 # v6
         with:
           python-version: "3.11"
-          cache: "pip"
 
       - name: Generate Binary
         run: >-
@@ -91,7 +94,7 @@ jobs:
           ./make.cmd freeze
 
       - name: Upload Artifact
-        uses: actions/upload-artifact@v4
+        uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4
         with:
           name: vyper-${{ runner.os }}
           path: dist/vyper.*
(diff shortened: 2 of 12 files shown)


.github/workflows/build.yml
pcaversaccio, line 7:
```suggestion
        default: ''
        type: string
```

pcaversaccio: just a nit improvement

pcaversaccio (commented):
some further idea:
- add `timeout-minutes` on every job (not sure if 45min would be enough). without it, a "hung" job can burn runner minutes indefinitely and can block other work tbh.

banteg:
Addressed in f7815da07.

Changes:
- added `timeout-minutes` to every workflow job;
- used 45 minutes for regular jobs and 120 minutes for the Era compiler tester because it can do a cache-miss toolchain build;
- applied the workflow_dispatch input nit by adding `type: string` for `tag`.

Validation:
- parsed all workflow YAML files and checked every job has `timeout-minutes`;
- `git diff --check`.

.github/workflows/build.yml
pcaversaccio: resolved via https://github.com/vyperlang/vyper/pull/5003/changes/f7815da07694368fb4b525d82d2fe62e240ad356

.github/workflows/bytecode-size-comment.yml
```diff
@@ -0,0 +1,62 @@
...
+    types: [completed]
+
+permissions: {}
+
+jobs:
+  comment:
+    if: >
+      github.event.workflow_run.conclusion == 'success' &&
+      github.event.workflow_run.event == 'pull_request' &&
+      github.event.workflow_run.pull_requests[0].number != null
```
harkal: this is null for PRs from forks, so we will probably not get the comments on out typical forked repo branches

harkal (dismissed):

.github/workflows/bytecode-size-comment.yml
banteg: Fixed in 662f385c6. The comment workflows no longer depend on workflow_run.pull_requests[0].number being populated; they use it when present and otherwise resolve the open PR from workflow_run.head_repository.owner.login + head_branch, matching head_sha before falling back to the first open match. Applied the same fix to both bytecode-size and gas-bench comment workflows. Validation: git diff --check; uv run --python 3.12 python -c YAML parse for both workflow files.

.github/workflows/era-tester.yml
```diff
@@ -19,30 +26,28 @@ concurrency:
...
+      with:
+        persist-credentials: false
 
     - name: Rust setup
-      uses: actions-rust-lang/setup-rust-toolchain@v1
+      uses: actions-rust-lang/setup-rust-toolchain@46268bd060767258de96ed93c1251119784f2ab6 # v1
       with:
         toolchain: nightly-2022-11-03
 
     - name: Set up Python ${{ matrix.python-version[0] }}
```
pcaversaccio: So maybe I missed it when reading the overall file, but `matrix.python-version[0]` references a matrix that does _not_ exist on this job here. So the Python version will silently fall back to the runner default. We should fix this imho.

.github/workflows/era-tester.yml
```diff
@@ -9,6 +9,13 @@ name: Era compiler tester
 
 on: [push, pull_request]
 
+permissions:
+  contents: read
+
+env:
+  ERA_HASH: 943fcc39d1173409fb4a3f53029b170fb7cb4e60
+  ERA_VYPER_HASH: 178d18fc81ab16bd76db0ce2bf89a87930ae9861
```
pcaversaccio: nit observation: `ERA_VYPER_HASH` has no effect on what code actually runs. It only influences the cache key from what I can tell, which means the cache could be invalidated by a SHA change that doesn't actually change anything in the build.

banteg: Fixed in aa47aa153 by removing `ERA_VYPER_HASH` from the workflow env and Era toolchain cache key. The cache now keys on the Era tester revision only, since the current Vyper binary is built and copied in after the toolchain cache is restored/saved.

Validation:
- `git diff --check`
- parsed `.github/workflows/era-tester.yml` with YAML assertions that `ERA_VYPER_HASH` is gone.

.github/workflows/era-tester.yml
banteg: Fixed in aa47aa153 by replacing the nonexistent `matrix.python-version[0]` reference with an explicit `PYTHON_VERSION: "3.11"` env value used by `actions/setup-python`.

Validation:
- `git diff --check`
- parsed `.github/workflows/era-tester.yml` with YAML assertions that the matrix reference is gone.

.github/workflows/era-tester.yml
```diff
@@ -19,30 +26,28 @@ concurrency:
...
 
     - name: Rust setup
-      uses: actions-rust-lang/setup-rust-toolchain@v1
+      uses: actions-rust-lang/setup-rust-toolchain@46268bd060767258de96ed93c1251119784f2ab6 # v1
       with:
         toolchain: nightly-2022-11-03
 
-    - name: Set up Python ${{ matrix.python-version[0] }}
-      uses: actions/setup-python@v6
+    - name: Set up Python ${{ env.PYTHON_VERSION }}
```
pcaversaccio, line 42: i like this `env.PYTHON_VERSION` approach, we don't do this in the other CI files (we just hardcode it there), can we also do this in the other files pls

banteg: Done in d330b6908. I added workflow-level `PYTHON_VERSION` env values to the other workflows using a single setup-python version (`build`, `bytecode-size`, `gas-bench`, `ghcr`, `release-pypi`, and `test`) and replaced the hardcoded setup-python inputs/names. I left the test matrix version entries as matrix data.

Validation:
- `git diff --check`
- parsed the touched workflow YAML files and asserted there are no remaining hardcoded setup-python `python-version: "3.x"` inputs.

pcaversaccio: @charles-cooper we currently have a mix of Python `3.11` and `3.12`. Let's make it consistent `3.12` agreed?

charles-cooper: sure!

pcaversaccio: ok clanker bitch do your work 🤣🤣🤣

banteg: Done in 5e04cae56. All workflow-level `PYTHON_VERSION` values are now `3.12`, and the default test matrix moved to py312 while keeping a dedicated py311 coverage entry.

Validation:
- `git diff --check`
- parsed the touched workflow YAML files and asserted the setup-python versions are centralized on `PYTHON_VERSION: "3.12"`.

.github/dependabot.yml
```diff
@@ -0,0 +1,8 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    cooldown:
+      default-days: 7
```
pcaversaccio, line 8: let's group everything together to reduce noise:

```yml
groups:
  github-actions:
    patterns:
      - "*"
```

pcaversaccio (approved):
I left one final comment to be addressed, but otherwise lgtm. Approving for now, but I'd like @charles-cooper or @harkal to do a proper PR review as well before merging.

.github/dependabot.yml
banteg: Done in b5914eb05. I added a `github-actions` Dependabot group with `patterns: ["*"]` under the existing GitHub Actions update entry.

Validation:
- `git diff --check`
- parsed `.github/dependabot.yml` and asserted the group pattern is present.

charles-cooper (commented):
so iiuc the main changes here are: adding `persist-credentials: false` to the `pull_request_target` workflows, replacing most (all?) action versions with commit hashes rather than versions (not sure if i like this since we will lose access to patch / bugfixes, but i can see that it reduces chance of supply-chain attacks), and a bunch of bash changes to reduce change of quoting / injection attacks. is that right?

banteg:
@charles-cooper yes, that is the main scope.

More precisely:
- untrusted PR jobs stay read-only; bytecode/gas comment writes moved to separate `workflow_run` follow-up jobs that only download the report artifact and write the comment body;
- token-bearing jobs now use narrower permissions and checkout credentials are disabled where checkout is needed;
- all current workflow `uses:` references are full-SHA pinned, with tag comments left in place for reviewability/Dependabot updates. The tradeoff is intentional to reduce mutable-tag supply-chain risk;
- shell/GitHub API handling was tightened around quoting, comment body generation, and fork PR lookup edge cases.

The later review follow-ups added job timeouts, Python 3.12 consistency, and grouped Dependabot action updates.

PR #5004, opened 2026-05-21
chore[docs]: improve details in skills

Sporarum:
### Commit message

```
this commit improves small details in the skill files for agents.
```

### Cute Animal Picture

### DIFF
--- a/skills/semantics.md
+++ b/skills/semantics.md
@@ -15,7 +15,7 @@ Source → pre_parser → Python AST → Vyper AST → semantic analysis → ann
 See [vyper/ast/README.md](../vyper/ast/README.md).
 
 - `pre_parser.py`: Vyper source → parseable Python (substitutes `struct`/`interface` → `class`, etc.)
-- `grammar.lark`: Lark grammar definition
+- `grammar.lark`: Lark grammar definition (but we mainly use the python parser)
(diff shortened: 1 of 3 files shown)


skills/semantics.md
```diff
@@ -15,7 +15,7 @@ Source → pre_parser → Python AST → Vyper AST → semantic analysis → ann
 See [vyper/ast/README.md](../vyper/ast/README.md).
 
 - `pre_parser.py`: Vyper source → parseable Python (substitutes `struct`/`interface` → `class`, etc.)
-- `grammar.lark`: Lark grammar definition
+- `grammar.lark`: Lark grammar definition (but we mainly use the python parser)
```
charles-cooper: the authoritative parser, which we also use with pre- and post-processing steps, is the CPython parser. the lark grammar is there mainly for reference; we test it to make sure we adhere to a bnf spec.

charles-cooper (commented):
some small comments, otherwise lgtm

skills/semantics.md
Sporarum: I opted for being very short, to not spend too many tokens on this since we use it so little

Sporarum: Updated, let me know what you think

charles-cooper (approved):

PR #5009, opened 2026-05-22
fix[venom]: phi single use fix

HodanPlodky:
### Commit message

```
Fixing the error which would trigger incorrect compile if the phi instruction did not follow single use assumption. This commit improved the SingleUseExpansion pass to ensure this assumption.
```

### Cute Animal Picture


harkal (approved):
lgtm

PR #5010, opened 2026-05-23
fix[tool]: serialize symbolic lengths in metadata

xrchz:
### What I did

Made symbolic type lengths JSON-serializable in compiler metadata.

### How I did it

Added a small helper for serializing `INF` and wildcard length sentinels,
and used it when bytestring and dynamic-array types emit `to_dict()`
metadata. Numeric bounds are still emitted as integers.

### How to verify it

Export test JSON for cases involving unbounded or wildcard lengths, or
check that annotated type dictionaries containing `String[INF]`,
`Bytes[INF]`, or `DynArray[..., INF]` can be passed through
`json.dumps()`.

### Commit message

```
symbolic length sentinels introduced for unbounded and wildcard types
can flow into annotated type dictionaries. those dictionaries are
expected to be JSON-serializable, but the raw enum members for INF and
wildcard are not, causing JSON export of compiled test metadata to fail
during final serialization.

serialize symbolic lengths at the same boundary as other AST/type
metadata: keep bounded lengths as integers, but emit INF and wildcard as
their source-level string forms. this mirrors the existing Ellipsis
serialization behavior and avoids changing the runtime sentinel objects
or adding broad exporter-specific fallback handling.
```

Fix JSON serialization of compiler metadata for symbolic type lengths.

### Cute Animal Picture

quokka

### DIFF
--- a/vyper/semantics/types/bytestrings.py
+++ b/vyper/semantics/types/bytestrings.py
@@ -2,7 +2,7 @@ from vyper import ast as vy_ast
 from vyper.abi_types import ABI_Bytes, ABI_String, ABIType
 from vyper.exceptions import CodegenPanic, StructureException, UnexpectedNodeType, UnexpectedValue
 from vyper.semantics.types.base import VyperType
-from vyper.semantics.types.infinity import INF, WILDCARD, LengthUpperBound
+from vyper.semantics.types.infinity import INF, WILDCARD, LengthUpperBound, length_to_json
 from vyper.semantics.types.utils import get_index_value
 from vyper.utils import ceil32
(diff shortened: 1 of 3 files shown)


charles-cooper (commented):
I see a serialization mechanism but no deserialization mechanism.. not sure if it's important for this scope

charles-cooper (approved):

xrchz:
> I see a serialization mechanism but no deserialization mechanism.. not sure if it's important for this scope

yeah i think it's not needed for this. we're not reading the json back in in this repo afaik.

PR #5016, opened 2026-05-27
fix[docs]: update interface return-bound note

lufa23:
### What I did
  
Rewrite the stale note in `docs/interfaces.rst` (Implementing an Interface section). The previous wording described the pre-243d6557600a contravariance, where the interface bound was treated as a lower bound on the implementation. After 243d6557600a, variance is covariant on returns, and `T[...]` is the way to spell "any length" in an interface.

### Commit message

```
The note in the implements section described the pre-243d6557600a
contravariance, where the interface bound was treated as a lower bound
on the implementation. Variance was inverted in 243d6557600a, so the
upper bound in the interface is now an actual upper bound on the
implementation, and the T[...] wildcard is the way to spell "any length"
in an interface. Rewrite the note to match current behavior.

commit 243d6557600a is GH 4967
```


docs/interfaces.rst
```diff
@@ -128,7 +128,9 @@ Multiple ``implements`` statements can be grouped into one:
 
 .. note::
 
-  Interfaces that implement functions with return values that require an upper bound (e.g. ``Bytes``, ``DynArray``, or ``String``), the upper bound defined in the interface represents the lower bound of the implementation. Assuming a function ``my_func`` returns a value ``String[1]`` in the interface, this would mean for the implementation function of ``my_func`` that the return value must have **at least** length 1. This behavior might change in the future.
+  For functions returning length-bounded types (``Bytes``, ``DynArray``, ``String``), the upper bound declared in the interface is the **upper bound** for the implementation: if the interface declares ``String[10]``, the implementation must return a value of length **at most** 10 (covariant return types). Parameter bounds go the other way — an implementation may accept *wider* parameter types than the interface declares (contravariant parameters).
```
Sporarum: I think it would be useful to add an example, and justify why this is correct (given contravariance is notoriously hard to grasp)

I personally like the funnel metaphor:
To replace a funnel, you need it to be *wider* at the input, and *narrower* at the output.
If it were narrower at the input, things that are meant to go inside the funnel could pass by it.
And if it were wider at the output, then things that are meant to go on the container below it might actually spill to the side.
Similarly to implement a given function, you need to accept everything it accepts, but can accept more, so you need a super-type of the parameter.
And for it to always returns something which the original function can return, it must have a type no bigger than the return type, in other words, a subtype.

docs/interfaces.rst
```diff
@@ -128,7 +128,9 @@ Multiple ``implements`` statements can be grouped into one:
...
+
+  To declare "any length" in an interface, use the wildcard syntax ``T[...]`` (e.g. ``Bytes[...]``, ``String[...]``, ``DynArray[uint256, ...]``). The wildcard is only valid inside interface declarations and is resolved at the external call site against the expected type.
```
Sporarum: I would discourage using `T[...]` to mean that, as it is a better fit to denote fixed-size lists, which do not allow `...` !

docs/interfaces.rst
lufa23: thanks, let me know if it's better now

docs/interfaces.rst
lufa23: how about now?

docs/interfaces.rst
```diff
@@ -128,7 +128,34 @@ Multiple ``implements`` statements can be grouped into one:
...
+      @external
+      @view
+      def foo(x: Bytes[20]) -> String[5]:  # wider parameter, narrower return
+          return ""
+
+  Matching bounds are also valid: an implementation declared as ``def foo(x: Bytes[10]) -> String[10]`` satisfies the same interface.
+
+  Think of a function as a funnel. To swap it for another, the input has to be at least as wide (the new function must accept everything the original did, and may accept more), and the output has to be no wider (whatever it produces must still fit where the original output went). A narrower input would let some valid arguments fall through, and a wider output would exceed what the caller is prepared to receive.
+
+  To declare "any length" in an interface, use the wildcard ``...`` with one of the length-bounded types: ``Bytes[...]``, ``DynArray[uint256, ...]``, or ``String[...]``. The wildcard is only valid inside interface declarations and only for these types; static arrays (e.g. ``uint256[5]``) still require a concrete length. At the call site, a wildcard in the return type is resolved against the expected type, or to an unbounded type otherwise.
```
Sporarum: We should maybe explain how it works wrt mismatches, like what if someone calls the method with `"hello"` which is of course a valid `String[...]`, but our implementation only handles up to `String[2]`
This will make the receiving contract revert


`...` is inherently less safe than a specific n or `INF`
(where "safe" means "program might revert unexpectedly", and not "program is vulnerable to attackers")

Sporarum: "an unbounded type otherwise"

I'm not sure whether unbounded type should mean "a length of INF" or "a length of INF or `...`", both have their merrits, what do you think ?

(Asking here because this choice impacts that sentence)

Sporarum (commented):
Very nice !

Here are a couple additional comments

docs/interfaces.rst
Sporarum: I think it's much better !

docs/interfaces.rst
Sporarum: Perfect

docs/interfaces.rst
lufa23: updated, lmk if this reads right

lufa23: It may be better not to name INF to avoid introducing it as a concept

Sporarum (approved):
I think it's a very clear explanation now, thanks !

PR #5020, opened 2026-05-29
feat[codegen]: changed the convert from bytes

HodanPlodky:
### What I did
changed behavior of `convert` when converting from bytes to int

### How I did it
Added step before sign extend so that the value passed is converting from max size and not from runtime size (example in tests).

### How to verify it
test added in this PR

### Commit message

### Cute Animal Picture


vyper/builtins/_convert.py
```diff
@@ -445,17 +452,17 @@ def _cast_bytestring(expr, arg, out_typ):
...
 @_input_types(StringT, BytesT)
-def to_bytes(expr, arg, out_typ):
+def to_bytes(expr, arg, in_typ, out_typ):
     return _cast_bytestring(expr, arg, out_typ)
 
 
 @_input_types(IntegerT)
-def to_flag(expr, arg, out_typ):
+def to_flag(expr, arg, in_typ, out_typ):
     if arg.typ != UINT256_T:
```
Sporarum, line 494: When is `arg.typ` different from `in_typ` ?

If the answer is "sometimes", this feels like something which should be fixed in the typer

Sporarum:
Does this bring venom towards current vyper, or away from it ?
In other words, how does the other codegen handle it

Sporarum:
Please add a test comparing `convert(b"\xff", int16)` and `convert(b"\x00\xff", int16)`

vyper/builtins/_convert.py
HodanPlodky: It should be only in the case the constant.
```
FOO: constant(Bytes[2]) = b'\xff'

@external
def foo() -> int16:
    return convert(FOO, int16)
```

Sporarum: In this example, what are the arg.typ and in_typ ?

HodanPlodky: `arg.typ` would be `Bytes[1]` and and `in_typ` would be `Bytes[2]`, since the argument is reduced to the `b'\xff'` at the start.

HodanPlodky: https://github.com/vyperlang/vyper/pull/5020/changes#diff-4efb65bec529eac2b88ea939ea97f7a333a051988d8bf25d0352fb8fb3cc1f6dL471 here is that reduce

Sporarum: I think it would be better to have `arg.typ` be `Bytes[2]`, which makes sense since the arg is `FOO` and its type is `Bytes[2]`

HodanPlodky: That would be some where in the typechecker I think. Do you think it should be done like that? Since there was discussion on what semantics the convert should have. Or maybe I could reduce it later in convert

Sporarum: Note that your implementation also assumes a specific semantics !
I'm not sure what the semantics of `convert` should be, but I am 100% sure the following should be equivalent:
```python
FOO: constant(Bytes[2]) = b'\xff'
FOO: constant(Bytes[2]) = b'\x00\xff'
```
Both `FOO`s should have type `Bytes[2]` even if their expression has a stricter type.

So we should fix this where appropriate, probably in the typer, since this is not related only to conversions.
(Conversions might be the only place for now where it produces an observable difference however)

Sporarum: Wait, what I said is wrong, I am 100% sure the following should be equivalent:
```python
FOO: constant(Bytes[2]) = b'\xff'

_FOO: constant(Bytes[1]) = b'\xff'
FOO: constant(Bytes[2]) = _FOO
```
Both FOOs should have type Bytes[2].

But in my previous example, both should still have type `Bytes[2]`

Sporarum: The `arg.typ` vs `in_typ` discrepancy remains, I think this should be fixed before this gets merged

Sporarum (changes requested):

Sporarum (approved):
LGTM !

How did you solve the in_typ vs arg.typ in the end ?

HodanPlodky:
> LGTM !
> 
> How did you solve the in_typ vs arg.typ in the end ?

after widening it was not necessary

PR #5022, opened 2026-05-30
fix[parser]: reject non-name for-loop targets early

banteg:
### What I did

Fixed a parser crash when a `for` loop uses an invalid tuple target, such as:

```vyper
for i, j in [(1, 2)]:
    pass
```

This is intentionally invalid Vyper syntax. The fix is that it now raises a clearer `SyntaxException`:

```text
invalid for loop syntax: not a name
```

instead of leaking an internal `AttributeError`.

### How I did it

The parser now rejects non-name loop targets explicitly before constructing the missing type annotation hint. Simple name targets still use the existing missing-annotation diagnostic.

### How to verify it

```bash
uv run --no-dev python -m py_compile vyper/ast/parse.py tests/functional/syntax/test_for_range.py
```

I also ran a focused compiler smoke check for the invalid tuple-target case:

```vyper
for i, j in [(1, 2)]:
    pass
```

### Commit message

```text
Reject non-name for-loop targets before constructing the missing type
annotation hint.

Add syntax coverage for tuple targets so this path raises a clear
SyntaxException instead of leaking AttributeError.
```

Fixed an internal parser error for `for` loops with tuple targets.

### Cute Animal Picture

red panda

### DIFF
--- a/tests/functional/syntax/test_for_range.py
+++ b/tests/functional/syntax/test_for_range.py
@@ -483,6 +483,7 @@ def bar():
             """
 @external
 def bar():
+    # Tuple targets are invalid Vyper syntax, but should raise SyntaxException.
     for i, j in [(1, 2)]:
         pass
     """,
--- a/tests/functional/syntax/test_for_range.py
+++ b/tests/functional/syntax/test_for_range.py
@@ -487,7 +487,7 @@ def bar():
     for i, j in [(1, 2)]:
         pass
     """,
-            "missing type annotation",
+            "invalid for loop syntax: not a name",
         ),
         (
             """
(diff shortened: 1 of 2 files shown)


charles-cooper:
i think we should just reject non-name targets explicitly. "missing type annotation" won't help you if you typed `for i, j in ...`

so instead the code should read like
```
invalid for loop syntax: not a name
```

banteg:
Addressed in c739fe20f: non-name for-loop targets are now rejected before the missing-annotation hint is constructed, with the explicit error:

```text
invalid for loop syntax: not a name
```

Updated the regression test and PR description accordingly.

charles-cooper (approved):

PR #5023, opened 2026-05-30
fix[parser]: improve missing for-loop annotation hints

banteg:
### What I did

Improved the hint for missing `for` loop type annotations when the compiler can infer the iterable element type from nearby syntax.

For example, this now suggests `address` instead of always suggesting `uint256`:

```vyper
@external
def foo(strategies: DynArray[address, 10]):
    for strategy in strategies:
        pass
```

The existing `uint256` hint remains for `range(...)` loops and for cases where the parser does not have enough local syntax to identify the iterable element type.

### How I did it

The parser annotation visitor now tracks simple annotations for module state, function arguments, and local annotated assignments. When the missing annotation diagnostic is built, it unwraps obvious container annotations such as `DynArray[T, N]`, static arrays, and `public(DynArray[T, N])` to suggest the loop item type.

### How to verify it

```bash
uv run --no-dev python -m py_compile vyper/ast/parse.py tests/functional/syntax/test_for_range.py
```

I also ran focused compiler smoke checks for `range(...)`, function argument arrays, local array variables, and `self.<state array>` cases.

### Commit message

```text
fix[parser]: improve missing for-loop annotation hints

Suggest the iterable element type for missing for-loop type annotations
when it can be recovered from simple parser-visible annotations.

Keep the existing uint256 hint for range loops and fallback cases where
the parser cannot identify an iterable item type.
```

Improved missing `for` loop type annotation hints for array iteration.

### DIFF
--- a/tests/functional/syntax/test_for_range.py
+++ b/tests/functional/syntax/test_for_range.py
@@ -495,6 +495,58 @@ def test_for_annotation_syntax_exceptions(code, error_message):
         compiler.compile_code(code)
 
 
+@pytest.mark.parametrize(
+    "code,expected_hint",
+    [
+        (
+            """
+@external
+def bar():
+    for i in range(10):
+        pass
+    """,
+            "did you mean something like `for i: uint256 in ...`?",
+        ),
+        (
+            """
+@external
+def bar(strategies: DynArray[address, 10]):
+    for strategy in strategies:
+        pass
+    """,
+            "did you mean something like `for strategy: address in ...`?",
+        ),
+        (
+            """
+@external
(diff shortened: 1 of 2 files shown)


charles-cooper (changes requested):
there's a deeper design issue here.. the parser should not really know anything about types (which is why we gave such a bad error message to begin with -- we don't have access yet to machinery like `type_from_annotation()` etc)

banteg:
Closing this based on the review: the current approach puts annotation/type interpretation into the parser, which is the wrong layer. The better fix needs a different diagnostic design outside the parser rather than patching this PR further. Leaving the branch intact for reference.

PR #5024, opened 2026-05-31
feat(parser): collect all syntax errors before raising

Olyno:
### What I did

Wired the existing ExceptionList mechanism into the parser so it collects all syntax errors in a single pass instead of stopping at the first one.

### How I did it

In _parse_to_ast, replaced the try/except/raise with a while True loop. On each SyntaxError, the offending line is blanked in the reformatted code and the parse is retried. After the parse succeeds, all collected errors are raised at once via ExceptionList.

### How to verify it

 Run the new test:
    
 ```bash
 pytest tests/functional/syntax/test_multi_error.py -v
 ```   
    
 Or compile any Vyper file with multiple syntax errors. All of them should appear in the output.

### Commit message

```
feat(parser): collect all syntax errors before raising
```

Report all syntax errors at once instead of stopping at the first one.

### Cute Animal Picture

### DIFF
--- a/vyper/ast/parse.py
+++ b/vyper/ast/parse.py
@@ -111,6 +111,15 @@ def _parse_to_ast(
                 lines[e.lineno - 1] = ""
             code = "\n".join(lines)
 
+    # If we collected syntax errors during parsing, raise them before
+    # AST annotation. Otherwise, blanked lines may leave unconsumed
+    # pre-parser metadata (e.g. for_loop_annotations) that would cause
+    # an internal AssertionError instead of the intended diagnostics.
+    if errors:
+        err_list = ExceptionList()
+        err_list.extend(errors)
+        err_list.raise_if_not_empty()
+
     annotate_python_ast(
         py_ast,
         vyper_source,
@@ -133,12 +142,6 @@ def _parse_to_ast(
     module.is_interface = is_interface
 
     module.settings = pre_parser.settings
-
-    # If we collected errors along the way, raise them now
-    if errors:
-        err_list = ExceptionList()
-        err_list.extend(errors)
-        err_list.raise_if_not_empty()
 
     return module
 
--- a/vyper/ast/parse.py
+++ b/vyper/ast/parse.py
@@ -8,7 +8,7 @@ from typing import Optional
 
 from vyper.ast import nodes as vy_ast
 from vyper.ast.pre_parser import PreParser
-from vyper.exceptions import CompilerPanic, ParserException, SyntaxException
+from vyper.exceptions import CompilerPanic, ExceptionList, ParserException, SyntaxException
 from vyper.utils import sha256sum
 from vyper.warnings import Deprecation, vyper_warn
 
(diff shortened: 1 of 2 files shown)


vyper/ast/parse.py
```diff
@@ -76,29 +76,49 @@ def _parse_to_ast(
...
-
-        likely_errors = ("staticall", "staticcal")
-        tmp = str(new_e)
-        for s in likely_errors:
-            if s in tmp:
-                new_e._hint = "did you mean `staticcall`?"
-                break
-
-        raise new_e from None
+    errors: list[SyntaxException] = []
```
Sporarum: Should be a `ExceptionList` directly

vyper/ast/parse.py
```diff
@@ -76,29 +76,49 @@ def _parse_to_ast(
...
+    code = pre_parser.reformatted_code
+
+    while True:
+        try:
+            py_ast = python_ast.parse(code)
```
Sporarum, line 84: How does this solve the problem, isn't `code` the same on every loop ?

Sporarum (changes requested):
Hello, thanks for your contribution !

I really like the idea, I think it is a limitation of the current system which is not ideal

It's however a very hard problem to fix in practice, as one early syntax mistake can cause many downstream ones, and therefore create a lot of unhelpful noise

Therefore to merge something like this, it would need to come with extensive testing to make sure the UX is not made worse

Olyno:
Hi @Sporarum  !

Thank you for the review! I agree this needs to be conservative to avoid making UX worse. I updated the implementation to stop recovery when it no longer makes progress, and to avoid  reporting indentation cascade errors caused by earlier syntax recovery.

I also added broader tests covering multiple recoverable errors, source-order reporting, no-progress recovery, cascade suppression, pre-parser metadata, resolved paths, and preserved hints.

---

To give a little bit more context, I'm interested in the Vyper project, and because I think it lacks modern tooling for IDEs like Zed or VS Code, I decided to create my own extensions (https://github.com/Olyno/serpent). This includes an LSP server.

To improve the UX, I wanted to get all errors for the user. That's why I'm creating this PR haha. In addition to the unit tests, I tried the extensions, mainly Zed, to make sure they worked fine.

Don't hesitate to tell me if you spot anything I could improve or anything I missed!

vyper/ast/parse.py
Olyno: `code` changes at the end of each loop: the line reported by `SyntaxError.lineno` is blanked, then the parser retries on that modified source.

vyper/ast/parse.py
Olyno: Updated! 👍🏻

vyper/ast/parse.py
Sporarum: Oh, I don't think that's the right approach ^^'

With that I mean that removing a faulty line can create even more downstream issues:
```python
#if True:
    line1
    line2
    line3
    ...
```
The excessive indentation would probably be reported at line1, then that would be deleted, and then the error would be on line2, and then on line3, etc

So instead of a single error, you would get an arbitrary number !

Sporarum (commented):
I understand now, super cool you're working on tooling !

I think to get multiple errors at the same time, we need to work with the parser, not fight against it
How does multiple syntax errors work for python ?
Because we use the cpython parser, so it would probably be very insightful

Sporarum:
Note to self: Don't reply to a discussion as part of a review, that creates weird things

Olyno:
Hi, thanks for the detailed feedback.

I agree with your concern. After thinking more about it, I see a few possible directions:

1. Keep the current best-effort approach, with strict stop conditions for unsafe cases.
2. Keep CPython fail-fast, and only collect multiple errors from Vyper’s own parsing/preprocessing layer before CPython.
3. Change the direction of this PR toward a Vyper-aware recovery layer, where Vyper controls safe recovery points instead of relying on CPython retries.

What approach would you prefer here?

charles-cooper:
> I understand now, super cool you're working on tooling !
> 
> I think to get multiple errors at the same time, we need to work with the parser, not fight against it How does multiple syntax errors work for python ? Because we use the cpython parser, so it would probably be very insightful

i don't think cpython supports multiple syntax errors...

Olyno:
> i don't think cpython supports multiple syntax errors...

@charles-cooper It doesn't, that's why it can be quite complicated to rely only on that.

Sporarum:
If cpython doesn't support multiple syntax errors, I don't think we should try to hack around it
For now we should leave it as-is, and improve what we can on the other side

And at a later point, we could consider using our own parser once enough divergences/shortcomings have accumulated (but we're not there yet)

PR #5025, opened 2026-05-31
fix: reject extcall/staticcall in assert/raise revert reason

### LINKED ISSUE #4966: extcall in assert message causes CodegenPanic
### Version Information

* OS: Linux Ubuntu Jammy
* Python Version: Python 3.11

### What's your issue about?

Using an `extcall` expression as the message of an `assert` statement crashes the compiler with a `CodegenPanic` whose text explicitly admits the typechecker missed it.

Reproducer:

```vyper
interface I:
    def get_msg() -> String[20]: nonpayable

@external
def f(x: bool, target: address):
    assert x, extcall I(target).get_msg()
```

Error:

```
CodegenPanic: unhandled exception typechecker missed this, parse_ExtCall
```

This is an asymmetric coverage gap in Vyper's "constant context" check for the assert message expression:

- `assert x, "literal"` → compiles
- `assert x, s` (a `String[...]` variable) → compiles
- `assert x, self.view_fn()` → compiles
- `assert x, staticcall I(t).fn()` → compiles
- `assert x, self.mutating()` (nonpayable internal) → cleanly rejected
- `assert x, extcall I(t).fn()` → `CodegenPanic` (this bug)

Internal mutating calls in the assert msg are rejected by the constant-context check at `vyper/codegen/self_call.py:53`:

```python
if context.is_constant() and func_t.is_mutable:
    raise StateAccessViolation(
        f"May not call state modifying function '{method_name}' "
        f"within {context.pp_constancy()}.",
        stmt_expr,
    )
```

External calls go through `parse_ExtCall`, which has no equivalent guard. The typechecker side, in `vyper/semantics/analysis/local.py:499-509`, only validates the type of the msg expression, not the expression kind:

```python
def _validate_revert_reason(self, msg_node: vy_ast.VyperNode) -> None:
    if isinstance(msg_node, vy_ast.Str):
        ...
    elif not (isinstance(msg_node, vy_ast.Name) and msg_node.id == "UNREACHABLE"):
        try:
            validate_expected_type(msg_node, StringT(1024))
        except TypeMismatch as e:
            raise InvalidType("revert reason must fit within String[1024]") from e
        self.expr_visitor.visit(msg_node, get_exact_type_from_node(msg_node))
```

So an `extcall` slips past the typechecker and reaches codegen, where `parse_ExtCall` panics with the "typechecker missed this" message.

### How can it be fixed?

Add a constant-context check to `parse_ExtCall` (analogous to the one at `vyper/codegen/self_call.py:53` for internal calls), or preferably, reject `extcall` expressions in the typechecker for `assert.msg` / `raise.exc` positions in `_validate_revert_reason` (`vyper/semantics/analysis/local.py:499-509`), so the error is caught at semantic analysis instead of leaking out of codegen as a panic.

wahajahmed010:
## Summary

Using `extcall` (or `staticcall`) as the message argument to `assert` or `raise` crashes the compiler with `CodegenPanic` because the codegen pipeline sets the context to constant mode when evaluating revert reasons, and external calls violate that constraint.

This adds an early guard in `_validate_revert_reason` in the typechecker to reject `ExtCall` and `StaticCall` nodes with a clear `StructureException` before they reach codegen.

## Changes

- `vyper/semantics/analysis/local.py`: Added two type checks at the top of `_validate_revert_reason` to reject `ExtCall` and `StaticCall` nodes
- `tests/functional/syntax/test_revert_reason_external_calls.py`: Added parametrized tests covering `assert` and `raise` with both `extcall` and `staticcall`

## Before

```
CodegenPanic: unhandled exception typechecker missed this, parse_ExtCall
```

## After

```
StructureException: Cannot use extcall in revert reason
  function "f", line 7:14
       6 def f(x: bool, target: address):
  ---> 7     assert x, extcall I(target).get_msg()
```

Fixes #4966

### DIFF
--- a/tests/functional/syntax/test_revert_reason_external_calls.py
+++ b/tests/functional/syntax/test_revert_reason_external_calls.py
@@ -1,0 +1,57 @@
+import pytest
+
+from vyper import compiler
+from vyper.exceptions import StructureException
+
+fail_list = [
+    (
+        """
+interface I:
+    def get_msg() -> String[20]: nonpayable
+
+@external
+def f(x: bool, target: address):
+    assert x, extcall I(target).get_msg()
+        """,
+        StructureException,
+    ),
+    (
+        """
+interface I:
+    def get_msg() -> String[20]: nonpayable
+
+@external
+def f(x: bool, target: address):
+    raise extcall I(target).get_msg()
+        """,
+        StructureException,
+    ),
+    (
+        """
+interface I:
+    def get_msg() -> String[20]: nonpayable
+
+@external
+def f(x: bool, target: address):
+    assert x, staticcall I(target).get_msg()
+        """,
+        StructureException,
+    ),
+    (
+        """
+interface I:
+    def get_msg() -> String[20]: nonpayable
+
+@external
+def f(x: bool, target: address):
+    raise staticcall I(target).get_msg()
+        """,
+        StructureException,
+    ),
+]
+
+
+@pytest.mark.parametrize("bad_code,exc", fail_list)
+def test_revert_reason_rejects_external_calls(bad_code, exc):
+    with pytest.raises(exc):
+        compiler.compile_code(bad_code)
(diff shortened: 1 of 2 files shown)


Sporarum (commented):
Hello, thank you for your contribution !

There are other places where staticcalls are forbidden, and I think these are done using a modifiability check
It would probably also work here, could you try ?

Inspecting asts directly tends to be fragile, as codex pointed out

PR #5028, opened 2026-06-02
fix[venom]: sccp eval compares ensure numeric value

HodanPlodky:
### Commit message

### Cute Animal Picture


charles-cooper (changes requested):
looks fishy, what is the bug this solves?

HodanPlodky:
> looks fishy, what is the bug this solves?

it was not a bug it was a bad interaction between sccp and other passes the algebraic opt in particular. Sccp could boolean into IR and that could prevent some optimization if the pass expects numbers

PR #5029, opened 2026-06-03
feat[lang]: allow wildcard container lengths in events

Sporarum:
### What I did

Allow wildcard container lengths (`Bytes[...]`, `String[...]`, `DynArray[T, ...]`) in event definitions. Previously this syntax was only accepted in interfaces.

### How I did it

Extended the context check in `get_index_value` (in `vyper/semantics/types/utils.py`) to also permit `vy_ast.EventDef` ancestors. Added a targeted error when `INF` is used in an event field, directing the user to use `...` instead.

Added comprehensive tests:
- wildcard Bytes, String, and DynArray in event fields
- indexed wildcard Bytes/String in events
- ABI signature verification (wildcards collapse to canonical unbounded types)
- rejection of `INF` in event fields

### How to verify it

Run the updated tests in `tests/unit/semantics/types/test_inf.py`.

### Commit message

```
event field types currently require exact lengths for dynamic containers
(Bytes, String, DynArray). this forces users to choose an arbitrary
upper bound even when they simply want to log data of unknown size.
interfaces already accept the `...` wildcard for this purpose, but
events reject it.

extend the wildcard check in `get_index_value` to also allow event
definitions. when a wildcard is used in an event field, the ABI
signature uses the canonical unbounded type (bytes, string, or T[]).

also add a specific error when `INF` is used in an event field,
directing the user to `...` instead. `INF` represents the unbounded type
internally, but event fields must have a concrete ABI encoding, so the
wildcard shortcut is the correct spelling.

note: DynArray[...] in events currently panics in legacy codegen. this
is an existing limitation and is marked xfail in the test suite.
```

Allow wildcard container lengths (`...`) in event field definitions.

### Cute Animal Picture

### DIFF
--- a/tests/unit/semantics/types/test_inf.py
+++ b/tests/unit/semantics/types/test_inf.py
@@ -115,7 +115,7 @@ def test_wildcard_rejected_outside_interface(build_node):
 
 
 @pytest.mark.parametrize(
-    "source,expected",
+    "source,expected,panics_in_legacy",
     [
         (
             """
@@ -127,6 +127,7 @@ def go():
     log Foo(x=b"abc")
 """,
             b"abc",
+            False,
         ),
         (
             """
@@ -138,8 +139,9 @@ def go():
     log Foo(x="abc")
 """,
             "abc",
+            False,
         ),
-        pytest.param(
+        (
             """
 event Foo:
     x: DynArray[uint256, ...]
@@ -149,11 +151,15 @@ def go():
     log Foo(x=[1, 2, 3])
 """,
             [1, 2, 3],
-            marks=pytest.mark.xfail(raises=CodegenPanic, strict=True),
+            True,
         ),
     ],
 )
-def test_event_wildcard_emits(get_contract, get_logs, source, expected):
+def test_event_wildcard_emits(
+    get_contract, get_logs, source, expected, panics_in_legacy, experimental_codegen, request
+):
+    if experimental_codegen or panics_in_legacy:
+        request.node.add_marker(pytest.mark.xfail(raises=CodegenPanic, strict=True))
     c = get_contract(source)
     c.go()
     (log,) = get_logs(c, "Foo")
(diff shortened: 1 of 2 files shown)


Sporarum:
I do not think we should do this, see [the first commit](github.com/vyperlang/vyper/pull/5029/commits/70ec64b9deb6ccb855b918c241e221afce769568) for what already works, both inside and outside interfaces

The fact we have to add extra logic to me highlights that this clashes with the design of unbounded types, where type containing `INF` are regular types allowed basically everywhere, and `...` are only a shortcut for interfaces

pcaversaccio (commented):
so my current understanding is that the wildcard `...` means that the dev is _free_ to do whatever they want -> perfect for interfaces, it's generic, no restrictions. `INF` means that the length is _unbounded_ at runtime so you might need to check it -> not generic enough for an interface. The wildcard `...` has already different semantics depending how it is used (e.g. in return types vs function arguments) and as an application dev I will try to stick to _one_ syntax whenever possible and won't add `INF` to an interface to not add further mental overhead. I second `...` for `event`s. How we're gonna solve this for `struct`s is a different conversation but fortunately no prominent EIPs uses `struct`s.

charles-cooper (approved):
lgtm. @pcaversaccio ?

tests/unit/semantics/types/test_inf.py
```diff
@@ -103,15 +103,110 @@ def test_dynarray_wildcard_from_annotation(build_node):
...
+""",
+            [1, 2, 3],
+            True,
+        ),
+    ],
+)
+def test_event_wildcard_emits(
+    get_contract, get_logs, source, expected, panics_in_legacy, experimental_codegen, request
+):
+    if experimental_codegen or panics_in_legacy:
```
charles-cooper, line 161: why does this panic?

pcaversaccio (commented):
quickly mirroring the offline discussion we had in the Vyper group. so this currently compiles:

```vy
event Foo:
    x: String[...]

@external
def bar():
    log Foo(x="vyper")
```

IMHO the best and most simplistic way is to only allow `...` in interfaces (and no `INF`) and only allow `INF` (and no `...`) in contracts. I understand that this will create a challenge with `struct`s in interfaces.

PR #5035, opened 2026-06-05
feat[lang]: access immutables through `self`

Sporarum:
### What I did

Make immutables accessible through `self`, consistent with storage and transient variables. Bare-name access (`foo` instead of `self.foo`) still works but now emits a `Deprecation` warning; it will be removed in a future release.

### How I did it

Mostly by removing special cases:

- Register immutables as members of the `self` type, like storage variables (they are also still added to the module namespace so the deprecated bare-name form keeps working for now).
- Emit a `Deprecation` warning (with a `use self.<name>` hint) whenever an immutable is referenced by bare name — every occurrence is reported so a contract can be migrated in one pass.
- Accept both `self.<name> = ...` and the deprecated `<name> = ...` in the constructor-assignment check, and drop the old special-cased "Immutable variables must be accessed without 'self'" error.
- Generate public getters with a standard `self.<name>` return expression instead of the immutable-specific `Name` node.
- While at it, reword the bare-name storage variable error to hint style: `did you mean self.<name>?`.

### How to verify it

This now compiles cleanly:

```vyper
foo: public(immutable(uint256))

@deploy
def __init__(x: uint256):
    self.foo = x

@external
def bar() -> uint256:
    return self.foo
```

while the old bare-name form still compiles but warns on every occurrence:

```
Deprecation: immutables should now be accessed through `self`

  contract "imm.vy:5", function "__init__", line 5:4
       4 def __init__(x: uint256):
  ---> 5     foo = x
  -----------^

  (hint: use `self.foo` instead)
```

Test coverage: `tests/functional/syntax/test_immutables.py`, `tests/functional/syntax/warnings/test_deprecation_warning.py`, `tests/functional/codegen/features/test_immutable.py`; the rest of the test suite was migrated to the `self.` form.

### Commit message

```
immutables were registered in the module namespace and accessed as bare
names, unlike storage and transient variables which are accessed through
`self`.

register immutables as members of the self type so `self.<name>` becomes
the canonical way to read and assign them. bare-name access keeps
working for now but emits a `Deprecation` warning with a `use
self.<name>` hint; every occurrence is reported so a contract can be
migrated in one pass. the constructor assignment check accepts both
forms. desugaring of public getters is also updated to use
`self.<name>`, matching storage getters.

bare-name access will be removed in a future release.
```

Immutables are now accessed through `self` (e.g. `self.foo`); bare-name access is deprecated and will be removed in a future release.

### Cute Animal Picture

### DIFF
--- a/vyper/semantics/analysis/local.py
+++ b/vyper/semantics/analysis/local.py
@@ -540,10 +540,21 @@ class FunctionAnalyzer(VyperNodeVisitorBase):
             raise ImmutableViolation("Cannot write to calldata")
 
         if info.modifiability == Modifiability.RUNTIME_CONSTANT:
-            if info.location != DataLocation.CODE:
+            if info.location == DataLocation.CODE:
+                if not func_t.is_constructor:
+                    raise ImmutableViolation("Immutable value cannot be written to")
+
+                # handle immutables
+                if info.var_info is not None:  # don't handle complex (struct,array) immutables
+                    # special handling for immutable variables in the ctor
+                    # TODO: maybe we want to remove this restriction.
+                    if info.var_info._modification_count != 0:
+                        raise ImmutableViolation(
+                            "Immutable value cannot be modified after assignment"
+                        )
+                    info.var_info._modification_count += 1
+            else:
                 raise ImmutableViolation("Environment variable cannot be written to")
-            if not func_t.is_constructor:
-                raise ImmutableViolation("Immutable value cannot be written to")
 
         if info.modifiability == Modifiability.CONSTANT:
             raise ImmutableViolation("Constant value cannot be written to.")
--- a/vyper/semantics/analysis/local.py
+++ b/vyper/semantics/analysis/local.py
@@ -540,21 +540,10 @@ class FunctionAnalyzer(VyperNodeVisitorBase):
             raise ImmutableViolation("Cannot write to calldata")
 
         if info.modifiability == Modifiability.RUNTIME_CONSTANT:
-            if info.location == DataLocation.CODE:
-                if not func_t.is_constructor:
-                    raise ImmutableViolation("Immutable value cannot be written to")
-
-                # handle immutables
-                if info.var_info is not None:  # don't handle complex (struct,array) immutables
-                    # special handling for immutable variables in the ctor
-                    # TODO: maybe we want to remove this restriction.
-                    if info.var_info._modification_count != 0:
-                        raise ImmutableViolation(
-                            "Immutable value cannot be modified after assignment"
-                        )
-                    info.var_info._modification_count += 1
-            else:
+            if info.location != DataLocation.CODE:
                 raise ImmutableViolation("Environment variable cannot be written to")
+            if not func_t.is_constructor:
+                raise ImmutableViolation("Immutable value cannot be written to")
 
         if info.modifiability == Modifiability.CONSTANT:
             raise ImmutableViolation("Constant value cannot be written to.")
--- a/vyper/semantics/analysis/module.py
+++ b/vyper/semantics/analysis/module.py
@@ -137,7 +137,7 @@ def _analyze_module_bodies(module_ast: vy_ast.Module) -> None:
     with override_global_namespace(namespace):
         analyze_functions(module_ast)
         _validate_exports_uses(module_ast, module_t)
-        _validate_initialized_modules(module_ast, module_t)
+        _validate_constructor(module_ast, module_t)
         _validate_used_modules(module_ast, module_t)
 
 
@@ -181,7 +181,7 @@ def _validate_used_modules(module_ast: vy_ast.Module, module_t: ModuleT) -> None
         err_list.raise_if_not_empty()
 
 
-def _validate_initialized_modules(module_ast: vy_ast.Module, module_t: ModuleT) -> None:
+def _validate_init_calls(module_ast: vy_ast.Module, module_t: ModuleT) -> None:
     """Check all `initializes:` modules have `__init__()` called exactly once."""
     # only call `__init__()` for modules which have an
     # `__init__()` function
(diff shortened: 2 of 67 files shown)


vyper/semantics/analysis/module.py
```diff
@@ -250,6 +250,63 @@ def _validate_initialized_modules(module_ast: vy_ast.Module, module_t: ModuleT)
...
+
+    constructor = module_t.init_function
+
+    set_immutables: dict[str, vy_ast.Assign] = {}
+
+    if constructor is not None:
+        assert isinstance(constructor.ast_def, vy_ast.FunctionDef)
+        self_assigns = constructor.ast_def.get_descendants(
+            vy_ast.Assign, {"target.value.id": "self"}
+        )
```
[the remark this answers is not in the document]
Sporarum: Not sure the previous system caught that either, but will fix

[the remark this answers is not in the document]
charles-cooper: yes it did

vyper/semantics/analysis/local.py
```diff
@@ -540,21 +540,10 @@ def _handle_modification(self, target: vy_ast.ExprNode):
...
-
-                # handle immutables
-                if info.var_info is not None:  # don't handle complex (struct,array) immutables
-                    # special handling for immutable variables in the ctor
-                    # TODO: maybe we want to remove this restriction.
-                    if info.var_info._modification_count != 0:
-                        raise ImmutableViolation(
-                            "Immutable value cannot be modified after assignment"
-                        )
-                    info.var_info._modification_count += 1
```
charles-cooper, line 555: uhhhhhh why was this entire (correct) section removed?

vyper/semantics/analysis/module.py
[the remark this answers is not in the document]
charles-cooper: https://github.com/vyperlang/vyper/pull/5035/changes#r3366118793

vyper/semantics/analysis/local.py
pcaversaccio: why are the tests passing while this correct section was removed? we need to add a test for this case!

pcaversaccio: btw, for this PR you can ignore the failing `gas-bench` CI run as this is a breaking change which is currently incompatible with the latest snekmate version.

Sporarum: Because it's not quite correct:
```python
@deploy
def __init__(b: bool) ->
    if b:
       self.my_var = 1
    else:
       self.my_var = 0 # ImmutableViolation: Immutable value cannot be modified after assignment
```
I'm working on a system which would solve that, and this is a first step in this direction
(Would also solve the same kind of thing with `init` calls)

vyper/semantics/analysis/module.py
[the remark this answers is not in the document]
Sporarum: Checking my new tests on master, it did indeed catch it, but reported a potentially confusing error message in some cases:
```python
VALUE: immutable(uint256)

@deploy
def __init__():
    VALUE += 1
```
Would fail with "Immutable definition requires an assignment in the constructor"
(due to only checking `Assign` in `_validate_exports_uses`)


And adding that assignment would then raise "Immutable value cannot be modified after assignment" on whichever came second.
(because we check all accesses in `_handle_modification`)

vyper/semantics/analysis/local.py
charles-cooper: i mean that's a known issue https://github.com/vyperlang/vyper/issues/3278

but i don't think you should introduce regressions to known behavior that are outside the scope of this PR, especially by accident. like we can maybe change immutables semantics later in a different PR, after discussion

Sporarum: Reverted

vyper/semantics/analysis/utils.py
```diff
@@ -391,7 +391,7 @@ def types_from_Name(self, node):
             and name in self.namespace["self"].typ.members
         ):
             raise InvalidReference(
-                f"'{name}' is a storage variable, access it as self.{name}", node
+                f"'{name}' is an assignable variable, access it as self.{name}", node
```
charles-cooper:
```suggestion
                f"'{name}' is a module-level variable, access it as self.{name}", node
```

charles-cooper: hm was just a thought, take it or leave it

vyper/semantics/analysis/module.py
```diff
@@ -756,19 +756,12 @@ def visit_VariableDecl(self, node):
             # mutability is checked automatically preventing assignment
             # outside of the constructor, here we just check a value is assigned,
             # not necessarily where
+            # Assign(target=Attribute(value=Name("self"), attr=<node.target.id>))
             assignments = self.ast.get_descendants(
-                vy_ast.Assign, filters={"target.id": node.target.id}
+                vy_ast.Assign, filters={"target.value.id": "self", "target.attr": node.target.id}
```
charles-cooper: we should have a more sophisticated analysis for this but i think it can be out of scope, make a new issue

vyper/semantics/analysis/utils.py
charles-cooper: i think we should actually do

```suggestion
                f"'{name}'", node, hint=f"did you mean self.{name}?"
```

vyper/semantics/analysis/module.py
```diff
@@ -756,19 +756,12 @@ def visit_VariableDecl(self, node):
...
             )
             if not assignments:
-                # Special error message for common wrong usages via `self.<immutable name>`
```
charles-cooper, line 763: this now leaves a UX bug. this program now produces a rather confusing error message:

```
FOO: immutable(uint256)

@deploy
def __init__():
    FOO = 5
```
=>
```
vyper.exceptions.ImmutableViolation: Immutable definition requires an assignment in the constructor

  contract "tmp/foo.vy:2", line 2:0 
       1
  ---> 2 FOO: immutable(uint256)
  -------^
       3
```

charles-cooper (changes requested):
https://github.com/vyperlang/vyper/pull/5035/changes#r3508092299

vyper/semantics/analysis/module.py
Sporarum: Done

Sporarum:
@charles-cooper 
The link in there leads me nowhere: https://github.com/vyperlang/vyper/pull/5035#pullrequestreview-4611774898
Did I handle the issue that was there ?

vyper/semantics/analysis/module.py
```diff
@@ -756,19 +757,21 @@ def visit_VariableDecl(self, node):
...
-                    "Immutable variables must be accessed without 'self'"
-                    if len(wrong_self_attribute) > 0
-                    else "Immutable definition requires an assignment in the constructor"
+                # `<name> = ...`: Assign(target=<name>)
+                wrong_assignments = self.ast.get_descendants(
+                    vy_ast.Assign, filters={"target.id": name}
                 )
+                if wrong_assignments:
+                    # same error that would be raised later
+                    node = wrong_assignments[0].target
```
charles-cooper: i'm guessing users may need to do this in bulk; can we raise an exception list with the invalid reference + hint for every node in wrong_assignments? also we don't know that they are supposed to be immutables.. do we have that info by this time?

charles-cooper:
> @charles-cooper The link in there leads me nowhere: [#5035 (review)](https://github.com/vyperlang/vyper/pull/5035#pullrequestreview-4611774898) Did I handle the issue that was there ?

weird, i think this is a gh regression -- same thing happens for the link in https://github.com/vyperlang/vyper/issues/5178#issue-4793219750, it just redirects to the base PR url

vyper/semantics/analysis/module.py
Sporarum: > also we don't know that they are supposed to be immutables.. 

We're in the following if-branch haha:
```
if node.is_immutable:
```

> can we raise an exception list with the invalid reference + hint for every node in wrong_assignments?

Sure, I just went for the simplest implementation

Sporarum:
@charles-cooper 

> weird, i think this is a gh regression

Funnily enough it works with middle-click instead of left-click ...

vyper/semantics/analysis/module.py
Sporarum: > > can we raise an exception list with the invalid reference + hint for every node in wrong_assignments?
> 
> Sure, I just went for the simplest implementation

Oh actually this is a bit tricky, `wrong_assignments` is only the wrong assignments for that specific immutable (we're in `visit_VariableDecl`)

So collecting the exceptions for all immutables would require storing some state in ModuleAnalyzer and then raising afterwards
And that also means we need to keep going after we already saw an error, and hope the broken invariants don't cause issues on the way
Let me know if you still want to do it, and I'll look into it tomorrow

charles-cooper (approved):

charles-cooper:
hm i wonder if we really want to go for a breaking change with this. 0.5.x is tempting for breaking change, but there is not really a language- or compiler-level reason we need to do this, we can deprecate it for 0.5.x and then flip the 'breaking change' switch in next series 0.6.x (or 1.x?)

pcaversaccio (commented):
so, hmm, historically we have done "deprecation first, break later" with other things (e.g. https://github.com/vyperlang/vyper/pull/3697, https://github.com/vyperlang/vyper/pull/4257), thus I'm personally fine to do this also here.

Sporarum:
Implements: #2695

PR #5066, opened 2026-06-12
fix[venom]: deploy runtime code on explicit constructor return

### LINKED ISSUE #5048: Venom: constructor `return` can deploy empty runtime bytecode
### Version Information

* OS: macOS

### What's your issue about?

A constructor path with an explicit `return` can terminate initcode before Venom emits the deploy epilogue, causing the deployed contract to have empty runtime bytecode.

The relevant path is `vyper/codegen_venom/module.py`'s constructor generation interacting with `vyper/codegen_venom/stmt.py`'s external-return lowering. Constructors are lowered with no `return_pc`, so a bare `return` routes to the external-return path and emits `stop()`. Because the builder is then terminated, constructor generation skips the deploy epilogue that would copy and return the runtime bytecode.

A minimal shape is:

```vyper
@deploy
def __init__():
    return

@external
def f() -> uint256:
    return 1
```

Under the legacy pipeline, the constructor body is still wrapped by the deploy sequence and returns runtime bytecode. Under the Venom path, the explicit constructor return can halt with `STOP`, returning zero-length initcode output. If the `return` is conditional, this becomes data-dependent: only that constructor path deploys an empty contract.

### How can it be fixed?

Route constructor `return` through the deploy epilogue instead of emitting a direct `stop()`. One approach is to give constructor lowering a shared epilogue block/return target and make `return` jump there, mirroring the way legacy funnels deploy exits through the deploy wrapper.

banteg:
### What I did

fixes #5048
Fixed a bug in the experimental codegen pipeline (`codegen_venom`) where a constructor containing an explicit `return` statement deploys a contract with **empty runtime bytecode**.

### How I did it

An explicit `return` in `__init__` routes to `_lower_external_return` (the constructor has no `return_pc`), which emits `stop`. `_generate_constructor` then skips the deploy epilogue because the builder is already terminated, so the init code halts without ever codecopy'ing/returning the runtime code. With a conditional `return`, only that path deploys empty, making the corruption data-dependent. The legacy pipeline doesn't have this problem because it always wraps the constructor in a `deploy` IR node that performs the copy+return regardless of how the body terminates.

The fix creates a shared constructor exit block holding the nonreentrant unlock and the deploy epilogue. Explicit returns jump to it, and the fall-through path does the same. The block is always appended after the constructor body; unreachable exit paths are left for the optimizer to remove.

This keeps the current language behavior intact: the frontend and legacy codegen already accept a bare `return` in `@deploy`, while Venom deployed empty runtime bytecode for the same source. Forbidding constructor `return` would be a separate language-level change across semantic validation and both pipelines.

### How to verify it

```vyper
x: public(uint256)

@deploy
def __init__():
    self.x = 42
    return
```

Before this fix, compiling with `--experimental-codegen` produces init code ending in `STOP` (no `CODECOPY`/`RETURN`), deploying empty runtime code:

```
venom (before): 0x3461000a57602a5f55 00 5b5f80fd...
venom (after):  0x3461001557602a5f55 602c5f6100199039602c5ff3 5b5f80fd...
```

The added tests deploy contracts with unconditional and conditional ctor returns and check state via a getter, under both pipelines.

### Commit message

```
in the codegen_venom pipeline, an explicit `return` in `__init__`
routes to `_lower_external_return` (the constructor has no
`return_pc`), which emits `stop`. `_generate_constructor` then
skips the deploy epilogue because the builder is already terminated, so
the init code halts without ever codecopy'ing and returning the runtime
code -- the contract deploys with empty runtime bytecode. with a
conditional `return`, only that path deploys empty, making the
corruption data-dependent.

the legacy pipeline does not have this problem because it always wraps
the constructor body in a `deploy` IR node which performs the
copy+return regardless of how the body terminates.

fix by creating a shared constructor exit block that holds the
nonreentrant unlock and the deploy epilogue. explicit returns jump to
it, and the fall-through path does the same. the block is always
appended after the constructor body; unreachable exit paths are left
for the optimizer to remove.
```

fix[venom]: deploy runtime code on explicit constructor return


Sporarum:
Maybe we should just forbid `return` in the constructor

banteg:
I think forbidding constructor `return` can be considered separately, but it is broader than this Venom bug fix. Today the frontend and legacy codegen accept a bare `return` in `@deploy`; legacy deploys runtime correctly, while Venom deployed empty runtime bytecode. I updated the PR description to make that scope explicit: this PR preserves current semantics and fixes Venom parity. A language-level ban would need semantic validation/tests for both pipelines.

Focused validation: `uv run --python 3.12 -m pytest tests/functional/codegen/features/test_constructor.py -k "constructor_explicit_return or constructor_conditional_return" -q` passes.

charles-cooper:
> Maybe we should just forbid `return` in the constructor

no i don't think this is good, fast path in ctor is a totally valid pattern

vyper/codegen_venom/context.py
```diff
@@ -96,6 +96,12 @@ class VenomCodegenContext:
     # used by deploy epilogue to copy staging area into bytecode.
     immutables_alloca: Optional[IRVariable] = None
 
+    # Shared exit block holding the deploy epilogue (for constructor context).
+    # Explicit `return` in __init__ must jump here instead of halting,
+    # otherwise the runtime code would never be returned.
+    ctor_exit_block: Optional[IRBasicBlock] = None
```
charles-cooper, line 102: why optional? i think we should create ctor exit block unconditionally. optimizing uses is left to the optimizer

charles-cooper (changes requested):
https://github.com/vyperlang/vyper/pull/5066/changes#r3415537986

vyper/codegen_venom/context.py
banteg: Done in 8fc8406ee. Removed `ctor_exit_used` and now append `ctor_exit` unconditionally; fallthrough and explicit constructor returns both target the same deploy epilogue. Validation: constructor explicit/conditional return pytest and black check passed.

vyper/codegen_venom/module.py
```diff
@@ -1379,18 +1379,27 @@ def _generate_constructor(
...
+    if fallthrough:
+        builder.jmp(exit_bb.label)
+
+    builder.append_block(exit_bb)
+    builder.set_block(exit_bb)
+
     # Unlock
-    if not builder.is_terminated():
-        codegen_ctx.emit_nonreentrant_unlock(func_t)
+    codegen_ctx.emit_nonreentrant_unlock(func_t)
```
charles-cooper, line 1403: looks good, although i don't think we can even have nonreentrant on ctors so idk how this code got in there in the first place

charles-cooper (approved):
lgtm, see https://github.com/vyperlang/vyper/pull/5066/changes#r3416094028 which could be considered out of scope. paging @harkal for addl review

banteg:
Refreshed this branch in f7fc5dfca with a normal merge of current master (no force-push). The only overlap was in codegen_venom/context.py; the resolution keeps both master’s IDENTITY_PRECOMPILE import and this PR’s IRBasicBlock import.

Validation:
- uv run --python 3.12 -m pytest tests/functional/codegen/features/test_constructor.py --experimental-codegen -q (18 passed)
- uv run --python 3.12 -m pytest tests/functional/codegen/features/test_constructor.py -q (18 passed)
- focused Ruff and Black checks passed

I also corrected the PR body to match the current implementation: the shared constructor exit block is always appended, with unreachable cleanup left to the optimizer.

harkal:
#5208 has the same constructor fix plus the `SimplifyCFGPass` canonicallization and tests for nested/loop/immutable returns. I suggest closing this in favour of #5208 rather than reconciling two conflicting versions of `_generate_constructor`, but happy to send just the CFG fix here instead if you'd rather keep this one.

harkal:
https://github.com/vyperlang/vyper/pull/5208 got merged, deprecating this.

PR #5067, opened 2026-06-12
fix[venom]: fix call argument evaluation order

### LINKED ISSUE #5050: Venom: call arguments can observe later side effects
### Version Information

* OS: macOS

### What's your issue about?

Venom internal and external call lowering can read an earlier argument after a later argument or keyword argument has already mutated it.

Both call paths first lower argument expressions into `VyperValue`s, then unwrap/copy them later into staging buffers. For storage, memory, and transient values, `lower()` can produce a pointer rather than snapshotting the value immediately. If a later argument has side effects, the earlier pointer is read after those side effects.

Minimal shape:

```vyper
x: uint256

@internal
def mutate() -> uint256:
    self.x = 999
    return 0

@internal
def pick(a: uint256, b: uint256) -> uint256:
    return a

@external
def run() -> uint256:
    self.x = 1
    return self.pick(self.x, self.mutate())
```

Legacy snapshots `self.x` before evaluating `self.mutate()` and returns `1`. The Venom lowering can pass the mutated value and return `999`. The same class applies to external call positional args when a later positional arg or extcall kwarg mutates the earlier argument's location.

This is related to the public evaluation-order work around #2863, #4596, #4985, #4988, and #5001, but those do not cover ordinary positional call arguments being held as lazy pointers.

### How can it be fixed?

Snapshot call arguments left-to-right at evaluation time. Complex/location-backed values need to be copied into stable temporaries before evaluating later arguments or kwargs, while still avoiding staging-buffer clobbering by nested calls.

banteg:
### What I did

fixes #5050
Fixed an argument evaluation-order bug in the experimental codegen pipeline (`codegen_venom`): internal and external call arguments were read *after* later arguments' (or kwargs') side effects executed, passing the mutated values.

### How I did it

Both call paths lower all argument expressions before unwrapping/staging them. But lowering a storage/memory variable yields a *pointer*, not a value — no read is emitted. If a later argument (or an extcall kwarg like `value=`) mutates the location, the staging copy reads the mutated value:

```vyper
self.x = 1
return self.foo(self.x, self.mutate())   # mutate() sets self.x = 999
```

passes `999` for the first argument under venom; legacy copies each argument into the args buffer before evaluating the next one and passes `1`. The same divergence affects complex arguments, e.g. `self.take(arr, arr.pop())` sees the shrunk array.

The fix freezes each argument as soon as it is evaluated — loading primitive words onto the stack, copying composites into a fresh temporary — whenever any later argument or kwarg expression can have side effects (conservatively: contains a call). Arguments without risky successors keep the lazy pointer, so the common case emits no extra copies. Staging buffers are still allocated only after all arguments are evaluated, preserving the existing protection against nested internal calls clobbering them.

### How to verify it

The added tests cover the internal-call scalar case, the internal-call complex (dynarray + `pop()`) case, and the extcall case, under both pipelines. All three fail on master with `--experimental-codegen` (e.g. `9990 != 10`, `203 != 303`) and pass with this fix.

### Commit message

```
in the codegen_venom pipeline, internal and external calls lower all
argument expressions before unwrapping and staging them. but lowering a
storage/memory variable yields a pointer, not a value -- no read is
emitted. if a later argument (or an extcall kwarg such as `value=`)
has side effects that mutate the location, the mutated value is read
during staging and passed to the callee:

    self.x = 1
    self.foo(self.x, self.mutate())  # mutate() sets self.x = 999

passes 999 for the first argument, whereas the legacy pipeline copies
each argument into the args buffer before evaluating the next one and
passes 1. the same divergence affects complex arguments, e.g.
`self.take(arr, arr.pop())` sees the shrunk array.

fix by freezing each argument as soon as it is evaluated -- loading
primitive words onto the stack and copying composites into a fresh
temporary -- whenever any later argument or kwarg expression can have
side effects (conservatively, contains a call). arguments without
risky successors keep the lazy pointer, avoiding gratuitous copies in
the common case. the staging buffers are still allocated only after all
arguments are evaluated, since nested internal calls may clobber them
(see the existing comment referencing legacy contains_self_call
handling).
```

fix[venom]: fix call argument evaluation order

### DIFF
--- a/vyper/codegen_venom/expr.py
+++ b/vyper/codegen_venom/expr.py
@@ -70,6 +70,15 @@ class _CallKwargs:
 ENVIRONMENT_VARIABLES = {"block", "msg", "tx", "chain"}
 
 
+def _may_have_side_effects(node: vy_ast.VyperNode) -> bool:
+    """Whether evaluating the expression can mutate observable state.
+
+    Conservative: any call (internal call, extcall, builtin, mutating
+    method call like `arr.append(...)`) counts as a side effect.
+    """
+    return len(node.get_descendants(vy_ast.Call, include_self=True)) > 0
+
+
 class Expr:
     """Lower Vyper expressions to Venom IR.
 
@@ -1355,9 +1364,18 @@ class Expr:
         # By evaluating all args first, we ensure nested calls complete before we
         # allocate staging buffers, avoiding corruption.
         # See legacy codegen: vyper/codegen/self_call.py (contains_self_call handling)
+        #
+        # Each argument is frozen (loaded to stack / copied to a fresh
+        # temporary) as soon as it is evaluated if any later argument can
+        # have side effects; otherwise the lazily-read location could be
+        # mutated before the staging copy, observably changing evaluation
+        # order vs the legacy pipeline.
         arg_vals: list[VyperValue] = []
-        for arg_node in all_arg_nodes:
-            arg_vals.append(Expr(arg_node, self.ctx).lower())
+        for i, arg_node in enumerate(all_arg_nodes):
+            arg_val = Expr(arg_node, self.ctx).lower()
+            if any(_may_have_side_effects(n) for n in all_arg_nodes[i + 1 :]):
+                arg_val = self.ctx.freeze_value(arg_val, annotation="internal call arg")
+            arg_vals.append(arg_val)
 
         # Now allocate staging buffers and copy evaluated values
         for i, arg_val in enumerate(arg_vals):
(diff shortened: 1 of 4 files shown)


vyper/codegen_venom/expr.py
```diff
@@ -1355,9 +1364,18 @@ def _lower_internal_call(self) -> VyperValue:
...
+        # temporary) as soon as it is evaluated if any later argument can
+        # have side effects; otherwise the lazily-read location could be
+        # mutated before the staging copy, observably changing evaluation
+        # order vs the legacy pipeline.
         arg_vals: list[VyperValue] = []
-        for arg_node in all_arg_nodes:
-            arg_vals.append(Expr(arg_node, self.ctx).lower())
+        for i, arg_node in enumerate(all_arg_nodes):
+            arg_val = Expr(arg_node, self.ctx).lower()
+            if any(_may_have_side_effects(n) for n in all_arg_nodes[i + 1 :]):
```
charles-cooper, line 1376: this is clever.. in an ideal world we would just freeze all subexprs so we don't need to worry about this? but idk how aggressively we are able to optimize (yet). cc @harkal thoughts?

charles-cooper (commented):
paging @harkal for review

banteg:
Superseded by #5127, now merged into master. I checked the current call-argument snapshot implementation and reran the extracted call-order/overlap regressions against master (25 passed, 4 expected failures across the combined coverage). Closing this redundant branch; no remaining change from this PR needs to be carried forward.

PR #5078, opened 2026-06-12
fix[tool]: make integrity sum invariant to layout file formatting

### LINKED ISSUE #5074: spurious integrity mismatch warning on solc_json round trip with storage layout overrides
### Version Information

* OS: osx

### What's your issue about?

A clean round trip through `-f solc_json` with a storage layout override produces a spurious integrity mismatch warning.

```bash
$ cat main.vy
x: uint256

@external
def set_x(v: uint256):
    self.x = v

$ cat layout.json   # pretty-printed, as a human would write it
{
    "x": {
        "type": "uint256",
        "n_slots": 1,
        "slot": 0
    }
}

$ vyper -f solc_json main.vy --storage-layout-file layout.json > solc.json
$ vyper-json solc.json
...
[warning] Mismatched integrity sum\! Expected 7a37e4d6c9d13d1812a45ec82731d97e364b3c09d8c6ba15e464f35bf800365b but got c6432e0cbb38b1c8907d2dad302bd8f068357052893a7549370f13c2466b3775. (This likely indicates a corrupted archive)
```

Root cause: the integrity sum hashes the layout override's **raw text** (`JSONInput.contents`, via `CompilerData._compute_integrity_sum`), but the two sides of the solc_json round trip disagree about what that text is:

* `SolcJSONWriter.write_storage_layout_overrides` exports the **parsed object** (`storage_layout_override.data`), losing the original formatting:
  https://github.com/vyperlang/vyper/blob/242a4fa7d/vyper/compiler/output_bundle.py#L190-L197
* `get_storage_layout_overrides` reconstructs contents as compact `json.dumps(override_data)`:
  https://github.com/vyperlang/vyper/blob/242a4fa7d/vyper/cli/vyper_json.py#L225-L231

So any layout file whose formatting differs from compact `json.dumps` output (pretty-printed, trailing newline, key spacing) triggers the false "corrupted archive" warning. With `-W error` the round trip fails outright.

The `.vyz` archive format is unaffected: `VyperArchiveWriter` stores the raw `contents` bytes, which round-trip exactly.

### How can it be fixed?

Either:

* hash a canonical form of JSON inputs (e.g. `json.dumps(data, sort_keys=True, separators=(",", ":"))`) where `JSONInput.sha256sum` feeds the integrity sum — this also removes the odd property that the same logical layout yields different integrity sums depending on input format; or
* make `SolcJSONWriter` emit the raw `contents` string so `vyper_json` can reconstruct it byte-exactly.

banteg:
### What I did

Fixes #5074.

A clean round trip through `-f solc_json` with a storage layout override produced a spurious `Mismatched integrity sum! ... (This likely indicates a corrupted archive)` warning (an outright failure with `-W error`).

### How I did it

The integrity sum hashed the layout override's raw file text, but `SolcJSONWriter` exports the *parsed* layout object and `vyper_json` reconstructs the contents with compact `json.dumps` — so any layout file whose formatting differs from compact `json.dumps` output (e.g. pretty-printed) broke the round trip. The `.vyz` archive format was unaffected since it stores raw bytes.

Fix: add `JSONInput.canonical_sha256sum`, which hashes a canonical serialization (`json.dumps(data, sort_keys=True)`) of the parsed data instead of the raw text, and use it for the layout component of the integrity sum. This makes the integrity sum invariant to layout file formatting across all input bundle types (filesystem, standard json, archive).

Note: this changes integrity sums for compilations that use a storage layout override, so archives produced by older compiler versions will produce a one-time integrity warning when recompiled.

### How to verify it

The `input_files` fixture now pretty-prints the layout file, and the new `test_solc_json_storage_layout_integrity` asserts a solc_json round trip produces no warnings. Both this test and the pre-existing `test_integrity_sum` fail on master with the pretty-printed fixture.

### Commit message

```
the integrity sum hashed the storage layout override's raw file text,
but the solc_json output format exports the parsed layout object and
`vyper_json` reconstructs the contents with compact `json.dumps`. any
layout file whose formatting differs from compact `json.dumps` output
(e.g. pretty-printed) therefore produced a spurious "Mismatched
integrity sum! (This likely indicates a corrupted archive)" warning on
a clean solc_json round trip -- or an outright failure with `-W error`.
(the .vyz archive format was unaffected, since it stores the raw file
bytes.)

fix by hashing a canonical serialization of the parsed layout data
instead of the raw text, making the integrity sum invariant to layout
file formatting across all input bundle types. note this changes
integrity sums for compilations which use a storage layout override,
so archives produced by older compiler versions will produce a
(one-time) integrity warning when recompiled with fixed versions.
```

Fix: spurious integrity sum mismatch warning when round-tripping a compilation with storage layout overrides through the `solc_json` format.

### Cute Animal Picture

cute animal

### DIFF
--- a/tests/unit/cli/vyper_compile/test_compile_files.py
+++ b/tests/unit/cli/vyper_compile/test_compile_files.py
@@ -337,7 +337,9 @@ def bar(x: uint256) -> uint256:
         "a": {"type": "uint256", "n_slots": 1, "slot": 5},
         "b": {"type": "uint256", "n_slots": 1, "slot": 0},
     }
-    storage_layout_source = json.dumps(storage_layout_overrides)
+    # pretty-print to test that the integrity sum is invariant to the
+    # formatting of the storage layout file
+    storage_layout_source = json.dumps(storage_layout_overrides, indent=4)
 
     tmpdir = tmp_path_factory.mktemp("fake-package")
     with open(tmpdir / "lib.vy", "w") as f:
@@ -353,7 +355,7 @@ def bar(x: uint256) -> uint256:
     library_hash = sha256sum(library_source)
     jsonabi_hash = sha256sum(json_source)
     resolved_imports_hash = sha256sum(contract_hash + sha256sum(library_hash) + jsonabi_hash)
-    storage_layout_hash = sha256sum(storage_layout_source)
+    storage_layout_hash = sha256sum(json.dumps(storage_layout_overrides, sort_keys=True))
     expected_integrity = sha256sum(storage_layout_hash + resolved_imports_hash)
 
     return (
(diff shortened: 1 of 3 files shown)


charles-cooper (commented):
hm, do we still want ("raw") sha256sum or should we just switch everything to "canonical" sha256sum?

charles-cooper:
my concern here is that it makes the sha256sum harder to predict (without running the compiler), but maybe we shouldn't care -- it's just the compiler's job to generate the integrity sum and that's that

banteg:
Rechecked the current diff and the reproducibility concern. The canonical input is the parsed layout serialized with `json.dumps(layout, sort_keys=True)` using its default separators, then hashed as UTF-8 with SHA-256. That makes whitespace and object-key ordering irrelevant while preserving array ordering and values. A consumer must reproduce those serialization settings rather than hash its original JSON bytes. The two focused integrity tests pass; I do not see a simpler implementation change needed here.

PR #5079, opened 2026-06-12
fix[tool]: validate storage layout overrides in standard json input

### LINKED ISSUE #5076: broken validation in get_storage_layout_overrides silently drops extra overrides
### Version Information

* OS: osx

### What's your issue about?

Operator precedence bug in `get_storage_layout_overrides` in `vyper/cli/vyper_json.py`:

https://github.com/vyperlang/vyper/blob/242a4fa7d/vyper/cli/vyper_json.py#L221-L223

```python
if not isinstance(value, dict) and len(value.items()) \!= 1:
    raise JSONError(f"invalid storage layout override: {value}")
override_path, override_data = next(iter(value.items()))
```

The condition should be `or`, not `and`. As written:

* a dict with **more than one entry** passes validation, and `next(iter(value.items()))` silently uses the first entry and drops the rest — the dropped overrides also escape the integrity sum;
* a **non-dict** value (e.g. a list) crashes with `AttributeError: 'list' object has no attribute 'items'` instead of raising a `JSONError`.

### How can it be fixed?

```python
if not isinstance(value, dict) or len(value) \!= 1:
    raise JSONError(f"invalid storage layout override: {value}")
```

banteg:
### What I did

Fixes #5076.

Fixed an operator precedence bug in `get_storage_layout_overrides` which made the storage layout override validation in standard json input ineffective.

### How I did it

```python
if not isinstance(value, dict) and len(value.items()) != 1:
```

short-circuits to `False` for any dict, so a multi-entry override dict passed validation and `next(iter(value.items()))` silently used the first entry, dropping the rest. A non-dict value (e.g. a list) crashed with `AttributeError` instead of raising a `JSONError`. Changed `and` to `or` (and `len(value.items())` to `len(value)`, which also rejects empty dicts before `next()` would raise `StopIteration`).

### How to verify it

New parametrized test `test_invalid_storage_layout_overrides` covers multi-entry dicts, non-dict values, and the empty dict — all four cases fail on master.

### Commit message

```
the validation in `get_storage_layout_overrides` used `and` where it
meant `or`, so the check never fired for dict values: an override with
multiple files for one target passed validation and all entries except
the first were silently dropped. non-dict values crashed with
`AttributeError` instead of raising `JSONError`.
```

Fix: storage layout overrides in standard json input with multiple override files per target are now rejected instead of silently dropping all but the first.

### Cute Animal Picture

cute animal

### DIFF
--- a/tests/unit/cli/vyper_json/test_compile_json.py
+++ b/tests/unit/cli/vyper_json/test_compile_json.py
@@ -326,16 +326,16 @@ def test_unknown_storage_layout_overrides(input_json):
 @pytest.mark.parametrize(
     "bad_override",
     [
-        # more than one override file for a single target
-        {
-            "foo_overrides.json": FOO_STORAGE_LAYOUT_OVERRIDES,
-            "bar_overrides.json": BAR_STORAGE_LAYOUT_OVERRIDES,
-        },
-        # not a dict
-        [FOO_STORAGE_LAYOUT_OVERRIDES],
-        "foo_overrides.json",
-        # empty dict
-        {},
+        pytest.param(
+            {
+                "foo_overrides.json": FOO_STORAGE_LAYOUT_OVERRIDES,
+                "bar_overrides.json": BAR_STORAGE_LAYOUT_OVERRIDES,
+            },
+            id="multiple-override-paths",
+        ),
+        pytest.param([FOO_STORAGE_LAYOUT_OVERRIDES], id="not-a-dict-list"),
+        pytest.param("foo_overrides.json", id="not-a-dict-string"),
+        pytest.param({}, id="empty-dict"),
     ],
 )
 def test_invalid_storage_layout_overrides(input_json, bad_override):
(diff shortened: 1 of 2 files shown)


tests/unit/cli/vyper_json/test_compile_json.py
```diff
@@ -323,6 +323,28 @@ def test_unknown_storage_layout_overrides(input_json):
     assert e.value.args[0] == f"unknown target for storage layout override: {unknown_contract_path}"
 
 
+@pytest.mark.parametrize(
+    "bad_override",
+    [
+        # more than one override file for a single target
```
charles-cooper: how is this more than one override file for a single target?

banteg: The bad value is installed as the value for the single target `contracts/foo.vy`; the two keys inside it are the two override paths for that target. I clarified the parametrization in 58865d534 by naming that case `multiple-override-paths` and giving the other invalid shapes explicit ids. Validation: `uv run --python 3.12 -m pytest tests/unit/cli/vyper_json/test_compile_json.py -k storage_layout_overrides -q` and the focused black check.

banteg:
Rechecked the validation and the earlier test-shape request against the current head. The individual invalid-layout cases and their expected errors are covered, and all five focused tests pass. The requested adjustment is already present; I did not find an additional code change needed.

PR #5080, opened 2026-06-12
fix[tool]: raise user errors for malformed input files

### LINKED ISSUE #5120: Missing event indexed field causes compiler panic
### Version Information

* OS: Linux Ubuntu Jammy
* Python Version: Python 3.11

### What's your issue about?

Importing a JSON ABI that contains an `event` entry whose `inputs[]` are missing the `indexed` field (even on a single input) crashes the compiler with a `CompilerPanic` and the "internal compiler error" banner.

The Solidity ABI spec treats `indexed` as a boolean that defaults to `false` when not present, so many real-world ABIs (older Solidity output, hand-written ABIs, minified/canonicalized ABIs from CLI tools) omit it for non-indexed parameters. Vyper's `EventT.from_abi` reads it with an unguarded list comprehension at `vyper/semantics/types/user.py:235`:

```python
members: dict = {}
indexed: list = [i["indexed"] for i in abi["inputs"]]
for item in abi["inputs"]:
    members[item["name"]] = type_from_abi(item)
return cls(abi["name"], members, indexed)
```

If any input lacks `indexed`, Python raises `KeyError: 'indexed'`, which propagates and is re-wrapped as a `CompilerPanic`. The error is in the semantic phase, so both the legacy and `--experimental-codegen` pipelines emit the same panic.

Reproduction:

`event_no_indexed.json` (typical ERC-20 `Transfer` event with `indexed` omitted):

```json
[{
  "type": "event",
  "name": "Transfer",
  "inputs": [
    {"name": "from", "type": "address"},
    {"name": "to", "type": "address"},
    {"name": "value", "type": "uint256"}
  ]
}]
```

`main.vy`:

```vyper
import event_no_indexed as ifc

@external
def f(addr: address):
    pass
```

Command and full output:

```
$ vyper main.vy
Error compiling: main.vy
KeyError: 'indexed'

vyper.exceptions.CompilerPanic: unhandled exception 'indexed'

```

### How can it be fixed?

One-line fix at `vyper/semantics/types/user.py:235`, aligning with the ABI spec default of `false`:

```python
indexed: list = [i.get("indexed", False) for i in abi["inputs"]]
```

Related unguarded lookups at `user.py:234-238` have the same shape and could fail similarly: `abi["inputs"]`, `item["name"]`, and `abi["name"]`.

### LINKED ISSUE #5123: Invalid JSON ABI stateMutability raises CompilerPanic
### Version Information

* OS: Linux Ubuntu Jammy
* Python Version: Python 3.11

### What's your issue about?

Importing a JSON ABI that contains an invalid `stateMutability` value raises a `CompilerPanic` — an "unhandled internal compiler error" — even though the actual cause is just user-provided invalid input in an imported `.json` ABI file.

Any unrecognized `stateMutability` value triggers it: a typo (`"pyure"`, `"viiew"`), a casing mistake (`"View"`, `"PURE"`), a future-EIP / non-standard keyword (`"static"`, `"readonly"`), or any other non-enum string. Realistic exposure includes JSON ABIs hand-written by users, ABIs from third-party tools that emit non-standard mutabilities, and case mismatches between web3 tooling versions.

Root cause: `StateMutability.from_abi` at `vyper/semantics/analysis/base.py:37` calls `cls(abi_dict["stateMutability"])`. For an unknown value, Python's enum lookup invokes the `StringEnum._missing_` override at `vyper/utils.py:160`, which raises `CompilerPanic` instead of a typed user-facing error:

```python
class StringEnum(enum.Enum):
    # Override ValueError with our own internal exception
    @classmethod
    def _missing_(cls, value):
        raise CompilerPanic(f"{value} is not a valid {cls.__name__}")
```

The override was intended for internal-error reporting (where `cls(value)` is called with a known-good value), but `from_abi` exposes it directly to untrusted user input at the JSON-ABI boundary. Both the legacy and `--experimental-codegen` pipelines emit the same panic, since the error occurs in the shared semantic phase.

**Command and code that caused the failure:**

`bad_mut.json` (alongside the `.vy`):
```json
[{"type":"function","name":"foo","inputs":[],"outputs":[],"stateMutability":"banana"}]
```

`main.vy`:
```vyper
import bad_mut as ifc

@external
def f(addr: address):
    extcall ifc(addr).foo()
```

**Full output:**
```
$ vyper main.vy
Error compiling: main.vy
vyper.exceptions.CompilerPanic: banana is not a valid StateMutability

  contract "main.vy:1", line 1:0
  ---> 1 import bad_mut as ifc
  -------^
       2

```

Invalid user input at the ABI boundary should produce a typed, user-facing error (e.g. `StructureException`), not a `CompilerPanic` asking the user to file an internal-compiler-error report.

### How can it be fixed?

Validate the value in `from_abi` before invoking the enum constructor, and raise a clean `StructureException` (the missing-field path already defaults to `NONPAYABLE`). At `vyper/semantics/analysis/base.py:32-43`:

```python
if "stateMutability" in abi_dict:
    sm = abi_dict["stateMutability"]
    if not cls.is_valid_value(sm):
        raise StructureException(
            f"Invalid stateMutability '{sm}': expected one of "
            f"{[o.value for o in cls]}"
        )
    return cls(sm)
```

The same pattern may affect other `StringEnum` instances used at user-input boundaries and is worth auditing (`OptimizationLevel` and `EVMVersion` are guarded by argparse; pragma values are parsed in `pre_parser.py`).

banteg:
### What I did

Fixes #5123.
Fixes #5120.

Storage layout override files and imported JSON ABIs are user input, but several malformed-yet-plausible inputs crashed the compiler with internal errors instead of compile errors:

| input error | before | after |
|---|---|---|
| extra entry in override file | `CompilerPanic` ("Please create an issue on Github") | `StorageLayoutException` + hint |
| wrong `n_slots` / missing `type` fields | `CompilerPanic` | `StorageLayoutException` + hint |
| slot as string (`"slot": "0"`) | `TypeError: unsupported operand type(s) for +` | `StorageLayoutException` |
| entry not a dict (`"a": 5`) | `TypeError: 'int' object is not subscriptable` | `StorageLayoutException` |
| `$.nonreentrant_key` entry missing `"slot"` | `KeyError: 'slot'` | `StorageLayoutException` |
| imported JSON ABI with invalid `stateMutability` | `CompilerPanic` | `StructureException` |

### How I did it

- `_fetch_path` now validates the traversal (each segment must land on a dict) and the fetched slot (must be an `int`, with `bool` explicitly excluded since it is an `int` subclass; a non-integral float would previously have been accepted by the reservation logic and only caught later by the roundtrip check).
- The `$.nonreentrant_key` slot is read through `_fetch_path` instead of bare indexing.
- The post-allocation roundtrip check now raises `StorageLayoutException` with a hint instead of `CompilerPanic` because it is reachable by ordinary user input.
- `StateMutability.from_abi` now validates `stateMutability` from imported JSON ABI files before invoking the internal enum constructor, so invalid user-provided mutability strings raise `StructureException` instead of falling through `StringEnum._missing_` to `CompilerPanic`.

No semantics change: valid files still compile, and malformed files are still rejected. The change is the exception type and diagnostic path.

### How to verify it

New tests in `tests/unit/cli/storage_layout/test_storage_layout_overrides.py` cover the storage-layout rows above. `test_json_interface_invalid_state_mutability` covers #5123; the same imported JSON ABI reproducer raises `CompilerPanic` on `origin/master` and `StructureException` on this branch.

Focused validation:

```
uv run --python 3.12 -m pytest tests/unit/cli/storage_layout/test_storage_layout_overrides.py -q
uv run --python 3.12 -m pytest tests/functional/codegen/test_interfaces.py -k invalid_state_mutability -q
uv run --python 3.12 black --check vyper/semantics/analysis/data_positions.py vyper/semantics/analysis/base.py tests/functional/codegen/test_interfaces.py tests/unit/cli/storage_layout/test_storage_layout_overrides.py
git diff --check
```

### Commit message

```
fix[tool]: raise user errors for malformed input files

Storage layout override files and imported JSON ABI files are user
input, but several malformed files crashed through internal-error paths:
bad layout overrides could raise CompilerPanic, TypeError or KeyError,
and invalid JSON ABI stateMutability values reached StringEnum._missing_
and raised CompilerPanic.

Validate storage-layout traversal and slot values before reserving
slots, downgrade reachable layout roundtrip mismatches to
StorageLayoutException, and validate JSON ABI stateMutability before
constructing the internal enum.

Valid files still compile, and malformed files are still rejected. The
change is that user input errors now produce typed user-facing
exceptions instead of internal compiler errors.
```

Fix: malformed storage layout override files and invalid JSON ABI `stateMutability` values now produce typed user-facing errors instead of internal compiler errors.

### Cute Animal Picture

cute animal


Sporarum:
Related: https://github.com/vyperlang/vyper/issues/5123

Would be nice if this fixed it as well

vyper/semantics/analysis/data_positions.py
```diff
@@ -27,13 +27,17 @@ def set_data_positions(
...
+        # note this is reachable by user error, e.g. an override file with
+        # extra entries or wrong `type`/`n_slots` fields.
         roundtrip = generate_layout_export(vyper_module).get(_LAYOUT_KEYS[DataLocation.STORAGE], {})
         if roundtrip != storage_layout_overrides:
             msg = "Computed storage layout does not match override file!\n"
             msg += f"expected: {json.dumps(storage_layout_overrides)}\n\n"
             msg += f"got:\n{json.dumps(roundtrip)}"
-            raise CompilerPanic(msg)
+            hint = "did you provide a complete storage layout file, "
+            hint += "as generated by `vyper -f layout`?"
```
Sporarum: I would say instead something like "for example generated by ..." to highlight the fact there can be other ways to get a layout file
(but anyways we're likely to drop support)

vyper/semantics/analysis/data_positions.py
```diff
@@ -153,18 +157,23 @@ def _fetch_path(path: list[str], layout: StorageLayout, node: vy_ast.VyperNode):
...
                 "Have you used the correct storage layout file?",
                 node,
             )
         tmp = tmp[segment]
 
-    try:
-        ret = tmp["slot"]
-    except KeyError as e:
-        raise StorageLayoutException(f"no storage slot for {qualified_path}", node) from e
+    if not isinstance(tmp, dict) or "slot" not in tmp:
```
Sporarum:
```suggestion
    if not (isinstance(tmp, dict) and "slot" in tmp):
```
For me at least it's easier to think about what should be true

vyper/semantics/analysis/data_positions.py
```diff
@@ -153,18 +157,23 @@ def _fetch_path(path: list[str], layout: StorageLayout, node: vy_ast.VyperNode):
...
+        raise StorageLayoutException(f"no storage slot for {qualified_path}", node)
+
+    ret = tmp["slot"]
+    # bool is an instance of int, disallow it explicitly
+    if not isinstance(ret, int) or isinstance(ret, bool):
```
Sporarum: Hmm confusing operator precedence, prefer:

```suggestion
    if not (isinstance(ret, int) or isinstance(ret, bool)):
```

or


```suggestion
    if isinstance(ret, bool) or not isinstance(ret, int):
```

Depending on which you mean

vyper/semantics/analysis/data_positions.py
banteg: Updated the hint wording to say the layout file can be, for example, one generated by `vyper -f layout`.

vyper/semantics/analysis/data_positions.py
banteg: Applied this as the positive condition: `if not (isinstance(tmp, dict) and "slot" in tmp):`.

vyper/semantics/analysis/data_positions.py
banteg: Changed this to `if isinstance(ret, bool) or not isinstance(ret, int):`, which is the intended check since `bool` is an `int` subclass but should be rejected.

banteg:
Included #5123 in this PR in `f3ec1e692`.

What changed:
- invalid `stateMutability` values in imported JSON ABIs now raise `StructureException` before calling the internal enum constructor
- added `test_json_interface_invalid_state_mutability`
- applied the three inline readability suggestions in `data_positions.py`
- updated the PR title/body to reflect the broader malformed-input-files scope

Validation:

```
uv run --python 3.12 -m pytest tests/unit/cli/storage_layout/test_storage_layout_overrides.py -q
uv run --python 3.12 -m pytest tests/functional/codegen/test_interfaces.py -k invalid_state_mutability -q
uv run --python 3.12 black --check vyper/semantics/analysis/data_positions.py vyper/semantics/analysis/base.py tests/functional/codegen/test_interfaces.py tests/unit/cli/storage_layout/test_storage_layout_overrides.py
git diff --check
```

tests/functional/codegen/test_interfaces.py
```diff
@@ -268,6 +274,29 @@ def bar() -> uint256:
...
+    code = """
+import bad_mut as ifc
+
+@external
+def f(addr: address):
+    extcall ifc(addr).foo()
+    """
+
+    input_bundle = make_input_bundle({"bad_mut.json": json.dumps(bad_abi)})
+    with pytest.raises(StructureException, match="Invalid stateMutability 'banana'"):
```
Sporarum: Oh please use `assert ==` instead of match

match only checks the string is included, we want to check the exact string to prevent any regressions

(for all tests, not only this one)

vyper/semantics/analysis/data_positions.py
```diff
@@ -153,18 +157,23 @@ def _fetch_path(path: list[str], layout: StorageLayout, node: vy_ast.VyperNode):
     qualified_path = ".".join(path)
 
     for segment in path:
-        if segment not in tmp:
+        if not (isinstance(tmp, dict) and segment in tmp):
```
Sporarum: Since tmp is a StorageLayout, can it sometimes be a dict, and sometimes not at runtime ?

tests/functional/codegen/test_interfaces.py
banteg: Done in 072a39cee. I converted the PR-added malformed-input tests from `pytest.raises(..., match=...)` to captured exceptions with exact `.message` / `.hint` equality checks. Focused validation: `uv run --python 3.12 -m pytest tests/unit/cli/storage_layout/test_storage_layout_overrides.py -q` and `uv run --python 3.12 -m pytest tests/functional/codegen/test_interfaces.py -k invalid_state_mutability -q`.

vyper/semantics/analysis/data_positions.py
banteg: Yes. `_fetch_path` starts with the full `StorageLayout`, but it reassigns `tmp = tmp[segment]` for each user-provided path segment. For malformed overrides, that nested value can be a non-dict, e.g. `{"a": 5}`. The `isinstance(tmp, dict)` guard is intentional: it keeps malformed user input on the `StorageLayoutException` path instead of letting a `TypeError`/internal error escape.

tests/unit/cli/storage_layout/test_storage_layout_overrides.py
```diff
@@ -472,14 +473,22 @@ def test_override_extra_entry():
...
             code,
             output_formats=["layout"],
             storage_layout_override=json_input(storage_layout_override),
         )
+    expected_layout = {"a": {"type": "uint256", "n_slots": 1, "slot": 0}}
+    assert exc_info.value.message == (
+        "Computed storage layout does not match override file!\n"
+        f"expected: {json.dumps(storage_layout_override)}\n\n"
+        f"got:\n{json.dumps(expected_layout)}"
+    )
```
Sporarum, line 486: I would even maybe put the json as text as well ?

vyper/semantics/analysis/data_positions.py
Sporarum: Oh I hadn't noticed that, maybe we can add a type hint to make it more visible ?
(Kinda surprised mypy is not already complaining actually)

vyper/semantics/analysis/data_positions.py
```diff
@@ -153,18 +157,23 @@ def _fetch_path(path: list[str], layout: StorageLayout, node: vy_ast.VyperNode):
...
             raise StorageLayoutException(
                 f"Could not find storage slot for {qualified_path}. "
                 "Have you used the correct storage layout file?",
                 node,
             )
         tmp = tmp[segment]
```
Sporarum, line 165: Maybe we can put the `isinstance` of `dict` here instead ?
Or do we want `tmp` to be a leaf value sometimes ?

tests/unit/cli/storage_layout/test_storage_layout_overrides.py
banteg: Done in 85e6f6f3a. I spelled out the expected JSON fragments as literal text in the exact-message tests, including the parametrized invalid-slot messages. Focused validation: `uv run --python 3.12 -m pytest tests/unit/cli/storage_layout/test_storage_layout_overrides.py -q`; formatting/whitespace checks are clean.

tests/functional/codegen/test_interfaces.py
charles-cooper: > match only checks the string is included, we want to check the exact string to prevent any regressions

i disagree btw but we don't have a very formal policy here. like i don't think we need error messages to be tested exactly, it's a style thing

tests/unit/cli/storage_layout/test_storage_layout_overrides.py
Sporarum: How does it look like with a multiline string ?

banteg:
Addressed the fresh layout feedback in df031a032. Changes: made the mixed traversal state explicit with `tmp: object`, documented that malformed overrides can leave a leaf value at that point, and added a multiline-string malformed slot case showing the JSON-escaped error text. Focused validation: `uv run --python 3.12 -m pytest tests/unit/cli/storage_layout/test_storage_layout_overrides.py -q` (24 passed); `git diff --check`; `uv run --python 3.12 black --check vyper/semantics/analysis/data_positions.py tests/unit/cli/storage_layout/test_storage_layout_overrides.py`.

Sporarum:
Does it/could it also solve https://github.com/vyperlang/vyper/issues/5120 ?

vyper/semantics/analysis/data_positions.py
Sporarum: @banteg Have you considered this ?

banteg: Addressed in d9563d205. I split the traversal guard so `tmp` must be a dict before checking membership or subscripting it; malformed intermediate leaf values now stay on the `StorageLayoutException` path, and the final leaf-value check still handles a resolved path without `slot`. Focused validation: `uv run --python 3.12 -m pytest tests/functional/codegen/test_interfaces.py::test_json_interface_event_without_indexed tests/unit/cli/storage_layout/test_storage_layout_overrides.py -q`; `uv run --python 3.12 black --check tests/functional/codegen/test_interfaces.py vyper/semantics/analysis/data_positions.py vyper/semantics/types/user.py`; `git diff --check`.

banteg:
Checked #5120 and folded the contained fix into this PR. In d9563d205, missing JSON ABI event `indexed` fields now default to `False`; in 2542a9d9, JSON ABI event field names such as `from`/`to` remain importable, matching the existing JSON ABI function-argument behavior where external ABI names are not necessarily Vyper identifiers. Added `test_json_interface_event_without_indexed` using the ERC20-style field names from the issue. Focused validation: `uv run --python 3.12 -m pytest tests/functional/codegen/test_interfaces.py::test_json_interface_event_without_indexed tests/unit/cli/storage_layout/test_storage_layout_overrides.py -q`; `uv run --python 3.12 black --check tests/functional/codegen/test_interfaces.py vyper/semantics/analysis/data_positions.py vyper/semantics/types/user.py`; `git diff --check`.

vyper/semantics/analysis/data_positions.py
Sporarum: My point was that the code is peppered with `isinstance(tmp, dict)` and `tmp` has type `object`.
Sometimes I point at a specific line to highlight a pattern that is present in other places as well, make sure to always consider that.

Sporarum:
I updated the PR description so that it auto-closes #5120

vyper/semantics/analysis/data_positions.py
banteg: Addressed in 128b6942c. I collapsed the repeated `isinstance(tmp, dict)` checks into `_ensure_layout_dict()`, so `_fetch_path()` has one explicit boundary for malformed leaf values and uses the returned mapping for membership/subscript access. This keeps the behavior unchanged while making the traversal pattern clearer across the function. Focused validation: `uv run --python 3.12 -m pytest tests/functional/codegen/test_interfaces.py::test_json_interface_event_without_indexed tests/unit/cli/storage_layout/test_storage_layout_overrides.py -q`; `uv run --python 3.12 black --check vyper/semantics/analysis/data_positions.py tests/functional/codegen/test_interfaces.py vyper/semantics/types/user.py`; `git diff --check`.

vyper/semantics/analysis/data_positions.py
```diff
@@ -152,26 +152,33 @@ def _reserve_slot(self, slot: int, var_name: str) -> None:
...
-        if segment not in tmp:
+        tmp_dict = _ensure_layout_dict(tmp, qualified_path, node)
+        if segment not in tmp_dict:
             raise StorageLayoutException(
                 f"Could not find storage slot for {qualified_path}. "
                 "Have you used the correct storage layout file?",
                 node,
             )
-        tmp = tmp[segment]
+        tmp = tmp_dict[segment]
```
Sporarum: If here you put:
```
tmp = tmp[segment]
if not isinstance(tmp, dict):
        raise StorageLayoutException(f"no storage slot for {qualified_path}", node)
```
It would also work, guarantee that `tmp` is always a dict, and allow you to remove the other checks, no ?

banteg: Done in `777db0fcb`: `_fetch_path()` now immediately normalizes each `tmp[segment]` back through `_ensure_layout_dict()`, so `tmp` remains dict-typed through traversal and the final slot check no longer needs a second normalized variable. Focused validation: `uv run --python 3.12 -m pytest tests/functional/codegen/test_interfaces.py::test_json_interface_event_without_indexed tests/unit/cli/storage_layout/test_storage_layout_overrides.py -q` (25 passed); `uv run --python 3.12 black --check ...`; `uv run --python 3.12 ruff check vyper/semantics/analysis/data_positions.py`; `git diff --check`.

vyper/semantics/analysis/data_positions.py
```diff
@@ -161,24 +161,21 @@ def _ensure_layout_dict(
...
-        tmp_dict = _ensure_layout_dict(tmp, qualified_path, node)
-        if segment not in tmp_dict:
+        if segment not in tmp:
             raise StorageLayoutException(
                 f"Could not find storage slot for {qualified_path}. "
                 "Have you used the correct storage layout file?",
                 node,
             )
-        tmp = tmp_dict[segment]
+        tmp = _ensure_layout_dict(tmp[segment], qualified_path, node)
```
Sporarum: Since this is the only use remaining(?), I think it should be inlined

banteg: Done in `0f146fd98`: inlined the dict guard at the only remaining use and removed `_ensure_layout_dict()`. Focused validation: `uv run --python 3.12 -m pytest tests/functional/codegen/test_interfaces.py::test_json_interface_event_without_indexed tests/unit/cli/storage_layout/test_storage_layout_overrides.py -q` (25 passed); `uv run --python 3.12 black --check ...`; `uv run --python 3.12 ruff check vyper/semantics/analysis/data_positions.py`; `uv run --python 3.12 python -m py_compile ...`; `git diff --check`.

vyper/semantics/types/user.py
```diff
@@ -232,10 +240,11 @@ def from_abi(cls, abi: dict) -> "EventT":
         Event object.
         """
         members: dict = {}
-        indexed: list = [i["indexed"] for i in abi["inputs"]]
+        indexed: list = [i.get("indexed", False) for i in abi["inputs"]]
         for item in abi["inputs"]:
             members[item["name"]] = type_from_abi(item)
-        return cls(abi["name"], members, indexed)
+        # JSON ABI argument names can be reserved Vyper identifiers, e.g. ERC20 `from`.
+        return cls(abi["name"], members, indexed, validate_member_names=False)
```
Sporarum: Not sure this is safe to do, not sure what we should do in this case.
@charles-cooper thoughts ?

My worry is something like this:
* abi has method `address` which gives the destination address of something
* vyper has method `address` on interfaces which gives the address of the interface instance
* one of the two is silently chosen by the compiler, and the user is never notified of the conflict

(`address` is the most illustrative example I could think of, but would work for any member we add to interfaces)

One way to solve this would be to scramble the reserved words like `address` into something like `__raw__address`.
Then even if an interface defined a `__raw__address`, it would instead be accessible as: `__raw____raw__address`

banteg: I checked the collision paths. The #5120 event-field change only affects `EventT.arguments`: those names are used for event kwargs/signature ordering and are not inserted into the interface instance member namespace, so they cannot shadow `some_interface.address`.

The concrete JSON ABI method-name example did expose an adjacent bug: a function named `address` already collided with the builtin interface member, but the error path raised `CompilerPanic` because `InterfaceT._id` was not set yet. I fixed that in banteg/vyper@bb9e8adea so it now raises a user-facing `NamespaceCollision`, and added a regression test. I left raw-name scrambling/support for such ABI methods out of this PR pending maintainer direction, since that would be a larger access-name design change.

Validation:
- `uv run --python 3.12 -m pytest tests/functional/codegen/test_interfaces.py::test_json_interface_event_without_indexed tests/functional/codegen/test_interfaces.py::test_json_interface_function_name_conflicts_with_builtin_member tests/unit/cli/storage_layout/test_storage_layout_overrides.py -q` (`26 passed`)
- `uv run --python 3.12 black --check vyper/semantics/types/module.py tests/functional/codegen/test_interfaces.py`
- `uv run --python 3.12 ruff check vyper/semantics/types/module.py tests/functional/codegen/test_interfaces.py`
- `uv run --python 3.12 python -m py_compile vyper/semantics/types/module.py tests/functional/codegen/test_interfaces.py`
- `git diff --check`

charles-cooper: > My worry is something like this:
> 
> abi has method address which gives the destination address of something
> vyper has method address on interfaces which gives the address of the interface instance
> one of the two is silently chosen by the compiler, and the user is never notified of the conflict

hm why would there be a silent choice -- don't we fail loudly here?

banteg: Yes, the current branch now fails loudly for method/member collisions.

The #5120 event-field path itself does not insert event argument names into the interface instance member namespace. For the adjacent JSON ABI function-name collision, banteg/vyper@bb9e8adea adds `test_json_interface_function_name_conflicts_with_builtin_member` and changes that path from `CompilerPanic` to a user-facing `NamespaceCollision`.

So I think this PR covers the loud-failure behavior; raw-name scrambling would be a separate access-name design change if we want imported ABI methods with reserved or builtin member names to stay callable.

banteg:
Also updated the branch in banteg/vyper@f4c5124d6 with a normal merge from current upstream `master` (no force push) to clear the dirty merge state.

The only semantic overlap Mergiraf resolved was in `InterfaceT.__init__`: the merge keeps this PR's earlier `_id` assignment before `super().__init__()` while carrying upstream's new `errors` member set into `members` / `_helper`.

Validation:

- `uv run --python 3.12 -m pytest tests/functional/codegen/test_interfaces.py::test_json_interface_event_without_indexed tests/functional/codegen/test_interfaces.py::test_json_interface_function_name_conflicts_with_builtin_member tests/unit/cli/storage_layout/test_storage_layout_overrides.py -q` (`26 passed`)
- `uv run --python 3.12 black --check vyper/semantics/types/module.py vyper/semantics/types/user.py tests/functional/codegen/test_interfaces.py tests/unit/cli/storage_layout/test_storage_layout_overrides.py`
- `uv run --python 3.12 ruff check vyper/semantics/types/module.py vyper/semantics/types/user.py tests/functional/codegen/test_interfaces.py tests/unit/cli/storage_layout/test_storage_layout_overrides.py`
- `uv run --python 3.12 python -m py_compile vyper/semantics/types/module.py vyper/semantics/types/user.py tests/functional/codegen/test_interfaces.py tests/unit/cli/storage_layout/test_storage_layout_overrides.py`
- `git diff --check origin/master...HEAD`

vyper/semantics/types/user.py
Sporarum: > hm why would there be a silent choice -- don't we fail loudly here?

Because the code explicitly does `validate_member_names=False`, which IIRC checks for collisions

banteg: Addressed in da844c211. I renamed the escape hatch to `validate_member_identifiers` so it no longer implies member collision checks are disabled, and JSON ABI event inputs now use the same `_abi_input_name` normalization as errors so duplicate ABI names do not silently overwrite earlier entries before the event type is built.

Validation:
- `uv run --python 3.12 -m pytest tests/functional/codegen/test_interfaces.py::test_json_interface_event_without_indexed tests/functional/codegen/test_interfaces.py::test_json_interface_event_duplicate_input_names tests/functional/codegen/test_interfaces.py::test_json_interface_function_name_conflicts_with_builtin_member tests/unit/cli/storage_layout/test_storage_layout_overrides.py -q`
- `uv run --python 3.12 black --check vyper/semantics/types/user.py tests/functional/codegen/test_interfaces.py tests/unit/cli/storage_layout/test_storage_layout_overrides.py`
- `uv run --python 3.12 ruff check vyper/semantics/types/user.py tests/functional/codegen/test_interfaces.py tests/unit/cli/storage_layout/test_storage_layout_overrides.py`
- `uv run --python 3.12 python -m py_compile vyper/semantics/types/user.py tests/functional/codegen/test_interfaces.py tests/unit/cli/storage_layout/test_storage_layout_overrides.py`
- `git diff --check`

Sporarum: @banteg I think it would have been better to wait for the discussion to have settled instead of adding more changes for us to review

banteg: Understood, sorry for moving ahead before the discussion had settled. I will pause further branch changes here and wait for maintainer direction before adding any more commits.

If `da844c211` ends up being more review noise than help, I can split or revert that follow-up once you decide the preferred direction.

charles-cooper: @banteg please revert the signature change, validate_member_names was fine

re the _abi_input_name normalization, is that invertible? if not, what are the consequences..?

vyper/semantics/analysis/data_positions.py
```diff
@@ -160,11 +163,18 @@ def _fetch_path(path: list[str], layout: StorageLayout, node: vy_ast.VyperNode):
...
+
+    if "slot" not in tmp:
+        raise StorageLayoutException(f"no storage slot for {qualified_path}", node)
 
-    try:
-        ret = tmp["slot"]
-    except KeyError as e:
-        raise StorageLayoutException(f"no storage slot for {qualified_path}", node) from e
+    ret = tmp["slot"]
+    # bool is an instance of int, disallow it explicitly
```
charles-cooper, line 173: oh man, that one almost got me

vyper/semantics/types/user.py
charles-cooper: oh never mind, i see why it was called validate_member_identifiers

vyper/semantics/types/user.py
```diff
@@ -29,8 +29,15 @@
 
 # user defined type
 class _UserType(VyperType):
-    def __init__(self, members=None):
-        super().__init__(members=members)
+    def __init__(self, members=None, *, validate_member_identifiers: bool = True):
+        if validate_member_identifiers:
+            super().__init__(members=members)
```
charles-cooper: shouldn't we just thread `validate_member_identifiers` to the base ctor instead of reaching into internal VyperType methods (`_check_add_member`)

banteg: Done in b1c9bccd3: VyperType.__init__ now takes validate_member_identifiers, and _UserType forwards the flag instead of reaching into the member internals itself. This still performs duplicate-member checks while skipping only identifier syntax validation for JSON ABI names. The _abi_input_name fallback is deterministic for internal member keys and is not meant to be inverted back into user-facing ABI names. Validation: focused JSON-interface/storage-layout tests passed; ruff passed.

vyper/semantics/types/user.py
```diff
@@ -292,10 +300,13 @@ def from_abi(cls, abi: dict) -> "EventT":
...
-        for item in abi["inputs"]:
-            members[item["name"]] = type_from_abi(item)
-        return cls(abi["name"], members, indexed)
+        indexed: list = [i.get("indexed", False) for i in abi["inputs"]]
+        for i, item in enumerate(abi["inputs"]):
+            name = _abi_input_name(item, i, members)
+            members[name] = type_from_abi(item)
+        # JSON ABI argument names can be reserved Vyper identifiers, e.g. ERC20 `from`.
+        # Skip identifier validation only; member additions still use the normal checks.
+        return cls(abi["name"], members, indexed, validate_member_identifiers=False)
```
charles-cooper: i'm still not sure that `validate_member_identifiers=False` is actually a good idea -- like you will be able to import the json but once you try to access the attribute (e.g. `foo(address)`) you will just get a panic or something? or am i missing something

banteg: I checked the event path on this branch. For JSON ABI events, the imported names are event argument keys, not interface-instance attributes, and they are consumed by `EventT.validate_call` when logging the event. A positional log with a reserved ABI name still compiles, and a keyword log with a non-reserved-but-builtin-ish name like `address` also compiles: `log Token.Foo(address=a)` succeeds. So I do not see a later panic from `validate_member_identifiers=False` on the event path. Names that are syntactically reserved in Vyper, like `from`, still cannot be written as kwargs, but can be used positionally; that is the same usability limitation this PR is trying to avoid for importing the ABI at all.

banteg: Revisited this in 747cd6d02. The bypass did not panic, but a reserved JSON event argument such as from created two concrete dead ends: the positional-call deprecation warning recommended unparseable from= syntax, and interface output emitted from: address and could not be compiled again. The shared _abi_input_name helper now maps invalid or reserved names to collision-safe _argN names, and the validate_member_identifiers bypass/plumbing is removed. Valid names such as address remain unchanged. Re-emitted cosmetic ABI argument names are sanitized, but the event selector is unaffected.

Validation: uv run --python 3.12 -m pytest tests/functional/codegen/test_interfaces.py -q (77 passed, 2 xfailed), plus focused Ruff, Black, and git diff --check.

charles-cooper:
@banteg some new merge conflicts, please check, thanks!

banteg:
Resolved the fresh merge conflicts in dd3258c28 with a normal merge of current `master` (`74d213fae`), preserving the PR history. The `EventT.from_abi` resolution keeps upstream’s missing-`inputs` fallback while retaining this PR’s missing-`indexed` default and collision-safe input-name normalization. I also removed the duplicate `_id` assignment produced by the automatic merge and added the narrow `FS003` suppression required for the deliberately literal JSON error text.

Validation:
- relevant storage-layout, semantic-type, and interface suites: `1496 passed, 9 xfailed` under legacy
- the same suites: `1496 passed, 9 xfailed` under experimental codegen
- focused storage-layout regression: `24 passed`
- `make lint`

tests/unit/cli/storage_layout/test_storage_layout_overrides.py
banteg: Covered in df031a032 and retained after the current-master merge in dd3258c28. The parametrized exact-message test includes a multiline slot value (`zero\none`) and asserts the escaped JSON spelling (`"zero\\none"`).

banteg:
Resolved the new master conflict in 710dde443, preserving `StructureException` and dropping the obsolete `CodegenPanic` import. The interface and storage-override suites pass: 102 passed, 1 expected failure. `make lint` passes.

vyper/semantics/analysis/data_positions.py
```diff
@@ -160,11 +163,18 @@ def _fetch_path(path: list[str], layout: StorageLayout, node: vy_ast.VyperNode):
                 node,
             )
         tmp = tmp[segment]
+        if not isinstance(tmp, dict):
+            raise StorageLayoutException(f"no storage slot for {qualified_path}", node)
```
Sporarum, line 167:
```suggestion
            raise StorageLayoutException(f"invalid storage slot for {qualified_path}: expected a dict, got a {type(tmp)}: {tmp}", node)
```

vyper/semantics/analysis/data_positions.py
```diff
@@ -160,11 +163,18 @@ def _fetch_path(path: list[str], layout: StorageLayout, node: vy_ast.VyperNode):
...
+    if "slot" not in tmp:
+        raise StorageLayoutException(f"no storage slot for {qualified_path}", node)
 
-    try:
-        ret = tmp["slot"]
-    except KeyError as e:
-        raise StorageLayoutException(f"no storage slot for {qualified_path}", node) from e
+    ret = tmp["slot"]
+    # bool is an instance of int, disallow it explicitly
+    if isinstance(ret, bool) or not isinstance(ret, int):
```
Sporarum, line 174:
```suggestion
    proper_int = isinstance(ret, int) and not isinstance(ret, bool)
    if not proper_int:
```

Sporarum (approved):
Other than that, LGTM

vyper/semantics/analysis/data_positions.py
banteg: Applied in a498a907b, including the qualified path, offending type and value. Updated the diagnostic regression; storage-layout and interface suites report 102 passed and 1 expected failure. make lint passes.

vyper/semantics/analysis/data_positions.py
banteg: Applied in a498a907b. The named proper_int check keeps bool excluded while accepting integer slots.

tests/unit/cli/storage_layout/test_storage_layout_overrides.py
```diff
@@ -551,7 +551,9 @@ def test_override_entry_not_dict():
             output_formats=["layout"],
             storage_layout_override=json_input(storage_layout_override),
         )
-    assert exc_info.value.message == "no storage slot for a"
+    assert exc_info.value.message == (
+        "invalid storage slot for a: expected a dict, got a <class 'int'>: 5"
```
Sporarum: The type should be pretty printed, in other words:
`got a int: 5` and not `got a <class 'int'>: 5`

banteg: Fixed in 14a2e0519 using type(tmp).__name__. The diagnostic now reads `expected a dict, got a int: 5`. All 24 storage-layout override tests and make lint pass.

tests/unit/cli/storage_layout/test_storage_layout_overrides.py
```diff
@@ -551,9 +551,7 @@ def test_override_entry_not_dict():
             output_formats=["layout"],
             storage_layout_override=json_input(storage_layout_override),
         )
-    assert exc_info.value.message == (
-        "invalid storage slot for a: expected a dict, got a <class 'int'>: 5"
-    )
+    assert exc_info.value.message == ("invalid storage slot for a: expected a dict, got a int: 5")
```
Sporarum, line 554: Parens no longer necessary

PR #5081, opened 2026-06-12
fix[lang]: fix off-by-one in storage allocator bounds check

banteg:
### What I did

Fixed an off-by-one in `SimpleAllocator.allocate_slot`: an allocation that exactly fills the available space was rejected, and the error message overstated the attempted range by one slot.

### How I did it

The allocator hands out slots `[_slot, _slot + n)`, so the last slot used is `_slot + n - 1` and the bound should be `_slot + n > _max_slot`, not `>=`. Concretely, on master:

- `X: immutable(uint256[768])` (exactly the `0x6000`-byte immutables limit) is rejected, while 767 words compiles;
- a storage array exactly filling slots `1 .. 2**256 - 1` is rejected with *"tried to allocate slots 1 through 115792…936"* — slot 2**256 was never part of the allocation.

With the fix, exact-fit allocations compile and the error message reports the true range (`_slot + n - 1`). One-slot-over allocations still fail, now with an accurate message.

### How to verify it

`test_allocator_exact_fit`, `test_allocator_overflow`, `test_immutables_exact_fit`, and `test_immutables_overflow` in `tests/unit/semantics/test_storage_slots.py`. The pre-existing `test_allocator_overflow` pinned the buggy boundary (its array exactly filled storage and was expected to fail); it is split into an exact-fit success case and a genuine overflow case.

### Commit message

```
the storage allocator hands out slots [_slot, _slot + n), so the last
slot used by an allocation is _slot + n - 1 -- but the bounds check
rejected _slot + n >= _max_slot, refusing allocations that exactly
fill the available space. concretely, immutables of exactly 0x6000
bytes (`uint256[768]`) and storage arrays exactly filling slots
1 through 2**256 - 1 were rejected. the error message also overstated
the attempted range by one slot.

relax the check to `>` and report the correct last slot in the error
message.
```

Fix: storage and immutable allocations that exactly fill the available space are no longer rejected with an off-by-one bounds error.

### Cute Animal Picture

cute animal

### DIFF
--- a/tests/unit/semantics/test_storage_slots.py
+++ b/tests/unit/semantics/test_storage_slots.py
@@ -1,5 +1,6 @@
 import pytest
 
+from vyper.compiler import compile_code
 from vyper.evm.opcodes import version_check
 from vyper.exceptions import StorageLayoutException
 
@@ -97,7 +98,7 @@ def test_reentrancy_lock(get_contract):
     assert c.h(0) == 123456789
 
 
-def test_allocator_overflow(get_contract):
+def test_allocator_exact_fit():
     # cancun allocates reeentrancy slot in transient storage,
     # so allocate an actual storage variable
     if version_check(begin="cancun"):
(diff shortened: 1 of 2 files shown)


charles-cooper (approved):
seems fine in principle but not merging just yet, i prefer being conservative on the storage allocator

banteg:
Rechecked the allocator-bound change and its four focused tests; all pass. The boundary checks remain small and appropriately placed. No further code changes from this pass.

Sporarum (approved):
I don't really know about storage slots, but LGTM

PR #5082, opened 2026-06-12
fix[venom]: make truthy or rule wrap-aware

### LINKED ISSUE #5072: Algebraic optimizer treats unwrapped zero literal as truthy in or
## Summary

The algebraic optimizer's truthy-position `or` rule checks raw literal truthiness instead of EVM-wrapped truthiness. A literal equal to `2**256` is zero as an EVM word, but `_rule_or` can treat it as nonzero and fold the expression to constant `1`.

## Reproducer

```venom
function main {
main:
    %x = source
    %y = or 115792089237316195423570985008687907853269984665640564039457584007913129639936, %x
    assert %y
    stop
}
```

Running `AlgebraicOptimizationPass` on current master produces:

```text
%x = source
%y = 1
assert %y
stop
```

But `115792089237316195423570985008687907853269984665640564039457584007913129639936 == 2**256`, which wraps to zero in EVM arithmetic. The expression is equivalent to `or 0, %x`, so in a truthy position it should still depend on `%x`.

## Root cause

In `vyper/venom/passes/algebraic_optimization.py`, `_rule_or` has a wrap-aware helper for zero and max-uint cases, but the truthy-position case uses the raw Python value:

```python
if is_truthy and self._is_lit(ops[0]) and ops[0].value != 0:
    self.updater.mk_assign(inst, IRLiteral(1))
```

That should use wrap-aware zero testing, e.g. `not lit_eq(ops[0], 0)` / `wrap256`, or rely on canonicalized literals as an enforced invariant before this pass.

## Impact

Latent IR-level correctness bug. Frontend-generated literals may normally be canonicalized, but the optimizer rule itself is not safe for valid 256-bit EVM literal semantics.

## Fix direction

Make the truthy-position `or` rule wrap-aware. A focused regression can parse the IR above, run `AlgebraicOptimizationPass`, and assert it does not fold `%y` to `1`.

banteg:
### What I did

Fixes #5072.

The algebraic optimizer's truthy-position `or` rule treated an unwrapped literal equal to `2**256` — which is zero as an EVM word — as nonzero, folding `or 2**256, %x` in an `assert` position to constant `1` instead of depending on `%x`.

An adjacent audit found the same raw-literal problem in `signextend` byte indexes: producer folding and no-op elimination compared unwrapped Python integers, so an index such as `2**256` was treated as at least 31 instead of byte index zero.

### How I did it

`_rule_or` now uses the existing wrap-aware `lit_eq` helper instead of raw Python truthiness. The adjacent audit adds `lit_word_value()` and uses it for both nested-`signextend` producer folding and local `signextend` elimination, so byte indexes are compared with 256-bit EVM word semantics.

### How to verify it

`test_or_truthy_unwrapped_zero_literal` parses the issue reproducer, runs `AlgebraicOptimizationPass`, and asserts the `or` reduces to `%x` rather than constant `1`. `test_signextend_unwrapped_zero_byte_index_not_eliminated` and `test_signextend_chain_uses_wrapped_byte_indexes` cover the adjacent byte-index rules. All three fail on master.

### Commit message

```
the truthy-position `or` rule checked raw literal truthiness
(`ops[0].value != 0`), so an unwrapped literal equal to 2**256 --
which wraps to zero as an EVM word -- was treated as nonzero and the
expression was folded to constant 1, even though it is equivalent to
`or 0, %x` and must depend on %x.

make the check wrap-aware via the existing `lit_eq` helper, consistent
with the `x | 0 -> x` rule below it.

the same raw-literal comparison affected `signextend` byte indexes in
both producer folding and local no-op elimination. normalize literal
indexes to EVM words before comparing them.
```

Fix: Venom algebraic optimization handles wrapped literals consistently in truthy `or` rules and `signextend` byte-index rewrites.

### Cute Animal Picture

cute animal


charles-cooper (approved):
are there other bugs of this class in algebraic opts?

banteg:
Audited the rest of `algebraic_optimization.py` for the same raw-literal vs wrapped-EVM-word class.

Found one adjacent issue in `signextend`: the byte index operand is also an EVM word, so `2**256` must be treated as byte index `0`, not as `>= 31`. I pushed follow-up commit `49cf212b4` to make the direct `signextend(n, x)` no-op rule and the nested `signextend(n, signextend(m, x))` rule use wrapped literal values, with regressions for both cases.

The remaining raw `.value` uses in the pass are either immediately normalized with `wrap256`/`lit_eq`, or are in the power-of-two strength reduction path after wrapped-zero/one cases have already been handled.

Checked with:

```
uv run --python 3.12 -m pytest tests/unit/compiler/venom/test_algebraic_binopt.py tests/unit/compiler/venom/test_algebraic_optimizer.py
uv run --python 3.12 ruff check vyper/venom/passes/algebraic_optimization.py tests/unit/compiler/venom/test_algebraic_binopt.py
```

banteg:
I did a quick sweep of the rest of `AlgebraicOptimizationPass` for the same raw-literal/wrap class. I do not see another confirmed correctness bug.

The remaining direct literal reads are in the power-of-two strength-reduction probe for `mul`/`div`/`mod`. The wrap-zero aliases are already caught by the preceding `lit_eq(..., 0)` rule; non-canonical aliases like a literal that wraps to `2` but is not syntactically a power of two would just miss the strength reduction rather than produce a wrong rewrite. The comparator and `signextend` paths now go through `wrap256`/`lit_word_value` where the literal value affects correctness.

HodanPlodky (approved):

banteg:
Review refresh: the code remains clean against current master (315 focused optimizer tests passed, 1 skipped; Ruff clean), and the adjacent raw-literal audit found no further unsafe comparisons. I updated the PR body so it now describes the already-pushed signextend byte-index fixes and their two regressions from 49cf212b4, rather than presenting the current branch as only the original one-line or-rule change.

HodanPlodky (approved):

PR #5083, opened 2026-06-12
fix[venom]: do not coalesce gas reads in CSE

### LINKED ISSUE #5040: Venom CSE coalesces distinct msg.gas reads
## Summary

The Venom CSE path treats `gas` as a pure/idempotent expression, so two `msg.gas` reads can be merged. `GAS` returns remaining gas, so two executions in one call are not equivalent: the later value should be lower after intervening execution.

## Reproducer

```vyper
@external
def f() -> (uint256, uint256):
    a: uint256 = msg.gas // 3
    b: uint256 = msg.gas // 3
    return a, b
```

Compiled with experimental codegen at `-O gas`, the optimized IR contains exactly one `gas` instruction and the contract returns `a == b`. Runtime semantics should perform two `GAS` reads, so `b < a` after the first read and intervening operations.

A focused checker:

```python
import vyper
from vyper.compiler.settings import Settings, OptimizationLevel

src = """
@external
def f() -> (uint256, uint256):
    a: uint256 = msg.gas // 3
    b: uint256 = msg.gas // 3
    return a, b
"""
out = vyper.compile_code(
    src,
    output_formats=["ir_runtime"],
    settings=Settings(experimental_codegen=True, optimize=OptimizationLevel.GAS),
)
assert str(out["ir_runtime"]).count(" gas") == 2
```

## Root cause

- `gas` has no read/write effect entry that prevents availability propagation.
- `gas` is not listed in `NONIDEMPOTENT_INSTRUCTIONS` in `vyper/venom/analysis/available_expression.py`.
- `gas` is treated as uninteresting in `common_subexpression_elimination.py`, alongside transaction-constant opcodes such as `timestamp`, `number`, and `chainid`.

## Impact

Gas-introspecting code can observe wrong values. Patterns such as `before = msg.gas; ...; spent = before - msg.gas` are the practical casualty.

`pc` should probably be audited under the same rule: it is also execution-position-dependent rather than idempotent.

## Fix direction

Treat `gas` as nonidempotent and exclude it from the available-expression set, the same way call-like nonidempotent operations are excluded.

banteg:
### What I did

Fixes #5040.

Venom CSE could coalesce expressions derived from two distinct `msg.gas` reads, e.g. `a: uint256 = msg.gas // 3` and `b: uint256 = msg.gas // 3` compiled to a single `GAS` read with `a == b`. `GAS` returns the remaining gas, which decreases between reads, so the two reads are not equivalent.

### How I did it

The merge happened indirectly: `gas` has no read/write effects, so each `gas` instruction entered the available-expressions set; the second `gas` resolved to the same `_Expression` instance, making the two derived `div` expressions compare equal. The fix classifies `gas` as non-idempotent in `available_expression.py` — the same treatment as `staticcall`, which is non-idempotent precisely *because* it depends on gas — so it never enters the available-expression set. Also removed `gas` from CSE's "uninteresting" list, where it was misclassified alongside transaction-constant opcodes like `timestamp` and `chainid`. (`msize`/`memtop` is already handled via the MEMORY effect.)

### How to verify it

`test_cse_gas_not_idempotent` (venom-IR-level: two `gas` reads feeding two `div`s, asserts no change) fails on master. The issue's source-level reproducer now keeps both `gas` reads. Full venom suite passes.

### Commit message

```
venom CSE treated `gas` as a transaction-constant opcode: it had no
effects and was not in NONIDEMPOTENT_INSTRUCTIONS, so two `msg.gas`
reads resolved to the same available expression and downstream
expressions derived from distinct reads could be coalesced. GAS
returns the remaining gas, which decreases as execution proceeds, so
distinct reads are not equivalent.

classify `gas` as non-idempotent in the available-expression analysis
(same treatment as `staticcall`, which is non-idempotent precisely
because it depends on gas), and remove it from CSE's uninteresting
transaction-constant list.
```

Fix: Venom CSE no longer merges distinct `msg.gas` reads.

### Cute Animal Picture

cute animal


charles-cooper:
this seems fine, @HodanPlodky wdyt?

charles-cooper (approved):
@Sporarum any thoughts here on semantics?

HodanPlodky (approved):

Sporarum:
@charles-cooper @harkal 
I think `msg.gas` should always have the value corresponding to the exact spot where it is in the source code, so this should be fixed, and furthermore we should not allow reordering around it

It's the kind of operation that really limits what optimization we can do, but it has to be that way to have good semantics
(aka "the user doesn't need to know how the compiler works to guess what will happen")

If we wanted to allow ourselves these kind of optimizations, we should find a different semantics for `msg.gas` which is amenable to them, for example "`msg.gas2` always refers to the amount of available gas at the start of the block" (but it feels like this would be less useful than the current semantics)

PR #5084, opened 2026-06-12
fix[venom]: register self in global analyses cache fallback

### LINKED ISSUE #5046: IRAnalysesCache fallback global cache does not register self
## Summary

`IRAnalysesCache._ensure_global_analyses_cache` has a fallback path that can create a parallel cache for `self.function` instead of registering `self` in the global cache.

The curated `run_passes_on` pipeline appears to avoid this because it builds the global cache from the same cache dict used by passes. But the fallback path exists for other analysis-cache uses, and there it can make invalidation and analysis requests diverge.

## Problem code

Current shape in `vyper/venom/analysis/analysis.py`:

```python
for fn in self.function.ctx.functions.values():
    if fn not in global_cache.function_analyses_caches:
        global_cache.function_analyses_caches[fn] = IRAnalysesCache(fn)
if self.function not in global_cache.function_analyses_caches:
    global_cache.function_analyses_caches[self.function] = self
```

The loop inserts a fresh `IRAnalysesCache(fn)` for every missing function, including `self.function`. That makes the trailing `if self.function not in ...` false, so `self` is never registered.

This contradicts the nearby `global_cache is None` branch, which correctly initializes the map with `self.function -> self`.

## Impact

On the fallback path, a global analysis can compute and store per-function analyses in a different `IRAnalysesCache` object than the one owned by the active pass. The active pass's `InstUpdater` and invalidations then do not necessarily invalidate the global-analysis cache entry that later consumers read.

That is a stale-analysis hazard. It is latent in the normal optimization pipeline but can affect direct analysis/pass use outside `run_passes_on`.

## Fix direction

Register `self` first, then fill in the rest:

```python
global_cache.function_analyses_caches[self.function] = self
for fn in self.function.ctx.functions.values():
    if fn not in global_cache.function_analyses_caches:
        global_cache.function_analyses_caches[fn] = IRAnalysesCache(fn)
```

A regression can construct an `IRAnalysesCache(fn)` without pre-seeding a global cache, request a global analysis, and assert that `global_cache.function_analyses_caches[fn] is ac` for the original cache.

banteg:
### What I did

Fixes #5046. Local passes and global analyses now share one canonical analysis cache per function. Previously a fallback registry could hold a different cache, so invalidating the active pass's DFG left global consumers reading stale results.

### How I did it

The global registry creates missing function caches when it is initialized. Each registered cache keeps its identity for the lifetime of that registry; later local consumers bind to it and delegate requests, invalidations and forced recomputations. There is no placeholder replacement state or inspection of another cache's analysis dictionary.

Consumers bind once per registry. A local cache hit only checks the registry identity, avoiding a scan of all functions. Replacing the registry between compiler phases causes consumers to rebind. Fresh global analyses fill caches for any newly added functions before running.

Semantic and post-lowering validators retain isolated caches so stale pass results cannot hide malformed IR after a missed invalidation.

### How to verify it

Nine focused regressions cover fallback registration, separate caches for separate functions, shared invalidation and forced refresh, preserved global invalidation hooks, isolated validation, registry replacement and newly added functions.

- Venom unit and functional codegen suites: 6,444 passed, 2 skipped, 24 expected failures.
- `make lint` passes.
- A bounded probe of 1,000 warm local cache hits with 1, 100 and 1,000 functions records zero registry function visits at every size. The previous PR head visited 1,000, 100,000 and 1,000,000 functions respectively. This measures registry traversal, not whole-compiler performance.

### Commit message

```text
Fallback global-analysis initialization can create a function cache
separate from the one used by the active pass. Invalidation then leaves
global consumers reading stale local analysis results.

Give each function a stable canonical cache within the global registry.
Bind local consumers once per registry and delegate requests,
invalidations and forced refreshes to the canonical cache. Rebind when
compiler phases replace the registry and register newly added functions
before fresh global analyses.

Use isolated caches in validators so stale pass results cannot hide
malformed IR after a missed invalidation.
```

Fix stale local analyses reached through the fallback global-analysis registry.

### Cute Animal Picture

cute animal


vyper/venom/analysis/analysis.py
```diff
@@ -78,11 +78,10 @@ def _ensure_global_analyses_cache(self) -> "IRGlobalAnalysesCache":
             self.function.ctx.global_analyses_cache = global_cache
             return global_cache
 
+        global_cache.function_analyses_caches[self.function] = self
```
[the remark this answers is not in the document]
banteg: Good catch — the unconditional overwrite traded one divergence hazard for another. Addressed in 33a528d:

`IRAnalysesCache` now tracks whether it was auto-created as a placeholder by `_ensure_global_analyses_cache`. The registration rule is: a requesting cache displaces a *placeholder* entry (or fills an absent one), but never a non-placeholder cache registered by a real consumer such as the pass pipeline.

"Register only when absent" alone wouldn't work here: both branches pre-fill caches for every function in the context, so a later real cache for one of those functions would find an entry already present (the placeholder) and the original #5046 bug — global analyses living in a parallel cache the active pass never invalidates — would persist. Distinguishing placeholders resolves both directions; `test_existing_authoritative_cache_not_displaced` pins the no-clobbering behavior alongside the original regression test.

tests/unit/compiler/venom/test_analyses_cache.py
```diff
@@ -0,0 +1,75 @@
...
+    ac = IRAnalysesCache(entry)
+    ac.request_analysis(FCGGlobalAnalysis)
+
+    global_cache = ctx.global_analyses_cache
+    assert global_cache is not None
+    # the cache which requested the global analysis must be the registered one
+    assert global_cache.function_analyses_caches[entry] is ac
+    # other functions get fresh caches
+    for fn in ctx.functions.values():
+        assert fn in global_cache.function_analyses_caches
```
harkal:
```suggestion
        assert global_cache.function_analyses_caches[fn] is not ac
```

tests/unit/compiler/venom/test_analyses_cache.py
```diff
@@ -0,0 +1,75 @@
+from tests.venom_utils import parse_venom
+from vyper.venom.analysis import IRAnalysesCache
+from vyper.venom.analysis.fcg import FCGGlobalAnalysis
+from vyper.venom.basicblock import IRLabel
+
+SRC = """
+function entry {
+entry:
+    %1 = invoke @f
```
harkal:
```suggestion
    invoke @f
```

harkal (changes requested):
I would add a _placeholders set in the cache and drop both _auto_created and _placeholder. THe bit "is this entry a placeholder?" describes the registration not the cache, and it would be nicer if IRAnalysesCache didn't have to access another's instance's private attribute to make a decision that belongs to the registry.

banteg:
Addressed the cache ownership and test feedback in 2a6dc3960. Placeholder state now lives in the global registry; the invoke fixture is void; tests explicitly assert that different functions have distinct caches. I also found and fixed a lifecycle hole: a superseded temporary cache could keep serving a stale DFG after the active cache invalidated it. Requests, invalidations and forced refreshes now delegate to the canonical cache. The new mutation/invalidation regression, all 1,990 Venom unit tests, and 4,450 functional codegen tests pass (plus expected skips/xfails). `make lint` passes.

banteg:
Addressed both new reports in 89117d8f0, with reproductions that failed on the previous pushed head:

- A populated placeholder now stays canonical, preserving CFG/MustHalt/global stack-safety invalidation hooks. The regression populates `StackCleanupSafety`, replaces a return with `stop`, invalidates through a later consumer, and verifies a fresh safety result.
- Semantic and post-lowering validators use isolated local caches. The regression deliberately removes a definition without invalidating the pass cache and verifies that validation still reports `VarNotDefined`.

All 1,992 Venom unit tests and 4,450 functional codegen tests pass, alongside the existing skips/xfails. `make lint` passes. The description is updated to include these lifecycle constraints.

harkal:
lgtm just one nit before @charles-cooper sees it :D analysis.py:187 uses truthy, prefer len() == 0 (or a has_analyses() accessor so the registry doesn't peek into the cache's dict).

banteg:
Updated in 37d11c9db. Reviewing the ownership model showed that the placeholder replacement path could be removed entirely: the registry now gives each function a stable canonical cache, and later consumers bind to it once per registry. This removes the truthiness check and all inspection of another cache's analysis dictionary. Local cache hits no longer scan the function registry. Registry replacement and newly added functions have regressions, and the invalidation-hook and isolated-validator regressions are retained.

The combined Venom unit and functional codegen suites report 6,444 passed, 2 skipped and 24 expected failures; make lint passes. A bounded 1,000-hit probe records zero registry function visits with 1, 100 or 1,000 functions (the previous head visited 1,000, 100,000 and 1,000,000 respectively). The PR description now describes the final ownership model.

PR #5085, opened 2026-06-12
fix[venom]: check operand position in mem2var promotion

### LINKED ISSUE #5070: Mem2Var promotes allocas used as stored pointer values
## Summary

`Mem2Var` decides whether an `alloca` is promotable by looking only at the opcodes of its DFG uses. It does not check which operand position uses the alloca.

That means an alloca used as the **value** being stored by `mstore` can be mistaken for a local memory-slot access and promoted to a stack variable, changing the stored pointer value.

## Reproducer

```venom
function main {
main:
    %a = alloca 32
    %b = alloca 32
    mstore %b, %a
    %x = mload %a
    return %b, 32
}
```

Running `MakeSSA` then `Mem2Var` on current master produces:

```venom
function main {
  main:
      %a = alloca 32
      %b = alloca 32
      %alloca_a_0 = %a
      %x = %alloca_a_0
      return %b, 32
}
```

The original `mstore %b, %a` is gone. `%a` was not only being used as a memory slot; it was the pointer value being stored into `%b`.

## Root cause

`vyper/venom/passes/mem2var.py` currently accepts an alloca when all uses have opcodes in `mstore`, `mload`, or `return`:

```python
if not all2(inst.opcode in ["mstore", "mload", "return"] for inst in uses):
    return
```

For `mstore`, the alloca is only a promotable memory-slot use when it appears in the address operand position. A use in the stored-value position is an escape/value use and should block promotion.

## Impact

IR-level correctness bug. Current frontend emission may avoid storing raw alloca pointer values in this shape, but the pass invariant is unsound and future Venom IR shapes can trigger it.

## Related

PR #4841 is already touching `Mem2Var`, but it addresses uninitialized memory reads. This is a separate operand-position/escape check.

## Fix direction

Require all promotable uses to be in valid memory-address positions:

- `mload`: pointer operand
- `mstore`: destination/address operand only, not stored value
- `return`: returned buffer pointer operand only

If an alloca is used as an ordinary value, operand to pointer arithmetic that escapes, or non-address operand, leave it in memory.

banteg:
### What I did

Fixes #5070.

`Mem2Var` decided whether an alloca is promotable by looking only at the *opcodes* of its uses, not the operand position. An alloca used as the **value** stored by an `mstore` (i.e. a pointer escaping into memory, `mstore %b, %a`) was mistaken for a memory-slot access and promoted, silently deleting the store of the pointer value.

### How I did it

Added an `_is_pointer_use` check requiring every use to place the alloca in the memory-address operand position: `mload` (single operand, always the address) is fine; for `mstore` (operands `[value, ptr]`) and `return` (operands `[size, ptr]`), the alloca must not appear in `operands[0]`. This also blocks the self-referential `mstore %a, %a` shape. Operand-order semantics confirmed against `memory_location.py` and `load_elimination.py` (venom operands are reversed relative to assembly).

### How to verify it

New `tests/unit/compiler/venom/test_mem2var.py`: the issue reproducer asserts `mstore %b, %a` survives MakeSSA+Mem2Var (fails on master, where the store is deleted), plus a positive case asserting ordinary promotion still happens. Full `tests/unit/compiler/` passes.

### Commit message

```
Mem2Var decided promotability by checking only the opcodes of an
alloca's uses (`mstore`/`mload`/`return`). an alloca appearing as the
*value* operand of an mstore -- i.e. a pointer being stored into
another memory location -- was therefore treated as a promotable
memory-slot access, and promotion deleted the store of the pointer
value.

require every use to place the alloca in the memory-address operand
position (mstore/return operands are [value/size, ptr]); a use in the
value/size position is an escape and blocks promotion.
```

Fix: Venom `Mem2Var` no longer promotes allocas whose pointer value escapes as the stored value of an `mstore`.

### Cute Animal Picture

cute animal


vyper/venom/passes/mem2var.py
```diff
@@ -51,7 +66,7 @@ def _process_alloca_var(self, dfg: DFGAnalysis, alloca_inst: IRInstruction, var:
 
         uses = dfg.get_uses(alloca_inst.output)
 
-        if not all2(inst.opcode in ["mstore", "mload", "return"] for inst in uses):
+        if not all2(self._is_pointer_use(inst, alloca_inst.output) for inst in uses):
             return
```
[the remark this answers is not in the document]
banteg: Verified empirically — the order-swapped shape is handled correctly. With `%b` visited first:

```
%b = alloca 32
%a = alloca 32
mstore %b, %a
%x = mload %a
%y = mload %b
sink %x, %y
```

promoting `%b` rewrites the mstore to `%alloca_b_0 = %a` via `InstUpdater`, which maintains the DFG — so when `%a` is examined, its uses are `[assign, mload]`, and the `assign` opcode fails `_is_pointer_use`, blocking promotion. Output after MakeSSA+Mem2Var:

```
%b = alloca 32
%a = alloca 32
%alloca_b_0 = %a
%x = mload %a
%y = %alloca_b_0
sink %x, %y
```

`%a` survives as a real allocation with a real load, and the pointer *value* flows into `%b`'s replacement variable — which is exactly the promotion semantics for `%b` (its slot held the pointer value). Note the escape check exists to prevent the pointer *bytes* from being silently dropped, not to prevent the destination slot from being promoted; once the destination is a stack variable, the stored pointer value is preserved in SSA.

Added `test_mem2var_stored_pointer_dest_promoted_first` (722a274) pinning this shape.

vyper/venom/passes/mem2var.py
```diff
@@ -35,11 +35,26 @@ def _mk_varname(self, varname: str):
         self.var_name_count += 1
         return varname
 
+    @staticmethod
+    def _is_pointer_use(inst: IRInstruction, ptr: IRVariable) -> bool:
```
charles-cooper, line 39: the usage here is fine, add a comment that there is slight conceptual duplication with memory_location.py (memory_read_ops, memory_write_ops)

charles-cooper (approved):
lgtm, paging @harkal and @HodanPlodky for review

vyper/venom/passes/mem2var.py
banteg: Done in 7bc55b055: added the note about the slight conceptual overlap with `memory_location.py`s `memory_read_ops`/`memory_write_ops`, while keeping the local non-address operand rejection explicit. Validated with `uv run --python 3.12 -m pytest tests/unit/compiler/venom/test_mem2var.py -q`, black check, and `git diff --check`.

HodanPlodky (approved):

PR #5086, opened 2026-06-12
fix[venom]: resolve chained label replacements transitively

### LINKED ISSUE #5043: SimplifyCFG label replacement can leave dangling data-section labels
## Summary

`SimplifyCFGPass` can leave dangling labels in data-section references when it schedules a chain of label replacements in one pass epoch.

The function CFG itself is patched eagerly during `_merge_jump`, but deferred label references in data sections go through `_replace_all_labels`, which applies the label map only once and does not resolve replacements transitively.

## Problem shape

```venom
function main {
main:
    jnz %c, @b1, @other
b1:
    jmp @b2
b2:
    jmp @b3
b3:
    ...
other:
    ...
}

data readonly {
    dbsection jumptable:
        db @b1
}
```

If `_merge_jump` schedules both `b1 -> b2` and `b2 -> b3` during the same epoch, the blocks `b1` and `b2` can be removed. CFG operands are updated as the merge happens, so the function body points at `b3`. But the data-section reference `@b1` is rewritten only one step, to `@b2`, leaving a reference to a removed block.

## Root cause

`vyper/venom/passes/simplify_cfg.py:_schedule_label_replacement` records direct replacements, and `base_pass.py:_replace_all_labels` applies the map non-transitively. A replacement map like this:

```python
{b1: b2, b2: b3}
```

rewrites a deferred `@b1` reference to `@b2`, not `@b3`.

`tail_merge.py` shares the replacement machinery, but it does not remove the same kind of intermediate keepers in the same run, so the observed hazard is in `SimplifyCFGPass`.

## Impact

At best, this produces an assembly-time failure due to an unresolvable symbol. If labels are later reused or otherwise collide, it could also point a data table entry at the wrong target.

This appears to need a jump-table/data-section reference into a merge chain, so it is more likely with direct Venom or future codegen patterns than with current ordinary frontend output.

## Fix direction

Resolve label replacements transitively before recording or applying them:

- when adding `old -> new`, first follow any existing `new -> final` mapping;
- update any existing map values equal to `old` or `new` so all scheduled replacements point at the final live label;
- add a regression with a data-section `db @b1` reference where both `b1` and `b2` are merged away in the same `SimplifyCFGPass` run.

banteg:
### What I did

Fixes #5043.

`SimplifyCFGPass` can schedule a *chain* of label replacements in one pass epoch (e.g. `{b1: b2, b2: b3}` when merging `b1 -> b2 -> b3`). The function CFG is patched eagerly, but deferred label references in data sections go through `_replace_all_labels`, which applied the map only once — rewriting a jumptable's `@b1` to the now-removed `@b2`.

### How I did it

Path-compress the replacement map (`{a: b, b: c}` → `{a: c, b: c}`) at the top of the shared `_replace_all_labels`, with a visited-set cycle guard. Placing the resolution in the shared helper covers both `SimplifyCFGPass` and `TailMergePass`; for tail-merge it is a no-op since keeper labels never appear as map keys.

### How to verify it

`test_data_section_label_chain` parses the issue's shape (jump chain merged in one epoch + a `data readonly` section with `db @b1`), runs the pass, and asserts every data-section label refers to a surviving block. Fails on master with a dangling `@b2` reference.

### Commit message

```
`SimplifyCFGPass` patches CFG operands eagerly while merging jump
chains, but deferred label references in data sections are rewritten
afterwards via `_replace_all_labels`, which applied the replacement
map only once. when a chain of merges is scheduled in one pass epoch
(map like {b1: b2, b2: b3}), a data-section reference to @b1 was
rewritten one step to @b2 -- a removed block -- leaving a dangling
label.

path-compress the label map (with a cycle guard) before applying it,
so every deferred reference resolves to its final surviving block.
the resolution lives in the shared `_replace_all_labels`, covering
tail_merge as well, where it is a no-op since keeper labels are never
map keys.
```

Fix: Venom `SimplifyCFG` no longer leaves dangling data-section label references when merging chains of jumps in a single pass.

### Cute Animal Picture

cute animal


charles-cooper (approved):
lgtm

charles-cooper:
paging @HodanPlodky for review

HodanPlodky (approved):

PR #5087, opened 2026-06-12
fix[venom]: do not elide self-overlapping mcopy

### LINKED ISSUE #5041: MemoryCopyElision treats self-overlapping mcopy as idempotent
## Summary

`MemoryCopyElisionPass` records every `mcopy` as a reusable copy fact, including partially self-overlapping copies. That is unsound because EVM `MCOPY` has memmove semantics: when source and destination ranges overlap, the copy can clobber bytes that are still needed as source bytes.

A self-overlapping `mcopy` is not idempotent, and after it executes the destination is not necessarily a byte-copy of the original source.

## Pass-level reproducer

```python
from vyper.venom.parser import parse_venom
from vyper.venom.analysis.analysis import IRAnalysesCache
from vyper.venom.passes.memory_copy_elision import MemoryCopyElisionPass

src = """
function main {
main:
    %buf = alloca 256
    %src = add %buf, 0
    %dst = add %buf, 32
    mcopy %dst, %src, 64
    mcopy %dst, %src, 64
    %p = add %buf, 64
    %1 = mload %p
    mstore %buf, %1
    return %buf, 32
}
"""
ctx = parse_venom(src)
fn = next(iter(ctx.functions.values()))
ac = IRAnalysesCache(fn)
MemoryCopyElisionPass(ac, fn).run_pass()
print(fn)
```

The second `mcopy %dst, %src, 64` is rewritten to `nop`, but it is not a runtime no-op. Correct execution of the second overlapping copy changes bytes in the tail of the copied region; the optimized version leaves bytes from the first smeared copy.

## Root cause

`MemoryCopyElisionPass` registers the copy fact unconditionally:

```python
self.copies[write_loc] = inst
```

Invalidation checks whether later writes clobber a copy source, but it does not invalidate a copy whose own write aliases its own read.

This affects both consumers:

- `_try_elide_redundant_copy` can remove a repeated overlapping copy.
- `_try_elide_copy` can forward a later copy through an earlier overlapping copy and read bytes that were already changed by the first copy.

## Reachability

Current frontend codegen appears to mostly emit identical or disjoint aggregate copies, so this is easiest to trigger with direct Venom input today. There is no IR rule enforcing non-overlap, though, so future codegen or pass changes can expose this silently.

Related but not duplicate: #4851 and #4976 touch `MemoryCopyElisionPass`; #4899 is another copy-forwarding aliasing issue. None of those report this exact self-overlap/idempotence bug.

## Fix direction

When registering or consuming an `mcopy` fact, skip it if the read and write locations may alias, unless the source and destination regions are exactly equal.

banteg:
### What I did

Fixes #5041.

`MemoryCopyElisionPass` recorded every fixed-location `mcopy` as a reusable copy fact, including copies whose source and destination ranges overlap. EVM `MCOPY` has memmove semantics: a self-overlapping copy clobbers bytes that are still needed as source bytes, so it is not idempotent — repeating it produces different memory. Eliding the repeat (issue reproducer: two identical `mcopy %dst, %src, 64` with 32-byte overlap, second one rewritten to `nop`) changed runtime memory contents.

### How I did it

The pass now also fetches the mcopy's read location and only registers the copy fact when the write location is fixed and `MemoryLocation.may_overlap(read_loc, write_loc)` is false — unknown offsets conservatively count as overlapping. If `_try_elide_copy` already rewrote the mcopy to e.g. `calldatacopy`, the MEMORY read location is empty and registration is unaffected (a non-memory source cannot self-overlap).

### How to verify it

Two new tests (literal-offset and alloca-relative reproducers) assert no change is made for overlapping repeated copies; both fail on master. The existing `test_repeated_mcopy_elision_with_intermediate_reads` covers the still-allowed non-overlapping case. Full venom suite passes.

### Commit message

```
MemoryCopyElisionPass recorded every fixed-location mcopy as a
reusable copy fact. EVM MCOPY has memmove semantics: when source and
destination ranges overlap, the copy clobbers bytes still needed as
source bytes, so a self-overlapping mcopy is not idempotent -- after
it executes, the destination is not a byte-copy of the original
source. recording it as a copy fact allowed a repeated overlapping
copy to be elided as a no-op, changing runtime memory contents.

skip registering the copy fact when the mcopy's memory read location
may overlap its write location, treating unknown offsets
conservatively as overlapping.
```

Fix: Venom memory copy elision no longer treats self-overlapping `mcopy` as idempotent.

### Cute Animal Picture

cute animal


charles-cooper (approved):
lgtm, paging @HodanPlodky and @harkal for review

HodanPlodky (approved):

PR #5088, opened 2026-06-12
fix[venom]: fix bad uses in dload/mstore merge

### LINKED ISSUE #5042: MemMerge dload/mstore merge ignores which mstore operand uses the dload
## Summary

`MemMergePass._merge_mstore_dload` assumes that when a `dload` result is used by an `mstore`, the `dload` value is the value being stored. It does not consistently check that the use is in the value operand rather than the destination/pointer operand.

That makes the dload/mstore-to-`dloadbytes` rewrite unsound for cases where the loaded word is used as the store destination, or where it appears in both destination and value positions.

## Problem shape

A minimal bad shape is:

```venom
%d = dload 5
mstore %d, 42
```

The only use of `%d` is an `mstore`, but it is the destination pointer, not the value that should be copied from transient/data memory. The single-use path currently classifies this as a simple dload/mstore merge and rewrites the instruction to `dloadbytes`, losing the original stored value `42` and/or the dynamic destination role.

There is a second multi-use variant:

```venom
%d = dload 5
mstore %d, %d
sink %d
```

The multi-use path checks one operand role before rewriting, but does not reject the case where the destination operand also depends on the same `dload` output. `move_uses(dload_out, mload)` can then rewrite operands of the newly inserted `dloadbytes` / replacement `mload` and produce malformed or self-referential IR.

## Root cause

In `vyper/venom/passes/memmerging.py:_merge_mstore_dload`:

- the single-use path only checks `mstore.opcode == "mstore"` before rewriting;
- the multi-use path checks that the value operand is the `dload` output, but does not also ensure the destination operand is independent of that output before moving uses.

The existing `test_merge_mstore_dload_disallowed` covers a multi-use disallowed case, but not the single-use destination-only case.

## Impact

This is currently caught in some paths by assertions or later malformed-IR behavior, but with assertions disabled it can become a wrong rewrite: a store of an arbitrary value to a dynamic destination can turn into a copy of a dload word to the wrong destination.

## Fix direction

Before merging, explicitly validate the operand roles after parser/repr normalization:

- the `dload` output must be the `mstore` value operand;
- the `mstore` destination operand must not be the same `dload` output and should not otherwise depend on the value being moved;
- if either condition fails, skip the optimization.

A regression should include the single-use destination-only shape, because that is not covered by the current disallowed test.

banteg:
### What I did

Fixes #5042.

`MemMergePass._merge_mstore_dload` assumed that when a `dload` result is consumed by an `mstore`, it is always the *value* being stored. When the dload output was actually the destination pointer (`mstore %d, 42`) — or both operands (`mstore %d, %d`) — the pass either rewrote the store to `dloadbytes`, dropping the stored value, and then crashed in `updater.nop` (the dload was still used), or silently emitted self-referential IR via `move_uses` (`%1 = mload %1`).

### How I did it

Both paths now validate operand roles (venom `mstore` operands are `[value, ptr]`): the single-use path unpacks the operands and skips unless the dload output is the stored value and is not the destination; the multi-use path's existing value-operand check is extended to also reject a destination that depends on the dload output.

### How to verify it

Three new tests next to `test_merge_mstore_dload_disallowed` cover destination-only, self-referential single-use, and self-referential multi-use shapes, all asserting no change. On master they fail with the assertion crash / malformed-IR mismatch described above. Full venom suite passes.

### Commit message

```
`MemMergePass._merge_mstore_dload` assumed the dload output is always
the value operand of the consuming mstore. when it was actually the
destination pointer (or both operands), the single-use path rewrote
the mstore to dloadbytes -- dropping the stored value -- and then
crashed in `updater.nop` because the dload still had uses; the
multi-use path could emit self-referential IR via `move_uses`.

validate operand roles in both paths: the dload output must be the
stored value and must not appear as the destination pointer.
```

Fix: Venom memory merging validates operand roles before rewriting `dload`/`mstore` pairs to `dloadbytes`, fixing crashes and malformed IR when the loaded word is used as a store destination.

### Cute Animal Picture

cute animal

### DIFF
--- a/tests/unit/compiler/venom/test_memmerging.py
+++ b/tests/unit/compiler/venom/test_memmerging.py
@@ -1687,3 +1687,54 @@ def test_merge_mstore_dload_disallowed():
     """
 
     _check_no_change(pre)
+
+
+def test_merge_mstore_dload_dst_disallowed():
+    """
+    Test that the dload/mstore merge does not fire when the only use of
+    the dload output is as the *destination* pointer of the mstore
+    (rather than the stored value).
+    """
+    pre = """
+    _global:
+        %par = source
+        %d = dload %par
+        mstore %d, 42
+        stop
+    """
+
+    _check_no_change(pre)
+
+
+def test_merge_mstore_dload_self_dst_disallowed():
+    """
+    Test that the dload/mstore merge does not fire when the dload output
+    is used both as the stored value and as the destination pointer of
+    the mstore (single-use path).
+    """
+    pre = """
+    _global:
+        %par = source
+        %d = dload %par
+        mstore %d, %d
+        stop
+    """
+
+    _check_no_change(pre)
+
+
+def test_merge_mstore_dload_self_dst_more_uses_disallowed():
+    """
+    Test that the dload/mstore merge does not fire when the dload output
+    is used both as the stored value and as the destination pointer of
+    the mstore (multi-use path).
+    """
+    pre = """
+    _global:
+        %par = source
+        %d = dload %par
+        mstore %d, %d
+        sink %d
+    """
+
+    _check_no_change(pre)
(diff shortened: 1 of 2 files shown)


charles-cooper (approved):
lgtm, paging @HodanPlodky for addl review

HodanPlodky (approved):

PR #5090, opened 2026-06-12
fix[venom]: assert staticcall in constant context

### LINKED ISSUE #5060: Venom: external calls in constant context lack staticcall sanity assert
### Version Information

* OS: macOS

### What's your issue about?

Legacy external-call lowering contains a codegen-layer sanity check that a constant/view context can only emit `staticcall`:

```python
if context.is_constant():
    assert use_staticcall, "typechecker missed this"
```

The Venom external-call path derives `use_staticcall` from the target function mutability but does not assert this against the enclosing codegen context. Today this should be masked by semantic analysis, but if the frontend ever misses a state-mutating external call from a constant context, Venom would emit a `call` instead of failing loudly in codegen.

This is a defense-in-depth parity issue rather than a currently reachable source-level bug.

### How can it be fixed?

Add the same `context.is_constant()` / `use_staticcall` assertion to Venom external-call lowering.

banteg:
### What I did

Fixes #5060.

Legacy external-call lowering carries a codegen-layer sanity assert that a constant (view/pure, or range-expression) context can only emit `STATICCALL` (`vyper/codegen/external_call.py`: `assert use_staticcall, "typechecker missed this"`). The Venom external-call lowering derived `use_staticcall` from the target's mutability but never cross-checked it against the enclosing context's constancy — a frontend miss would silently emit `CALL` instead of failing loudly in codegen.

### How I did it

Added the equivalent two-line assert in `Expr._lower_external_call`, immediately after `use_staticcall` is derived and before the CALL/STATICCALL dispatch. `VenomCodegenContext.is_constant()` mirrors legacy `Context.is_constant()` semantics (per-function constancy from view/pure mutability, plus forced constancy inside range expressions). The assert message matches legacy verbatim for greppability.

### How to verify it

The assert is unreachable by construction (semantic analysis rejects state-mutating external calls in constant contexts), matching the untested legacy assert — this is defense-in-depth parity. Suites pass: `tests/unit/compiler/`, `tests/functional/codegen/calling_convention/` and `test_interfaces.py` under `--experimental-codegen`.

### Commit message

```
legacy external-call lowering asserts at codegen time that a constant
(view/pure or range-expression) context only emits staticcall. the
venom lowering derived `use_staticcall` from the target's mutability
but never cross-checked it against the enclosing context's constancy,
so a frontend miss would silently emit CALL instead of failing loudly.

add the equivalent assert to the venom external-call lowering,
matching the legacy message.

the assert is unreachable by construction; this is defense-in-depth
parity.
```

Fix: Venom external-call lowering asserts that constant contexts only emit `STATICCALL`, matching legacy defense-in-depth.

### Cute Animal Picture

cute animal


vyper/codegen_venom/expr.py
```diff
@@ -1793,6 +1793,8 @@ def _lower_external_call(self) -> VyperValue:
 
         # === Dispatch CALL or STATICCALL ===
         use_staticcall = fn_type.mutability in (StateMutability.VIEW, StateMutability.PURE)
+        if self.ctx.is_constant():
+            assert use_staticcall, "typechecker missed this"
```
Sporarum: We tend to avoid reason strings (unless they add some runtime debug information)

Instead prefer a comment before the assert

banteg: Addressed in a1b81206f. I moved the explanation into a source comment before the assertion and left the assert itself bare. Focused validation: `uv run --python 3.12 -m pytest tests/functional/codegen/calling_convention/test_external_contract_calls.py::test_constant_external_contract_call_cannot_change_state -q`; `uv run --python 3.12 black --check vyper/codegen_venom/expr.py`; `git diff --check`.

Sporarum (approved):

charles-cooper (approved):

vyper/codegen_venom/expr.py
```diff
@@ -1793,6 +1793,9 @@ def _lower_external_call(self) -> VyperValue:
 
         # === Dispatch CALL or STATICCALL ===
         use_staticcall = fn_type.mutability in (StateMutability.VIEW, StateMutability.PURE)
+        if self.ctx.is_constant():
+            # Non-static calls in constant context should be rejected by the typechecker.
+            assert use_staticcall
```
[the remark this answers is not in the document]
charles-cooper: lame, stop being a pedant

PR #5091, opened 2026-06-12
fix[venom]: track multi-output phi origins by variable

### LINKED ISSUE #5039: Venom PhiElimination collapses phi over different outputs of one invoke
## Summary

`PhiEliminationPass` can treat a phi over two different outputs of the same multi-output `invoke` as trivial because `_get_phi_origins_r` keys origins by producing instruction, not by produced variable.

For this shape:

```venom
%r1, %r2 = invoke @f, ...
%z = phi @t, %r1, @e, %r2
```

both operands trace to the same invoke instruction. `_process_phi` then sees a single origin and tries to assign from `src.output`. In normal Python this raises because the invoke has multiple outputs; under `python -O` it would silently pick the first output and miscompile the `%r2` path.

## Reproducer

This reproduces from plain Vyper with experimental codegen and `-O gas`. The callee needs to stay uninlined, and there need to be two call sites so the `invoke` survives.

```python
import vyper
from vyper.compiler.settings import Settings, OptimizationLevel

body = "\n".join(f"    x{i}: uint256 = p + {i}" for i in range(40))
src = f"""
@internal
def two(p: uint256) -> (uint256, uint256):
{body}
    return p + x0, p * 2

@external
def f(c: bool, p: uint256) -> uint256:
    a: uint256 = 0
    b: uint256 = 0
    a, b = self.two(p)
    x: uint256 = b
    if c:
        x = a
    return x

@external
def g(p: uint256) -> uint256:
    a: uint256 = 0
    b: uint256 = 0
    a, b = self.two(p)
    return a + b
"""
vyper.compile_code(src, settings=Settings(experimental_codegen=True, optimize=OptimizationLevel.GAS))
```

Observed failure:

```text
AssertionError: expected single output for %24, %25 = invoke @"internal 0 two(uint256)_runtime", %19
```

## Root cause

`vyper/venom/passes/phi_elimination.py:_get_phi_origins_r` returns origins by instruction. For multi-output producers, instruction identity is not enough to distinguish `%r1` from `%r2`.

## Fix direction

Key phi origins by `(producer_instruction, produced_variable)`, or conservatively bail out of `_process_phi` when the apparent single source instruction has `num_outputs != 1`.

banteg:
### What I did

Fixes #5039 by making `PhiEliminationPass` distinguish outputs of the same multi-output producer.

Current master already avoids the original #5039 crash and `python -O` miscompile via the conservative multi-output bailout added in #4922. This PR keeps the distinct-output safety, but replaces the bailout with precise origin tracking so trivial phis over the same output slot can still be eliminated.

### How I did it

Phi origins are now tracked by root operand. The operand identifies both its producer and the produced output slot, so `%r1` and `%r2` from the same `invoke` remain distinct origins without storing an extra producer-instruction key. When all paths reach the same output operand, `_process_phi` assigns from that operand. When a phi is a real multi-origin join, its own output operand remains the barrier origin.

### How to verify it

`test_phi_elim_multi_output_invoke_distinct_outputs_kept` covers the #5039 safety case: a phi over two different outputs of the same `invoke` survives.

`test_phi_elim_multi_output_invoke_same_output_eliminated` covers the master-distinguishing case: a phi over the same output variable of a multi-output `invoke` is eliminated. On current master, the same shape leaves `phis_after_pass=1`; this branch eliminates it.

Validation run:

```bash
uv run --python 3.12 -m pytest tests/unit/compiler/venom/test_phi_elimination.py -q
uv run --python 3.12 black --check vyper/venom/passes/phi_elimination.py tests/unit/compiler/venom/test_phi_elimination.py
git diff --check
```

### Commit message

```text
fix[venom]: track multi-output phi origins by variable

PhiEliminationPass used producer-instruction identity as the phi origin. That is safe only for single-output producers. For multi-output producers such as invoke, the produced SSA variable is part of the origin identity.

Track origins by root operand. Distinct outputs remain distinct roots because their operands are distinct, while phis that only join the same multi-output slot can still be folded.
```

Fix: Venom phi elimination now tracks multi-output origins by produced operand, keeping distinct-output phis while still eliminating trivial same-output phis.


tests/unit/compiler/venom/test_phi_elimination.py
```diff
@@ -319,3 +322,40 @@ def test_phi_elim_two_phi_merges():
     """
 
     _check_pre_post(pre, post, hevm=True)
+
+
+def test_phi_elim_multi_output_invoke():
```
HodanPlodky: this test does not fail on master, there was a fix for similar issue in the dyn alloc PR

banteg: Thanks, yes, the #4922 dyn-alloc merge added the conservative bailout, so the old distinct-output test no longer fails on current master.

I merged current master into this branch, removed that now-redundant bailout now that origins carry the produced variable, and added `test_phi_elim_multi_output_invoke_same_output_eliminated`. That new case distinguishes current master: the same shape leaves `phis_after_pass=1` there, while this branch eliminates it. I kept the distinct-output safety test as coverage for #5039 and updated the PR title/body to describe the post-#4922 scope.

Validation:
- `uv run --python 3.12 -m pytest tests/unit/compiler/venom/test_phi_elimination.py -q`
- `uv run --python 3.12 black --check vyper/venom/passes/phi_elimination.py tests/unit/compiler/venom/test_phi_elimination.py`
- `git diff --check`

vyper/venom/passes/phi_elimination.py
```diff
@@ -1,10 +1,14 @@
...
 from vyper.venom.passes.base_pass import InstUpdater, IRPass
 
 
 class PhiEliminationPass(IRPass):
-    phi_to_origins: dict[IRInstruction, set[IRInstruction]]
+    # phi -> set of (root instruction, variable produced by it). origins
+    # are keyed by the produced variable and not just the instruction,
+    # since different outputs of one multi-output instruction (e.g.
+    # invoke) are distinct origins.
+    phi_to_origins: dict[IRInstruction, set[tuple[IRInstruction, IROperand]]]
```
HodanPlodky: I think it would be better to only keep account of operands since operand identifies the inst

banteg: Done. `phi_to_origins` is now `set[IROperand]`: the root operand carries both producer identity and output slot identity, so we no longer store `(inst, operand)` pairs. Real multi-origin phis use their own output operand as the barrier marker, and `_process_phi` assigns from the sole root operand unless it is the phi output itself.

Validation:
- `uv run --python 3.12 -m pytest tests/unit/compiler/venom/test_phi_elimination.py -q`
- `uv run --python 3.12 black --check vyper/venom/passes/phi_elimination.py tests/unit/compiler/venom/test_phi_elimination.py`
- `git diff --check`

I also updated the PR body to describe the operand-only representation.

HodanPlodky (approved):

HodanPlodky (approved):

PR #5092, opened 2026-06-12
fix[venom]: reject `abi_encode` with no arguments

### LINKED ISSUE #5059: Venom: abi_encode accepts zero arguments
### Version Information

* OS: macOS

### What's your issue about?

Venom `abi_encode` / `_abi_encode` accepts calls with zero positional arguments, while legacy codegen rejects them with a structure error.

The semantic layer appears to allow the zero-argument shape through, so the rejection currently lives in legacy codegen. The Venom path builds an empty tuple payload and can encode zero bytes, or a bare selector if `method_id` is supplied.

Representative shapes:

```vyper
x: Bytes[0] = abi_encode()
```

```vyper
x: Bytes[4] = abi_encode(method_id=0x12345678)
```

Legacy raises `StructureException("abi_encode expects at least one argument")`; experimental codegen should not silently accept a larger source set unless this is an intentional language change.

### How can it be fixed?

Add the same `len(node.args) > 0` validation to Venom `lower_abi_encode`, or move the rule into shared semantic/builtin validation so both codegen paths agree.

banteg:
### What I did

Fixes #5059.

The experimental Venom codegen accepted `abi_encode()` / `_abi_encode()` with zero positional arguments, building an empty tuple payload (or a bare selector with `method_id`), while legacy codegen rejects the shape with `StructureException("abi_encode expects at least one argument")`. Experimental codegen should not silently accept a larger source set.

### How I did it

Added the same `len(node.args) < 1` guard at the top of `lower_abi_encode` in `vyper/codegen_venom/builtins/abi.py`, raising the identical exception type and message as legacy (`vyper/builtins/functions.py`).

### How to verify it

`test_abi_encode_no_args` in `tests/functional/builtins/codegen/test_abi_encode.py`, parametrized over both pipelines via `Settings(experimental_codegen=...)`. The venom case fails on master ("DID NOT RAISE"); the legacy case pins existing behavior. Full `test_abi_encode.py` passes under both pipelines.

### Commit message

```
the venom codegen accepted `abi_encode()` with zero positional
arguments, building an empty tuple payload (or a bare selector with
method_id), while legacy codegen rejects the shape with a
StructureException. the rejection lives in codegen (semantic analysis
allows the shape through), so the venom path silently accepted a
larger source set.

add the same validation to `lower_abi_encode`, with the identical
exception type and message.
```

Fix: experimental codegen rejects `abi_encode()` with no arguments, matching legacy.

### Cute Animal Picture

cute animal

### DIFF
--- a/tests/functional/builtins/codegen/test_abi_encode.py
+++ b/tests/functional/builtins/codegen/test_abi_encode.py
@@ -3,7 +3,6 @@ from eth.codecs import abi
 
 from tests.utils import decimal_to_int
 from vyper.compiler import compile_code
-from vyper.compiler.settings import Settings
 from vyper.exceptions import StructureException
 
 
@@ -397,13 +396,11 @@ def foo(ensure_tuple: bool) -> Bytes[96]:
     assert c.foo(True) == expected_output
 
 
-@pytest.mark.parametrize("use_venom", [False, True])
-def test_abi_encode_no_args(use_venom):
+def test_abi_encode_no_args():
     code = """
 @external
 def foo() -> Bytes[32]:
     return abi_encode()
     """
-    settings = Settings(experimental_codegen=use_venom)
     with pytest.raises(StructureException, match="abi_encode expects at least one argument"):
-        compile_code(code, settings=settings)
+        compile_code(code)
--- a/tests/functional/builtins/codegen/test_abi_encode.py
+++ b/tests/functional/builtins/codegen/test_abi_encode.py
@@ -2,6 +2,9 @@ import pytest
 from eth.codecs import abi
 
 from tests.utils import decimal_to_int
+from vyper.compiler import compile_code
+from vyper.compiler.settings import Settings
+from vyper.exceptions import StructureException
 
 
 # @pytest.mark.parametrize("string", ["a", "abc", "abcde", "potato"])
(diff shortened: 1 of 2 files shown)


tests/functional/builtins/codegen/test_abi_encode.py
```diff
@@ -392,3 +395,15 @@ def foo(ensure_tuple: bool) -> Bytes[96]:
     assert c.foo(False) == expected_output
     expected_output = b"\x00" * 31 + b"\x20" + b"\x00" * 32
     assert c.foo(True) == expected_output
+
+
+@pytest.mark.parametrize("use_venom", [False, True])
```
charles-cooper: we don't need to parametrize because it's already parametrized at the CI level

banteg:
Addressed in `d50e880e1`: removed the local `use_venom` parametrization from the `abi_encode()` no-args regression and let the suite-level pipeline parametrization cover it.

Validation:

- `uv run --python 3.12 -m pytest tests/functional/builtins/codegen/test_abi_encode.py -k abi_encode_no_args -q`
- `uv run --python 3.12 ruff check tests/functional/builtins/codegen/test_abi_encode.py`

charles-cooper (approved):

Issue #5012, opened 2026-05-25
Experimental backend miscompile: `x < 0` branch wrongly eliminated for `x = (negative_literal & v)`


jubnzv:
### Version Information

* vyper 6a348ae7
* OS: Linux

### What's your issue about?
The venom backend proves the `x < 0` arm unreachable and DCEs it when `x = NEG_LITERAL & v` (literal `!= -1`) is first narrowed by an `x > 0` test. Known-bits analysis treats `neg_literal & v` as non-negative (it ignores the literal's all-ones high bits), so with the `x>0`-false fact (`x<=0`) it concludes `x==0`.

MRE:
```vyper
@external
@view
def test(a1: int256) -> int256:
    x: int256 = -2 & a1
    if x > 0:
        return 1
    elif x < 0:
        return -1
    else:
        return 0
```

Reproduction:
```
vyper --experimental-codegen -f bytecode_runtime mre.vy   # venom
vyper                        -f bytecode_runtime mre.vy   # legacy
```
Call `test(-1)`  (`x = -2 & -1 = -2`, which is `< 0`).

- Expected: `-1`
- venom:    `0` 

Trigger needs all of: literal `!= -1` (`-1 & v` folds to identity), AND with a runtime value (not const-const, not var-var), result in a local, and an `x > 0` test before the `x < 0` test.

### How can it be fixed?
- [`eval_and`](https://github.com/vyperlang/vyper/blob/6a348ae71bcb0ef1b59fb73a7e2cbe20f30759a7/vyper/venom/analysis/variable_range/evaluators.py#L96-L145) -- root cause ([L138-L140](https://github.com/vyperlang/vyper/blob/6a348ae71bcb0ef1b59fb73a7e2cbe20f30759a7/vyper/venom/analysis/variable_range/evaluators.py#L138-L140): returns `iv(0, mask)` for a negative mask).
- [`_try_range_cmp`](https://github.com/vyperlang/vyper/blob/6a348ae71bcb0ef1b59fb73a7e2cbe20f30759a7/vyper/venom/passes/algebraic_optimization.py#L355) -- consumer that folds the comparison from that range.

PR #5093, opened 2026-06-12
fix[venom]: guard sdiv/smod range eval at sign boundary

### LINKED ISSUE #5045: Signed range evaluators for sdiv and smod ignore sign-boundary ranges
## Summary

The Venom variable-range domain can represent intervals with `lo >= 0` and `hi > SIGNED_MAX`. Members above `SIGNED_MAX` are negative EVM words when interpreted by signed opcodes.

The comparison-folding code has sign-boundary guards, but the signed division/modulo evaluators reason about sign directly from the raw integer `lo`/`hi`. That is unsound when a range crosses from small non-negative unsigned values into high unsigned / signed-negative words.

## Example

For `SMOD`, the range `[0, 2**256 - 32]` includes the word `2**256 - 32`, which is signed `-32`.

With divisor `10`:

```text
SMOD(-32, 10) = -2
```

As an EVM word, that result is `2**256 - 2`, not in `[0, 9]`.

But `_eval_smod` can reason from raw `lo >= 0` and return a small non-negative modulo range such as `[0, 9]`, excluding the actual result for high unsigned members.

The same sign-boundary issue applies to `_eval_sdiv` when the dividend range spans the signed boundary.

## Root cause

`vyper/venom/analysis/variable_range/evaluators.py` treats signed-opcode operand ranges as if `lo`/`hi` already describe a single monotonic signed interval. That is not true for ranges that include values above `SIGNED_MAX`.

The comparator optimizations already have a `_range_spans_sign_boundary` style guard. `eval_sdiv` and `eval_smod` need an equivalent guard or a conversion into a signed-domain representation before they make sign-based claims.

## Reachability

No type-correct frontend route was found in the audit: frontend `int256` values tend to stay in signed-range intervals. This is still an analysis-domain soundness gap, because other evaluators can produce high unsigned ranges and future optimization consumers may rely on these ranges for folding.

Related but not duplicate:

- #5012 reports a concrete signed range miscompile rooted in `eval_and`.
- #4827 tightens comparison folding and sign-boundary handling around comparisons, but does not fix the signed `sdiv`/`smod` evaluators themselves.

## Fix direction

For `eval_sdiv` and `eval_smod`, either:

- bail out to TOP when the dividend or divisor range spans the signed boundary; or
- split/normalize ranges into signed-domain pieces before computing the result range.

A focused regression can call the evaluator directly with a range containing both small unsigned values and high unsigned signed-negative words.

banteg:
### What I did

Fixes #5045.

The Venom variable-range domain can represent intervals like `[0, 2**256 - 32]` whose members above `SIGNED_MAX` are negative EVM words under signed interpretation. `eval_sdiv` and `eval_smod` reasoned about sign from raw `lo`/`hi`, so e.g. `smod([0, 2**256 - 32], 10)` folded to `[0, 9]` even though the member `2**256 - 32` (signed `-32`) yields `-2` (`2**256 - 2` as a word) — outside the claimed range.

### How I did it

Both evaluators now guard ranges extending above `SIGNED_MAX`, using the same boundary idiom `eval_sar` already has: `eval_sdiv` bails to `ValueRange.top()`, and `eval_smod` falls back to the magnitude bound `[-(|d|-1), |d|-1]`, which is sound for any dividend word (it is the same result the existing spans-zero branch returns, so `smod(TOP, d)` keeps its precision). Constant-dividend paths are unaffected (literals are normalized to signed representation before evaluation). Siblings audited: `eval_sar` already guarded; `eval_and` is tracked separately in #5012.

While auditing, one further candidate surfaced for a follow-up: the signed compare folding (`slt`/`sgt`) in `eval_compare` reasons from raw `lo`/`hi` with no `hi > SIGNED_MAX` guard — not touched here to keep the diff scoped.

### How to verify it

Four new analysis-level tests (per the `test_eval_byte.py` convention) cover spanning and entirely-above-boundary dividend ranges for both opcodes; all fail on master (e.g. smod returned `[0, 9]`). Full `tests/unit/venom/` and `tests/unit/compiler/venom/` pass.

### Commit message

```
the variable-range domain can represent intervals with lo >= 0 and
hi > SIGNED_MAX, whose high members are negative EVM words under
signed interpretation. eval_sdiv and eval_smod reasoned about sign
from raw lo/hi, so e.g. smod([0, 2**256 - 32], 10) was folded to
[0, 9] even though the member 2**256 - 32 (signed -32) yields -2,
i.e. 2**256 - 2 as a word.

guard ranges extending above SIGNED_MAX, matching the existing
eval_sar boundary guard: sdiv bails to TOP, and smod falls back to
the magnitude bound [-(|d|-1), |d|-1], which is sound for any
dividend word.
```

Fix: Venom range analysis no longer makes unsound sign-based claims for `sdiv`/`smod` when the dividend range crosses the signed boundary.

### Cute Animal Picture

cute animal


HodanPlodky (approved):

tests/unit/venom/test_variable_range_analysis.py
```diff
@@ -2248,3 +2249,44 @@ def test_smod_by_zero():
     smod_inst = next(inst for inst in entry.instructions if inst.opcode == "smod")
     rng = analysis.get_range(smod_inst.output, entry.instructions[-1])
     assert rng.lo == 0 and rng.hi == 0
+
+
+def test_smod_dividend_spanning_sign_boundary():
+    """
+    Bug: smod narrowed based on raw lo/hi signs for ranges crossing the
```
charles-cooper, line 2256: are there any venom optimization passes which were previously unsound but are sound after this PR? i would like to see those tests

banteg: Added optimizer-level coverage in d8e60d864. The new `AssertEliminationPass` tests cover both `sdiv` and `smod`: `%x = and source, 2**256 - 32` creates a range that includes the signed word `-32`, and the pass must not remove `assert (sgt result, -1)`. Validated with `uv run --python 3.12 -m pytest tests/unit/venom/test_assert_elimination.py tests/unit/venom/test_variable_range_analysis.py -q`, black check, and `git diff --check`.

HodanPlodky (approved):

PR #5094, opened 2026-06-12
fix[venom]: reject statically undersized abi_decode input

### LINKED ISSUE #5055: Venom: abi_decode accepts statically undersized input buffers
### Version Information

* OS: macOS

### What's your issue about?

Venom `abi_decode` does not appear to port legacy's compile-time check that the input bytestring max length can fit the requested output type.

Legacy `ABIDecode.build_IR` rejects code where:

```text
data.typ.maxlen < wrapped_typ.abi_type.size_bound()
```

That catches cases such as decoding a large static tuple from `Bytes[4]` at compile time. The Venom path has runtime decoder bounds checks, so this does not look like a memory safety issue, but it changes the accepted program set: code that legacy rejects compiles under `--experimental-codegen` into code that can only revert at runtime.

Representative shape:

```vyper
@external
def f(data: Bytes[4]) -> (uint256, uint256):
    return abi_decode(data, (uint256, uint256))
```

### How can it be fixed?

Port the legacy compile-time `maxlen >= abi_type.size_bound()` check into Venom `lower_abi_decode`, before emitting the decoder.

banteg:
### What I did

Fixes #5055.

Reject `abi_decode` calls when the input bytestring's maximum length cannot fit
the ABI size bound of the requested output type. The validation now happens
during semantic analysis, so analysis-only consumers and both codegen backends
accept the same program set.

### How I did it

Moved the user-facing size check into `ABIDecode.infer_arg_types`, including the
`unwrap_tuple` behavior used to calculate the external ABI type. Invalid source
continues to raise `StructureException` at the input argument.

The equivalent checks remain in legacy and Venom codegen as `CodegenPanic`
invariants. Reaching either one now indicates that semantic validation failed to
enforce the shared contract rather than a user error discovered late in codegen.

### How to verify it

The undersized-buffer regression compiles the same source in three modes:

- analysis-only `annotated_ast_dict`
- legacy bytecode
- experimental/Venom bytecode

All three must reject the source with `StructureException`.

Focused validation:

```
uv run --python 3.12 -m pytest tests/functional/builtins/codegen/test_abi_decode.py -q
uv run --python 3.12 -m pytest tests/functional/builtins/codegen/test_abi_decode.py --experimental-codegen -q
uv run --python 3.12 make lint
git diff --check
```

### Commit message

```
fix[venom]: reject statically undersized abi_decode input

Venom previously accepted `abi_decode` calls whose input bytestring could
never fit the requested ABI type, while legacy codegen rejected them.

Validate the size relationship during semantic analysis so analysis-only
consumers and both codegen backends reject the same invalid source. Keep
backend checks as internal invariants in case semantic validation is ever
bypassed.
```

Fix: reject `abi_decode` calls with statically undersized input buffers during
semantic analysis.

### Cute Animal Picture

cute animal


tests/functional/builtins/codegen/test_abi_decode.py
```diff
@@ -491,6 +491,18 @@ def test_abi_decode_length_mismatch(get_contract, assert_compile_failed, bad_cod
     assert_compile_failed(lambda: get_contract(bad_code), exception)
 
 
+def test_abi_decode_undersized_buffer_venom():
```
charles-cooper: do we not already have a test for this which is just not tripped on venom pipeline?

charles-cooper (approved):
ok in principle; see my testing comment. paging @HodanPlodky for review

tests/functional/builtins/codegen/test_abi_decode.py
banteg: Yes, there was already a length-mismatch compile-failure table. I folded the public `abi_decode(data, (uint256, uint256))` undersized-buffer case into `test_abi_decode_length_mismatch` and removed the separate Venom-named test in b0de8d88f. Validation: `uv run --python 3.12 -m pytest tests/functional/builtins/codegen/test_abi_decode.py -k length_mismatch -q` and the focused black check.

tests/functional/builtins/codegen/test_abi_decode.py
```diff
@@ -478,6 +478,14 @@ def foo(x: Bytes[32]):
     (
         """
 @external
+def foo(data: Bytes[4]) -> (uint256, uint256):
```
HodanPlodky: this test passes also in the master why was the change in logic necessary?

banteg: You're right - the folded table row passes on master in a default local run because that path uses legacy codegen. The regression is only in Venom.

I added `test_abi_decode_undersized_buffer_venom`, which forces `Settings(experimental_codegen=True)`: on `origin/master` the reproducer compiles, while this branch raises `StructureException`. I also updated the PR body to make that distinction explicit.

Validation:

```
uv run --python 3.12 -m pytest tests/functional/builtins/codegen/test_abi_decode.py -k "length_mismatch or undersized_buffer_venom" -q
uv run --python 3.12 -m pytest tests/functional/builtins/codegen/test_abi_decode.py -k "length_mismatch or undersized_buffer_venom" --experimental-codegen -q
uv run --python 3.12 black --check tests/functional/builtins/codegen/test_abi_decode.py vyper/codegen_venom/builtins/abi.py
git diff --check
```

charles-cooper: @HodanPlodky was your concern resolved?

vyper/codegen_venom/builtins/abi.py
```diff
@@ -214,6 +214,19 @@ def lower_abi_decode(node: vy_ast.Call, ctx: VenomCodegenContext) -> VyperValue:
     if unwrap_tuple:
         wrapped_typ = calculate_type_for_external_return(output_typ)
 
+    # Compile-time check that the input buffer can fit the decoded type
+    abi_size_bound = wrapped_typ.abi_type.size_bound()
+    input_max_len = data_node._metadata["type"].maxlen
+    if input_max_len < abi_size_bound:
+        raise StructureException(
```
charles-cooper: this should be a compiler or codegen panic.

charles-cooper (changes requested):

vyper/codegen_venom/builtins/abi.py
banteg: Addressed in 822a4bb79. The user-facing StructureException now comes from ABIDecode.infer_arg_types, so invalid source is rejected during semantic analysis. Both legacy and Venom retain the same max-length check as a CodegenPanic invariant if that semantic contract is ever bypassed. The regression covers analysis-only output plus legacy and experimental bytecode.

tests/functional/builtins/codegen/test_abi_decode.py
banteg: The regression is now explicit about the phase and backend coverage: annotated_ast_dict exercises semantic analysis only, while bytecode is compiled with legacy and experimental codegen settings. All three reject with StructureException, so the test no longer depends on the default backend or the CI matrix.

banteg:
Addressed the August 25 feedback in 822a4bb79. The invalid-input StructureException is now raised from ABIDecode.infer_arg_types, shared by analysis-only consumers and both codegen backends; legacy and Venom keep CodegenPanic invariant checks. The regression now covers annotated_ast_dict, legacy bytecode, and experimental bytecode. Validation: 121 tests passed for the full abi_decode test file on each backend, and make lint plus git diff --check passed. I also refreshed the PR description and resolved the addressed review threads.

harkal:
Thanks for this. This was a real gap when you opened it. It's since been fixed on master by #5127, which added the same check in `lower_abi_decode` (guarded for unbounded `INF` lengths), and `test_abi_decode_length_mismatch`'s `FAIL_LIST` already covers the case under both pipelines.

Closing as superseded, happy to review a fresh branch if you still want the check hoisted into semantic analysis.

PR #5095, opened 2026-06-12
fix[venom]: order memtop after memory-expanding reads

### LINKED ISSUE #5071: DFT can schedule memtop before memory-expanding reads
## Summary

`memtop` lowers to `MSIZE`, so it must observe prior memory growth. DFT currently prevents `memtop` from crossing memory writes, but it can still move `memtop` before an earlier memory read such as `mload`.

EVM memory reads can expand memory. If an `mload` touches beyond the current high-water mark, moving `MSIZE` before that read changes the returned value.

## Reproducer

```venom
function main {
main:
    %ptr = source
    %v = mload %ptr
    %m = memtop
    sink %v, %m
}
```

Running `DFTPass` on current master changes instruction order from:

```text
source, mload, memtop, sink
```

to:

```text
source, memtop, mload, sink
```

The exact repro used:

```python
from vyper.venom.parser import parse_venom
from vyper.venom.analysis import IRAnalysesCache
from vyper.venom.basicblock import IRLabel
from vyper.venom.passes.dft import DFTPass

src = '''
function main {
main:
    %ptr = source
    %v = mload %ptr
    %m = memtop
    sink %v, %m
}
'''
ctx = parse_venom(src)
fn = ctx.get_function(IRLabel("main"))
print([inst.opcode for bb in fn.get_basic_blocks() for inst in bb.instructions])
DFTPass(IRAnalysesCache(fn), fn).run_pass()
print([inst.opcode for bb in fn.get_basic_blocks() for inst in bb.instructions])
```

Observed:

```text
['source', 'mload', 'memtop', 'sink']
['source', 'memtop', 'mload', 'sink']
```

## Root cause

`memtop` has a `MEMORY` read effect, and DFT adds read-after-write dependencies for that effect. But there is no ordering dependency between two memory reads, even when the earlier read can expand memory and thereby affect the later `MSIZE` result.

## Impact

IR-level correctness bug. Current frontend/user-level reachability was not established in the audit, but the pass-local semantics are wrong for hand-written or future Venom IR using `memtop` after a memory-expanding read.

PR #4922 removes the old `memtop`/`MSIZE` dynamic allocation model, so that PR may obsolete this issue if it lands first. Until then, current master can reorder the IR above.

## Fix direction

Either model memory-expanding reads as affecting the memory high-water mark, or give `memtop` a stronger dependency on prior memory accesses, not only prior memory writes. A narrower hardening would treat `mload` and copy-like reads with dynamic/high offsets as barriers for `memtop` scheduling.

banteg:
### What I did

Fixes #5071.

`memtop` lowers to `MSIZE`, but DFT only serialized it against memory *writes*, so it could be scheduled before a memory-expanding *read* — `[source, mload, memtop, sink]` reordered to `[source, memtop, mload, sink]` (and reads could equally be hoisted above a memtop). EVM memory reads expand memory when they touch beyond the high-water mark, so the reorder changes the observed value.

### How I did it

Added an `MSIZE` effect with an intentionally "inverted" orientation: `memtop` *writes* MSIZE, and instructions that read memory without writing it (`mload`, `iload`, `sha3`, `log`, `return`, `revert`; the call/create/invoke families inherit via `ALL`) *read* MSIZE. Memory expansion commutes — msize is a monotonic high-water mark — so reads stay freely reorderable with each other, while `memtop` is serialized against them in both directions (reads-before-write and read-after-write edges). Memory writers were already ordered against `memtop` through its existing MEMORY read effect, which also covers `mcopy` and the copy opcodes.

The "honest" model (readers write MSIZE, memtop reads it) was rejected: DFT's write-after-write rule would chain every pair of memory-touching instructions — a real scheduling regression — and it would pessimize CSE and DSE. Audited all `get_read_effects`/`get_write_effects` consumers (DSE, CSE/available-expression, load elimination, mem-SSA, memmerging, copy elision, assert combiner): no store opcode gains effects and `mload` keeps empty overlap effects, so eligibility and CSE-ability are untouched; existing memtop CSE tests still pass.

### How to verify it

`test_dft_memtop_stays_after_memory_read` and `test_dft_memory_read_stays_after_memtop` in `tests/unit/compiler/venom/test_dft.py`; both fail on master. Full `tests/unit/compiler/` passes.

### Commit message

```
`memtop` lowers to MSIZE, which observes the memory high-water mark.
DFT serialized memtop only against memory writes, so it could be
scheduled across a memory-expanding *read* (e.g. mload past the
high-water mark), changing the observed value.

add an MSIZE effect: memtop writes it, and memory-reading
instructions read it. memory expansion commutes (msize is monotonic),
so reads remain freely reorderable with each other, while memtop is
ordered against them in both directions. memory writers were already
ordered against memtop via its MEMORY read effect.

the inverse orientation (readers write MSIZE) would serialize every
pair of memory-touching instructions via the write-after-write rule
and pessimize CSE/DSE; this orientation adds no new constraints
between ordinary instructions.
```

Fix: Venom instruction scheduling no longer moves `msize` observations across memory-expanding reads.

### Cute Animal Picture

cute animal

### DIFF
--- a/tests/unit/compiler/venom/test_dft.py
+++ b/tests/unit/compiler/venom/test_dft.py
@@ -4,8 +4,8 @@ from vyper.venom.passes import DFTPass
 _check_pre_post = PrePostChecker([DFTPass])
 
 
-def _check_no_change(pre):
-    _check_pre_post(pre, pre)
+def _check_no_change(pre, hevm=None):
+    _check_pre_post(pre, pre, hevm=hevm)
 
 
 def test_dft():
@@ -25,3 +25,35 @@ def test_dft():
         return %x, %y
     """
     _check_pre_post(pre, post)
+
+
+def test_dft_memtop_stays_after_memory_read():
+    """
+    `memtop` lowers to MSIZE, and memory reads (like mload) can expand
+    memory, advancing MSIZE. DFT must not schedule `memtop` before a
+    prior memory read.
+    """
+    pre = """
+    main:
+        %ptr = source
+        %v = mload %ptr
+        %m = memtop
+        sink %v, %m
+    """
+    # hevm does not model msize-dependent equivalence
+    _check_no_change(pre, hevm=False)
+
+
+def test_dft_memory_read_stays_after_memtop():
+    """
+    The converse direction: a memory read scheduled before a prior
+    `memtop` could expand memory and change the observed MSIZE.
+    """
+    pre = """
+    main:
+        %ptr = source
+        %m = memtop
+        %v = mload %ptr
+        sink %m, %v
+    """
+    _check_no_change(pre, hevm=False)
--- a/vyper/venom/effects.py
+++ b/vyper/venom/effects.py
@@ -8,6 +8,7 @@ class Effects(Flag):
     STORAGE = auto()
     TRANSIENT = auto()
     MEMORY = auto()
+    MSIZE = auto()
     IMMUTABLES = auto()
     RETURNDATA = auto()
     LOG = auto()
@@ -30,6 +31,7 @@ ALL = ~EMPTY
 STORAGE = Effects.STORAGE
 TRANSIENT = Effects.TRANSIENT
 MEMORY = Effects.MEMORY
+MSIZE = Effects.MSIZE
 IMMUTABLES = Effects.IMMUTABLES
 RETURNDATA = Effects.RETURNDATA
 LOG = Effects.LOG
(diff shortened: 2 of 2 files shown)


harkal (commented):
Fixed after [#4922](https://github.com/vyperlang/vyper/pull/4922)

banteg:
@harkal makes sense — #4922 removes `memtop`/MSIZE-based allocation from the IR surface entirely, which retires this whole bug class. Treat this PR as an interim fix for current master; happy to close it in favor of #4922, or keep it as a stopgap if #4922's timeline is longer — your call.

On the Codex comment about CSE: confirmed, and it's worth noting it is a *pre-existing* gap, not something this PR regresses or claims to fix — #5071 was specifically about DFT scheduling. CSE merges `%m2 = memtop` into `%m1` across a memory-expanding `mload` on master *without* this PR too:

```
%m1 = memtop
%v = mload %ptr      ; can expand memory
%m2 = %m1            ; <- CSE, both on master and with this PR
```

Within the available-expression framework the only sound complete fix in the current model would be treating `memtop` as non-idempotent (like `gas` in #5083), at the cost of no longer merging back-to-back memtops — the framework can't express "invalidated by memory reads" without making readers effect-writers, which pessimizes DFT/DSE. Given #4922 deletes `memtop`, that extra surgery only makes sense if this interim fix is wanted at all. Say the word and I'll extend this PR; otherwise leaving the CSE gap documented here.

charles-cooper:
@banteg what is your diagnosis now that #4922 is merged? are there any opcodes in master which have incorrect dft or cse treatment?

harkal:
I think we should close this as obsolete. `gas` CSE is still covered by #5083, but I don’t see a remaining `memtop`/MSIZE DFT or CSE issue on current master

banteg:
Closing as obsolete after #4922, per the current diagnosis in the thread. The remaining `gas` CSE issue is covered by #5083, and I do not see a remaining `memtop`/MSIZE DFT or CSE issue on current master.

PR #5096, opened 2026-06-12
fix[venom]: apply write effects of multi-output instructions in CSE

### LINKED ISSUE #5068: Venom CSE forwards stale storage load across tuple-returning invoke
## Summary

Venom CSE can forward a storage load across a tuple-returning internal `invoke`. The invoked function writes the same storage slot, but the optimized caller reuses the pre-call `sload` value for the post-call read.

This is separate from #5039: that issue is PhiElimination collapsing different outputs of one invoke. This one is available-expression/CSE treating the multi-output `invoke` boundary as insufficiently invalidating for storage expressions.

## Reproducer

The filler keeps the internal function uninlined so the `invoke` survives. Compile with experimental codegen and `-O gas`.

```python
from vyper.compiler import compile_code
from vyper.compiler.settings import Settings, OptimizationLevel, VenomOptimizationFlags

filler = "\n".join([f"    y = y + {i}" for i in range(1, 35)])
src = f'''
s: public(uint256)

@internal
def set_and_pair(x: uint256) -> (uint256, uint256):
    y: uint256 = x
{filler}
    self.s = x
    return y, x + 1

@external
def f(x: uint256) -> uint256:
    before: uint256 = self.s
    a: uint256 = 0
    b: uint256 = 0
    a, b = self.set_and_pair(x)
    after: uint256 = self.s
    return after - before + a - a + b - b

@external
def g(x: uint256) -> uint256:
    a: uint256 = 0
    b: uint256 = 0
    a, b = self.set_and_pair(x)
    return a + b
'''

def sload_count(flags=None):
    out = compile_code(
        src,
        output_formats=["ir_runtime"],
        settings=Settings(
            experimental_codegen=True,
            optimize=OptimizationLevel.GAS,
            venom_flags=flags,
        ),
    )["ir_runtime"]
    return str(out).count("sload")

print("default", sload_count())
print("disable_cse", sload_count(VenomOptimizationFlags(level=OptimizationLevel.GAS, disable_cse=True)))
```

Observed on current master:

```text
default 2
disable_cse 3
```

The missing `sload` is the post-`invoke` `self.s` read in `f()`. With CSE disabled, the post-call load is present.

## Impact

Source-triggerable incorrect codegen under experimental/Venom codegen. A storage read after an internal function call can observe the old value even when the callee writes the slot.

## Fix direction

Audit `vyper/venom/analysis/available_expression.py` and the CSE handling around multi-output `invoke`. Any `invoke` with storage effects needs to invalidate available storage loads regardless of the number of return values or output binding shape.

banteg:
### What I did

Fixes #5068.

Venom CSE could forward a storage load across a tuple-returning internal `invoke` whose callee writes the same slot — the caller reused the pre-call `sload` value for the post-call read.

### How I did it

Root cause: `AvailableExpressionAnalysis._handle_bb` skipped instructions with more than one output with a bare `continue` *before* the `remove_effect(write_effects)` step. `invoke` declares `writes = ALL`, and a single-output invoke goes through the normal path and invalidates everything correctly — but a tuple-returning invoke never invalidated anything, so pre-call expressions (e.g. `sload(slot)`) stayed available across the call. The CSE pass itself was fine (it already refuses multi-output instructions as substitution candidates); only the analysis failed to model their effects.

The fix applies the instruction's declared write effects before the multi-output skip — fully general for any future multi-output opcode.

### How to verify it

`test_cse_multi_output_invoke_effects` (venom-IR-level: `sload 0; %a, %b = invoke @f; sload 0` with the callee sstoring slot 0; asserts CSE makes no change) fails on master. The issue's source-level reproducer now keeps the same `sload` count with CSE on and off. Full `tests/unit/compiler/` passes.

### Commit message

```
the available-expression analysis skipped multi-output instructions
before applying their write effects. `invoke` declares writes = ALL,
and the single-output case invalidates available expressions
correctly -- but a tuple-returning invoke never invalidated anything,
so a pre-call sload stayed available across a call whose callee
writes the same slot, and CSE forwarded the stale value to the
post-call read.

apply the instruction's declared write effects before the
multi-output skip, so any multi-output opcode's effects are honored.
```

Fix: Venom CSE no longer forwards stale expressions (e.g. storage loads) across tuple-returning internal calls.

### Cute Animal Picture

cute animal


charles-cooper (approved):
lgtm, @HodanPlodky for addl review

HodanPlodky (approved):

PR #5097, opened 2026-06-12
perf[venom]: fast path for empty dynarray assignment

### LINKED ISSUE #5062: Venom: empty DynArray assignment misses zero-store fast path
### Version Information

* OS: macOS

### What's your issue about?

Venom assignment lowering special-cases empty bytestring assignment, but does not appear to special-case empty `DynArray` assignment.

Legacy lowers an empty dynamic-array assignment to a simple length store:

```text
STORE(dst, 0)
```

The Venom path sends `self.arr = []` through the more general copy path. This is correct semantically, but produces larger/slower code for a common reset operation.

Representative shape:

```vyper
arr: DynArray[uint256, 10]

@external
def clear():
    self.arr = []
```

### How can it be fixed?

Add the same empty-`DynArray` fast path used by legacy: store zero to the dynamic array length slot and skip the general copy machinery.

banteg:
### What I did

Fixes #5062.

Venom assignment lowering special-cased empty *bytestring* assignment but routed `self.arr = []` through the general dynarray copy machinery — a staging buffer plus a three-block runtime copy loop — where legacy emits a single zero store to the length slot.

### How I did it

Widened the two existing empty-bytestring checks (`lower_Assign` fast path and `_assign_value`) to `(_BytestringT, DArrayT)`. The existing `_is_empty_value` / `_get_target_ptr` helpers already do the right thing; `ptr_store` dispatches the correct store per data location (`sstore`/`tstore`/`mstore`/`istore`). `_is_empty_value` also matches `empty(DynArray[...])`, which is exact parity with legacy's `is_empty_intrinsic` fast path in `_dynarray_make_setter` (`[]` lowers to `~empty` in legacy too). Non-empty lists are untouched.

Before/after for `self.a = []` (storage, unoptimized venom IR): a 352-byte alloca plus a `dyn_cond`/`dyn_body`/`dyn_exit` copy loop → a single `sstore 0, 0`.

### How to verify it

New functional clear-array tests (storage/memory/transient, both pipelines) and unoptimized-IR shape tests (`tests/unit/compiler/venom/test_empty_dynarray_assign.py`) asserting the copy loop is gone — the shape tests assert at lowering time, not after optimizer cleanup. Full `tests/unit/compiler/` and `test_dynamic_array.py` pass under both pipelines.

### Commit message

```
venom assignment lowering special-cased empty bytestring assignment
but routed empty dynarray assignment (`self.arr = []`) through the
general copy machinery, emitting a staging buffer and a runtime copy
loop where a single zero store to the length slot suffices.

extend the existing empty-value fast path to DArrayT, matching legacy
codegen's `is_empty_intrinsic` fast path (which also covers
`empty(DynArray[...])`). the store dispatches per data location.
```

Perf: experimental codegen lowers empty `DynArray` assignment to a single length store, matching legacy.

### Cute Animal Picture

cute animal


tests/functional/codegen/types/test_dynamic_array.py
```diff
@@ -2340,3 +2340,68 @@ def foo():
     assert len(logs) == 1
     expected = [b"\xee\xb5", b"\xee\xb5", b"\xee\xb5", b"\xee\xb5"]
     assert logs[0].args.data == expected
+
+
+def test_assign_empty_list_clears_storage(get_contract, tx_failed):
```
charles-cooper: do we not already have tests for this?

vyper/codegen_venom/stmt.py
```diff
@@ -95,8 +95,9 @@ def lower_Assign(self) -> None:
         if isinstance(target, vy_ast.Tuple):
             return self._lower_tuple_unpack()
 
-        # Special case: empty Bytestring assignment — just zero the length word.
-        if isinstance(target_typ, _BytestringT) and self._is_empty_value(node.value):
+        # Special case: empty Bytestring/DynArray assignment — just zero the
+        # length word.
+        if isinstance(target_typ, (_BytestringT, DArrayT)) and self._is_empty_value(node.value):
```
charles-cooper: don't we have `is_dynamically_sized_type` or something? (if we don't, we should add it)

charles-cooper (changes requested):

charles-cooper:
looks ok in principle, left a couple comments to address

tests/functional/codegen/types/test_dynamic_array.py
banteg: Done in 97f7568ee. Removed the duplicate functional tests here; the existing `test_dynarray_length_no_clobber` cases already cover the clearing/oob behavior, and this PR keeps only the Venom-unit fast-path assertion. Validation: `test_empty_dynarray_assign.py` and `test_dynamic_array.py -k dynarray_length_no_clobber` passed.

vyper/codegen_venom/stmt.py
banteg: Done in 97f7568ee. Reused the existing `has_length_word()` predicate instead of spelling out `(_BytestringT, DArrayT)` in the empty-assignment fast path. Validation: `test_empty_dynarray_assign.py`, `test_dynamic_array.py -k dynarray_length_no_clobber`, black check, and `git diff --check` passed.

tests/unit/compiler/venom/test_empty_dynarray_assign.py
charles-cooper, line 1: there's nothing really wrong with these tests they are just kind of overzealous lol

tests/unit/compiler/venom/test_empty_dynarray_assign.py
```diff
@@ -0,0 +1,58 @@
...
+    self.a = []
+    """
+    ctx = _lower(code)
+
+    opcodes = [inst.opcode for inst in _instructions(ctx)]
+    # fast path: a single store of the (zero) length word, no copy loop
+    assert opcodes.count(store_opcode) == 1
+
+    labels = [str(bb.label) for fn in ctx.functions.values() for bb in fn.get_basic_blocks()]
+    assert not any("dyn_cond" in label for label in labels)
```
charles-cooper: too brittle

tests/unit/compiler/venom/test_empty_dynarray_assign.py
banteg: Trimmed this in f6e4a7db0 by deleting the bespoke lowering-shape unit test file. The PR now relies on the existing functional `dynarray_length_no_clobber` coverage through the codegen matrix instead of asserting raw Venom opcode/label shape. Validation: `uv run --python 3.12 -m pytest tests/functional/codegen/types/test_dynamic_array.py -k dynarray_length_no_clobber -q` and the focused black check.

charles-cooper (approved):

PR #5099, opened 2026-06-12
fix[venom]: port augassign read/write overlap guard

### LINKED ISSUE #5051: Venom: augmented assignment missing read/write overlap guard
### Version Information

* OS: macOS

### What's your issue about?

Venom `AugAssign` lowering does not appear to port the legacy read/write overlap guard that was added for GHSA-4w26-8p97-f4jp.

Legacy codegen rejects augmented assignments whose RHS can mutate the LHS target. That protection lives in codegen, not semantic analysis. In the Venom path, `lower_AugAssign` computes the target pointer and loads the current value before lowering the RHS, with no equivalent `read_write_overlap` / `contains_writeable_call` check.

Two bad consequences follow from the same ordering:

* For a dynamic-array target such as `self.a[1] += self.side_effect()`, if the side effect shrinks/pops `self.a`, the element pointer can be computed and bounds-checked against the old length and then written after the array has changed.
* For a scalar target such as `self.x += self.f()` where `f()` writes `self.x`, Venom can load the old target before the RHS side effect. Legacy rejects this class outright.

This looks like a regression of the codegen-layer guard from the advisory class.

### How can it be fixed?

Port the legacy overlap check to Venom augmented assignment lowering and raise before codegen for disallowed RHS/LHS combinations.

banteg:
### What I did

Fixes #5051.

The Venom pipeline's `lower_AugAssign` was missing the read/write overlap guard from GHSA-4w26-8p97-f4jp: the target pointer is computed (and bounds-checked) before the RHS executes, so programs like `self.a[1] += self.a.pop()` compiled under `--experimental-codegen` and could write out of bounds after the array shrank. Legacy rejects this class at codegen time (`parse_AugAssign`, added in #4487, relaxed in #4497).

### How I did it

Legacy's check reads legacy-IRnode properties (`referenced_variables`, `variable_writes`, `contains_writeable_call`), which don't exist in the AST-based venom lowering — but each has a faithful AST-level equivalent seeded from the same frontend analysis legacy itself seeds from:

- `referenced_variables` → recursive collection of `_expr_info._reads` over the reduced target AST (the frontend folds internal callees' state reads into call nodes);
- `variable_writes` → `get_expr_writes(rhs)` (the exact function legacy uses to seed IRnodes; internal-call writes are tracked transitively by the frontend);
- `contains_writeable_call` → `extcall` nodes, non-static `raw_call` (via `get_mutability_at_call_site`, so `is_static_call=True` stays allowed like legacy), `send`, and create builtins, recursing transitively through internal callees via `reachable_internal_functions`.

Same prim-word-target relaxation, same `CodegenPanic("unreachable")`.

### How to verify it

The GHSA source lists in `test_assignment.py` are extracted into module constants and exercised under `Settings(experimental_codegen=True)`: 6 rejection cases (all fail on master — they compile when they shouldn't) plus legacy-accepted overlap-OK sources still compiling. A 12-case differential confirmed venom now matches legacy exactly (6 rejected / 6 accepted), including a state-writing extcall hidden behind a transitive internal call. `tests/unit/compiler/` plus a 2094-test venom functional sweep pass with no over-rejection.

### Commit message

```
the venom pipeline's augassign lowering was missing the read/write
overlap guard from GHSA-4w26-8p97-f4jp: the target pointer is
computed and bounds-checked before the RHS is evaluated, so e.g.
`self.a[1] += self.a.pop()` compiled and could write out of bounds
after the array shrank. legacy rejects this class at codegen time.

port the legacy guard using AST-level equivalents of the legacy
IRnode properties, seeded from the same frontend analysis legacy
seeds from: target reads via `_expr_info._reads`, rhs writes via
`get_expr_writes`, and writeable-call detection covering extcall,
non-static raw_call, send and create builtins, recursing transitively
through internal callees. the same source programs are rejected with
the same exception, and everything legacy accepts still compiles.
```

Fix: experimental codegen rejects augmented assignments whose RHS can mutate the assignment target (GHSA-4w26-8p97-f4jp parity).

### Cute Animal Picture

cute animal

### DIFF
--- a/tests/functional/codegen/features/test_assignment.py
+++ b/tests/functional/codegen/features/test_assignment.py
@@ -1,5 +1,7 @@
 import pytest
 
+from vyper import compile_code
+from vyper.compiler.settings import Settings
 from vyper.evm.opcodes import version_check
 from vyper.exceptions import CodegenPanic, ImmutableViolation, InvalidType, TypeMismatch
 
@@ -40,16 +42,16 @@ def augmod(x: int128, y: int128) -> int128:
     print("Passed aug-assignment test")
 
 
-@pytest.mark.parametrize(
-    "source",
-    [
-        """
+# augassigns which can write to the target out from under us;
+# rejected with a (compile-time) panic - GHSA-4w26-8p97-f4jp
+AUGASSIGN_OOB_SOURCES = [
+    """
 @external
 def poc():
     a: DynArray[uint256, 2] = [1, 2]
     a[1] += a.pop()
     """,
-        """
+    """
 a: DynArray[uint256, 2]
 
 def side_effect() -> uint256:
@@ -60,7 +62,7 @@ def poc():
     self.a = [1, 2]
     self.a[1] += self.side_effect()
     """,
-        """
+    """
 a: DynArray[uint256, 2]
 
 def side_effect() -> uint256:
@@ -72,7 +74,7 @@ def poc():
     self.a = [1, 2]
     self.a[1] += self.side_effect()
     """,
-        """
+    """
 a: DynArray[uint256, 2]
 
 interface Foo:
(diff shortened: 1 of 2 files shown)


charles-cooper:
i think for venom we should actually just do the right thing rather than rejecting the program. i'd be ok with this as an interim fix tho

banteg:
Looked at this alongside #5100. I left this PR as the interim augassign guard for now: unlike rvalue subscript lowering, augassign writes back after RHS evaluation, so making it non-rejecting needs a broader write-semantics decision rather than the stale-base-pointer fix. #5100 now implements the narrower stale-base handling requested there.

banteg:
Superseded by #5127, now merged into master, which includes the augmented-assignment overlap guard. I checked the implementation and reran the extracted call-order/overlap regressions against master (25 passed, 4 expected failures across the combined coverage). Closing this redundant branch.

PR #5100, opened 2026-06-12
fix[venom]: reject costly storage overlap snapshots

### LINKED ISSUE #5054: Venom: array-subscript index side effects can stale the base pointer
### Version Information

* OS: macOS

### What's your issue about?

Venom array-subscript lowering evaluates the base before the index expression and does not appear to port legacy's `read_write_overlap(sub, index)` guard.

Legacy rejects cases where the index expression can mutate the array being indexed. Venom computes the base pointer/metadata, then lowers the index expression. If that index expression has side effects that mutate the same array, the base pointer or length assumptions can become stale before the load/store completes.

Representative shape:

```vyper
# index expression mutates the array being read through the already-lowered base
x: uint256 = self.a[0][self.mutate_a_and_return_index()]
```

This is the read-side version of the same aliasing/evaluation-order discipline that legacy codegen defends in augmented assignment.

### How can it be fixed?

Port the legacy `read_write_overlap(sub, index)` check to Venom subscript lowering and reject index expressions that can mutate the indexed base.

banteg:
### What I did

Prevent capacity-sized snapshots for nested storage/transient array reads whose index can mutate the previously evaluated base.

For example, `self.a[0][self.mutate_and_return_ix()]` needs stable data while the index runs. Current master snapshots it, but copying a storage `DynArray[uint256, 100000]` visits its declared capacity even when its runtime length is one. This PR conservatively rejects that overlap until a bounded-cost snapshot implementation is available.

### How I did it

Use master's existing overlap analysis. Reject an overlapping read when the delayed base is in storage or transient storage. Keep memory snapshots and the existing rejection for overlapping assignment targets. Ordinary direct storage indexing remains accepted; no recursive AST mirror of risky opcodes is added.

### How to verify it

Run `tests/functional/codegen/types/test_array_indexing.py` with and without `--experimental-codegen`.

The Venom run passes 14 tests with 7 expected failures. Added coverage includes storage/transient bounds of 5, 100000 and 2**64, plus an overlapping memory read that remains supported. `make lint` passes.

### Commit message

```text
Nested array reads may need to preserve the selected base while an index
expression mutates its owner. The storage snapshot currently copies the
entire declared capacity, causing disproportionate gas and memory use
even for short runtime arrays.

Reject these delayed storage and transient reads using the existing
overlap analysis. Retain memory snapshots and ordinary direct indexing
until storage snapshots can copy only the live value at bounded cost.
```

Experimental codegen rejects nested storage/transient subscript overlap that would require a capacity-sized snapshot.

### Cute Animal Picture

cute animal


charles-cooper:
maybe we should just freeze the base pointer; see my thought at https://github.com/vyperlang/vyper/pull/5099#issuecomment-4711384788

banteg:
Updated in c39659127: Venom now snapshots only rvalue subscript bases whose precomputed pointer can become stale (for example, a nested dynamic-array element), while leaving mapping-derived dynamic-array bases live so existing external-call mutation behavior is preserved. Assignment targets now lower with as_ptr=True so they still write through to the original target. Validation: uv run --python /opt/homebrew/bin/python3.11 pytest tests/functional/codegen/types/test_array_indexing.py -q --experimental-codegen; uv run --python /opt/homebrew/bin/python3.11 pytest tests/functional/codegen/types/test_array_indexing.py -q; uv run --python /opt/homebrew/bin/python3.11 ruff check vyper/codegen_venom/expr.py vyper/codegen_venom/stmt.py tests/functional/codegen/types/test_array_indexing.py.

tests/functional/codegen/types/test_array_indexing.py
```diff
@@ -3,7 +3,27 @@
...
+    return {
+        "type": "function",
+        "name": name,
+        "inputs": [],
+        "outputs": [{"name": "", "type": return_type}],
+        "stateMutability": "nonpayable",
+    }
+
+
+def _deploy_contract(env, code, compiler_settings, abi):
```
charles-cooper: no please don't reimplement `get_contract`

tests/functional/codegen/types/test_array_indexing.py
```diff
@@ -3,7 +3,27 @@
 import pytest
 
 from vyper.compiler import compile_code
-from vyper.exceptions import CompilerPanic
+
+
+def _xfail_legacy_codegen(experimental_codegen):
```
charles-cooper: i would put this in conftest or tests/utils.py

tests/functional/codegen/types/test_array_indexing.py
```diff
@@ -205,70 +224,23 @@ def calculate_index() -> uint256:
     self.a[0] = [1]
     return 0
     """
-    c = get_contract(code)
+    c = _deploy_contract(env, code, compiler_settings, [_abi_fn("bar", "uint256")])
 
     assert c.bar() == 1
 
 
-# sources rejected by the read/write overlap guard in subscript lowering:
```
charles-cooper: instead of deleting these let's promote them to functional tests (actually running bytecode and checking results)

vyper/codegen_venom/expr.py
```diff
@@ -163,6 +163,14 @@ def _subscript_read_write_overlap(
     return _contains_risky_call(index_node)
 
 
+def _subscript_base_pointer_can_stale(base_node: vy_ast.VyperNode) -> bool:
```
charles-cooper: rename: subscript_base_length_can_stale

vyper/codegen_venom/expr.py
```diff
@@ -927,17 +935,23 @@ def _lower_array_subscript(self, bounds_check: bool = True) -> VyperValue:
         """
         node = self.node
         assert isinstance(node, vy_ast.Subscript)
-        base_vv = Expr(node.value, self.ctx).lower()
+        base_vv = Expr(node.value, self.ctx, as_ptr=self.as_ptr).lower()
+        base_typ = node.value._metadata["type"]
+
+        # For rvalue reads, snapshot a base whose pointer can be invalidated
```
charles-cooper: ehhhh i don't think we can or want to do this. for example this will materialize huge dynarrays which don't fit in memory (but are valid in storage)

tests/functional/codegen/types/test_array_indexing.py
banteg: Done in c0663c653: removed the local compile/deploy helper. The promoted runtime repros now go through the shared get_contract fixture, with a supplied ABI so they can compile bytecode only and avoid the legacy-IR ABI side path.

tests/functional/codegen/types/test_array_indexing.py
banteg: Done in c0663c653: removed the local xfail helper. The legacy-codegen xfail is now explicit at each overlap runtime test.

tests/functional/codegen/types/test_array_indexing.py
banteg: Done in c0663c653: the three prior xfail compile cases are runtime tests now, and the pop case remains a runtime revert check. Validation: focused overlap tests pass with --experimental-codegen and xfail under legacy.

vyper/codegen_venom/expr.py
banteg: Done in c0663c653: renamed this to _subscript_base_length_can_stale.

tests/functional/codegen/types/test_array_indexing.py
```diff
@@ -2,7 +2,15 @@
 
 import pytest
 
-from vyper.exceptions import CompilerPanic
+
+def _abi_fn(name, return_type):
```
charles-cooper: dead code?

tests/conftest.py
```diff
@@ -292,7 +295,21 @@ def fn(source_code, *args, **kwargs):
             )["bytecode_runtime"]
             tests.hevm.hevm_check_bytecode(bytecode1, bytecode2, addl_args=_HEVM_MARKER.args)
 
-        return env.deploy_source(source_code, output_formats, *args, **kwargs)
+        if abi is not None:
```
charles-cooper: what on earth?

charles-cooper: why are we overriding the abi?

banteg: Removed in 087d6f998. I dropped the get_contract ABI override entirely and kept the bytecode-only deployment local to these regression tests via env.deploy([], bytecode). Focused Venom tests pass; legacy remains xfailed.

tests/functional/codegen/types/test_array_indexing.py
banteg: Removed in 087d6f998; the tests now use raw selector calls plus ABI decoding, so this helper is gone.

charles-cooper: sorry why do we need it all? like why can't we just use get_contract?

tests/conftest.py
charles-cooper: ditto - https://github.com/vyperlang/vyper/pull/5100#discussion_r3486822946

banteg: Same reason as above: there is no ABI override left in get_contract, and plain get_contract still goes through ABI generation, which currently forces legacy IR and trips the legacy overlap guard for these Venom-only runtime repros.

tests/functional/codegen/types/test_array_indexing.py
banteg: Plain get_contract still fails for these sources because it requests ABI+bytecode, and ABI output currently forces legacy ir_runtime to populate _ir_info. Even with experimental_codegen=True, that path hits legacy parse_Subscript and raises CompilerPanic("risky overlap") before the Venom bytecode can be deployed. I checked the direct compile path: compile_code(..., output_formats=("abi", "bytecode"), settings=Settings(experimental_codegen=True)) fails at build_abi_output -> compiler_data.ir_runtime -> legacy parse_Subscript. The raw selector call is only to keep these as Venom runtime tests without changing shared fixtures; the conftest ABI override is gone.

charles-cooper: oh, oof. maybe we should not force population of ir_info for abi output? i would rather guard against that in compiler/output.py:build_abi_output. like either:
1) only "ensure ir_info is generated" in the `if compiler_data.show_gas_estimates` branch
2) only allow `show_gas_estimates` if experimental_codegen is off, or
3) (likely best) switch to `_ = compiler_data.venom_runtime  # ensure ir_info is generated` if the venom pipeline populates ir_info. this can be further hidden behind a helper `_ensure_ir_info` for hygiene

banteg: Done in 3b25fba52. `build_abi_output` now ensures `_ir_info` via `venom_runtime` when experimental codegen is enabled, and via `ir_runtime` otherwise; gas-estimate ABI output uses the same pipeline-aware function signatures. I also added an ABI+bytecode regression for this overlap source. Focused validation: overlap tests pass under `--experimental-codegen` and xfail under legacy; ruff and `git diff --check` pass.

tests/functional/codegen/types/test_array_indexing.py
```diff
@@ -2,7 +2,20 @@
 
 import pytest
 
-from vyper.exceptions import CompilerPanic
+from tests.evm_backends.abi import abi_decode
+from vyper.compiler import compile_code
+from vyper.utils import method_id
+
+
+def _deploy_bytecode(env, compiler_settings, source_code):
```
charles-cooper: this still can't be removed? @banteg

vyper/compiler/output.py
```diff
@@ -302,18 +302,28 @@ def build_method_identifiers_output(compiler_data: CompilerData) -> dict:
...
+
+    fs = compiler_data.annotated_vyper_module.get_children(vy_ast.FunctionDef)
+    return {f.name: f._metadata["func_type"] for f in fs}
+
+
 def build_abi_output(compiler_data: CompilerData) -> list:
     module_t = compiler_data.annotated_vyper_module._metadata["type"]
     if not compiler_data.annotated_vyper_module.is_interface:
-        _ = compiler_data.ir_runtime  # ensure _ir_info is generated
+        _function_signatures_with_ir_info(compiler_data)
```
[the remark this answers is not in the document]
charles-cooper: @banteg

tests/functional/codegen/types/test_array_indexing.py
banteg: Done in 3f10e9159. I removed the raw bytecode deploy/call helpers from this test file; the overlap execution regressions now deploy through `env.deploy_source` with ABI/bytecode/metadata and call the ABI wrapper directly. Focused validation: overlap tests pass under `--experimental-codegen` and xfail under legacy; ruff and `git diff --check` pass.

vyper/compiler/output.py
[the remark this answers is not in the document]
banteg: Done in 3f10e9159. `build_metadata_output` now uses the same pipeline-aware IR-info path, forcing deploy-time Venom when metadata needs full reachable function info, and it omits legacy-only `frame_info` when Venom has no equivalent. I also added metadata to the overlap deployment regression so this path is covered. Focused validation: overlap tests pass under `--experimental-codegen` and xfail under legacy; `test_symbol_map.py::test_simple_map`, `test_output_selection.py::test_metadata_contain_all_reachable_functions`, ruff, and `git diff --check` pass.

tests/functional/codegen/types/test_array_indexing.py
banteg: Follow-up in 18fd09838: the overlap execution tests now use `get_contract` directly. To keep the override scoped to these Venom-only regressions, `get_contract` accepts a per-call `output_formats` override; the rest of this file still uses the default broad output-format fixture. Re-ran the focused overlap tests under Venom and legacy, plus one existing default-output `get_contract` test under both pipelines, ruff, and `git diff --check`.

vyper/codegen_venom/expr.py
banteg: Addressed in c67bb29f5 by restoring the original conservative CompilerPanic overlap guard and removing the later snapshot/output/test plumbing. The snapshot approach was not salvageable with a pointer-only freeze: a DynArray[DynArray[uint256, 100000], 1] repro emitted a 100001-slot storage copy plus 3200032 bytes of memory, nested lvalue cases could drop or write through stale assignments, and pop-on-index behavior disagreed with the direct-array live semantics. Rejection keeps both pipelines aligned until the evaluation semantics can be designed explicitly.

Validation: the full array-indexing file passes under both settings (10 passed, 3 expected xfails each); focused Ruff, Black, and git diff --check pass.

banteg:
Addressed the large-storage-array concern in 68c74e055. Master now handles these reads with snapshots, but a length-one array with a 100000-element bound still generated a full-capacity storage copy. This follow-up rejects only the delayed nested storage/transient overlap that needs that copy; it keeps memory snapshots and ordinary direct indexing. No broad risky-call mirror or fixture workarounds remain. Added regressions at bounds 5, 100000 and 2**64 and a supported overlapping memory read. Array-indexing tests pass on both backends (Venom: 14 passed, 7 expected failures), and `make lint` passes. The PR body now describes this narrower tradeoff.

banteg:
I also checked the older discussion against the final diff: no `get_contract`, ABI, or metadata overrides remain. Master already supplies the field/tuple lvalue propagation, stale-base traversal and pipeline-aware ABI/metadata generation. The new change is limited to rejecting costly storage/transient snapshots and testing that boundary. I resolved the threads about the superseded implementation.

The old Venom ABI gas-estimate limitation is unchanged from master and is not claimed fixed by this PR; I have left that thread open rather than marking it addressed.

PR #5102, opened 2026-06-12
chore[docs]: use an SVG for the deep verification layers diagram

lufa23:
The "Layers" section rendered the verification stack as an ASCII-art block, which is hard to read and does not scale. Replace it with an SVG that mirrors the same layers, stays legible on both the light and dark docs themes, and is capped to a sensible width so it does not span the full content column.
                                                                          
### How I did it

Added `docs/_static/deep-verification-layers.svg` (same layers as the
old block) and swapped the literal block in `deep-verification.rst` for
an image capped to a sensible width and centered.

### Commit message

```
The "Layers" section rendered the verification stack as an ASCII-art
block, which is hard to read and does not scale. Replace it with an SVG
that mirrors the same layers, stays legible on both the light and dark
docs themes, and is capped to a sensible width so it does not span the
full content column.
```


Sporarum (approved):
LGTM !

PR #5107, opened 2026-06-12
fix[lang]: fix `in` and `not in` bytesM folding

### LINKED ISSUE #5104: `in`/`not in` folding of `bytesM` literals is case-sensitive
### Version Information

* OS: osx

### What's your issue about?

`bytesM` literals are case-insensitive (`0xAB` and `0xab` denote the same `bytes1` value), but constant folding of `in`/`not in` compares the raw source strings case-sensitively. #4254 fixed this for `==`/`!=`, but the `In`/`NotIn` branch of `ConstantFolder.visit_Compare` was missed, so a membership test over literals can silently fold to the wrong boolean.

```vyper
@external
def folded() -> bool:
    return 0xAB00000000000000000000000000000000000001 in [0xab00000000000000000000000000000000000001]

@external
def runtime() -> bool:
    x: bytes20 = 0xAB00000000000000000000000000000000000001
    return x in [0xab00000000000000000000000000000000000001]
```

`folded()` returns `False` (the folded constant), `runtime()` returns `True` — same expression, opposite results. In `-f ir_runtime` output the folded function is just a constant `0`.

A more realistic trigger: `method_id()` folds to a lowercase hex node, so an uppercase selector literal never matches:

```vyper
@external
def is_transfer() -> bool:
    # folds to False, but 0xa9059cbb IS the transfer selector
    return 0xA9059CBB in [method_id("transfer(address,uint256)", output_type=bytes4)]
```

Notes:

* `address` literals are not affected (checksum validation makes the spelling of a given address unique).
* all `bytesM` widths are affected.

### How can it be fixed?

Lowercase `Hex` values in the `In`/`NotIn` branch of `visit_Compare`, mirroring what the `Eq`/`NotEq` path does since #4254. PR incoming.

banteg:
### What I did

Fixes #5104. Membership folding compared the source spelling of Hex literals case-sensitively. For example, `0xA9059CBB in [method_id("transfer(address,uint256)", output_type=bytes4)]` incorrectly folded to `False`.

### How I did it

Normalize `Hex.value` to lowercase at construction, establishing one representation for semantic comparisons. Equality and membership folding use that canonical value directly, without comparison-local case conversion.

Preserve the literal's spelling in `Hex.original_value` for address checksum validation, mixed-case bytes validation and diagnostics. The normal AST serializer exposes both fields under their own names; dictionary fields match node fields:

```json
{"ast_type": "Hex", "value": "0xa1aab33f", "original_value": "0xA1AAB33F"}
```

This changes Hex AST output: `value` is normalized and `original_value` carries the source spelling. Dictionary round trips preserve both fields. Older input dictionaries that contain only the source-spelled `value` remain accepted and are normalized on reconstruction.

### How to verify it

Regressions compare folded membership with the equivalent runtime operation, check normalization and serialization on both comparison operands, and cover mixed-case diagnostics, checksummed-address conversion and AST round trips. The reconstructed checksum literal is explicitly validated, including reconstruction from an older dictionary without `original_value`.

- AST, constant syntax and conversion suites: 1,719 passed per backend (legacy and Venom), after merging master at `4c04d6b29` and preserving its improved checksum diagnostics.
- `make lint` and `git diff --check` pass.

### Commit message

```text
Hex membership folding compares raw source strings, so equivalent
literals with different casing can produce different results.

Normalize Hex values at construction and use the canonical values in
equality and membership folding. Keep source spelling in original_value
for checksum validation, mixed-case validation and diagnostics.

Serialize both fields under their own names so AST dictionaries match
the nodes and preserve checksum-sensitive round trips. Continue to
accept older input dictionaries containing only the source value.
```

Fix case-sensitive Hex membership folding. Hex AST output now contains a normalized `value` and a source-spelled `original_value`.

### Cute Animal Picture

dog


vyper/semantics/analysis/constant_folding.py
```diff
@@ -169,7 +169,14 @@ def visit_Compare(self, node):
                 raise UnfoldableNode("Node contains invalid field(s) for evaluation")
             if len(set([type(i) for i in right.elements])) > 1:
                 raise UnfoldableNode("List contains multiple literal types")
-            value = node.op._op(left.value, [i.value for i in right.elements])
+            lvalue = left.value
+            rvalues = [i.value for i in right.elements]
+            if isinstance(left, vy_ast.Hex):
+                # Hex values are str, convert to be case-unsensitive.
+                lvalue = lvalue.lower()
+            if any(isinstance(i, vy_ast.Hex) for i in right.elements):
```
charles-cooper: i don't think this works for recursive lists? like list of list of hex elements. i think it would be architecturally better to have an earlier normalization step (would have to be after address checksum validation though)

charles-cooper (changes requested):
https://github.com/vyperlang/vyper/pull/5107/changes#r3415477115

vyper/semantics/analysis/constant_folding.py
banteg: Done in 504a5065a. Replaced the top-level hex special case with recursive literal comparison keys for lists/tuples. Hex values are lowercased only while deriving comparison keys, so checksum-sensitive AST literals are not rewritten globally. Added nested-list membership/equality coverage. Validation: compare-folding pytest and black check passed.

vyper/semantics/analysis/constant_folding.py
```diff
@@ -9,6 +9,22 @@ def constant_fold(module_ast: vy_ast.Module):
     ConstantFolder(module_ast).run()
 
 
+def _comparison_key(node):
```
charles-cooper: kind of baroque and non-performant. i would prefer normalization in `Hex` ctor and perhaps save the original value on `Hex` value for later address checksum validation

banteg: Done in 3aec4127f. `Hex.value` is now normalized at AST construction and `Hex.original_value` preserves the source spelling for address checksum validation. The compare fold now uses a simple recursive literal equality helper instead of the previous hex-specific comparison-key path. Validation: `test_fold_compare.py`, `test_constants.py`, and black check passed.

vyper/ast/nodes.py
```diff
@@ -865,14 +865,23 @@ class Hex(Constant):
         Value of the node, represented as a string taken directly from the contract source.
     """
 
-    __slots__ = ()
+    __slots__ = ("original_value",)
+
+    def __init__(self, parent: Optional["VyperNode"] = None, **kwargs: dict):
+        super().__init__(parent, **kwargs)
+        self.original_value = getattr(self, "original_value", self.value)
```
charles-cooper, line 880: i assume this happens in like copy operations, but there should be a comment

vyper/semantics/analysis/constant_folding.py
```diff
@@ -180,12 +201,15 @@ def visit_Compare(self, node):
...
-            # Hex values are str, convert to be case-unsensitive.
-            lvalue, rvalue = lvalue.lower(), rvalue.lower()
-        value = node.op._op(lvalue, rvalue)
-        return vy_ast.NameConstant.from_node(node, value=value)
+        if isinstance(node.op, (vy_ast.Eq, vy_ast.NotEq)):
+            value = _literal_eq(left, right)
+            if isinstance(node.op, vy_ast.NotEq):
+                value = not value
+            return vy_ast.NameConstant.from_node(node, value=value)
+        else:
```
charles-cooper: how does this work? like does `0x0A` compare gt `0x0a`?

not sure we need any of the `literal_eq` machinery at all

vyper/semantics/analysis/constant_folding.py
```diff
@@ -9,6 +9,25 @@ def constant_fold(module_ast: vy_ast.Module):
...
+    if type(left) is not type(right):
+        return False
+
+    if isinstance(left, vy_ast.Constant):
+        return left.value == right.value
+
+    return len(left.elements) == len(right.elements) and all(
+        _literal_eq(l, r) for l, r in zip(left.elements, right.elements)
+    )
+
```
charles-cooper: i think this machinery is superfluous and confusing

banteg: Agreed, removed in d052cee73. The PR now relies on normalized `Hex.value` and the existing scalar constant-folding path instead of adding recursive literal comparison support. Same focused fold-compare, constants syntax, and black checks passed.

vyper/semantics/analysis/constant_folding.py
banteg: Simplified in d052cee73. I removed the `_literal_eq` / recursive list-tuple comparison machinery and restored scalar-only folding here. Hex literals now normalize at `Hex.value`, so `in` / `not in` and `==` / `!=` compare normalized values directly; ordered hex comparisons remain unfoldable, with a small guard test for `0x0A > 0x0a`. Validation: focused fold-compare tests, constants syntax tests, and black check passed.

charles-cooper (approved):
lgtm, thanks!

charles-cooper:
hmm, maybe address literals got fucked up now? https://github.com/vyperlang/vyper/actions/runs/27589755492/job/83742065481?pr=5107 and https://github.com/vyperlang/vyper/pull/5107#issuecomment-4813151177

banteg:
Confirmed. `Hex.value` is normalized lowercase on this branch, and legacy `parse_Hex` was still checksum-validating against that normalized spelling for `AddressT` literals.

Pushed `a4294f34d` to keep using the normalized value for codegen numerics, but validate address checksum spelling against `Hex.original_value`. I also added a non-fuzz regression for `convert(0xffffFFFfFFffffffffffffffFfFFFfffFFFfFFfE, bool)` so this path is covered outside the fuzz split.

Validation:
- `uv run --python /opt/homebrew/bin/python3.12 pytest -q tests/functional/builtins/codegen/test_convert.py::test_convert_checksum_address_literal_bool`
- `uv run --python /opt/homebrew/bin/python3.12 pytest -q -m fuzzing 'tests/functional/builtins/codegen/test_convert.py::test_convert_passing[i_typ5-o_typ5-0xffffFFFfFFffffffffffffffFfFFFfffFFFfFFfE]'`
- `uv run ruff check vyper/codegen/expr.py tests/functional/builtins/codegen/test_convert.py`

vyper/ast/nodes.py
```diff
@@ -862,17 +862,30 @@ class Hex(Constant):
...
 
-    __slots__ = ()
+    __slots__ = ("original_value",)
+
+    def __init__(self, parent: Optional["VyperNode"] = None, **kwargs: dict):
+        super().__init__(parent, **kwargs)
+        # AST copies and folded-value snapshots can rebuild Hex from existing fields.
+        # Preserve the source spelling when it is already available.
+        self.original_value = getattr(self, "original_value", self.value)
+        self.value = self.value.lower()
```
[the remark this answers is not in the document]
banteg: Fixed in b780703b4. BytesM_T.validate_literal now checks mixed-case spelling from Hex.original_value before constant folding uses the normalized value, restoring the existing invalid-literal coverage including checksum-shaped bytes20 literals.

Validation: uv run --python 3.12 -m pytest tests/unit/ast tests/functional/syntax/test_constants.py tests/functional/builtins/codegen/test_convert.py -q (1715 passed).

vyper/ast/nodes.py
```diff
@@ -55,7 +55,7 @@
     "src",
 )
 
-DICT_AST_SKIPLIST = ("full_source_code", "node_source_code")
+DICT_AST_SKIPLIST = ("full_source_code", "node_source_code", "original_value")
```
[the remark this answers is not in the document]
banteg: Fixed in b780703b4. Hex.to_dict now emits the original source spelling as value, preserving the public AST format and allowing dict_to_ast to reconstruct checksum-sensitive literals. Added a checksummed-address round-trip regression.

Validation: uv run --python 3.12 -m pytest tests/unit/ast tests/functional/syntax/test_constants.py tests/functional/builtins/codegen/test_convert.py -q (1715 passed), plus focused Ruff, Black, and git diff --check.

tests/unit/ast/nodes/test_fold_compare.py
```diff
@@ -127,3 +127,46 @@ def foo(a: bytes4, b: bytes4) -> bool:
...
+        ("0xA1AAB33F", ["0xa1aab33f", "0x12345678"]),
+        ("0xa1aab33f", ["0xA1AAB33F", "0x12345678"]),
+        ("0xA1AAB33F", ["0x12345678", "0xabcdef01"]),
+    ],
+)
+def test_compare_in_bytes(get_contract, op, left, right):
+    source = f"""
+@external
+def foo(a: bytes4, b: bytes4[{len(right)}]) -> bool:
+    c: bytes4[{len(right)}] = b
```
Sporarum, line 147: Why the local variable ?

(not saying this is a mistake, genuine question)

vyper/ast/nodes.py
```diff
@@ -862,17 +862,37 @@ class Hex(Constant):
     Attributes
     ----------
     value : str
+        Value of the node, represented as a normalized lowercase string.
+    original_value : str
```
Sporarum, line 870: Is `original_value` ever different from `node_source_code` ?
If not, maybe it would make the code cleaner to remove `original_value` entirely ?

vyper/ast/nodes.py
```diff
@@ -862,17 +862,37 @@ class Hex(Constant):
...
+    def __init__(self, parent: Optional["VyperNode"] = None, **kwargs: dict):
+        super().__init__(parent, **kwargs)
+        # AST copies and folded-value snapshots can rebuild Hex from existing fields.
+        # Preserve the source spelling when it is already available.
+        self.original_value = getattr(self, "original_value", self.value)
+        self.value = self.value.lower()
+
+    @classmethod
+    def get_comparison_fields(cls) -> set:
+        return super().get_comparison_fields() - {"original_value"}
```
Sporarum: Is this correct ?

What is `get_comparison_fields` used for ?
Depending we might or might not want that

vyper/ast/nodes.py
```diff
@@ -862,17 +862,37 @@ class Hex(Constant):
...
+
+    def to_dict(self):
+        ast_dict = super().to_dict()
+        # Preserve the source spelling in public AST output so a dict round-trip
+        # can reconstruct checksum-sensitive literals.
```
Sporarum, line 886: That seems slightly weird, I would expect the dict output to stick as close as possible to the nodes (including field names and their values)

vyper/semantics/types/primitives.py
```diff
@@ -116,7 +116,9 @@ def validate_literal(self, node: vy_ast.Constant) -> None:
 
         assert isinstance(node, vy_ast.Hex)  # keep mypy happy
 
-        val = node.value
+        # Case is semantically irrelevant to constant folding, but mixed-case
+        # bytesM literals are invalid and must be checked before normalization.
```
Sporarum: Comment adds confusion, we only use `val` for the validity check below, this has nothing to do with constant folding, and it is obvious we should check the source value, not the sanitized one

vyper/semantics/types/primitives.py
Sporarum: Nit, and not due to this PR:

```suggestion
        nibbles = node.original_value[2:]  # strip leading 0x
```
Would make it more understandable, since the above `if node.n_bytes != self.m ...` does not use `val`

Sporarum (changes requested):

tests/unit/ast/nodes/test_fold_compare.py
banteg: Kept the local c and documented it in d9a392784. It deliberately materializes the calldata array in memory so the runtime side exercises the same memory-array membership path as the existing in/not in tests above, rather than only comparing against an inline literal.

vyper/ast/nodes.py
banteg: Keeping original_value. node_source_code is omitted by public AST dictionary round-trips, and synthesized Hex nodes can inherit source text for an enclosing expression rather than the literal being constructed. original_value travels with the Hex node itself, so validation reliably sees the spelling that produced that value.

vyper/ast/nodes.py
banteg: Removed the override in d9a392784. original_value affects mixed-case and checksum validation, so excluding it from AST comparison could conceal lossy reconstruction. The checksum-address dictionary round-trip now also asserts full deepequals with the preserved field included.

vyper/ast/nodes.py
banteg: Keeping to_dict value equal to original_value intentionally. value is the existing public AST field, and emitting the normalized lowercase value would lose checksum casing; dict_to_ast would then reconstruct and reject a previously valid checksummed address. The round-trip regression locks the public shape and full AST equality.

vyper/semantics/types/primitives.py
banteg: Agreed, removed the comment in d9a392784.

vyper/semantics/types/primitives.py
banteg: Applied in d9a392784: nibbles now comes directly from node.original_value[2:].

banteg:
Addressed the August 25 review in d9a392784. I kept Hex.original_value because public AST round-trips omit node_source_code and synthesized nodes can carry enclosing source text; I also kept public to_dict value source-faithful for checksummed-address compatibility. The get_comparison_fields override is removed, the validation cleanup is applied, and the runtime test now explains its deliberate memory-array materialization. Validation: 1,715 AST/constants/convert tests passed, 1,420 targeted experimental-codegen tests passed, and make lint plus git diff --check passed. The PR description is refreshed and all addressed threads are resolved.

harkal:
With this pr the error message quoting back a spelling the user didn't write (the actual correct one that would have compiled in the first place). Other than that lgtm.

banteg:
Fixed in cde625599. The fallback literal diagnostic now uses `Hex.original_value`, so it quotes the exact spelling from the source instead of the normalized lowercase value. I also added a regression for the mixed-case `bytes20` case.

Validation:
- `1719 passed` across `tests/unit/ast`, constants, and convert tests
- `1422 passed` for the folding/convert coverage with `--experimental-codegen`
- `make lint`

banteg:
Rechecked the literal normalization and `original_value` handling, including dictionary/AST round trips and checksum-sensitive behavior. The current implementation preserves the original spelling where needed while comparing normalized byte values. Focused tests pass on both pipelines (36 legacy, 37 Venom). I did not find another code change needed; the outstanding workflow run is awaiting approval rather than reporting a test failure.

vyper/ast/nodes.py
Sporarum: Explain this properly, what would change, and why do we care ?

banteg: You are right that the dictionary should match the node. The old approach lowercased Hex.value globally, then serialized a separate original_value to preserve address checksum casing. Serializing the lowercased value would lose information needed by checksum validation on dict_to_ast, but that mismatch was caused by an unnecessarily broad fix. In 7e008f769 I removed the representation split: Hex.value and the dictionary both retain the original spelling, and only temporary Hex values inside membership folding are normalized. Equality already follows this local approach. The checksum, mixed-case diagnostic, AST round-trip and runtime membership regressions remain; 1,679 tests pass per backend and make lint passes.

tests/unit/ast/nodes/test_fold_compare.py
```diff
@@ -156,12 +156,12 @@ def foo(a: bytes4, b: bytes4[{len(right)}]) -> bool:
...
-def test_hex_literal_value_normalized_after_preserving_original():
+def test_hex_comparison_preserves_source_spelling():
     vyper_ast = parse_and_fold("0xA1AAB33F == 0xa1aab33f")
     old_node = vyper_ast.body[0].value
 
-    assert old_node.left.value == "0xa1aab33f"
-    assert old_node.left.original_value == "0xA1AAB33F"
+    assert old_node.left.value == "0xA1AAB33F"
+    assert old_node.left.to_dict()["value"] == "0xA1AAB33F"
     assert old_node.get_folded_value().value is True
```
Sporarum: Why is this called old_node ?

tests/unit/ast/nodes/test_fold_compare.py
```diff
@@ -156,12 +156,12 @@ def foo(a: bytes4, b: bytes4[{len(right)}]) -> bool:
...
 
-def test_hex_literal_value_normalized_after_preserving_original():
+def test_hex_comparison_preserves_source_spelling():
     vyper_ast = parse_and_fold("0xA1AAB33F == 0xa1aab33f")
     old_node = vyper_ast.body[0].value
 
-    assert old_node.left.value == "0xa1aab33f"
-    assert old_node.left.original_value == "0xA1AAB33F"
+    assert old_node.left.value == "0xA1AAB33F"
+    assert old_node.left.to_dict()["value"] == "0xA1AAB33F"
```
Sporarum: Should check the right one as well

vyper/semantics/types/primitives.py
Sporarum: You have undone the fix

vyper/semantics/analysis/constant_folding.py
```diff
@@ -181,6 +188,9 @@ def visit_Compare(self, node):
                 f"Invalid literal types for {node.op.description} comparison", node
             )
         lvalue, rvalue = left.value, right.value
+        if isinstance(left, vy_ast.Hex):
+            # Hex values are str, convert to be case-unsensitive.
+            lvalue, rvalue = lvalue.lower(), rvalue.lower()
         value = node.op._op(lvalue, rvalue)
```
Sporarum: vy_ast.Hex absolutely needs to be in a standard form, we absolutely don't want this

Why did you revert your previous work ?

Sporarum (changes requested):

vyper/semantics/analysis/constant_folding.py
banteg: I changed it to preserve the existing AST dictionary shape, but that failed to preserve the canonical-Hex invariant you wanted. Corrected in 668a6a244: Hex.value is normalized at construction again, and both equality and membership use it directly. original_value preserves spelling for validation and diagnostics. The serializer now emits both fields under their own names, so it no longer substitutes original_value for value. Checksum round trips and older dictionary inputs are covered; 1,720 tests pass per backend and make lint passes. The description now explicitly documents the AST output change.

vyper/semantics/types/primitives.py
banteg: Restored in 668a6a244: nibbles is taken directly from node.original_value[2:], without the unnecessary val temporary.

tests/unit/ast/nodes/test_fold_compare.py
banteg: Renamed it to comparison in 668a6a244; it is the comparison AST node, not an old version of a literal.

tests/unit/ast/nodes/test_fold_compare.py
banteg: Added matching assertions for the right operand in 668a6a244. Both sides now check canonical value, original spelling, and both serialized fields.

PR #5108, opened 2026-06-12
fix[tool]: make decimal precision thread-independent

### LINKED ISSUE #5105: decimal precision is thread-local: same source compiles to different bytecode on a worker thread
### Version Information

* OS: osx

### What's your issue about?

`vyper/utils.py` configures the 78-digit decimal context with

```python
decimal.setcontext(DecimalContextOverride(prec=78))
```

but `decimal.setcontext()` is **thread-local**: it only applies to the thread which imports `vyper.utils`. Any other thread gets a fresh context copied from `decimal.DefaultContext` (prec=28, ROUND_HALF_EVEN). Vyper decimals span up to ~51 significant digits, so decimal constant arithmetic performed on a worker thread silently rounds at digit 28 — and the same source produces different bytecode depending on which thread runs the compiler:

```python
import threading
import vyper
from vyper.compiler.settings import Settings

src = """
A: constant(decimal) = 12345678901234567890.1234567890
B: constant(decimal) = 0.0000000001
C: constant(decimal) = A + B

@external
def foo() -> decimal:
    return C
"""

settings = Settings(enable_decimals=True)
results = {}
def compile_it(key):
    results[key] = vyper.compile_code(src, output_formats=["bytecode"], settings=settings)["bytecode"]

compile_it("main")
t = threading.Thread(target=lambda: compile_it("worker")); t.start(); t.join()
print(results["main"] == results["worker"])  # False!
```

* main thread folds `C` to `12345678901234567890.1234567891` (exact)
* worker thread folds `C` to `12345678901234567890.1234567900` (rounded half-even to 28 significant digits)

No error or warning is raised. This affects any embedder compiling in a thread pool (language servers, web compile services, framework plugins). Besides wrong values, the prec=28 context also changes which constant expressions raise `decimal.InvalidOperation` in `quantize()` (surfaced as `OverflowException`), so valid in-range constants can also be spuriously rejected on worker threads. The `DecimalContextOverride` guard which blocks precision downgrades doesn't exist on those threads either.

### How can it be fixed?

Also mutate the `decimal.DefaultContext` template (which threads created later copy their context from on first use):

```python
decimal.DefaultContext.prec = 78
```

This covers the common case where `vyper` is imported before worker threads first touch `decimal`. (A thread whose decimal context was already initialized before the import keeps prec=28 — fully closing that would require running all decimal arithmetic under an explicit context.) PR incoming.

banteg:
### What I did

Fix #5105: decimal contexts are thread-local, so `decimal.setcontext(DecimalContextOverride(prec=78))` in `vyper/utils.py` only configures the thread which imports vyper. Compiling on any other thread (e.g. in a thread pool) silently used the stdlib default of 28 significant digits, rounding decimal constants wider than 28 digits -- the same source compiled to different bytecode depending on the calling thread.

### How I did it

Also mutate the `decimal.DefaultContext` template. Threads which haven't called `setcontext()` copy their context from this template on first use, so threads spawned after vyper is imported get prec=78 as well. (A thread whose decimal context was already initialized before the vyper import still keeps the default precision -- fully closing that would require running all decimal arithmetic under an explicit context, which seems not worth the invasiveness.)

### How to verify it

`tests/unit/utils/test_decimal_context.py` folds `12345678901234567890.1234567890 + 0.0000000001` on a fresh worker thread and checks the exact result. Without the fix it folds to `...67900` (rounded half-even at 28 significant digits) instead of `...67891`, and the repro in #5105 produces different bytecode on main vs. worker threads.

### Commit message

```text
decimal.setcontext() only applies to the thread which imports
vyper.utils. compiling on any other thread (e.g. in a thread pool)
silently used the stdlib default context of 28 significant digits,
rounding decimal constants wider than 28 digits and producing
different bytecode for the same source depending on the calling
thread.

also mutate the decimal.DefaultContext template, which threads
created later copy their context from on first use.
```

Decimal constants are now evaluated at full precision regardless of which thread runs the compiler.

### Cute Animal Picture

dog

### DIFF
--- a/vyper/utils.py
+++ b/vyper/utils.py
@@ -223,11 +223,10 @@ class DecimalContextOverride(decimal.Context):
 decimal.setcontext(DecimalContextOverride(prec=78))
 
 # contexts are thread-local; a thread which has not called setcontext()
-# gets a context copied from the `DefaultContext` template on first use.
-# mutate the template so that threads spawned later (e.g. compiling in a
-# thread pool) also get 78 digits of precision instead of the stdlib
-# default of 28, which silently rounds decimal constants wider than 28
-# significant digits.
+# gets a context copied from the original `DefaultContext` template on
+# first use. Mutate that template in place instead of rebinding
+# `decimal.DefaultContext`, which the C decimal module does not use for
+# seeding new thread contexts.
 decimal.DefaultContext.prec = 78
 
 
--- a/vyper/utils.py
+++ b/vyper/utils.py
@@ -219,7 +219,16 @@ class DecimalContextOverride(decimal.Context):
         super().__setattr__(name, value)
 
 
+# set the context for the thread which imports vyper
 decimal.setcontext(DecimalContextOverride(prec=78))
+
+# contexts are thread-local; a thread which has not called setcontext()
+# gets a context copied from the `DefaultContext` template on first use.
+# mutate the template so that threads spawned later (e.g. compiling in a
+# thread pool) also get 78 digits of precision instead of the stdlib
+# default of 28, which silently rounds decimal constants wider than 28
+# significant digits.
+decimal.DefaultContext.prec = 78
 
 
 def keccak256(x):
--- a/tests/unit/utils/test_decimal_context.py
+++ b/tests/unit/utils/test_decimal_context.py
@@ -1,0 +1,25 @@
+import decimal
+import threading
+
+from tests.utils import parse_and_fold
+
+
+def test_decimal_precision_in_worker_thread():
+    # decimal contexts are thread-local; threads other than the one which
+    # imported vyper must also evaluate decimals at full precision, not
+    # the stdlib default of 28 significant digits
+    results = {}
+
+    def fold():
+        try:
+            vyper_ast = parse_and_fold("12345678901234567890.1234567890 + 0.0000000001")
+            results["value"] = vyper_ast.body[0].value.get_folded_value().value
+        except Exception as e:  # pragma: nocover
+            results["error"] = e
+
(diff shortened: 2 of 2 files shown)


vyper/utils.py
```diff
@@ -219,8 +219,17 @@ def __setattr__(self, name, value):
...
+# set the context for the thread which imports vyper
 decimal.setcontext(DecimalContextOverride(prec=78))
 
+# contexts are thread-local; a thread which has not called setcontext()
+# gets a context copied from the `DefaultContext` template on first use.
+# mutate the template so that threads spawned later (e.g. compiling in a
+# thread pool) also get 78 digits of precision instead of the stdlib
+# default of 28, which silently rounds decimal constants wider than 28
+# significant digits.
+decimal.DefaultContext.prec = 78
```
charles-cooper, line 230: why not set DefaultContext = DecimalContextOverride(prec=78)?

banteg:
I checked this on Python 3.12 before changing it. Rebinding `decimal.DefaultContext` to a `DecimalContextOverride(prec=78)` does not seed new thread contexts in the C decimal module: new threads still get a plain `Context(prec=28)`. Mutating the existing `decimal.DefaultContext.prec` is the part that propagates to new threads.

Addressed in `7d7745d2e` by keeping the working mutation and adding a short comment explaining why this is not a `decimal.DefaultContext = ...` rebind.

Validation:

- `uv run --python 3.12 -m pytest tests/unit/utils/test_decimal_context.py -q`
- `uv run --python 3.12 ruff check vyper/utils.py tests/unit/utils/test_decimal_context.py`

charles-cooper (approved):

PR #5109, opened 2026-06-12
fix[builtins]: fix panic in `shift()` negative value folds

### LINKED ISSUE #5106: `shift()` folding of negative values produces an out-of-bounds unsigned constant
### Version Information

* OS: osx

### What's your issue about?

Folding the (deprecated) `shift()` builtin computes left shifts modulo 2**256:

```python
value = (value << shift) % (2**256)
```

For a negative first argument this produces an unsigned constant, but `fetch_call_return` types the result as the type of the first argument (signed), so the folded literal fails the bounds check:

```vyper
@external
def foo() -> int256:
    return shift(-1, 1)
```

```
vyper.exceptions.TypeMismatch: Expected int256 but literal can only be cast as uint256.
```

The same operation with a runtime operand compiles and returns `-2`:

```vyper
@external
def bar() -> int256:
    x: int256 = -1
    return shift(x, 1)   # ok, returns -2
```

So a valid constant expression fails to compile only because it was folded. (Right shifts of negative values are unaffected — Python's `>>` is arithmetic and matches `SAR`.)

### How can it be fixed?

The fold can't know the operand's signedness (folding runs before type checking), so don't fold left shifts of negative values — let them go through runtime codegen, which handles signed values correctly. PR incoming.

banteg:
### What I did

Fix #5106: folding the deprecated `shift()` builtin computed left shifts modulo 2**256, producing an unsigned constant which is out of bounds for the signed output type. `shift(-1, 1)` failed to compile with "Expected int256 but literal can only be cast as uint256" while the same expression with a runtime operand compiles and returns `-2`.

### How I did it

Folding runs before type checking, so the fold can't know the operand's signedness. Raise `UnfoldableNode` for left shifts of negative values and let them go through runtime codegen, which handles signed values correctly. Right shifts of negative values keep folding — Python's `>>` is arithmetic and matches `SAR`.

### How to verify it

```vyper
@external
def foo() -> int256:
    return shift(-1, 1)
```

now compiles and returns `-2`, matching the runtime-operand behavior. Added a regression test in `tests/functional/builtins/codegen/test_bitwise.py`.

### Commit message

```
folding `shift()` with a negative first argument computed the left
shift modulo 2**256, producing an unsigned constant which is out of
bounds for the signed output type. `shift(-1, 1)` failed to compile
with "Expected int256 but literal can only be cast as uint256" while
the same expression with a runtime operand compiles and returns -2.

leave negative left shifts unfolded so they go through runtime
codegen, which handles signed values correctly.
```

Fix compile error when folding `shift()` with a negative literal value.

### Cute Animal Picture

dog


charles-cooper (approved):

vyper/builtins/functions.py
```diff
@@ -1335,6 +1335,11 @@ def _try_fold(self, node):
         if shift < 0:
             value = value >> -shift
         else:
+            if value < 0:
```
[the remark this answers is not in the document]
charles-cooper: @banteg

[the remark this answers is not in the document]
banteg: Fixed in b63c36082. For non-negative shifts, _try_fold() now computes the EVM 256-bit SHL result and, when the folded input was negative, converts that word back to the signed twos-complement interpretation instead of rejecting every negative left shift. I added module-constant coverage for shift(-1, 256), shift(-2, 255), and adjacent signed cases. Validation: uv run --python 3.12 -m pytest tests/functional/builtins/codegen/test_bitwise.py -q (206 passed, 2 xfailed).

vyper/builtins/functions.py
```diff
@@ -1332,10 +1332,15 @@ def _try_fold(self, node):
...
 
+        is_signed_shift = value < 0
         if shift < 0:
             value = value >> -shift
         else:
             value = (value << shift) % (2**256)
+            if is_signed_shift and value > SizeLimits.MAX_INT256:
+                # Runtime SHL works on the 256-bit word and the signed return
+                # type interprets that word as two's complement.
+                value -= 2**256
```
charles-cooper: vyper.utils.wrap256()

vyper/builtins/functions.py
```diff
@@ -1332,10 +1333,11 @@ def _try_fold(self, node):
             # have been validated anyway
             raise InvalidLiteral("Shift must be between -256 and 256", node.args[1])
 
+        is_signed_shift = value < 0
```
charles-cooper: move inside else branch please

vyper/builtins/functions.py
banteg: Done in 00ac68485: replaced the manual signed wrap with `wrap256(value << shift, signed=is_signed_shift)`. Re-ran `uv run --python 3.12 -m pytest tests/functional/builtins/codegen/test_bitwise.py -k "shift_builtin_negative_literal or signed_shift" -q`, black check, and `git diff --check`.

vyper/builtins/functions.py
banteg: Done in 12240fcd0: moved `is_signed_shift` into the left-shift `else` branch where it is used. Re-ran `uv run --python 3.12 -m pytest tests/functional/builtins/codegen/test_bitwise.py -k "shift_builtin_negative_literal or signed_shift" -q`, black check, and `git diff --check`.

PR #4987, opened 2026-05-20
fix[venom]: add missing scalar conversion validation

banteg:
### What I did

Added missing Venom validation for scalar conversion paths:

- `as_wei_value(int*, "wei")` now rejects negative signed values
- signed integer to `address` conversion is rejected
- `address` to signed integer conversion is rejected
- integer/address/decimal to too-small `bytesM` conversion is rejected before Venom bytecode generation

### How I did it

Mirrored the scalar validation already enforced by legacy codegen and ABI-oriented compile paths before Venom bytecode lowering accepts these conversions.

Also added the runtime clamp for the `as_wei_value(a: int*, "wei")` fast path, where Venom previously returned the raw signed word for denominator `1` and allowed `-1` to become `2**256 - 1`.

### How to verify it

- `uv run --python 3.12 pytest tests/functional/builtins/codegen/test_as_wei_value.py tests/functional/builtins/codegen/test_convert.py::test_venom_bytecode_output_rejects_invalid_conversions --experimental-codegen -q`
- `uv run --python 3.12 pytest tests/functional/builtins/codegen/test_as_wei_value.py::test_negative_int_wei_reverts tests/functional/builtins/codegen/test_convert.py::test_venom_bytecode_output_rejects_invalid_conversions -q`

### Commit message

```
the venom codegen path for convert() and as_wei_value() was missing
validation that the legacy pipeline and type checker enforce for other
compilation targets.

three gaps in convert.py:

- signed integer to address: _to_address did not reject signed inputs,
  allowing negative values through as garbled addresses.
- address to signed integer: _to_int allowed the conversion, which is
  semantically invalid (addresses are unsigned 160-bit values).
- integer/address/decimal to bytesM: _to_bytes_m did not check that the
  source width fits in the target, so narrowing conversions silently
  truncated.

additionally, lower_as_wei_value had a fast-path bug: when the
denominator is 1 (e.g. as_wei_value(a: int128, "wei")), the signed value
was returned unclamped. a negative input like -1 would be interpreted as
2**256 - 1 in the unsigned return type. add a non-negative assertion for
signed integers in the fast path, matching the existing clamp in the
general (non-denominator-1) branch.
```

### Description for the changelog

Fix Venom validation for invalid scalar conversions and negative signed `as_wei_value(..., "wei")` inputs.

PR #5114, opened 2026-06-12
fix[venom]: share convert() validation rules across pipelines

### LINKED ISSUE #5019: Venom convert address/flag to decimal omits divisor
### Version Information

* vyper Version: 0.4.4+commit.bdf00d31
* OS: Linux Ubuntu Jammy
* Python Version: Python 3.11

### What's your issue about?

Venom-pipeline `convert(<address|flag>, decimal)` ACCEPTS what legacy rejects, but emits bytecode that returns the raw uint160/flag-bitmask value WITHOUT multiplying by the decimal divisor (10^10). Result: `convert(address(0x1), decimal)` returns `0.0000000001` instead of `1.0`. Legacy correctly refuses with `TypeMismatch: Can't convert address to decimal` via the `@_input_types(IntegerT, BoolT, BytesM_T, BytesT)` decorator at `vyper/builtins/_convert.py:330`. Venom's `_to_decimal` falls through to `return val` at `vyper/codegen_venom/builtins/convert.py:293` for both `AddressT` and `FlagT`. This is a correctness bug, not merely acceptance divergence.

Trigger (address):

```vyper
#pragma enable-decimals

@external
@view
def f(a: address) -> decimal:
    return convert(a, decimal)
```

Trigger (flag):

```vyper
#pragma enable-decimals

flag Color:
    RED
    BLUE

@external
@view
def f() -> decimal:
    c: Color = Color.RED
    return convert(c, decimal)
```

Observed behavior:

- `vyper t.vy` → `TypeMismatch: Can't convert address to decimal` (or `Can't convert Color to decimal`)
- `vyper --experimental-codegen t.vy` → compiles, emits bytecode returning the raw uint160/flag value with no divisor multiplication
- A caller of the venom-compiled function gets `0.0000000001` instead of `1.0` for input `address(0x1)` / `Color.RED`

IR-level evidence:

```
$ vyper --experimental-codegen -f ir_runtime t.vy
# Inspection shows NO `mul` and NO `0x2540be400` (= 10^10) anywhere in the IR
```

Source-level evidence:

`vyper/codegen_venom/builtins/convert.py:241-293`:
```python
def _to_decimal(val, in_t, out_t, arg_node, ctx):
    _check_bytes(in_t, out_t, 32, arg_node)
    b = ctx.builder
    divisor = out_t.divisor

    if isinstance(in_t, _BytestringT):
        # ... load bytes, sar to align, return val (already in decimal scale)
        return val

    if isinstance(in_t, BytesM_T):
        # ... sar to align, return val
        return val

    if isinstance(in_t, IntegerT):
        val = _clamp_numeric_convert(val, ..., (scaled_lo, scaled_hi), ..., ctx)
        result = b.mul(val, IRLiteral(divisor))    # MULTIPLIES by divisor
        return result

    if isinstance(in_t, BoolT):
        return b.mul(val, IRLiteral(divisor))      # MULTIPLIES by divisor

    return val   # FALL-THROUGH — address, flag — NO DIVISOR MULTIPLICATION
```

`vyper/builtins/_convert.py:329-331` (legacy):
```python
@_input_types(IntegerT, BoolT, BytesM_T, BytesT)
def to_decimal(expr, arg, out_typ):
    _check_bytes(expr, arg, out_typ, 32)
    ...
```

The decorator lists allowed source types. `AddressT` and `FlagT` are not included, so legacy raises `TypeMismatch` at compile time. Venom has no equivalent allowlist and falls through to `return val`, returning the input unchanged for `AddressT` / `FlagT`. The caller then interprets the uint160/flag value as a fixed-point decimal (divisor 10^10), so `address(0x1)` decodes as `0.0000000001` instead of `1.0`, and `Color.RED` (= 1) decodes the same way.

### How can it be fixed?

Two options:

- Minimal fix: before the fall-through `return val` at `vyper/codegen_venom/builtins/convert.py:293`, add a branch that multiplies by `divisor` for `AddressT` / `FlagT` sources, e.g. `if isinstance(in_t, (AddressT, FlagT)): return b.mul(val, IRLiteral(divisor))`.
- Preferred fix: port the `@_input_types(IntegerT, BoolT, BytesM_T, BytesT)` decorator from `vyper/builtins/_convert.py:330` to the venom `_to_decimal` entry so that `address`/`flag` sources are rejected at compile time, matching legacy behavior.

### LINKED ISSUE #5110: venom: `int -> decimal` lower-bound clamp off by one, returns out-of-range decimal
### Version Information

* vyper Version (output of `vyper --version`): 0.5.0a3, master @ 9b5baa003
* OS: macos

### What's your issue about?

In the venom pipeline (`--experimental-codegen`), the int → decimal conversion computes its pre-multiplication bounds with floor division, while legacy truncates toward zero:

```python
# vyper/codegen_venom/builtins/convert.py:287 (_to_decimal, integer branch)
pre_lo = out_lo // divisor   # floors toward -inf
pre_hi = out_hi // divisor
```

```python
# vyper/builtins/_convert.py:147 (_int_to_fixed)
out_lo = round_towards_zero(out_lo / decimal.Decimal(DIVISOR))   # truncates toward zero
out_hi = round_towards_zero(out_hi / decimal.Decimal(DIVISOR))
```

For the negative lower bound (`out_lo = -2**167`, `divisor = 10**10`) these differ by exactly 1, so venom's clamp is one value too permissive on the negative side. The admitted extra value scales below `MIN_DECIMAL`, producing a decimal whose underlying int168 is `< -2**167` — a type-invariant violation (legacy reverts; venom returns the corrupt value).

Trigger:

```vyper
#pragma enable-decimals

@external
def f(x: int256) -> decimal:
    return convert(x, decimal)
```

Call with `x = -(2**167) // 10**10 = -18707220957835557353007165858768422651596`:

* legacy: clamps at `-18707220957835557353007165858768422651595`, reverts ✅
* venom: clamps at `-18707220957835557353007165858768422651596`, passes, multiplies by `10**10`, and returns a raw int168 below `-2**167`. ABI decoders reject it as out-of-bounds; on-chain it is a corrupt decimal.

IR-level evidence — venom's lower-bound clamp constant is one larger in magnitude than the (correct) upper-bound constant, instead of equal:

```
$ vyper --enable-decimals --experimental-codegen -f ir_runtime t.vy | grep -E 'sgt|slt'
    %39 = -0x36f9bfb3af7b756fad5cd10396a21346cc   # lower clamp = -...596 (floor) ← off by one
    %16 = sgt %39, %38
    %41 = 0x36f9bfb3af7b756fad5cd10396a21346cb    # upper clamp = +...595 (correct)
    %18 = slt %41, %40
```

Legacy IR for the same contract emits `±18707220957835557353007165858768422651595` for both bounds.

Note the reverse direction (decimal → int, `_to_int` at convert.py:209) is fine — it scales the int bounds with an exact multiplication.

### How can it be fixed?

Use `round_towards_zero` (already in `vyper.utils`) for `pre_lo`/`pre_hi` in `_to_decimal`, matching legacy `_int_to_fixed`.

### LINKED ISSUE #5111: venom: `convert()` missing input-type validation accepts conversions legacy rejects
### Version Information

* vyper Version (output of `vyper --version`): 0.5.0a3, master @ 9b5baa003
* OS: macos

### What's your issue about?

Legacy `convert()` gates every conversion target with an `@_input_types(...)` allowlist plus explicit `_FAIL` checks (`vyper/builtins/_convert.py`). The venom pipeline's `lower_convert` (`vyper/codegen_venom/builtins/convert.py:30`) dispatches purely on the **output** type and never validates the input type, so several conversions that legacy rejects with `TypeMismatch` compile cleanly under `--experimental-codegen`:

| source → target | legacy | venom | legacy rule that is lost |
|---|---|---|---|
| `String[20] → uint256` | `TypeMismatch` | compiles | `to_int` allowlist excludes `StringT` |
| `flag → uint8` | `TypeMismatch` | compiles | flag may only convert to `uint256` (`_convert.py:313-315`) |
| `uint8 → flag` | `TypeMismatch` | compiles | `to_flag` requires `uint256` input (`_convert.py:459`) |
| `flag → bool` | `TypeMismatch` | compiles | `to_bool` allowlist excludes `FlagT` |
| `bool → address` | `TypeMismatch` | compiles | `to_address` allowlist excludes `BoolT` |

Each row verified on master @ 9b5baa003 with standalone contracts, e.g.:

```vyper
@external
def f(b: bool) -> address:
    return convert(b, address)
```

```
$ vyper t.vy
TypeMismatch: Can't convert bool to address
$ vyper --experimental-codegen t.vy
<bytecode>
```

These are compile-time soundness holes (accepts-invalid): programs the language rejects get bytecode. Runtime semantics of the accepted programs are at least clamped — e.g. `flag → uint8` goes through `_int_to_int`'s narrowing clamp so out-of-range flag values revert rather than truncate — but the acceptance divergence itself can mask type errors and makes the pipelines disagree on what is a valid program.

#5019 (`address/flag → decimal` accepted and miscompiled) is another instance of the same root cause: `_to_decimal` also never checks `in_t`. More generally, every `_to_*` helper in `codegen_venom/builtins/convert.py` trusts whatever input it is handed instead of enforcing the legacy conversion matrix.

### How can it be fixed?

Validate `in_t` against the same input-type allowlists legacy uses, ideally driven from a single shared table rather than a second hand-maintained copy in the venom path (the divergence here is exactly what happens when the matrix is forked).

banteg:
### What I did

Fixes #5110, fixes #5111, fixes #5019.

Three rounds of venom `convert()` validation fixes in three weeks (#4987, #5019, #5111) all share one root cause: `vyper/codegen_venom/builtins/convert.py` forked the legacy conversion logic and dropped details in the hand-port — the per-target input-type allowlists, and the truncating division used for the int→decimal clamp bounds. This PR stops the whack-a-mole: it extracts the conversion matrix and numeric clamp bounds into a backend-neutral module that both pipelines consume, so the rules can no longer diverge.

### How I did it

* New `vyper/builtins/_convert_rules.py` — single source of truth:
  * `validate_convertibility(in_typ, out_typ, node)`: the full conversion matrix (per-target input allowlists + the parametric rules: flag↔uint256 only, address↔unsigned only, no narrowing into bytesM, bytestring word-size limits, bytestring upcast ban). Only rules that are pure functions of the type pair live here; value-dependent checks (literal ranges, literal bytestring downcasts) stay in codegen.
  * `int_to_fixed_clamp_bounds` / `fixed_to_int_clamp_bounds`: the numeric clamp windows, using exact truncating integer division (`evm_div`) — this kills the floor-vs-truncate off-by-one of #5110 structurally.
  * The module imports only from `vyper.semantics` and `vyper.utils`, so either backend (or, in the future, the frontend typechecker) can use it without import cycles.
* Legacy `_convert.py`: the `@_input_types(...)` decorator allowlists are replaced by a `@_validate_inputs` decorator that delegates to the shared matrix (same-type `InvalidType` precedence preserved); `_int_to_fixed`/`_fixed_to_int` use the shared bounds helpers. Body-level checks are kept as defense in depth — the matrix is strictly at-least-as-strict, and the existing legacy test suite passing confirms it is not stricter.
* Venom `convert.py`: `lower_convert()` validates against the matrix once at dispatch. The partial checks accreted by #4987 and the dead `_check_bytes` helper are removed; the int→decimal and decimal→int clamps use the shared bounds helpers. The `_to_decimal` fallthrough that produced #5019's miscompile is now an unreachable `CompilerPanic`.

Why the existing matrix tests never caught this: several output formats (`ir_dict`, `metadata`, `abi`) run the **legacy** codegen even under `--experimental-codegen`, and the test fixtures compile all formats — so legacy's validation masked venom's holes in every venom CI lane. The new regression tests compile bytecode-only to dodge this (broader fix tracked separately).

### How to verify it

```
$ pytest tests/functional/builtins/codegen/test_convert.py -m "not fuzzing"                          # legacy
$ pytest tests/functional/builtins/codegen/test_convert.py -m "not fuzzing" --experimental-codegen   # venom
```

New tests: `test_illegal_conversions_blocked` (13 accepts-invalid pairs × both pipelines, bytecode-only) and `test_int_to_decimal_clamp_bounds` (exact boundary behavior at ±MIN/MAX_DECIMAL ÷ divisor). On master, the former fails for every venom-accepted pair and the latter fails under `--experimental-codegen`; on this branch all pass. Also verified: full `tests/functional/builtins` and `tests/functional/syntax` (non-fuzzing) green under both pipelines, and the `test_convert.py` fuzzing matrix green under legacy (86,458 cases). Note the fuzzing matrix under `--experimental-codegen` has 15,750 pre-existing failures (every `test_conversion_failures` case: venom's SCCP statically proves the constant clamp reverts and raises `StaticAssertionException` where the test expects a runtime revert) -- verified identical on master's compiler, unrelated to this change; that lane has never run in CI.

### Commit message

```
the venom-direct codegen for convert() forked the legacy conversion
logic and dropped two kinds of knowledge in the port: the per-target
input type allowlists (GH 5019, GH 5111) and the truncating (rather
than floor) division used to compute the int->decimal clamp bounds
(GH 5110). the result was a cluster of accepts-invalid bugs -- programs
which legacy rejects with TypeMismatch compiled to bytecode -- and an
off-by-one lower clamp which admits one extra negative value whose
scaled result lands below the decimal type's lower bound, producing an
out-of-range decimal at runtime.

this is the third round of such fixes in this file (after GH 4987 and
GH 5019); patching individual instances does not converge. instead,
extract the conversion matrix and the numeric clamp bounds into a new
backend-neutral module, vyper/builtins/_convert_rules.py, used by both
pipelines: legacy's @_input_types decorator delegates to it, and venom's
lower_convert() validates against it at dispatch before lowering. the
checks the venom helpers had accreted piecemeal (GH 4987) are removed
in favor of the single dispatch-time validation; value-dependent checks
(literal ranges, literal bytestring downcasts) stay in codegen.

the module deliberately imports only from vyper.semantics and
vyper.utils so that either backend (and potentially the frontend
typechecker, in the future) can use it without import cycles.

note: the existing conversion matrix tests did not catch these
divergences because several output formats (ir_dict, metadata, abi)
run the legacy codegen even under --experimental-codegen, so the test
harness's multi-format compilation lets legacy validation mask venom
bugs. the new regression tests compile bytecode-only.
```

Fix venom pipeline `convert()` bugs: missing input-type validation accepted illegal conversions (e.g. `bool` → `address`, `flag` → `uint8`), and the int→decimal clamp admitted one out-of-range negative value. Conversion rules are now shared between the legacy and venom pipelines.

### Cute Animal Picture

\dog

### DIFF
--- a/tests/functional/builtins/codegen/test_convert.py
+++ b/tests/functional/builtins/codegen/test_convert.py
@@ -847,19 +847,15 @@ ILLEGAL_CONVERSIONS = [
 
 
 @pytest.mark.parametrize("i_typ,o_typ", ILLEGAL_CONVERSIONS)
-@pytest.mark.parametrize("use_venom", [False, True])
-def test_illegal_conversions_blocked(i_typ, o_typ, use_venom):
+def test_illegal_conversions_blocked(i_typ, o_typ):
     preamble = FLAG_PREAMBLE if "Roles" in (i_typ, o_typ) else ""
     code = f"""{preamble}
 @external
 def foo(x: {i_typ}) -> {o_typ}:
     return convert(x, {o_typ})
     """
-    settings = Settings(experimental_codegen=use_venom, enable_decimals=True)
-    # compile bytecode only: other output formats run the legacy pipeline
-    # even when experimental_codegen is set, masking venom-only bugs
     with pytest.raises(TypeMismatch):
-        compile_code(code, output_formats=("bytecode",), settings=settings)
+        compile_code(code, output_formats=("bytecode",))
 
 
 def test_int_to_decimal_clamp_bounds(get_contract, tx_failed):
--- a/tests/functional/builtins/codegen/test_convert.py
+++ b/tests/functional/builtins/codegen/test_convert.py
@@ -818,3 +818,69 @@ def test_convert() -> uint256:
     """  # noqa: E501
     c = get_contract(code)
     assert c.test_convert() == 42
+
+
+FLAG_PREAMBLE = """
+flag Roles:
+    ADMIN
+    USER
+"""
+
+# conversions which the venom pipeline accepted due to missing input
+# validation (GH 5111, GH 5019). the conversion matrix is shared between
+# pipelines, so both must reject these at compile time.
+ILLEGAL_CONVERSIONS = [
+    ("String[20]", "uint256"),
+    ("String[20]", "address"),
+    ("Roles", "uint8"),
+    ("Roles", "bool"),
+    ("Roles", "decimal"),
+    ("uint8", "Roles"),
+    ("bytes32", "Roles"),
+    ("bool", "Roles"),
+    ("bool", "address"),
+    ("address", "decimal"),
+    ("address", "int256"),
+    ("decimal", "address"),
+    ("decimal", "Roles"),
+]
+
+
+@pytest.mark.parametrize("i_typ,o_typ", ILLEGAL_CONVERSIONS)
+@pytest.mark.parametrize("use_venom", [False, True])
+def test_illegal_conversions_blocked(i_typ, o_typ, use_venom):
+    preamble = FLAG_PREAMBLE if "Roles" in (i_typ, o_typ) else ""
+    code = f"""{preamble}
+@external
+def foo(x: {i_typ}) -> {o_typ}:
+    return convert(x, {o_typ})
+    """
+    settings = Settings(experimental_codegen=use_venom, enable_decimals=True)
+    # compile bytecode only: other output formats run the legacy pipeline
+    # even when experimental_codegen is set, masking venom-only bugs
+    with pytest.raises(TypeMismatch):
+        compile_code(code, output_formats=("bytecode",), settings=settings)
+
+
+def test_int_to_decimal_clamp_bounds(get_contract, tx_failed):
+    code = """
+@external
+def foo(x: int256) -> decimal:
+    return convert(x, decimal)
+    """
+    c = get_contract(code)
+
+    # the largest-magnitude integers whose promotion stays within the
+    # decimal range: MIN/MAX raw decimal divided by the divisor, rounded
+    # towards zero. flooring instead admits one extra negative value
+    # whose scaled result lands below the decimal lower bound (GH 5110).
+    bound = 18707220957835557353007165858768422651595
+    assert -(2**167) // DECIMAL_DIVISOR == -bound - 1  # floor differs from truncation
+
+    assert c.foo(bound) == bound * DECIMAL_DIVISOR
+    assert c.foo(-bound) == -bound * DECIMAL_DIVISOR
+
+    with tx_failed():
+        c.foo(bound + 1)
+    with tx_failed():
+        c.foo(-bound - 1)
--- a/vyper/builtins/_convert.py
+++ b/vyper/builtins/_convert.py
@@ -3,6 +3,11 @@ import functools
 import math
 
 from vyper import ast as vy_ast
+from vyper.builtins._convert_rules import (
+    fixed_to_int_clamp_bounds,
+    int_to_fixed_clamp_bounds,
+    validate_convertibility,
+)
 from vyper.codegen.core import (
     LOAD,
     IRnode,
@@ -31,20 +36,11 @@ from vyper.exceptions import (
     StructureException,
     TypeMismatch,
 )
-from vyper.semantics.types import (
-    AddressT,
-    BoolT,
-    BytesM_T,
-    BytesT,
-    DecimalT,
-    FlagT,
-    IntegerT,
-    StringT,
-)
+from vyper.semantics.types import AddressT, BoolT, BytesT, StringT
 from vyper.semantics.types.bytestrings import _BytestringT
 from vyper.semantics.types.infinity import is_bounded_length
 from vyper.semantics.types.shortcuts import INT256_T, UINT160_T, UINT256_T
-from vyper.utils import DECIMAL_DIVISOR, round_towards_zero, unsigned_to_signed
+from vyper.utils import DECIMAL_DIVISOR, unsigned_to_signed
 
 
 def _FAIL(ityp, otyp, source_expr=None):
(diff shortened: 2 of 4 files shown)


Sporarum:
Since #5126 is built on top of this (if my understanding is correct), is there still value in this PR ?

banteg:
Yes, #5126 is stacked on top of this branch. I kept #5114 as the smaller base PR because it only shares the existing `convert()` legality rules with Venom and switches the rejection coverage to rely on the CI codegen matrix. #5126 then moves the validation entrypoint into the builtin/typechecker layer on top of that. If you prefer reviewing only the full refactor in #5126, I can close/de-scope #5114, but the intended value is to keep the shared-rules piece reviewable independently.

Sporarum:
Please close this then, the other one is small enough to review, and it's hard to keep track of where feedback should go otherwise

banteg:
Closing as requested so review feedback can stay concentrated on #5126, which already includes this branch and is small enough to review directly.

PR #5115, opened 2026-06-12
chore[test]: add convert() pipeline parity and boundary tests

banteg:
### What I did

Adds the enforcement layer for #5110/#5111/#5019 (fixed in #5114): a differential test that the legacy and venom pipelines agree on which `convert()` programs are valid, plus exact boundary-value tests for every numeric conversion window.

**Depends on #5114** (branched on top of it; without it, the parity test correctly fails on the venom side).

### How I did it

Two test layers in `tests/functional/builtins/codegen/test_convert_pipeline_parity.py`:

* `test_convert_acceptance_parity` — compiles the full type-pair matrix (representative integers, bytesM widths straddling the boundaries, decimal, bool, address, bytestrings straddling 32 bytes, flags; 462 pairs) under **both** pipelines, bytecode-only, and asserts identical accept/reject verdicts including exception type. Bytecode-only matters: several output formats (`ir_dict`, `metadata`, `abi`) run the legacy codegen even under `--experimental-codegen`, which is precisely how the existing matrix tests (`test_type_conversion_blocked`) missed the #5111 cluster — legacy validation masked venom's. Pointed at master, this test immediately catches the divergences.
* `test_numeric_convert_boundaries` — for each numeric conversion pair, computes the legal input window and asserts the exact edge behavior: the boundary value converts to the expected result, one-past-boundary reverts. Runs under the session pipeline, so both CI lanes exercise it. This is the structural regression test for #5110 (which survived because boundary coverage lives in fuzzing-marked tests, and the fuzzing CI lane never runs `--experimental-codegen`).

The infra gaps themselves (format masking, fuzzing lane) are tracked in #5116.

### How to verify it

```
$ pytest tests/functional/builtins/codegen/test_convert_pipeline_parity.py
$ pytest tests/functional/builtins/codegen/test_convert_pipeline_parity.py --experimental-codegen
```

476 tests, ~7s. To see it catch real bugs: check out master's `vyper/` directory and rerun — the parity test fails on the first venom-accepted illegal pair.

### Commit message

```
the legacy and venom pipelines maintain separate codegen for convert();
nothing previously enforced that they agree on which programs are
valid. the existing conversion matrix tests (test_type_conversion_blocked,
test_flag_conversion) did not catch venom accepts-invalid bugs (GH 5019,
GH 5111) because the test fixtures compile with many output formats, and
several of those (ir_dict, metadata, abi) run the legacy codegen even
under --experimental-codegen -- legacy validation masked the venom holes.
boundary-value coverage had the same gap from a different angle: it
lives in fuzzing-marked tests, and the fuzzing CI lane never runs with
--experimental-codegen (which is how GH 5110 survived).

add two test layers:

- an acceptance parity test which compiles the full type-pair matrix
  (integers, bytesM, decimal, bool, address, bytestrings, flags)
  bytecode-only under both pipelines and asserts identical verdicts,
  immune to the output-format masking. pointed at master, it
  immediately catches the GH 5111 cluster.

- numeric boundary tests asserting exact clamp behavior at the edges
  of every conversion window (boundary converts, one-past-boundary
  reverts), running under the session pipeline so both CI lanes
  exercise them. regression test for GH 5110.
```

Add differential tests enforcing that the legacy and venom pipelines agree on `convert()` validity, plus boundary-value tests for numeric conversion clamps.

### Cute Animal Picture

\dog

### DIFF
--- a/tests/functional/builtins/codegen/test_convert.py
+++ b/tests/functional/builtins/codegen/test_convert.py
@@ -818,3 +818,69 @@ def test_convert() -> uint256:
     """  # noqa: E501
     c = get_contract(code)
     assert c.test_convert() == 42
+
+
+FLAG_PREAMBLE = """
+flag Roles:
+    ADMIN
+    USER
+"""
+
+# conversions which the venom pipeline accepted due to missing input
+# validation (GH 5111, GH 5019). the conversion matrix is shared between
+# pipelines, so both must reject these at compile time.
+ILLEGAL_CONVERSIONS = [
+    ("String[20]", "uint256"),
+    ("String[20]", "address"),
+    ("Roles", "uint8"),
+    ("Roles", "bool"),
+    ("Roles", "decimal"),
+    ("uint8", "Roles"),
+    ("bytes32", "Roles"),
+    ("bool", "Roles"),
+    ("bool", "address"),
+    ("address", "decimal"),
+    ("address", "int256"),
+    ("decimal", "address"),
+    ("decimal", "Roles"),
+]
+
+
+@pytest.mark.parametrize("i_typ,o_typ", ILLEGAL_CONVERSIONS)
+@pytest.mark.parametrize("use_venom", [False, True])
+def test_illegal_conversions_blocked(i_typ, o_typ, use_venom):
+    preamble = FLAG_PREAMBLE if "Roles" in (i_typ, o_typ) else ""
+    code = f"""{preamble}
+@external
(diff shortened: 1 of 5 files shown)


banteg:
Closing this one as de-scoped. It adds direct legacy-vs-Venom differential tests, and after the arithmetic test feedback we should rely on the CI backend matrix plus targeted oracle/boundary tests instead of asking reviewers to evaluate another pipeline-parity PR. The remaining convert legality work is covered in the focused convert PRs.

Issue #5116, opened 2026-06-12
venom test lanes validate against legacy codegen: output formats mask venom bugs, fuzzing lane never runs venom


banteg:
### Version Information

* vyper Version (output of `vyper --version`): 0.5.0a3, master @ 9b5baa003
* OS: macos
* Python Version (output of `python --version`): 3.12.13

### What's your issue about?

Two related CI/test-infrastructure gaps mean the venom pipeline's compile-time validation and runtime clamp behavior are largely untested, even though the test suite appears to cover them.

**1. Legacy codegen masks venom bugs in `--experimental-codegen` test runs.**

Several output formats run the *legacy* codegen even when `experimental_codegen` is set:

```
$ python -c "..."  # compile bool->address convert, experimental_codegen=True, per format
ir_dict             TypeMismatch: Can't convert bool to address
ir_runtime_dict     TypeMismatch: Can't convert bool to address
metadata            TypeMismatch: Can't convert bool to address
abi                 TypeMismatch: Can't convert bool to address
bytecode            OK        <-- venom accepts what legacy rejects
```

The test harness (`tests/conftest.py::output_formats`) compiles nearly all output formats for every `get_contract()` call. So in the `--experimental-codegen` CI lanes, every compile-rejection test (e.g. `test_type_conversion_blocked`, which covers the full BASE_TYPES x BASE_TYPES matrix) is actually validated by the *legacy* pipeline's checks, not venom's. This is how the accepts-invalid cluster in #5111 / #5019 survived: the venom pipeline compiles `convert(b, address)` to bytecode, but the suite never notices because `ir_dict`/`metadata`/`abi` raise first.

(That `abi` triggers legacy codegen at all is arguably its own bug — ABI generation should be derivable from semantic analysis alone.)

**2. The fuzzing lane never runs with `--experimental-codegen`.**

[...]

PR #5122, opened 2026-06-12
chore[test]: add arithmetic pipeline parity and boundary tests

banteg:
### What I did

Same enforcement layer as #5115, for the arithmetic surface: a differential harness asserting that the legacy pipeline, the venom pipeline, and an exact Python oracle agree on every arithmetic operation at type boundaries. Unlike #5115 this is standalone (no fix PR required) — the review of `vyper/codegen_venom/arithmetic.py` and the numeric builtins against legacy found **no divergences**; this PR locks that in.

Motivation: the safe/unsafe arithmetic codegen is hand-ported between pipelines (the same shape that produced #5110/#5111 in `convert()`), and numeric boundary coverage lives in fuzzing-marked tests which never run under `--experimental-codegen` in CI (#5116). Until that lane exists, venom's clamps at type boundaries have no CI coverage at all.

### How I did it

One test file, `tests/functional/codegen/types/numbers/test_arithmetic_pipeline_parity.py`. For each op it compiles one contract per type **bytecode-only** under both pipelines (immune to the #5116 format masking), deploys both, and runs boundary-rich operand sets, asserting each pipeline matches an exact Python oracle (value or revert):

* `+ - * // %` over 16 integer widths — including 136/248, which straddle the >128-bit `safemul` overflow-check threshold — plus decimal (scale-then-truncate ordering); full boundary cross-products, including int256 `MIN * -1` and `MIN // -1`
* unary minus (`-MIN` reverts), `abs`-class behavior via boundary probes
* `**` with literal exponent and literal base (incl. `(-2) ** y`, exercising folded-negative-literal extraction); bounds from `calculate_largest_base/power` probed at ±1
* `unsafe_add/sub/mul/div` two's-complement wrap semantics, `x/0 == 0`
* `<< >>` at shift amounts 0/1/8/255/256/257/2^256−1 incl. `sar` sign-fill
* `floor`/`ceil` negative rounding at ±MIN/MAX
* `uint2str` vs `str()`, `math.isqrt` vs `math.isqrt`, `math.sqrt` bit-identical across pipelines

~50k checks, ~3s wall clock. The harness is mutation-tested: disabling venom's int256 `MIN // -1` special case makes it fail at exactly that operand pair.

### How to verify it

```
$ pytest tests/functional/codegen/types/numbers/test_arithmetic_pipeline_parity.py
```

To see it catch a real bug, comment out the int256 special case in `vyper/codegen_venom/arithmetic.py::safe_floordiv` and rerun.

### Commit message

```
the safe/unsafe arithmetic codegen is hand-ported between the legacy and
venom pipelines, and the fuzzing CI lane (where numeric boundary
coverage lives) never runs with --experimental-codegen -- so clamp
behavior at type boundaries had no venom CI coverage at all (see GH 5116
for the general problem).

add a differential harness: for every arithmetic op (safe binops, unary
minus, pow with literal base/exponent, unsafe_*, shifts, floor/ceil,
uint2str, math.isqrt/math.sqrt), compile bytecode-only under BOTH
pipelines and execute boundary-rich operand sets, comparing each
pipeline against an exact python oracle (or against each other where the
spec is the implementation, e.g. math.sqrt). bytecode-only compilation
matters: several output formats run legacy codegen even under
--experimental-codegen, which masks venom-only bugs in fixture-based
tests.

covers all int widths straddling the >128-bit safemul threshold
(136/248), the int256 MIN*-1 / MIN//-1 special cases, decimal
scale-then-truncate ordering, calculate_largest_base/power bounds
probed at +/-1, and two's-complement wrap semantics for unsafe ops.
~50k checks run in about 3 seconds. the harness was mutation-tested:
disabling venom's int256 MIN//-1 special case fails at exactly that
operand pair.
```

Add differential boundary tests asserting the legacy and venom pipelines produce identical arithmetic behavior (values and reverts) at type boundaries.

### Cute Animal Picture

dog

### DIFF
--- a/tests/functional/codegen/types/numbers/test_arithmetic_pipeline_parity.py
+++ b/tests/functional/codegen/types/numbers/test_arithmetic_pipeline_parity.py
@@ -1,0 +1,364 @@
+"""
+Differential tests for safe/unsafe arithmetic: for every operation and a
+boundary-rich operand set, the legacy pipeline, the venom pipeline, and an
+exact Python oracle must all agree (same value, or revert in both).
+
+Compiles bytecode-only under each pipeline explicitly, so this is immune
+to the output-format masking where legacy codegen runs (and validates)
+even under --experimental-codegen.
+"""
+
+import itertools
+import math
+
+import pytest
+
+from tests.evm_backends.base_env import EvmError
+from vyper.compiler import compile_code
+from vyper.compiler.settings import Settings
+from vyper.semantics.types import DecimalT, IntegerT
+from vyper.utils import evm_div, evm_mod
+
+REVERT = object()
+
+INT_WIDTHS = [8, 16, 64, 128, 136, 160, 248, 256]
+INT_TYPES = [IntegerT(s, w) for w in INT_WIDTHS for s in (True, False)]
+
+# exponent/base literals baked into the per-type contract
+POW_EXPS = [2, 3, 7]
+POW_BASES = [2, 7]
+POW_NEG_BASES = [-2]
+
+
+def _boundary_values(typ):
+    lo, hi = typ.int_bounds
+    vals = {lo, lo + 1, 0, 1, 2, 3, 7, hi - 1, hi, hi // 2, hi // 2 + 1}
+    s = math.isqrt(hi)
+    vals |= {s, s + 1}
+    if typ.is_signed:
+        vals |= {-1, -2, -3, -7, -s, -s - 1, lo // 2, lo // 2 - 1}
+    if isinstance(typ, DecimalT):
+        d = typ.divisor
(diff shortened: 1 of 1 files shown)


charles-cooper (changes requested):
i don't think we really want differential tests btwn pipelines. the CI is designed to test each pipeline separately. in this case, legacy<=>oracle and oracle<=>venom should imply legacy<=>venom transitively

charles-cooper:
moving to draft so we remember not to re-review

banteg:
Closing this one since the requested pipeline-differential approach is not wanted here. The pipeline-specific CI/oracle coverage is the better place for these checks.

PR #5126, opened 2026-06-12
refactor[lang]: share conversion rules and bounds

banteg:
### What I did

Remove duplicate `convert()` legality checks from legacy and Venom lowering, and share numeric clamp bounds, including the negative decimal boundary fix for #5110.

Current master already validates type pairs in `Convert.infer_arg_types` through `_convert_rules.validate_convertibility` (merged in #5127). This PR uses that implementation instead of adding a second semantic matrix. It does not claim a new change to analysis-only rejection timing.

### How I did it

Keep the shared semantic validator as the authority for legal source/target pairs. Remove redundant backend type restrictions, retain value-dependent bounds checks and the runtime length check for unbounded bytes, and share integer/decimal clamp calculations through `_convert_bounds.py`.

The conversion-matrix test checks the shared validator against the existing test oracle. Direct literal inference and full-compile behavior remain covered.

### How to verify it

Conversion syntax and functional suites pass on both pipelines: 11,287 passed per backend.

The Venom unbounded bytes/string/dynamic-array runtime suite also passes: 209 passed. `make lint` passes.

### Commit message

```text
Conversion type-pair legality is already checked during semantic
analysis. Remove duplicate backend restrictions so the shared rules
remain authoritative, while preserving runtime bounds checks for value-
dependent conversions and unbounded bytestring lengths.

Share integer and decimal clamp windows between both lowerers, including
truncating division at the negative decimal boundary.
```

Fix the negative decimal conversion clamp boundary and share conversion bounds between backends.

### Cute Animal Picture

shiba


banteg:
note that this doesn't touch the legacy pipeline as per charles ask, so it's redundant and it meant to demonstrate the shape for where things could more logically belong. this pr assumes a buy-in and some further careful cleanup work.

vyper/codegen_venom/builtins/convert.py
```diff
@@ -144,12 +152,9 @@ def _to_address(
     """
     Convert to address (160-bit unsigned).
 
-    From signed integers: disallowed (type checker handles this)
+    From signed integers: disallowed (rejected by the conversion matrix)
```
Sporarum: I remember seeing something about type checking converts in the type checker, so this was probably not incorrect

(Not sure if it was better before however)

tests/functional/builtins/codegen/test_convert.py
```diff
@@ -818,3 +818,69 @@ def test_convert() -> uint256:
...
+    code = f"""{preamble}
+@external
+def foo(x: {i_typ}) -> {o_typ}:
+    return convert(x, {o_typ})
+    """
+    settings = Settings(experimental_codegen=use_venom, enable_decimals=True)
+    # compile bytecode only: other output formats run the legacy pipeline
+    # even when experimental_codegen is set, masking venom-only bugs
+    with pytest.raises(TypeMismatch):
+        compile_code(code, output_formats=("bytecode",), settings=settings)
```
Sporarum: Should check the error message as well (`assert ==` not `matches` on `raises`)

tests/functional/syntax/test_convert.py
```diff
@@ -0,0 +1,200 @@
...
+    types += [BytesT(32), BytesT(33), StringT(32), StringT(33)]
+    return types
+
+
+@pytest.mark.parametrize("in_typ,out_typ", itertools.product(_type_instances(), repeat=2))
+def test_frontend_matches_shared_rules(in_typ, out_typ):
+    # the frontend must reject exactly the pairs the shared conversion
+    # matrix rejects (modulo the same-type check, which raises earlier)
+    if in_typ.is_subtype_of(out_typ):
+        return  # same-type conversions are blocked before the matrix check
```
Sporarum: Maybe should check they are blocked instead ?

tests/functional/syntax/test_convert.py
```diff
@@ -0,0 +1,200 @@
...
+    return convert(x, {out_typ})
+    """
+
+    try:
+        validate_convertibility(in_typ, out_typ)
+        legal = True
+    except (TypeMismatch, StructureException):
+        legal = False
+
+    if legal:
```
Sporarum: This test feels a bit circular, since we check the output is what we expect based on the internals which produce it

This feels like it should instead be lists of valid/invalid conversions
(And that is also added above)

vyper/builtins/_convert_rules.py
```diff
@@ -0,0 +1,115 @@
+"""
```
Sporarum: I don't think here is really "in the typer", to me it feels like this logic should be in the definition of the convert built-in, and the inputs checked in the same way other function inputs are checked

Sporarum (changes requested):
Nice work !

To me the most important change is to move the check actually in the type checker, currently it looks like it was just moved to a shared helper
(I explain in more details in one of the comments)

banteg:
Addressed in `2b27c42e4`:

- Moved the type-pair legality matrix into the `Convert` builtin itself (`Convert._validate_type_pair()`), so this is visibly part of builtin/typechecker validation.
- Removed codegen-side conversion-pair revalidation; the shared helper is now only `_convert_bounds.py` for numeric clamp bounds used by both lowerers.
- Tightened `test_illegal_conversions_blocked` to assert the exact `TypeMismatch.message`.
- Replaced the circular frontend-vs-helper product test with explicit valid/invalid frontend cases and added an exact same-type invalid case.

Validation:

- `uv run --python 3.12 -m pytest tests/functional/syntax/test_convert.py tests/functional/builtins/codegen/test_convert.py -k "convert_fail_at_typechecking or convert_pass or illegal_conversions_blocked" -q`
- `uv run --python 3.12 ruff check vyper/builtins/_convert_bounds.py vyper/builtins/_convert.py vyper/builtins/functions.py vyper/codegen_venom/builtins/convert.py tests/functional/syntax/test_convert.py tests/functional/builtins/codegen/test_convert.py`
- `uv run --python 3.12 python -m py_compile vyper/builtins/_convert_bounds.py vyper/builtins/_convert.py vyper/builtins/functions.py vyper/codegen_venom/builtins/convert.py tests/functional/syntax/test_convert.py tests/functional/builtins/codegen/test_convert.py`

vyper/builtins/functions.py
```diff
@@ -194,6 +194,68 @@ def build_IR(self, expr, args, kwargs, context):
 
 class Convert(BuiltinFunctionT):
     _id = "convert"
+    _allowed_inputs: dict[type, tuple[type, ...]] = {
```
Sporarum: I think this deserves a docstring that it goes from destination type to source type (in other words, second parameter value to first parameter expected type)

Probably not on the value itself, but on the enclosing class

vyper/builtins/functions.py
```diff
@@ -194,6 +194,68 @@ def build_IR(self, expr, args, kwargs, context):
...
+            if value_type.maxlen > max_bytes:
+                cls._convert_fail(value_type, target_type, node)
+
+        if isinstance(target_type, _BytestringT):
+            # widening a bytestring within the same class is not a real
+            # conversion -- the assignment is already legal without convert()
+            if isinstance(value_type, type(target_type)) and value_type.maxlen <= target_type.maxlen:
+                cls._convert_fail(value_type, target_type, node)
+
+        # flags only convert to and from uint256
```
Sporarum: One of these could easily be redundant, right ?
By changing `FlagT: (IntegerT,)` to `FlagT: (UINT256_T,)`

Maybe _allowed_inputs should be a function instead, housing the full "what types can be converted to X?" logic
That way both of these would be redundant, as well as similar logic in this function
(Technically would remove the need for this whole function ?)

vyper/builtins/functions.py
```diff
@@ -234,9 +296,7 @@ def infer_arg_types(self, node, expected_return_typ=None):
         if isinstance(value_type, DArrayT) and not is_bounded_length(value_type.count):
             raise CodegenPanic("convert not yet implemented for unbounded sequence type")
 
-        # reject illegal conversion pairs at typechecking time, for all
-        # backends. codegen re-validates as defense-in-depth.
-        validate_convertibility(value_type, target_type, node)
+        self._validate_type_pair(value_type, target_type, node)
```
Sporarum: Since this validates the argument types, shouldn't it be in `infer_arg_types` ?

vyper/codegen_venom/builtins/convert.py
Sporarum: Has not been addressed

banteg:
Addressed the fresh convert feedback in ba43a560c. Changes: added a Convert class docstring explaining the target-to-source direction, moved exact FlagT <-> uint256 handling into the target-keyed source-type lookup, clarified that the Venom codegen path relies on typechecker rejection, and left an explicit comment at the infer_arg_types validation point. Focused validation: `uv run --python 3.12 -m pytest tests/functional/syntax/test_convert.py tests/functional/builtins/codegen/test_convert.py -q` (1432 passed); `git diff --check`; `uv run --python 3.12 black --check vyper/builtins/functions.py vyper/codegen_venom/builtins/convert.py`.

vyper/builtins/functions.py
```diff
@@ -193,7 +193,96 @@ def build_IR(self, expr, args, kwargs, context):
 
 
 class Convert(BuiltinFunctionT):
+    """
+    Typecheck convert(value, target_type).
```
Sporarum: This node represents convert as a whole, not just typechecking it

vyper/builtins/functions.py
```diff
@@ -193,7 +193,96 @@ def build_IR(self, expr, args, kwargs, context):
...
+
+    The allowed-source lookup is target-keyed: the second argument's
```
Sporarum: Please mention the field by name

vyper/builtins/functions.py
```diff
@@ -193,7 +193,96 @@ def build_IR(self, expr, args, kwargs, context):
...
+        BytesT: (StringT, BytesT),
+        StringT: (BytesT, StringT),
+    }
+
+    @staticmethod
+    def _convert_fail(value_type, target_type, node):
+        raise TypeMismatch(f"Can't convert {value_type} to {target_type}", node)
+
+    @classmethod
+    def _source_types_for_target(cls, target_type):
```
Sporarum: Is it possible to refactor all of `_validate_type_pair` into this format ?

For example:
```
# addresses are unsigned
        if (
            isinstance(value_type, AddressT)
            and isinstance(target_type, IntegerT)
            and target_type.is_signed
        ):
            cls._convert_fail(value_type, target_type, node)
```
Can be replaced by this function only returning positive types when inputted with AddressT

vyper/builtins/functions.py
```diff
@@ -193,7 +193,96 @@ def build_IR(self, expr, args, kwargs, context):
...
+        if isinstance(target_type, FlagT):
+            return (UINT256_T,)
+
+        allowed = cls._source_types_by_target_type.get(type(target_type))
+        if allowed is None:
+            return None
+
+        # flags only convert to uint256
+        if isinstance(target_type, IntegerT) and target_type != UINT256_T:
+            return tuple(i for i in allowed if i is not FlagT)
```
Sporarum: This condition should be the other way arround:
Add FlagT if the target type is uint256

vyper/builtins/functions.py
banteg: Done in `a2a8ef75f`: the class docstring now describes the `convert(value, target_type)` builtin, not just typechecking.

vyper/builtins/functions.py
banteg: Done in `a2a8ef75f`: the docstring now names `_source_types_by_target_type` explicitly.

vyper/builtins/functions.py
```diff
@@ -224,32 +224,50 @@ def _source_types_for_target(cls, target_type):
         if allowed is None:
             return None
 
-        # flags only convert to uint256
-        if isinstance(target_type, IntegerT) and target_type != UINT256_T:
-            return tuple(i for i in allowed if i is not FlagT)
+        if isinstance(target_type, AddressT):
+            # addresses are unsigned, so only unsigned integer inputs are valid
+            return (BytesM_T,) + IntegerT.unsigneds() + (BytesT(32),)
```
Sporarum: I'm not sure, but `BytesM_T` (e.g. `bytes2`) being bare here feels weird, especially since `BytesT` (e.g. `Bytes[2]`) is parametrized

So I would expect something like `BytesM_T(32)` here

vyper/builtins/functions.py
```diff
@@ -224,32 +224,50 @@ def _source_types_for_target(cls, target_type):
...
+
+        if isinstance(target_type, IntegerT):
+            allowed = (IntegerT, DecimalT, BytesM_T)
+            if target_type.is_signed:
+                # addresses are unsigned
+                allowed += (BoolT,)
+            else:
+                allowed += (AddressT, BoolT)
```
Sporarum:
```suggestion
            if target_type.is_signed:
                # addresses cannot be converted to signed integers
                allowed += (BoolT,)
            else:
                allowed += (AddressT, BoolT)
```

Sporarum: BoolT is added in both branches, should be equivalent to:

```suggestion
            allowed = (IntegerT, DecimalT, BytesM_T, BoolT)
            if not target_type.is_signed:
                allowed += (AddressT,)
```
(unless order matters)

If you merge this change, don't forget the `# addresses cannot be converted to signed integers` proposal above as well

vyper/builtins/functions.py
```diff
@@ -224,32 +224,50 @@ def _source_types_for_target(cls, target_type):
...
+            if target_type == UINT256_T:
+                # flags only convert to uint256
+                allowed += (FlagT,)
+            return allowed + (BytesT(32),)
```
Sporarum: This is added no matter what, it should also be with the initial `allowed`

Again, unless order matters, please tell me if that is the case

vyper/builtins/functions.py
Sporarum: Note that this applies to the method as a whole, not only this line

vyper/builtins/functions.py
```diff
@@ -224,32 +224,50 @@ def _source_types_for_target(cls, target_type):
...
+
+        if isinstance(target_type, DecimalT):
+            return (IntegerT, BoolT, BytesM_T, BytesT(32))
+
+        if isinstance(target_type, BytesM_T):
+            allowed = []
+            allowed.extend(i for i in IntegerT.all() if i.bits <= target_type.m_bits)
+            if DecimalT().bits <= target_type.m_bits:
+                allowed.append(DecimalT)
+            allowed.append(BytesM_T)
```
Sporarum: Isn't there a restriction on which `bytes<a>` can convert to which other `bytes<b>` ?

banteg: Current behavior has no size restriction for `bytesM -> bytesN`; `can_convert()` accepts any `BytesM_T` pair. Size restrictions remain for integer/decimal/address inputs to `bytesM`, and dynamic `Bytes` inputs are bounded through `BytesT(target_type.m)`. `0ec16ba0e` keeps that behavior and expresses the `bytesM` sources concretely with `BytesM_T.all()`.

vyper/builtins/functions.py
banteg: Done in `0ec16ba0e`: `_source_types_for_target()` now expands allowed `bytesM` sources with `BytesM_T.all()` instead of using a bare `BytesM_T`, including the bool/address/integer/decimal/bytesM target cases. Focused validation: `uv run --python 3.12 -m pytest tests/functional/syntax/test_convert.py tests/functional/builtins/codegen/test_convert.py -q` (1432 passed); `uv run --python 3.12 black --check ...`; `uv run --python 3.12 ruff check vyper/builtins/functions.py`; `uv run --python 3.12 python -m py_compile ...`; `git diff --check`.

vyper/builtins/functions.py
banteg: I kept this expanded intentionally to preserve source-order. `BoolT` is common to both branches, but `AddressT` belongs before it for unsigned integer targets in the source matrix.

vyper/builtins/functions.py
Sporarum: This seems equally wrong, why do we check `Bytes[32]` specifically, but any `bytes<n>` is allowed ?
Passing all tests can also mean a test-case is missing !

vyper/builtins/functions.py
Sporarum: Does this have any measurable effect ? If not, change it to be more readable as suggested.

Also the source matrix is a concept created inside this PR, so it feels weird to use it a justification for something being the way it is
(again unless there was a good reason for it to be done that way)

vyper/builtins/functions.py
Sporarum: This seems supported by `can_convert` in `test_convert.py`
(That's very weird, and should maybe be changed, but that is outside of the scope of this PR)

vyper/builtins/functions.py
Sporarum: According to `can_convert` in `test_convert.py`:
(Note that it is the opposite logic, we check the input type first, and then the output type)
```python
    if isinstance(i_typ, BytesM_T):
        if isinstance(o_typ, BytesT):
            # bytesN must be of equal or smaller size to the input
            return bytes_of_type(i_typ) <= bytes_of_type(o_typ)

        return isinstance(o_typ, (DecimalT, BytesM_T, IntegerT, AddressT))

    if isinstance(i_typ, BytesT):
        return isinstance(o_typ, (IntegerT, DecimalT, AddressT))
```
So it seems all `Bytes[N]` and all `BytesN` can convert to AddressT, so `Bytes[32]` should be replaced by `Bytes.all()`

Sporarum:
Please add a test that checks that `_validate_type_pair` and `can_convert` in `tests/functional/builtins/codegen/test_convert.py` agree
(Note that one raises and the other returns a boolean)

banteg:
Addressed in banteg/vyper@287a21415.

Changes:
- added a codegen matrix test that checks `Convert._validate_type_pair()` agrees with `can_convert()` over the existing `BASE_TYPES` convert matrix, treating same-type/subtype cases as invalid because `convert()` raises `InvalidType` there;
- replaced the remaining `Bytes[32]`-only dynamic-bytes source checks for address/integer/bool/decimal targets with an explicit `Bytes[1]` through `Bytes[32]` source set;
- added regression coverage for accepted word-sized dynamic bytes sources and rejected `Bytes[33]` sources.

Validation:
- `uv run --python 3.12 -m pytest tests/functional/syntax/test_convert.py tests/functional/builtins/codegen/test_convert.py -q` (`11253 passed`)
- `uv run --python 3.12 black --check vyper/builtins/functions.py tests/functional/builtins/codegen/test_convert.py`
- `uv run --python 3.12 ruff check vyper/builtins/functions.py tests/functional/builtins/codegen/test_convert.py`
- `uv run --python 3.12 python -m py_compile vyper/builtins/functions.py tests/functional/builtins/codegen/test_convert.py tests/functional/syntax/test_convert.py`
- `git diff --check`

vyper/builtins/functions.py
banteg: Updated in banteg/vyper@287a21415: dynamic bytes sources are now accepted through an explicit `Bytes[1]` through `Bytes[32]` tuple, covering narrower dynamic bytes without changing the existing `Bytes[33]` rejection. There is no `BytesT.all()` helper today, so I kept this local to `Convert`.

vyper/builtins/functions.py
banteg: Updated in banteg/vyper@287a21415. I simplified the integer-target branch and kept source order: integer/decimal, `bytesM`, address when allowed, bool, flag for `uint256`, then dynamic bytes sources. The bool/address conditional is now only the signedness-specific part.

vyper/builtins/functions.py
Sporarum: I meant BytesT.any() and BytesM_T.any()

vyper/builtins/functions.py
Sporarum: Please explain why you are keeping the order!

vyper/builtins/functions.py
banteg: Updated in banteg/vyper@f0694a483: source rules now use `BytesM_T.any()` and `BytesT.any()` where all fixed/dynamic bytes sizes are valid sources. I kept the target-sized `BytesT(target_type.m)` only for dynamic bytes into `bytesM`, where the destination width is part of the legality rule.

Validation:
- `uv run --python 3.12 -m pytest tests/functional/syntax/test_convert.py tests/functional/builtins/codegen/test_convert.py -q` (`11284 passed`)
- `uv run --python 3.12 black --check vyper/builtins/functions.py tests/functional/builtins/codegen/test_convert.py tests/functional/syntax/test_convert.py`
- `uv run --python 3.12 ruff check vyper/builtins/functions.py tests/functional/builtins/codegen/test_convert.py tests/functional/syntax/test_convert.py`
- `uv run --python 3.12 python -m py_compile vyper/builtins/functions.py tests/functional/builtins/codegen/test_convert.py tests/functional/syntax/test_convert.py`
- `git diff --check`

vyper/builtins/functions.py
banteg: The order is not a runtime or precedence dependency. I am keeping source order so the target-keyed rules remain easy to compare against `can_convert()` and the declared source classes in `_source_types_by_target_type`: integer/decimal, bytesM, address when legal, bool, flag for uint256, then dynamic bytes. That is reviewability only; the agreement test is the behavior guard.

`f0694a483` also removes the per-size bytes expansions by switching to `BytesM_T.any()` / `BytesT.any()`, so this should be less noisy while preserving that source-order layout.

charles-cooper:
@banteg merge conflict, i'm guessing due to the new flag->bytes32 rule

banteg:
Resolved in banteg/vyper@93d90c2f0 with a normal merge commit from current upstream `master` (no force push).

The conflict was the upstream `feat[lang]: allow conversions from flag to bytes32` change against this PR's move from `_input_types(...)` to `_validate_inputs` in legacy `_convert.py`. Resolution:

- kept `_validate_inputs` in `_convert.py`;
- retained upstream's `is_flag_type` codegen branch for `to_bytes_m`;
- added `FlagT` to `Convert._source_types_for_target()` for `bytes32` targets so frontend validation accepts the new rule.

Validation:

- `uv run --python 3.12 -m pytest tests/functional/syntax/test_convert.py tests/functional/builtins/codegen/test_convert.py -q` (`11286 passed`)
- `uv run --python 3.12 black --check vyper/builtins/_convert.py vyper/builtins/functions.py tests/functional/builtins/codegen/test_convert.py tests/functional/syntax/test_convert.py`
- `uv run --python 3.12 ruff check vyper/builtins/_convert.py vyper/builtins/functions.py tests/functional/builtins/codegen/test_convert.py tests/functional/syntax/test_convert.py`
- `uv run --python 3.12 python -m py_compile vyper/builtins/_convert.py vyper/builtins/functions.py tests/functional/builtins/codegen/test_convert.py tests/functional/syntax/test_convert.py`
- `git diff --check`

vyper/builtins/functions.py
```diff
@@ -231,6 +336,10 @@ def infer_arg_types(self, node, expected_return_typ=None):
         if isinstance(value_type, DArrayT) and not is_bounded_length(value_type.count):
             raise CodegenPanic("convert not yet implemented for unbounded sequence type")
 
+        # Keep conversion legality in argument inference so callers cannot get
+        # a return type for an invalid source/target pair.
+        self._validate_type_pair(value_type, target_type, node)
```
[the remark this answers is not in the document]
Sporarum: I think this is a positive change, but should be documented

vyper/builtins/functions.py
```diff
@@ -192,15 +193,119 @@ def build_IR(self, expr, args, kwargs, context):
...
+
+        if isinstance(target_type, IntegerT):
+            allowed = (IntegerT, DecimalT, BytesM_T.any())
+            if not target_type.is_signed:
+                allowed += (AddressT,)
+            allowed += (BoolT,)
+            if target_type == UINT256_T:
+                # flags only convert to uint256
+                allowed += (FlagT,)
+            return allowed + (BytesT.any(),)
```
Sporarum:
```suggestion
        if isinstance(target_type, IntegerT):
            allowed = (IntegerT, DecimalT, BytesM_T.any(), BoolT, BytesT.any())
            if not target_type.is_signed:
                allowed += (AddressT,)
            if target_type == UINT256_T:
                # flags only convert to uint256
                allowed += (FlagT,)
```

vyper/builtins/functions.py
Sporarum: Thank you.
This is not a good enough reason not to make the code simpler
And making the code simpler would also make the comparison easier
See https://github.com/vyperlang/vyper/pull/5126/changes#r3480621574 for an example

vyper/builtins/functions.py
banteg: Applied in 0c3807e02. The integer target branch now uses the simpler base tuple shape from this suggestion and returns it directly after adding the target-specific cases.

Validation:
- `uv run --python 3.12 -m pytest tests/functional/syntax/test_convert.py tests/functional/builtins/codegen/test_convert.py -q`
- `uv run --python 3.12 black --check vyper/builtins/functions.py tests/functional/syntax/test_convert.py tests/functional/builtins/codegen/test_convert.py`
- `uv run --python 3.12 ruff check vyper/builtins/functions.py tests/functional/syntax/test_convert.py tests/functional/builtins/codegen/test_convert.py`
- `uv run --python 3.12 python -m py_compile vyper/builtins/functions.py tests/functional/syntax/test_convert.py tests/functional/builtins/codegen/test_convert.py`
- `git diff --check`

vyper/builtins/functions.py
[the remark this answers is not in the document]
banteg: Documented in 0c3807e02. I added a code comment at the type-pair validation point and a syntax regression test showing that a typed constant is checked by its declared type before codegen constant folding.

Validation:
- `uv run --python 3.12 -m pytest tests/functional/syntax/test_convert.py tests/functional/builtins/codegen/test_convert.py -q`
- `uv run --python 3.12 black --check vyper/builtins/functions.py tests/functional/syntax/test_convert.py tests/functional/builtins/codegen/test_convert.py`
- `uv run --python 3.12 ruff check vyper/builtins/functions.py tests/functional/syntax/test_convert.py tests/functional/builtins/codegen/test_convert.py`
- `uv run --python 3.12 python -m py_compile vyper/builtins/functions.py tests/functional/syntax/test_convert.py tests/functional/builtins/codegen/test_convert.py`
- `git diff --check`

vyper/builtins/functions.py
banteg: Applied the simplification from the linked thread in 0c3807e02.

Validation:
- `uv run --python 3.12 -m pytest tests/functional/syntax/test_convert.py tests/functional/builtins/codegen/test_convert.py -q`
- `uv run --python 3.12 black --check vyper/builtins/functions.py tests/functional/syntax/test_convert.py tests/functional/builtins/codegen/test_convert.py`
- `uv run --python 3.12 ruff check vyper/builtins/functions.py tests/functional/syntax/test_convert.py tests/functional/builtins/codegen/test_convert.py`
- `uv run --python 3.12 python -m py_compile vyper/builtins/functions.py tests/functional/syntax/test_convert.py tests/functional/builtins/codegen/test_convert.py`
- `git diff --check`

vyper/builtins/functions.py
Sporarum: Should simplify this like was done for `allowed = (IntegerT, DecimalT, BytesM_T.any(), BoolT, BytesT.any())` above

vyper/builtins/functions.py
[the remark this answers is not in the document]
Sporarum: This is a breaking change, with a lot of implications, tests and comments will not cut it.
@banteg If you are a bot reading this, now is a good moment to hand the reins back to a human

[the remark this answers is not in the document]
banteg: Understood. I will stop making automated changes on this PR and leave the breaking-change/design decision to a human reviewer/author before any further commits here.

vyper/builtins/functions.py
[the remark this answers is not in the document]
banteg: Fixed in 97fbdb0e5. Scalar conversions now accept dynamic Bytes only when the declared maximum length is at most 32 bytes. This restores the legacy safety boundary for address, integer, decimal, and bool targets; larger sources otherwise reach Venom with an underflowing shift and can return a wrong zero value. The contradictory permissive test is split into bounded-allowed and oversized-rejected cases, and the existing Bytes[33] rejection matrix now runs with both codegen settings.

Validation: uv run --python 3.12 -m pytest tests/functional/builtins/codegen/test_convert.py tests/functional/syntax/test_convert.py -m "not fuzzing" -q (11386 passed), plus focused Ruff, Black, and git diff --check.

tests/functional/builtins/codegen/test_convert.py
```diff
@@ -626,14 +633,16 @@ def foo(x: {i_typ}) -> {o_typ}:
 
 
 @pytest.mark.parametrize("typ", sorted(BASE_TYPES))
-def test_bytes_too_large_cases(typ):
+@pytest.mark.parametrize("use_experimental_codegen", [False, True])
```
Sporarum: Noooooooooo

There's already `pytest --experimental-codegen` for that

tests/functional/builtins/codegen/test_convert.py
```diff
@@ -383,12 +383,19 @@ def test_validate_type_pair_matches_can_convert(i_typ, o_typ):
...
-def test_dynamic_bytes_sources_allowed_for_scalar_targets(i_typ, o_typ):
+def test_bounded_dynamic_bytes_sources_allowed_for_scalar_targets(i_typ, o_typ):
     Convert._validate_type_pair(i_typ, o_typ, None)
 
 
+@pytest.mark.parametrize("i_typ", [BytesT(33), BytesT(64)])
+@pytest.mark.parametrize("o_typ", [AddressT(), UINT256_T, DecimalT(), BoolT()])
+def test_oversized_dynamic_bytes_sources_rejected_for_scalar_targets(i_typ, o_typ):
+    with pytest.raises(TypeMismatch):
+        Convert._validate_type_pair(i_typ, o_typ, None)
```
Sporarum: Why does this TypeMismatch now ?

What is the behaviour on current master ?

This PR is supposed to be a strict refactor, right ?

vyper/builtins/functions.py
Sporarum: @banteg this still hasn't been done

vyper/builtins/functions.py
Sporarum: @banteg never answered

vyper/builtins/functions.py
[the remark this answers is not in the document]
Sporarum: @banteg use `!` in the PR name to document it being a breaking change, and update the PR description by describing exactly how it is a breaking change

Also add a comment to https://github.com/vyperlang/vyper/issues/4572#issue-2988277766, it should link to this PR

banteg:
handed reins back to the bot

vyper/builtins/functions.py
banteg: Yes. `_validate_type_pair()` is called directly from `Convert.infer_arg_types()` after source-type inference and the same-type check. The earlier top-level update did not answer this thread directly; codegen now retains only value-dependent checks.

vyper/builtins/functions.py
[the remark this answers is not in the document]
banteg: Done. The title is now `refactor[lang]!: validate convert() legality in the typechecker`; the PR body has a dedicated breaking-change section with the exact current-master vs PR behavior, and #5126 is linked from #4572 at https://github.com/vyperlang/vyper/issues/4572#issuecomment-4934702868.

vyper/builtins/functions.py
banteg: Done in 01189b8c4. The bool-target branch now returns one direct allowed-source tuple, matching the simpler tuple shape used by the neighboring branches.

tests/functional/builtins/codegen/test_convert.py
banteg: Current master already raises `TypeMismatch` for `Bytes[33] -> {uint256,address,bool,decimal}` during bytecode generation under both legacy and Venom; `annotated_ast_dict` still succeeds. This PR preserves the maxlen <= 32 rule but moves it into semantic analysis, so the direct validator and analysis-only output reject too. It is not a new full-compile rejection; the PR is marked breaking because the analysis-only behavior changes.

tests/functional/builtins/codegen/test_convert.py
banteg: Fixed in 01189b8c4. The local `Settings(experimental_codegen=...)` parametrization is removed; the test now uses the active compiler settings, so the existing `pytest --experimental-codegen` CI matrix covers the Venom lane.

banteg:
Addressed the fresh review feedback in 01189b8c4 after merging current master normally (a44cdc3ec): simplified the bool source tuple, restored CI-matrix backend coverage for the oversized-Bytes test, marked and documented the semantic-analysis compatibility break, and linked #5126 from #4572. Validation: focused convert suites 11,287 passed per backend; full functional syntax+builtins 13,737 passed / 16 xfailed per backend; `uv run --python 3.12 make lint` passed.

vyper/builtins/functions.py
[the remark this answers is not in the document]
Sporarum: @banteg I do not consider a breaking change an error moving from `-f annotated_ast_dict` to `<no -f param>`.
With this in mind, please clarify whether there is a breaking change here.

In particular please indicate if `FOO: constant(uint256) = 1; return convert(FOO, bytes1)` compiles on each combination of (legacy, venom) x (master, this PR)
(Again, by compile I mean, fully compiles, passing through every step of compilation.)

vyper/builtins/functions.py
Sporarum: That commit does not address my comment. (it is good however)
The issue is that we make the code more complicated than it has to be to preserve an order which does not matter.
For example `allowed.extend((BytesT(target_type.m), BoolT))` should instead be done by setting the initial `allowed`, at the top of this branch, to contain these types

vyper/builtins/functions.py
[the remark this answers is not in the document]
banteg: Confirmed against the exact base `bcb3daa53` and current head `201a239ad` using the CLI's default full-compile path (no `-f`):

| Revision | Legacy | Venom |
| --- | --- | --- |
| `master` | `TypeMismatch` | `TypeMismatch` |
| this PR | `TypeMismatch` | `TypeMismatch` |

So this example has no full-compile behavior change. The difference is limited to analysis-only compilation: `annotated_ast_dict` succeeds on `master` and raises on this PR because validation now happens during semantic analysis. I removed the `!` from the title and rewrote the PR's compile-behavior section, squash message, and changelog text accordingly.

vyper/builtins/functions.py
banteg: Fixed in `201a239ad`. The `bytesM` target branch now seeds `allowed` with `BytesM_T.any()`, `BytesT(target_type.m)`, and `BoolT`, then appends only the width-dependent integer/decimal/address/flag cases. This removes the unnecessary order-preservation steps.

Validation: both focused conversion suites pass under legacy and Venom (`11287 passed` each); Black, isort, flake8, mypy (`201` files), and `git diff --check` pass.

banteg:
Remaining review feedback is addressed in `201a239ad`:

- Simplified the `bytesM` target source list so unconditional types are declared up front.
- Verified the requested full-compile matrix at base `bcb3daa53` and head `201a239ad`: the typed-constant example raises `TypeMismatch` in all four master/PR × legacy/Venom lanes.
- Updated the title and PR description to state that there is no full-compile behavior break; the observable change is limited to earlier rejection for analysis-only output.
- Resolved the previously addressed stale threads.

Validation: `11287 passed` in each legacy and Venom conversion suite; Black, isort, flake8, mypy (`201` files), and `git diff --check` pass. Ready for re-review.

vyper/builtins/functions.py
```diff
@@ -193,15 +194,110 @@ def build_IR(self, expr, args, kwargs, context):
...
 class Convert(BuiltinFunctionT):
+    """
+    Built-in type for convert(value, target_type).
+
+    `_source_types_by_target_type` is target-keyed: the second argument's
+    destination type determines which first-argument source types it accepts.
+    """
+
     _id = "convert"
+    _source_types_by_target_type: dict[type, tuple[type, ...]] = {
```
Sporarum: Some of these are unused, right ?
for example `AddressT` is defined here, but the result is overridden in _source_types_for_target
(please check and confirm)

If that is the case, remove all entries which get overridden
If there is only a couple of entries remaining, remove `_source_types_by_target_type` all together by inlining the logic instead

vyper/codegen_venom/builtins/convert.py
```diff
@@ -208,30 +203,24 @@ def _to_int(
 
     # From decimal: divide by divisor
     if isinstance(in_t, DecimalT):
-        divisor = in_t.divisor
         # Clamp first to avoid overflow in intermediate
-        out_lo, out_hi = out_t.int_bounds
+        scaled_lo, scaled_hi = fixed_to_int_clamp_bounds(in_t, out_t)
```
Sporarum:
```suggestion
        out_lo, out_hi = fixed_to_int_clamp_bounds(in_t, out_t)
```
(and update use-sites accordingly)

"scaled" only made sense in the previous version, out_ is a much clearer name for it

vyper/codegen_venom/builtins/convert.py
```diff
@@ -208,30 +203,24 @@ def _to_int(
...
         )
         # Now divide
-        val = b.sdiv(val, IRLiteral(divisor))
+        val = b.sdiv(val, IRLiteral(in_t.divisor))
         return val
 
-    # From flag: treat as uint256, use int-to-int rules
+    # From flag: the conversion matrix guarantees out_t is uint256,
+    # so this is a no-op
     if isinstance(in_t, FlagT):
```
Sporarum, line 249: Shouldn't this be moved to the end ?
```
    # From bool: already 0 or 1, fits in any integer
    return val
```
To this part
The comment would need to be updated

vyper/codegen_venom/builtins/convert.py
```diff
@@ -281,11 +268,9 @@ def _to_decimal(
...
     # From integer: multiply by divisor
     if isinstance(in_t, IntegerT):
-        # Clamp input to valid range before scaling
-        out_lo, out_hi = out_t.int_bounds
-        # Scale bounds for pre-multiplication check
-        pre_lo = out_lo // divisor
-        pre_hi = out_hi // divisor
+        # Clamp input to valid range before scaling. Note the bounds are
+        # computed with truncating division (GH 5110).
+        pre_lo, pre_hi = int_to_fixed_clamp_bounds(out_t)
```
Sporarum:
```suggestion
        out_lo, out_hi = int_to_fixed_clamp_bounds(out_t)
```
Same as other comment.

Sporarum (changes requested):

vyper/builtins/functions.py
banteg: Confirmed and fixed in 87f5371c4. AddressT, IntegerT, BoolT, DecimalT, and BytesM_T were all overridden before their table entries could be used. Only the BytesT and StringT entries remained, so I removed _source_types_by_target_type entirely and inlined those two cases in _source_types_for_target.

vyper/codegen_venom/builtins/convert.py
banteg: Applied in 87f5371c4: the clamp result and all use sites now use out_lo and out_hi.

vyper/codegen_venom/builtins/convert.py
banteg: Applied in 87f5371c4. The bool case is now the final explicit source-type branch with an updated comment, followed by an unreachable CompilerPanic instead of an implicit catch-all return.

vyper/codegen_venom/builtins/convert.py
banteg: Applied in 87f5371c4: the integer-to-fixed clamp result and use sites now use out_lo and out_hi.

banteg:
Addressed the complete August 7 changes-requested review in 87f5371c4. I confirmed that five source-map entries were always overridden; only BytesT and StringT remained, so the map is removed and those two cases are inline. The Venom clamp variables now use out_lo/out_hi, and bool is the final explicit integer-source branch with an unreachable panic after it. Validation: the full conversion suites passed under legacy and Venom (11,287 each), and make lint plus git diff --check passed. The PR description is refreshed and all four review threads are resolved.

banteg:
Refreshed against master in 363ae3e07. Since #5127 already added `_convert_rules.validate_convertibility`, I removed this branch’s duplicate semantic matrix and reused that shared validator. The remaining change removes redundant backend legality checks and shares the clamp bounds, including the negative decimal boundary fix. Runtime checks for unbounded bytes are retained. Conversion suites pass on both pipelines (11,287 each), the Venom unbounded bytes/string/dynamic-array suite passes (209), and `make lint` passes. Updated the description to remove stale claims about master and rejection timing.

vyper/codegen_venom/builtins/convert.py
```diff
@@ -273,8 +260,11 @@ def _to_int(
     if isinstance(in_t, IntegerT):
         return _int_to_int(val, in_t, out_t, ctx)
 
-    # From bool: already 0 or 1, fits in any integer
-    return val
+    # From bool: the value is already 0 or 1 and fits in any integer
+    if isinstance(in_t, BoolT):
```
Sporarum, line 264:
```suggestion
    if isinstance(in_t, (BoolT, FlagT)):
```

Sporarum (approved):
LGTM

vyper/codegen_venom/builtins/convert.py
banteg: Applied in a066e5443, combining BoolT and FlagT in the final integer-conversion branch and removing the earlier flag-only branch. Frontend validation still restricts flags to uint256. All 11,287 conversion tests pass under Venom; make lint passes.

banteg:
The fuzzing-lane follow-up is now #5270 (draft, with this PR as its merge prerequisite). Enabling Venom exposes the negative integer-to-decimal boundary defect on current master. All 48 affected boundary cases pass with this fix, and all 15,750 reverting conversion cases pass on current master with #5270's harness changes and this PR's source patch applied. The new lane keeps those assertions strict.

PR #5127, opened 2026-06-15
feat[venom]: wire up dyn allocations

harkal:
### What I did

Add support for unbounded sequence types — `Bytes[INF]`, `String[INF]`, and
`DynArray[T, INF]` — in the Venom pipeline. These have
no compile-time maximum length; their size is computed and allocated at runtime.

### How I did it

Frontend-only, on top of the existing Venom runtime-allocation backend
(`dalloca`/`dret); the only backend change is a `dret` builder emitter.

- INF locals are stable 32-byte **pointer cells** holding the current payload
  pointer (the payload lives in relocatable scratch) — required because Venom
  codegen has no branch-aware variable-environment merge. Reads double-deref;
  reassignment and `DynArray` append reallocate an exact-size buffer.
- Runtime size math is overflow-checked (`checked_add`/`checked_mul`); untrusted
  ABI input (INF calldata/returndata/`abi_decode`) is bounds-checked against the
  readable region before allocating; trusted constructor (CODE) args keep the
  existing lenient decode.
- Bounded types stay on their existing, legacy-matching fast paths — non-INF
  contracts are unaffected.
- Internal returns forward `(ptr, runtime-size)` via `dret` with FMP adoption;
  tuples mixing word and INF members use a Venom-only dynamic tuple frame.
- Unsupported positions (nested/recursive unbounded, storage/struct/event/error
  members, ABI-dynamic `DynArray` elements, non-experimental codegen) are
  rejected at the frontend with typed errors.

Two intentional riders, called out explicitly: (1) mutating a temporary now
raises `ImmutableViolation` instead of an internal assert (a general frontend
diagnostic fix); (2) the venom metadata `IDGenerator`/`frame_info` fix for
stable, collision-free `function_id`s across the deploy and runtime passes.

### How to verify it

```bash
uv run pytest \
  tests/functional/codegen/features/test_inf_bytes_string_local_runtime.py \
  tests/functional/codegen/features/test_inf_dynarray_runtime.py \
  tests/unit/semantics/types/test_inf.py \
  tests/unit/compiler/venom/test_inf_bytes_string_locals.py \
  tests/functional/builtins/codegen/test_abi_decode.py \
  --experimental-codegen
```

Plus the full suite under `--experimental-codegen`. Non-INF gas is neutral except
`create*` (runtime-sized create buffer is a small win).

### Commit message

```
Bounded sequence types (Bytes[N], String[N], DynArray[T, N]) require a
compile-time maximum length, so a value whose size is only known at
runtime — calldata of unknown length, an external call's return, a loop-
accumulated array — had to be modeled with a conservative cap that
wastes memory and gas, or could not be expressed at all.

The Venom (--experimental-codegen) pipeline already allocates memory at
runtime through dalloca/dret and a free-memory-pointer virtual register.
This adds the frontend on top of that machinery so Bytes[INF],
String[INF] and DynArray[T, INF] lower to exact-size runtime
allocations. The backend is unchanged apart from a dret builder emitter.

An INF local is a stable 32-byte pointer cell holding the current
payload pointer, not a fixed buffer. Venom codegen has no branch-aware
variable-environment merge, so routing every read and write through a
memory cell is what keeps branch and loop semantics correct without
inventing one. The payload is relocatable: reassignment and DynArray
append allocate a fresh exact-size buffer, since an INF value carries no
spare capacity.

Because INF has no maxlen cap to bound it, all runtime size arithmetic
is overflow-checked, and untrusted ABI input (INF calldata, returndata,
abi_decode) is bounds-checked against the readable region before any
allocation; trusted constructor (CODE) args keep the existing lenient
decode. Bounded types are deliberately left on their existing, legacy-
matching fast paths, so non-INF contracts are unaffected. Internal
functions return INF values via dret, forwarding a (pointer, runtime-
size) pair and having the caller adopt the callee's free-memory pointer;
tuples mixing word and INF members use a Venom-only dynamic tuple frame.
Nested or recursive unbounded layouts, INF in storage/struct/event/error
positions, and DynArray[T, INF] with ABI-dynamic elements are rejected
at the frontend rather than partially lowered.

Two changes ride along and are intentional, not INF-specific:

- mutating a temporary value now raises ImmutableViolation instead of
  tripping an internal assert — a general frontend diagnostic fix.
- venom metadata output seeds the function-id generator from already-
  assigned ids so the deploy and runtime passes produce stable,
  collision-free function_ids; frame_info is emitted as an empty dict
  for venom, which has no legacy frame.

This also tightens the create_from_blueprint code-offset guard to
compare before the wrapped subtraction, closing an out-of-bounds read
the legacy path left open.
```

Support unbounded sequence types `Bytes[INF]`, `String[INF]`, and
`DynArray[T, INF]` under `--experimental-codegen`.

### Cute Animal Picture

image-25


tests/functional/codegen/features/test_inf_bytes_string_local_runtime.py
```diff
@@ -0,0 +1,1966 @@
...
+
+
+def _call(env, contract, signature, args_schema=None, args=None):
+    calldata = method_id(signature)
+    if args_schema is not None:
+        calldata += abi_encode(args_schema, args)
+    return env.message_call(contract.address, data=calldata)
+
+
+def test_inf_bytes_local_from_bounded(env):
```
HodanPlodky: is there a reason this is not using get_contract fixture?

tests/functional/codegen/features/test_inf_dynarray_runtime.py
```diff
@@ -0,0 +1,851 @@
...
+
+
+def _call(env, contract, signature, args_schema=None, args=None):
+    calldata = method_id(signature)
+    if args_schema is not None:
+        calldata += abi_encode(args_schema, args)
+    return env.message_call(contract.address, data=calldata)
+
+
+def test_inf_dynarray_local_from_literal(env):
```
HodanPlodky: same question as test_inf_bytes_string_local_runtime file

vyper/builtins/functions.py
```diff
@@ -226,10 +231,13 @@ def infer_arg_types(self, node, expected_return_typ=None):
             hint = "remove convert()"
             raise InvalidType(f"Already a '{target_type}' !", node, hint=hint)
 
-        if isinstance(value_type, _BytestringT) and not is_bounded_length(value_type.maxlen):
-            raise CodegenPanic("convert not yet implemented for unbounded sequence type")
         if isinstance(value_type, DArrayT) and not is_bounded_length(value_type.count):
-            raise CodegenPanic("convert not yet implemented for unbounded sequence type")
+            raise TypeMismatch(f"Can't convert {value_type} to {target_type}", node.args[0])
```
Sporarum, line 240: Why is that ?

vyper/builtins/functions.py
```diff
@@ -1598,12 +1618,24 @@ def build_IR(self, expr, args, kwargs, context):
...
     _has_varargs = True
 
     def _add_gas_estimate(self, args, should_use_create2):
         return _create_addl_gas_estimate(EIP_170_LIMIT, should_use_create2)
 
+    def infer_arg_types(self, node, expected_return_typ=None):
+        self._validate_arg_types(node)
+        bytecode_type = get_possible_types_from_node(node.args[0]).pop()
+        if is_bounded_length(bytecode_type.length) and bytecode_type.length > EIP_3860_LIMIT:
+            raise TypeMismatch(f"initcode length cannot exceed {EIP_3860_LIMIT}", node.args[0])
```
Sporarum, line 1635: This change requires checking it at runtime as well, as this been added to the codegen ?

Before this change, the check was done at value-construction-site, and then threaded through type checking

vyper/builtins/functions.py
```diff
@@ -1760,6 +1792,25 @@ class CreateFromBlueprint(_CreateBase):
...
 
+    def infer_arg_types(self, node, expected_return_typ=None):
+        self._validate_arg_types(node)
+        arg_types = [AddressT()]
+        arg_types.extend(get_exact_type_from_node(arg) for arg in node.args[1:])
+
+        raw_args = False
+        raw_args_kwarg = next(
+            (kw.value.reduced() for kw in node.keywords if kw.arg == "raw_args"), None
+        )
```
Sporarum: This feels very round-about, couldn't this be done with `get` ?

vyper/builtins/functions.py
```diff
@@ -1760,6 +1792,25 @@ class CreateFromBlueprint(_CreateBase):
     }
     _has_varargs = True
 
+    def infer_arg_types(self, node, expected_return_typ=None):
+        self._validate_arg_types(node)
+        arg_types = [AddressT()]
+        arg_types.extend(get_exact_type_from_node(arg) for arg in node.args[1:])
```
Sporarum: IIRC kwargs on builtins always need to be called as keyword arguments, and there is only a single input, so this should be equivalent to:

```suggestion
        arg_types = [AddressT()]
```

vyper/builtins/functions.py
```diff
@@ -1760,6 +1792,25 @@ class CreateFromBlueprint(_CreateBase):
...
+
+        raw_args = False
+        raw_args_kwarg = next(
+            (kw.value.reduced() for kw in node.keywords if kw.arg == "raw_args"), None
+        )
+        if isinstance(raw_args_kwarg, vy_ast.NameConstant):
+            raw_args = raw_args_kwarg.value
+
+        if raw_args and (len(arg_types) != 2 or not isinstance(arg_types[1], BytesT)):
```
Sporarum: I would prefer:

```suggestion
        if raw_args and not (len(arg_types) == 2 and isinstance(arg_types[1], BytesT)):
```

vyper/builtins/functions.py
```diff
@@ -2146,6 +2197,17 @@ def fetch_call_return(self, node):
...
 
+        arg_types = [get_exact_type_from_node(arg) for arg in node.args]
+        if any(type_contains_unbounded_sequence(arg_t) for arg_t in arg_types):
+            for arg, arg_t in zip(node.args, arg_types):
+                if type_contains_nested_unbounded_sequence(arg_t):
+                    raise StructureException(
+                        "print arguments cannot contain unbounded sequence types "
+                        "inside aggregate types",
+                        arg,
+                    )
```
Sporarum: Why ?

vyper/codegen_venom/abi/abi_decoder.py
```diff
@@ -159,15 +167,18 @@ def clamp_bytestring(
...
-        # item_end = src + 32 + length
-        item_end = b.add(src.operand, IRLiteral(32))
-        item_end = b.add(item_end, length)
-        b.assert_(b.iszero(b.gt(item_end, hi)))
+        if ctx.is_unbounded_bytestring_type(typ):
+            ctx.assert_abi_bytes_payload_in_bounds(src.operand, length, hi)
+        else:
+            item_end = b.add(src.operand, IRLiteral(32))
+            item_end = b.add(item_end, length)
+            b.assert_(b.iszero(b.gt(item_end, hi)))
```
Sporarum:
```suggestion
        else:
            # Check item_end <= hi
            # item_end = src + 32 + length
            item_end = b.add(src.operand, IRLiteral(32))
            item_end = b.add(item_end, length)
            b.assert_(b.iszero(b.gt(item_end, hi)))
```

vyper/codegen_venom/abi/abi_decoder.py
```diff
@@ -187,20 +198,26 @@ def clamp_dyn_array(
...
         elem_static_size = typ.value_type.abi_type.embedded_static_size()
-        payload_size = b.mul(count, IRLiteral(elem_static_size))
-        payload_size = b.add(payload_size, IRLiteral(32))
-        item_end = b.add(src.operand, payload_size)
-        b.assert_(b.iszero(b.gt(item_end, hi)))
+        if ctx.is_unbounded_dynarray_type(typ):
+            ctx.assert_abi_dynarray_payload_in_bounds(src.operand, count, elem_static_size, hi)
+        else:
+            payload_size = b.mul(count, IRLiteral(elem_static_size))
+            payload_size = b.add(payload_size, IRLiteral(32))
```
Sporarum, line 220:
```suggestion
        else:
            # Check payload_end <= hi
            # payload_end = src + 32 + count * elem_static_size
            payload_size = b.mul(count, IRLiteral(elem_static_size))
```

vyper/codegen_venom/abi/abi_decoder.py
```diff
@@ -236,7 +254,8 @@ def _getelemptr_abi(
 def _make_ptr_value(operand, location: DataLocation, typ) -> VyperValue:
     """Create a VyperValue with Ptr for a computed pointer."""
     if location == DataLocation.MEMORY:
-        buf = Buffer(_ptr=operand, size=typ.memory_bytes_required, annotation="abi_decoder")
+        size = None if is_unbounded_sequence_type(typ) else typ.memory_bytes_required
```
Sporarum, line 292: Maybe we could make typ.memory_bytes_required return None directly for unbounded sequence types ?

vyper/codegen_venom/abi/abi_decoder.py
```diff
@@ -273,10 +292,52 @@ def _decode_bytestring(
...
+        return
+
+    # INF destinations are exact-sized, so copy only the present ABI payload.
+    # When `hi` is provided, it proves the runtime payload is readable.
+    assert isinstance(dst, IRVariable)
+    length = ctx.builder.load(src.operand, src.location)
+    ctx.zero_bytestring_padding(dst, length)
+    copy_size = ctx.builder.add(IRLiteral(32), length)
+    ctx.builder.copy_to_memory(dst, src.operand, copy_size, src.location)
+
```
Sporarum, line 348: I would rather the conditions are swapped, so we have `if ctx.is_...`

The less double negatives the better !

vyper/codegen_venom/builtins/abi.py
```diff
@@ -83,27 +92,65 @@ def _parse_method_id(method_id_node: vy_ast.VyperNode) -> Optional[int]:
     return None
 
 
-def _create_tuple_in_memory(
-    ctx: VenomCodegenContext, args: list[IROperand], types: list
-) -> tuple[IROperand, TupleT]:
-    """Create a tuple in memory from individual args."""
+def _finish_abi_encoded_bytes(
```
Sporarum: What does finish mean in this context ?

vyper/semantics/types/utils.py
```diff
@@ -98,6 +98,8 @@ def type_from_annotation(
     # Gate uses of decimal outside of built-ins behind a flag
     if isinstance(typ, DecimalT) and node.module_node.source_id != BUILTIN:
         # is there a better place to put this check?
+        # Type construction runs during analysis, before a codegen Context
+        # exists, so read the global compiler settings directly here.
```
Sporarum: Feels like a remnant of AI self-notes

vyper/semantics/analysis/local.py
```diff
@@ -70,9 +70,26 @@
...
+    is_unbounded_sequence_type,
+    type_contains_nested_unbounded_sequence,
+    type_contains_unbounded_dynarray_with_dynamic_elements,
+    type_contains_unbounded_sequence,
+    type_contains_unsupported_unbounded_sequence,
+)
 from vyper.semantics.types.utils import type_from_annotation
 
 
+def _expr_contains_unbounded_sequence(node: vy_ast.VyperNode) -> bool:
```
Sporarum, line 82: Not sure this works if the unbounded sequence is inside the node, for example:

```python
msg.data[0]
```
(since `msg.data` is unbounded)

vyper/semantics/analysis/local.py
```diff
@@ -473,6 +490,11 @@ def visit_AnnAssign(self, node):
             )
 
         typ = type_from_annotation(node.annotation, DataLocation.MEMORY)
+        if type_contains_nested_unbounded_sequence(typ):
+            raise StructureException(
+                "Memory variables cannot contain unbounded sequence types inside aggregate types",
```
Sporarum, line 494: I'm not sure I understand what this means

vyper/semantics/analysis/local.py
```diff
@@ -995,6 +1018,24 @@ def visit_Call(self, node: vy_ast.Call, typ: VyperType) -> None:
...
+                    has_nested_unbounded = _expr_contains_unbounded_sequence(arg)
+                else:
+                    try:
+                        actual_arg_typ = get_exact_type_from_node(arg)
+                    except VyperException:
+                        has_nested_unbounded = False
+                    else:
+                        has_nested_unbounded = _expr_contains_unbounded_sequence(
+                            arg
+                        ) and not is_unbounded_sequence_type(actual_arg_typ)
```
Sporarum: Isn't that what _expr_contains_unbounded_sequence does ?

vyper/semantics/analysis/module.py
```diff
@@ -806,6 +810,20 @@ def visit_VariableDecl(self, node):
...
+                node.annotation,
+            )
+
+        if type_contains_unbounded_sequence(type_) and location in (
+            DataLocation.STORAGE,
+            DataLocation.TRANSIENT,
+            DataLocation.CODE,
+        ):
+            raise StructureException(
+                "Module variables cannot use unbounded sequence types", node.annotation
```
Sporarum, line 824: I like "module variables" I think it's quite clear, but I'm not sure it's documented anywhere ?

vyper/semantics/types/function.py
```diff
@@ -869,7 +873,13 @@ def _parse_return_type(funcdef: vy_ast.FunctionDef) -> Optional[VyperType]:
     if funcdef.returns is None:
         return None
     # note: consider, for cleanliness, adding DataLocation.RETURN_VALUE
-    return type_from_annotation(funcdef.returns, DataLocation.MEMORY)
+    ret = type_from_annotation(funcdef.returns, DataLocation.MEMORY)
+    if type_contains_unsupported_unbounded_sequence(ret):
+        raise StructureException(
+            "Function returns cannot contain unbounded sequence types inside aggregate types",
```
Sporarum, line 885: The same error is raised in `vyper/semantics/analysis/local.py`, how are the two different ?

vyper/semantics/types/function.py
```diff
@@ -1089,6 +1099,11 @@ def _parse_args(
             raise ArgumentException(f"Function argument '{argname}' is missing a type", arg)
 
         type_ = type_from_annotation(arg.annotation, DataLocation.CALLDATA)
+        if type_contains_nested_unbounded_sequence(type_):
+            raise StructureException(
+                "Function arguments cannot contain unbounded sequence types inside aggregate types",
```
Sporarum, line 1110: Same question, why is this error raised in two different spots ?

vyper/semantics/types/infinity.py
```diff
@@ -39,8 +39,102 @@ def is_bounded_length(_lengthval: LengthUpperBound) -> TypeGuard[int]:
...
+
+
 def length_to_json(length: LengthUpperBound) -> int | str:
     """Return a JSON-serializable representation of a length value."""
     if length is INF or length is WILDCARD:
         return str(length)
     return length
+
+
+def type_contains_unbounded_sequence(typ) -> bool:
```
Sporarum, line 90: This could be `is_unbounded_type` in line with the following idea from the VIP:
> (optional) Unbounded structs can be elements of fixed-size lists, the generated type is unbounded.
>
>Unbounded types are unbounded sequence types and unbounded structs.

Maybe it's too close to `is_unbounded_sequence_type` tho

vyper/semantics/types/subscriptable.py
```diff
@@ -356,6 +362,16 @@ def from_annotation(cls, node: vy_ast.Subscript) -> "DArrayT":
         if not value_type._as_darray:
             raise StructureException(f"Arrays of {value_type} are not allowed", value_node)
 
+        if type_contains_unbounded_sequence(value_type):
+            raise StructureException(
+                "Nested unbounded sequence types are not supported", value_node
```
Sporarum: They're not ?

vyper/semantics/types/user.py
```diff
@@ -111,6 +112,29 @@ def _abi_input_name(item: dict, index: int, members: dict[str, VyperType]) -> st
     return name
 
 
+def _metadata_types(node: vy_ast.VyperNode) -> Iterator[VyperType]:
```
Sporarum: This probably needs a better name, as-is I would expect it to list all "metadata types", as opposed to "all possible types for this node"

Furthermore this logic should not be needed

vyper/semantics/types/user.py
```diff
@@ -111,6 +112,29 @@ def _abi_input_name(item: dict, index: int, members: dict[str, VyperType]) -> st
...
+    typ = node._metadata.get("type")
+    if typ is not None:
+        yield typ
+
+    for key in ("possible_types_from_node_False", "possible_types_from_node_True"):
```
Sporarum: One of this is a superset of the other, and both are caches, this does not feel right

vyper/semantics/types/user.py
```diff
@@ -111,6 +112,29 @@ def _abi_input_name(item: dict, index: int, members: dict[str, VyperType]) -> st
...
+        yield from node._metadata.get(key, ())
+
+
+def _validated_expr_contains_unbounded_sequence(node: vy_ast.ExprNode) -> bool:
+    if isinstance(node, (vy_ast.Tuple, vy_ast.List)):
+        return any(_validated_expr_contains_unbounded_sequence(item) for item in node.elements)
+
+    # Event/error call validation invokes validate_expected_type/validate_kwargs first.
+    # Use that cached type metadata here instead of running expression resolution again.
```
Sporarum: Isn't this premature optimization ?
The cache is there to make that operation faster, so doing it again should not be expensive

vyper/semantics/types/user.py
```diff
@@ -416,6 +448,10 @@ def from_ErrorDef(cls, base_node: vy_ast.ErrorDef) -> "ErrorT":
 
         for member_name, node in _iter_user_type_members(base_node, "Error"):
             member_type = type_from_annotation(node.annotation)
+            if type_contains_unbounded_sequence(member_type):
+                raise StructureException(
+                    "Custom errors cannot contain unbounded sequence types", node.annotation
```
Sporarum: Why not ?

Sporarum (changes requested):

vyper/builtins/functions.py
harkal: this is handled in the Venom create lowering

vyper/builtins/functions.py
harkal: Not equivalent, this builtin has varargs. `node.args[1:]` are ctor args, not kwargs, and still need to be nferred/validated. I switched this to `super().infer_arg_types()` and split out `ctor_arg_types`.

vyper/builtins/functions.py
harkal: done, changed to the positive form

vyper/semantics/types/subscriptable.py
harkal: good catch!

tests/functional/codegen/features/test_inf_bytes_string_local_runtime.py
harkal: mostly convenience here due to raw ctor data deploys and manual ABI calls/decoding for INF values

tests/functional/codegen/features/test_inf_dynarray_runtime.py
harkal: same

vyper/builtins/functions.py
Sporarum: Oh I see

vyper/semantics/analysis/local.py
Sporarum: In other words, it feels like the try logic is superfluous, and the inside _expr_contains_unbounded_sequence already does it internally

vyper/semantics/analysis/utils.py
```diff
@@ -358,7 +363,7 @@ def types_from_List(self, node):
             else:
                 # empty list literal `[]`
                 # subtype can be anything
-                subtypes = types.PRIMITIVE_TYPES.values()
+                subtypes = _primitive_types()
```
Sporarum: Shouldn't it be something like `VyperType.any()` ?
Otherwise this would not work:
```python
flag Foo:
  pass

tmp: DynArray[Foo, 10] = [] 
```
(or if it does, we're doing something weird elsewhere)

vyper/codegen_venom/expr.py
```diff
@@ -664,15 +699,27 @@ def lower_Attribute(self) -> VyperValue:
...
 
-        # .code on address is an adhoc node handled by slice() - should not be lowered directly
-        # But "code" can be a valid struct field name, so only check for address types
-        if attr == "code" and isinstance(
-            node.value._metadata.get("type"), AddressT
-        ):  # pragma: nocover
-            raise CompilerPanic(".code requires slice() context")
+        # .code on address materializes the full runtime code as Bytes[INF].
+        # Struct fields named "code" are handled above before this address check.
+        if attr == "code" and isinstance(node.value._metadata.get("type"), AddressT):
```
HodanPlodky, line 801: this is semantic change right? I could not find similar change in legacy is it there?

Sporarum:
Note: github folds review comments, but these can contain un-resolved ones !

vyper/codegen_venom/expr.py
harkal: Yes, for Venom only. Legacy can't do INF in general and I don't think it is going to anymore.

vyper/semantics/analysis/utils.py
harkal: Good point. I changed bare `[]` to infer as `DynArray[VyperType.any(), 1]` and added a flag regression test.

vyper/semantics/types/user.py
harkal: It's a current codegen limitiationl. custom error still allocates fixed size payload

vyper/builtins/functions.py
harkal: `convert()` does not support `DynArray` at all. I updated it to reject `DynArray` source/target directly with TypeMismatch.

vyper/semantics/analysis/utils.py
Sporarum: Not related to this PR, but what is going on here ?

vyper/semantics/analysis/utils.py
```diff
@@ -363,7 +363,7 @@ def types_from_List(self, node):
             else:
                 # empty list literal `[]`
                 # subtype can be anything
-                subtypes = _primitive_types()
+                return [DArrayT(VyperType.any(), 1)]
```
Sporarum: Was it wrong before ?
Before it returned valid element types, and now we return valid list types

Also why a length of 1 ?

harkal: Doesn't seem to support 0 for dynarray, so I use 1 for validation of arraytype and no elements to infere anyway. I might be way off, my frontend knowledge is limited so feel free to push fixes directly if you know better ways for this.

charles-cooper: length of 1 seems weird, would wildcard be better? @Sporarum

Sporarum: Currently the minimum length of sequences (in their types) is 1, so this is the most precise type which can be given at the current time.

If we allowed a length of 0 it would be a better fit.

Making the length a wildcard would only make the type less precise which is undesirable.
(And it would not give any additional desirable properties)

vyper/semantics/analysis/utils.py
Sporarum: Looked into it: `_is_empty_list` is a predicate which is true for all recursively-empty lists, not just `[]`, for example: `[[], []]`

vyper/semantics/analysis/global_.py
```diff
@@ -1,12 +1,74 @@
...
+
+        if func_t.ast_def is None:  # pragma: nocover
+            raise CompilerPanic("function missing declaration node")
+        for node in func_t.ast_def.get_descendants(vy_ast.AnnAssign):
+            typ = node.target._metadata.get("type")
+            if typ is None:  # pragma: nocover
+                raise CompilerPanic("local variable missing analysis metadata")
+            _reject_legacy_unbounded_sequence(typ, node.annotation)
+
+        _validate_legacy_function_body_no_unbounded_sequences(func_t)
```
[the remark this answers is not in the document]
harkal: false

harkal:
Notes on the implementation:  https://gist.github.com/harkal/d4fad8226b19b8b76227f54285d77340

vyper/codegen_venom/stmt.py
```diff
@@ -95,6 +118,13 @@ def lower_Assign(self) -> None:
         if isinstance(target, vy_ast.Tuple):
             return self._lower_tuple_unpack()
 
+        if isinstance(target, vy_ast.Name) and target.id in self.ctx.variables:
```
HodanPlodky, line 163: why only for idents (is it because we dont have inf types as a part of other types?)

vyper/codegen_venom/stmt.py
```diff
@@ -1219,13 +1545,23 @@ def _revert_with_custom_error(self, msg: vy_ast.VyperNode, error_t: ErrorT) -> N
         assert isinstance(msg, vy_ast.Call)
 
         arg_nodes = self._custom_error_arg_nodes(msg, error_t)
-        arg_types = tuple(error_t.arguments.values())
-        args_tuple_t = TupleT(arg_types)
+        old_constancy = self.ctx.constancy
+        try:
```
HodanPlodky, line 1627: What can raise here?

vyper/codegen_venom/calling_convention.py
```diff
@@ -19,22 +26,70 @@
 MAX_STACK_RETURNS = 2
 
 
+def is_dynamic_tuple_return_type(typ: VyperType | None) -> bool:
+    return isinstance(typ, TupleT) and type_contains_unbounded_sequence(typ)
+
+
+def is_dynamic_tuple_dynamic_member_type(typ: VyperType) -> bool:
+    return is_unbounded_sequence_type(typ) or not typ._is_prim_word
```
HodanPlodky: isnt this just `not typ._is_prim_word` check?

tests/unit/compiler/venom/test_dalloca.py
HodanPlodky: this was already present but leaving comment here that it should follow the `--evm-backend` cli argument

vyper/codegen_venom/stmt.py
harkal: Yes, pointer cells are only used for direct `Bytes/``String`/`DynArray[…, INF]` locals and arguments, so the target must be a `Name`.

vyper/codegen_venom/stmt.py
harkal: `Expr.lower()` can raise, for example on a state-changing call. The finally restores `ctx.constancy`

vyper/codegen_venom/calling_convention.py
harkal: Yes. Unbounded sequences are already non-primitive, so the first check is redundant. I’ll simplify it.

HodanPlodky:
I just found issue with testing this since the INF is not present in the legacy. There are some output formats that require/work only legacy output (I think it should be only `ir_dict` and `ir_runtime_dict`), that means that test like that fails even though you pass `--experimental-codegen` to `pytest`.

```python
def test_concat2(get_contract):
    test_concat2 = """
@external
def foo(inp: Bytes[50]) -> Bytes[1000]:
    x: Bytes[50] = inp
    return concat(x, inp, x, inp, x, inp, x, inp, x, inp)
    """

    c = get_contract(test_concat2)
    assert c.foo(b"horse" * 9 + b"vyper") == (b"horse" * 9 + b"vyper") * 10
```

So it would be nice to disable those output formats for venom so you dont need to jump through hoops to test this

harkal:
@charles-cooper -> Updated the gist too: https://gist.github.com/harkal/d4fad8226b19b8b76227f54285d77340

charles-cooper (approved):

tests/functional/builtins/codegen/test_abi_decode.py
```diff
@@ -1006,9 +1012,21 @@ def run(x: Bytes[{buffer_size}]) -> Bytes[{buffer_size}]:
...
+def test_abi_decode_merge_head_and_length(env, get_contract):
+    data, expected = _run_abi_decode_merge_head_and_length(get_contract)
+
     _test_ctor_decode(env, "Bytes[256]", data, expected)
 
 
+def test_abi_decode_merge_head_and_length_venom(get_contract, compiler_settings):
+    settings = copy.copy(compiler_settings)
+    settings.experimental_codegen = True
+    _run_abi_decode_merge_head_and_length(get_contract, compiler_settings=settings)
```
Sporarum: Don't we already run all tests both with and without venom ?

tests/functional/codegen/features/test_custom_errors.py
```diff
@@ -163,3 +163,28 @@ def boom():
     data = _revert_data(excinfo)
     assert data[:4] == method_id("Fancy(string,uint256)")
     assert abi_decode("(string,uint256)", data[4:]) == ("hi", 3)
+
+
+@pytest.mark.parametrize("use_experimental_codegen", [False, True])
```
Sporarum, line 168: Same question: Don't we already do this implicitly ?

tests/functional/codegen/types/test_dynamic_array.py
```diff
@@ -1922,9 +1921,7 @@ def foo():
     self.a = [[1]]
     self.a.pop().append(2)
     """
-    c = get_contract(code)
-    with tx_failed():
-        c.foo()
+    assert_compile_failed(lambda: get_contract(code), ImmutableViolation)
```
Sporarum: We should not use this anymore, see:
```
# TODO: this should not be a fixture.
# remove me and replace all uses with `with pytest.raises`.
@pytest.fixture
def assert_compile_failed(tx_failed):
```

tests/functional/syntax/test_dynamic_array.py
```diff
@@ -74,19 +74,43 @@ def foo(x: DynArray[uint256, INF]) -> DynArray[uint256, 5]:
...
+    """,
+        StructureException,
+    ),
+    (
+        """
+@external
+def foo(x: DynArray[DynArray[uint256, 5], INF]):
+    pass
     """,
         StructureException,
```
Sporarum: Curious: What is the link between these tests and this branch ?
They look like they test a missing return more than anything

tests/functional/syntax/test_dynamic_array.py
```diff
@@ -208,54 +251,74 @@ def test_dynarray_inf_state_var():
...
-    return self.a
-
-@external
-def bar(other_b: DynArray[DynArray[uint256, INF], 5]) -> DynArray[DynArray[uint256, INF], 5]:
-    return self.b
-    """
-    compile_code(code)
+        """,
+    ):
+        with pytest.raises(StructureException):
```
Sporarum: I'd like a `as e ... assert e.value.message == "..."` here, so that we clearly document what the error looks like

tests/functional/syntax/test_msg_data.py
```diff
@@ -235,23 +235,29 @@ def foo(_value: uint256) -> uint256:
...
 def foo() -> uint256:
     bar: uint256 = convert(msg.data, uint256)
     return bar
     """
-    compiler.compile_code(code)
+    if not experimental_codegen:
+        with pytest.raises(StructureException):
+            compiler.compile_code(code)
+    else:
+        compiler.compile_code(code)
```
Sporarum: Could use `_compile_inf_dynarray_code`, no ?
(if moved somewhere more centralized)

tests/functional/syntax/test_dynamic_array.py
```diff
@@ -189,17 +226,23 @@ def foo():
     compile_code(code)
 
 
-@pytest.mark.xfail(raises=CodegenPanic, reason="unbounded sequence types not yet fully supported")
-def test_dynarray_inf_param():
+def _compile_inf_dynarray_code(code, experimental_codegen):
```
Sporarum: Is there a way to turn this into a fixture, so that we don't have to pass `experimental_codegen` manually every time ?

tests/unit/semantics/types/test_inf.py
```diff
@@ -285,6 +294,174 @@ def foo() -> Bytes[...]:
...
     ),
+    # Unbounded sequence types are not supported inside structs
+    (
+        """
+struct S:
+    x: Bytes[INF]
+    """,
+        StructureException,
+    ),
+    # Unbounded sequence types are not supported inside custom errors
```
Sporarum: Why not ?

Errors never exist at values, they don't seem different from `def foo(x: Bytes[INF])`

tests/unit/semantics/types/test_inf.py
```diff
@@ -285,6 +294,174 @@ def foo() -> Bytes[...]:
...
+    (
+        """
+error E:
+    x: Bytes[INF]
+    """,
+        StructureException,
+    ),
+    # Unbounded sequence types are not supported inside events
```
Sporarum: Why not ?

Events never exist at values, they don't seem different from `def foo(x: Bytes[INF])`

tests/unit/semantics/types/test_inf.py
```diff
@@ -285,6 +294,174 @@ def foo() -> Bytes[...]:
...
+        StructureException,
+    ),
+    # Unbounded sequence types are not supported in HashMap values
+    (
+        """
+a: HashMap[uint256, Bytes[INF]]
+    """,
+        StructureException,
+    ),
+    # Nested unbounded sequence types are not supported inside tuples.
```
Sporarum, line 348: Isn't it just that tuples are not allowed as parameter types ?

tests/unit/semantics/types/test_inf.py
```diff
@@ -294,7 +471,6 @@ def test_inf_fail(bad_code, exc):
         compiler.compile_code(bad_code)
 
 
-@pytest.mark.xfail(raises=CodegenPanic)
 def test_dynarray_inf():
```
Sporarum: Given `test_inf_module_variable_locations_rejected`, I think this test can be removed

tests/unit/semantics/types/test_inf.py
```diff
@@ -303,56 +479,490 @@ def test_dynarray_inf():
...
+        ),
+    ],
+)
+def test_inf_deferred_dynarray_shapes_rejected(typ, message):
+    code = f"""
+@external
+def foo(x: {typ}):
+    pass
+    """
+    with pytest.raises(StructureException, match=message):
```
Sporarum: There might be a way to check against something which is not a regex, avoiding the escaping, but I'm not sure if I remember right

tests/unit/semantics/types/test_inf.py
```diff
@@ -364,15 +974,337 @@ def _bar(x: Bytes[INF]) -> Bytes[INF]:
...
+def test_wildcard_return_dynamic_element_requires_expected_bound():
+    rejected = """
+interface I:
+    def foo() -> DynArray[Bytes[10], ...]: view
+
+@external
+def f(a: address) -> uint256:
+    return len(staticcall I(a).foo())
+    """
+    with pytest.raises(StructureException):
```
Sporarum: Why does this not compile ?

tests/unit/semantics/types/test_inf.py
```diff
@@ -364,15 +974,337 @@ def _bar(x: Bytes[INF]) -> Bytes[INF]:
...
+
+@external
+def emit(x: Bytes[INF]):
+    log JSONInterface.Foo(x=x)
+    """
+    input_bundle = make_input_bundle({"JSONInterface.json": json.dumps(abi)})
+    with pytest.raises(StructureException):
+        compiler.compile_code(
+            code, input_bundle=input_bundle, settings=Settings(experimental_codegen=True)
+        )
```
Sporarum: Isn't this a regression ?

tests/conftest.py
```diff
@@ -63,10 +63,15 @@ def pytest_addoption(parser):
...
 @pytest.fixture(scope="module")
-def output_formats():
+def output_formats(experimental_codegen):
     output_formats = compiler.OUTPUT_FORMATS.copy()
 
     to_drop = ("cfg", "cfg_runtime", "archive", "archive_b64", "solc_json")
+    if experimental_codegen:
+        # These formats require legacy IR, which is not available for
+        # Venom-only language features such as unbounded sequence types.
+        to_drop += ("ir_dict", "ir_runtime_dict")
```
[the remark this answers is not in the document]
Sporarum: @harkal I have the same question, why is this ok to do ?

vyper/builtins/_convert_rules.py
```diff
@@ -0,0 +1,86 @@
+"""
+Backend-independent validation rules for ``convert()``.
+
+Keeping type-pair validation in semantic analysis ensures that both the
+legacy and direct-to-Venom pipelines accept and reject the same programs.
+Value-dependent checks remain in code generation.
+"""
```
Sporarum, line 7: Does this supercede some @banteg PRs ?

vyper/builtins/_convert_rules.py
```diff
@@ -0,0 +1,86 @@
...
+    raise TypeMismatch(f"Can't convert {input_type} to {output_type}", node)
+
+
+def validate_convertibility(input_type, output_type, node=None):
+    allowed = _ALLOWED_CONVERSIONS.get(type(output_type))
+    if allowed is None:
+        raise StructureException(f"Conversion to {output_type} is invalid.", node)
+
+    if not isinstance(input_type, allowed):
+        _fail(input_type, output_type, node)
```
Sporarum: I think I'd prefer allowlist type logic if possible

By that I mean `_ALLOWED_CONVERSIONS` should store conversions which are always valid, and the logic below should allow additional special cases.
(As opposed to allowing by default, and then filtering out cases)

vyper/builtins/functions.py
```diff
@@ -1625,6 +1628,18 @@ def fetch_call_return(self, node):
...
 
+    def infer_arg_types(self, node, expected_return_typ=None):
+        self._validate_arg_types(node)
+        bytecode_type = get_possible_types_from_node(node.args[0]).pop()
+        if is_bounded_length(bytecode_type.length) and bytecode_type.length > EIP_3860_LIMIT:
+            raise TypeMismatch(f"initcode length cannot exceed {EIP_3860_LIMIT}", node.args[0])
+        ctor_arg_types = [get_exact_type_from_node(arg) for arg in node.args[1:]]
+        if any(type_contains_nested_unbounded_sequence(t) for t in ctor_arg_types):
+            raise StructureException(
+                "constructor arguments cannot contain nested unbounded sequence types", node
```
Sporarum, line 1639: If I understand correctly, nested unbounded sequence types are not allowed anywhere, maybe we should catch this much earlier then
(I'm thinking when we parse types)

Sporarum (changes requested):
Hello, here is my review

I'm not able to comment on the venom side of things, so I'll let others review that part

There are still some comments from my previous review which have not been addressed (github has a tendency to fold even unanswered comments away)

This PR adds a lot of tests, and it feels like there is some duplication among them
Some would also possibly benefit from using `pytest.mark.parametrize` to reduce boilerplate

Sporarum:
Oh and I forgot to say:

Amazing work !
I'm super excited for unbounded types to actually be able to compile !

tests/unit/semantics/types/test_inf.py
harkal: I scoped those for a follow up when I couldn't gauge the impact. I will see if I can pull it in.

tests/unit/semantics/types/test_inf.py
harkal: tuple params are fine (`(Bytes[10], uint256)` compiles). This fails on Function arguments cannot contain unbounded sequence types inside aggregate types: an aggregate with an INF member has no fixed memory layout, same rule as memory locals and structs. Tuple returns with INF are allowed since they're encoded straight into the return buffer.

tests/unit/semantics/types/test_inf.py
harkal: With no expected type, the wildcard in `DynArray[Bytes[10], ...]` resolves to INF, and `DynArray[T, INF]` requires ABI-static `T`: the decoder bounds the element count by dividing the remaining payload by a constant element size, which `Bytes[10]` doesn't have. Binding the wildcard via the return type (the accepted case) avoids it. Test now asserts that message.

tests/unit/semantics/types/test_inf.py
harkal: In what sense?

vyper/builtins/_convert_rules.py
harkal: like?

vyper/builtins/functions.py
harkal: Not quite anywhere -> `(Bytes[INF], uint256)` is a supported return shape, and tuple values with INF members flow out of such calls. Where it's position-independent it's already rejected at type construction (DynArray elements, struct/event/error members). This site is a value check: (x, y) is a tuple literal, never parsed as an annotation, so it has to live at the use site (same for abi_encode/print).

tests/conftest.py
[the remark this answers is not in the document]
harkal: INF is gated in production (ir_dict under venom raises). The fixture drop is only because the sweep also trips legacy's pre-existing risky overlap panic on contracts venom now compiles; venom bytecode never derives from that IR

tests/conftest.py
```diff
@@ -93,6 +100,23 @@
...
 
 
+@pytest.fixture
+def compile_inf_code(experimental_codegen):
+    """
+    Compile code that uses unbounded sequence types. Only venom can lower
+    them; legacy codegen rejects them at analysis time.
+    """
+
+    def _compile(code, **kwargs):
```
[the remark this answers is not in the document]
harkal: fixed

tests/functional/codegen/features/test_inf_dynarray_runtime.py
```diff
@@ -1,29 +1,27 @@
...
+def _venom_only(experimental_codegen):
+    if not experimental_codegen:
+        pytest.skip("unbounded sequence types require --experimental-codegen")
 
 
-def _deploy_venom(env, code, *, settings=None):
-    out = _compile_venom(code, ["bytecode"], settings=settings)
-    return env.deploy([], bytes.fromhex(out["bytecode"].removeprefix("0x")))
+@pytest.fixture
+def no_inlining_settings(compiler_settings):
```
Sporarum: Is this the same as in tests/functional/codegen/features/test_inf_bytes_string_local_runtime.py ?
If yes, it should probably be shared

tests/conftest.py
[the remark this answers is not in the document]
Sporarum: Oh I see, thanks !

vyper/builtins/_convert_rules.py
Sporarum: https://github.com/vyperlang/vyper/pull/5126

vyper/builtins/_convert_rules.py
```diff
@@ -16,71 +18,93 @@
...
+_CONDITIONAL_CONVERSIONS: list[
+    tuple[type, type | tuple[type, ...], Callable[[VyperType, VyperType], bool]]
+] = [
+    (BytesT, (BoolT, AddressT, IntegerT, DecimalT, BytesM_T), _fits_word),
+    (StringT, BoolT, _fits_word),
+    (IntegerT, BytesM_T, _fits_bytesm),
+    (DecimalT, BytesM_T, _fits_bytesm),
+    (AddressT, BytesM_T, _address_fits_bytesm),
+    # Addresses are unsigned.
+    (AddressT, IntegerT, _unsigned_output),
```
Sporarum: For the ones that are only used once, how does it look to use lambdas ?

Sporarum (commented):
Looking very nice !

Here are some previous comments which have slipped by you:

https://github.com/vyperlang/vyper/pull/5127#discussion_r3452179025
https://github.com/vyperlang/vyper/pull/5127#discussion_r3452188211
https://github.com/vyperlang/vyper/pull/5127#discussion_r3452214391
https://github.com/vyperlang/vyper/pull/5127#discussion_r3452472493
https://github.com/vyperlang/vyper/pull/5127#discussion_r3452502024
https://github.com/vyperlang/vyper/pull/5127#discussion_r3452520270
https://github.com/vyperlang/vyper/pull/5127#discussion_r3452533288
https://github.com/vyperlang/vyper/pull/5127#discussion_r3452545068
https://github.com/vyperlang/vyper/pull/5127#discussion_r3452548871
https://github.com/vyperlang/vyper/pull/5127#discussion_r3452571285

vyper/builtins/functions.py
harkal: same restriction as abi_encode. Unbounded types inside aggregates aren't ABI-encodable. Top-level ones are fine and print(x, y) already covers that.

vyper/codegen_venom/abi/abi_decoder.py
harkal: already applied

vyper/codegen_venom/abi/abi_decoder.py
harkal: Not so sure this is the right decision. Maybe open a separate issue for this to discuss and develope accordingly

vyper/semantics/analysis/local.py
harkal: msg.data in not indexable and would have been rejected. A slice of it will yield a bounded type which is fine

vyper/semantics/analysis/local.py
harkal: Good catch! I will fetch the type once and call the helper directly. The `try` itself has to stay however: the nested-check needs the actual type, and resolving it can raise here.

vyper/semantics/types/function.py
harkal: This one catches interface wildcards (`Bytes[...]`) that resolve to `INF` only at the call site. Will add a comment noting it.

vyper/semantics/types/function.py
harkal: Different objects: this one rejects nested `INF` written in a parameter annotation, the local.py one rejects argument values whose type only becomes nested-unbounded at the call site.

vyper/semantics/analysis/module.py
harkal: "module variables" here is just shorthand for storage/transient/immutable declarations -> types.rst

vyper/semantics/types/infinity.py
harkal: Keeping the name cause `contains` vs `is` distinguishes the recursive check from the direct one.

vyper/builtins/_convert_rules.py
harkal: Partially... whichever lands second rebases the other

vyper/builtins/_convert_rules.py
harkal: I think it looks good. I will "take it"! 😄

Sporarum (approved):
LGTM !

PR #5134, opened 2026-06-18
fix[ux]: panic explicitly in `safe_pow()` for two-variable case

### LINKED ISSUE #5026: codegen: safe_pow() silently returns None for two-variable exponentiation instead of raising TypeCheckFailure
## Summary

`safe_pow()` in `vyper/codegen/arithmetic.py` contains an `else` branch 
that silently returns `None` when both operands of an exponentiation 
expression are runtime variables. This creates a fragile implicit contract 
with the caller in `expr.py` — if the semantic guard in `primitives.py` 
is ever bypassed, the result is a cryptic downstream crash rather than a 
clean, actionable compiler error.

## Location

- `vyper/codegen/arithmetic.py` — `safe_pow()`, else branch (~line 375)
- `vyper/codegen/expr.py` — caller, line 447

## Current behavior

```python
# arithmetic.py
def safe_pow(x, y):
    ...
    else:
        # `a ** b` where neither `a` or `b` are known
        # TODO this is currently unreachable, once we implement a way 
        # to do it safely remove the check in 
        # `vyper/context/types/value/numeric.py`
        return  # ← returns None silently
```

The caller in `expr.py` passes this return value directly into IR 
construction:

```python
elif isinstance(op, vy_ast.Pow):
    ret = arithmetic.safe_pow(x, y)
# ...
return IRnode.from_list(b1.resolve(b2.resolve(ret)), typ=out_typ)
```

If `ret` is `None`, `b2.resolve(None)` propagates `None` into 
`IRnode.from_list()`, causing a `TypeError` or `AttributeError` with 
no meaningful message pointing to the root cause.

## Why this matters

The `return None` path is protected by a single semantic guard in 
`primitives.py` that raises `InvalidOperation` at compile time. 
This is currently effective, but:

1. The codegen layer has **no independent protection**
2. The TODO itself acknowledges the path will eventually need a real 
   implementation
3. If any future change to the analysis pipeline weakens or removes 
   the semantic guard, the failure mode is an **opaque internal crash** 
   rather than a clean `TypeCheckFailure`
4. The branch has no `# pragma: nocover` annotation, inconsistent 
   with the "currently unreachable" claim

## Expected behavior

`safe_pow()` should fail explicitly and immediately with a clear message 
if it receives two non-literal operands, making the invariant 
self-documenting and the failure easy to diagnose.

## Suggested fix

```python
else:
    # `a ** b` where neither `a` or `b` are known
    # TODO this is currently unreachable, once we implement a way
    # to do it safely remove the check in primitives.py
    raise TypeCheckFailure(
        "unreachable: safe_pow() called with two non-literal operands. "
        "The semantic guard in primitives.py should have prevented this."
    )
```

This converts a silent fragile contract into an explicit, 
self-documenting invariant with a meaningful error message.

## Environment

- Vyper source: master branch (commit b4145b7)
- File: `vyper/codegen/arithmetic.py`
```

---

**Labels à ajouter** (si tu en as accès) :
- `bug`
- `codegen`

### PR DESCRIPTION
**Pull request by UnknownHacker1**
### What I did
Fixes #5026. `safe_pow()` in `vyper/codegen/arithmetic.py` had an `else` branch
(neither operand a compile-time constant) that silently `return`ed `None`. The
caller in `expr.py` feeds that into IR construction, so if the front-end guard
were ever bypassed you'd get a cryptic downstream crash instead of a clear error.
### How I did it
Replaced the bare `return` with `raise CodegenPanic("unreachable")`, matching the
defensive `raise` style used a few lines above in the same function. The branch is
unreachable in practice (the type checker rejects two-variable exponentiation
with `InvalidOperation`), so it's marked `# pragma: nocover`. Also added
`CodegenPanic` to the `vyper.exceptions` import list.
### How to verify it
- `a ** 2` (literal exponent) still compiles unchanged.
- `a ** b` (both variables) is still rejected at the front-end (`InvalidOperation`).
- `pytest tests/functional/codegen/types/numbers/test_exponents.py` passes (23).
- pre-commit (isort / black / flake8 / mypy) clean.

### Commit message
```text
the else branch of `safe_pow()` (neither base nor exponent is a compile-
time constant) silently `return`ed `None`, which propagates into IR
construction in `expr.py` and would surface as a cryptic downstream
crash if the front-end guard were ever bypassed.

raise `CodegenPanic("unreachable")` instead of returning `None`,
matching the defensive `raise` style used a few lines above. the branch
is unreachable today because the type checker rejects two-variable
exponentiation with `InvalidOperation`, so mark it `# pragma: nocover`.

fixes GH 5026
```
Make `safe_pow()` raise a clear compiler error instead of silently returning
`None` when both exponentiation operands are runtime values.
### Cute Animal Picture
image

### DIFF
--- a/vyper/codegen/arithmetic.py
+++ b/vyper/codegen/arithmetic.py
@@ -10,7 +10,7 @@ from vyper.codegen.core import (
     is_numeric_type,
 )
 from vyper.codegen.ir_node import IRnode
-from vyper.exceptions import CompilerPanic, TypeCheckFailure, UnimplementedException
+from vyper.exceptions import CodegenPanic, CompilerPanic, TypeCheckFailure, UnimplementedException
 
 
 def calculate_largest_power(a: int, num_bits: int, is_signed: bool) -> int:
@@ -370,9 +370,10 @@ def safe_pow(x, y):
                 ok = ["le", x, upper_bound]
     else:
         # `a ** b` where neither `a` or `b` are known.
-        # this branch is currently unreachable; raise instead of silently
-        # returning None so a clear compiler error surfaces if it is ever hit.
-        raise TypeCheckFailure("pow requires at least one literal operand")  # pragma: nocover
+        # this is already caught by the type checker, so this branch is
+        # unreachable; panic instead of silently returning None so we fail
+        # loudly if it is ever hit.
+        raise CodegenPanic("pow requires at least one literal operand")  # pragma: nocover
 
     assertion = IRnode.from_list(["assert", ok], error_msg="safepow")
     return IRnode.from_list(["seq", assertion, ["exp", x, y]])
--- a/vyper/codegen/arithmetic.py
+++ b/vyper/codegen/arithmetic.py
@@ -368,12 +368,9 @@ def safe_pow(x, y):
                 ok = ["and", ["sge", x, lower_bound], ["sle", x, upper_bound]]
             else:
                 ok = ["le", x, upper_bound]
-    else:
-        # `a ** b` where neither `a` or `b` are known.
-        # this is already caught by the type checker, so this branch is
-        # unreachable; panic instead of silently returning None so we fail
-        # loudly if it is ever hit.
-        raise CodegenPanic("pow requires at least one literal operand")  # pragma: nocover
+    else:  # pragma: nocover
+        # type checker guarantees pow has at least one literal operand
+        raise CodegenPanic("unreachable")
 
     assertion = IRnode.from_list(["assert", ok], error_msg="safepow")
     return IRnode.from_list(["seq", assertion, ["exp", x, y]])
--- a/vyper/codegen/arithmetic.py
+++ b/vyper/codegen/arithmetic.py
@@ -369,10 +369,10 @@ def safe_pow(x, y):
             else:
                 ok = ["le", x, upper_bound]
     else:
-        # `a ** b` where neither `a` or `b` are known
-        # TODO this is currently unreachable, once we implement a way to do it safely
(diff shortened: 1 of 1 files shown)


Sporarum:
Given this is already caught by the type checker, I would raise a CodegenPanic here instead

FYI (since this seem to be your first contribution): Please don't force push on this branch, as this makes it harder for us to review

UnknownHacker1:
Hello, thank you for your advice. I have switched it to CodegenPanic. And noted on the force pushing, won't do it again. Thanks for the review!

vyper/codegen/arithmetic.py
```diff
@@ -369,10 +369,11 @@ def safe_pow(x, y):
...
     else:
-        # `a ** b` where neither `a` or `b` are known
-        # TODO this is currently unreachable, once we implement a way to do it safely
-        # remove the check in `vyper/context/types/value/numeric.py`
-        return
+        # `a ** b` where neither `a` or `b` are known.
+        # this is already caught by the type checker, so this branch is
+        # unreachable; panic instead of silently returning None so we fail
+        # loudly if it is ever hit.
+        raise CodegenPanic("pow requires at least one literal operand")  # pragma: nocover
```
Sporarum: The following is more common in this codebase:

```suggestion
    else: #pragma: nocover
        # type checker guarantees pow has at least one literal operand
        raise CodegenPanic("unreachable")
```

(note: I think github's preview of the suggestion is broken, I mean replacing the code from the `else:` to the `raise CodegenPanic...`)

UnknownHacker1:
Done, applied your suggestion. Thanks @Sporarum !

Sporarum (approved):

PR #5135, opened 2026-06-19
chore[ci]: remove Era compiler tester workflow

banteg:
### What I did

Removed the Era compiler tester GitHub Actions workflow.

### How I did it

Deleted `.github/workflows/era-tester.yml`. The workflow was already documented as a diagnostic downstream spot check rather than a merge requirement, and the tester steps were configured as non-blocking.

The workflow now fails before reaching its final success step because its setup still installs the stale `clang-13` / `lld-13` packages on `ubuntu-latest`, where those packages are no longer available.

### How to verify it

```bash
git diff --check
ruby -e 'require "yaml"; ARGV.each { |f| YAML.load_file(f) }' .github/workflows/*.yml .github/workflows/*.yaml
```

### Commit message

```text
the era-compiler-tester workflow was a diagnostic spot check against
matter labs' downstream backend, explicitly not a merge requirement (so
we are not blocked when our IR changes break downstream). it has been
failing before reaching its final success step because the setup still
installs `clang-13` / `lld-13`, which are no longer available on
`ubuntu-latest`. since the job was already non-blocking and now cannot
even reach a useful result, drop the workflow rather than leave a
perpetually-red diagnostic in CI.
```

No changelog entry required; CI-only maintenance.


charles-cooper (approved):

Sporarum:
@charles-cooper is there anything blocking this PR ?

oe99901-ship-it:
Fhh

PR #5149, opened 2026-06-23
fix[lang]: missing no-return implements check

### LINKED ISSUE #5145: Conflicting implements interfaces trigger compiler panic
### Version Information

* OS: Linux Ubuntu Jammy
* Python Version: Python 3.11

### What's your issue about?

When a contract declares `implements: IA` AND `implements: IB`, where both interfaces declare a function `f` but with conflicting signatures (e.g. `IA: def f(): nonpayable` vs `IB: def f() -> uint256: view`), and the contract implements the IB-shape (`@view def f() -> uint256: return 1`), Vyper panics with `CompilerPanic: unhandled exception 'NoneType' object has no attribute 'is_subtype_of'`.

The interface-conformance check attempts to type-compare the implementation's return type (`uint256`) against IA's return type (`None` for the nonpayable no-return signature) and the `.is_subtype_of()` call crashes on `None`. Implementing the IA-shape (`@external def f(): pass`) or neither produces a clean `InterfaceViolation: Contract does not implement all interface functions: f()`. Only the asymmetric IB-satisfying configuration triggers the panic.

Code that caused the failure:

```vyper
interface IA:
    def f(): nonpayable

interface IB:
    def f() -> uint256: view

implements: IA
implements: IB

@external
@view
def f() -> uint256:
    return 1
```

Command run:

```
$ vyper expect/V597418_conflict_iface_panic.input.vy
```

Full output of the error:

```
AttributeError: 'NoneType' object has no attribute 'is_subtype_of'

vyper.exceptions.CompilerPanic: unhandled exception 'NoneType' object has no attribute 'is_subtype_of'
```

Source-level evidence: the interface-conformance check in `vyper/semantics/types/function.py` calls `.is_subtype_of()` on the implementation's return type, comparing against the interface's declared return type. For nonpayable-no-return interfaces, the return type is `None`. The check doesn't guard for `None` before calling `.is_subtype_of()` — the AttributeError surfaces and is re-raised as a `CompilerPanic`.

### How can it be fixed?

Guard the interface-conformance check in `vyper/semantics/types/function.py` against a `None` return type before calling `.is_subtype_of()`. The expected behavior would be a clean `InterfaceViolation: f() in interface IA expects no return but implementation returns uint256`, naming both interfaces and the mismatch.

Sporarum:
### What I did

Fix: #5145 

### How I did it

Fast-path on None-ness mismatch of return types in `compare_signature`,
then apply the covariance check only when both sides return. See commit
message below for the rationale.

### How to verify it

Two tests added under `tests/functional/syntax/test_interfaces.py`; both
fail on the prior code (one with a compiler panic, one by accepting an
invalid `implements`).

### Commit message

```
`compare_signature` only enforced return-type covariance when the
implementing function had a non-None return type: the `if return_type
and ...` guard was truthy, so a function declaring no return value would
silently satisfy an interface that does return. this led to a compiler
panic (NoneType has no is_subtype_of) when the mismatch was detected
downstream, and in other cases accepted an incorrect implements
declaration without raising.

require the None-ness of both return types to match as a fast-path, then
apply the covariance check only when both sides actually return. the
assert on `other_return_t` is there to narrow the type for mypy after
the symmetry check, which a plain `!=` comparison will not do.
```
### Cute Animal Picture

### DIFF
--- a/vyper/semantics/types/function.py
+++ b/vyper/semantics/types/function.py
@@ -660,15 +660,24 @@ class ContractFunctionT(VyperType):
         arguments, return_type = self._iface_sig
         other_arguments, other_return_type = other._iface_sig
 
-        # Contravariant
         if len(arguments) != len(other_arguments):
             return False
         for atyp, btyp in zip(arguments, other_arguments):
+            # Contravariant
             if not btyp.is_subtype_of(atyp):
                 return False
 
-        # Covariant
-        if return_type and not return_type.is_subtype_of(other_return_type):  # type: ignore
+        neither_returns = return_type is None and other_return_type is None
+
+        both_return_and_match = (
+            return_type is not None
+            and other_return_type is not None
+            and return_type.is_subtype_of(other_return_type)  # Covariant
+        )
+
+        return_types_match = both_return_and_match or neither_returns
+
+        if not return_types_match:  # type: ignore
             return False
 
         return self.mutability == other.mutability
--- a/vyper/semantics/types/function.py
+++ b/vyper/semantics/types/function.py
@@ -657,28 +657,24 @@ class ContractFunctionT(VyperType):
 
         assert self.visibility == other.visibility
 
-        arguments, return_type = self._iface_sig
-        other_arguments, other_return_type = other._iface_sig
+        arguments, return_t = self._iface_sig
+        other_arguments, other_return_t = other._iface_sig
 
         if len(arguments) != len(other_arguments):
             return False
         for atyp, btyp in zip(arguments, other_arguments):
-            # Contravariant
+            # argument checking is contravariant
             if not btyp.is_subtype_of(atyp):
                 return False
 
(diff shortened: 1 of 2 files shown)


charles-cooper:
@codex review

vyper/semantics/types/function.py
```diff
@@ -660,15 +660,24 @@ def implements(self, other: "ContractFunctionT") -> bool:
         arguments, return_type = self._iface_sig
         other_arguments, other_return_type = other._iface_sig
 
-        # Contravariant
         if len(arguments) != len(other_arguments):
             return False
         for atyp, btyp in zip(arguments, other_arguments):
+            # Contravariant
```
charles-cooper:
```suggestion
            # argument typechecking is contravariant
```

charles-cooper:
@codex review

charles-cooper (approved):

Issue #5153, opened 2026-06-24
`empty` does not constant fold


Sporarum:
### Version Information

* vyper Version (output of `vyper --version`): 0.5.0a4+commit.751931ff
* OS: NixOS 26.11
* Python Version (output of `python --version`): Python 3.11.15

### What's your issue about?

Using `empty` in a location where constant folding is required fails with `vyper.exceptions.UnfoldableNode: not foldable`:
```python
@external
def f(x: Bytes[32]) -> uint256:
    return abi_decode(x, uint256, unwrap_tuple=empty(bool))
```

### How can it be fixed?

Add a `_try_fold` to `Empty`, this is more complicated than it might seem, as some types do not have good constructors for this purpose.
For example custom flags, where additionally `empty` has a special meaning not available without it.

Sporarum:
https://github.com/vyperlang/vyper/pull/5154

Adds some xfail tests which document this

Sporarum:
One of the reasons this was not caught sooner is that:

```python
class FoldedFunctionT(BuiltinFunctionT):
    # Base class for nodes which should always be folded

    _modifiability = Modifiability.CONSTANT
```
a parent class of `Empty`, does not actually enforce having a `_try_fold` method (which `Empty` lacks)

Hopefully with this issue fixed, we would be able to add some logic that all calls to `_try_fold` on a `FoldedFunctionT` succeed (meaning method is present and it doesn't raise).

charles-cooper:
i think in general we are not guaranteed that the ast-level constant folder will fold everything that can be folded. some things are just marked as "known to be constant". if we wanted to guarantee this we would do compile-time elaboration (compile-to-venom and then interpret)

Sporarum:
> we are not guaranteed that the ast-level constant folder will fold everything that can be folded

Yes but:
1. `empty` itself is a `FoldedFunctionT`, which according to its docstring "Base class for nodes which should always be folded"
2. it also has `Constant` mutability, and a bunch of places in the compiler assume `Constant` = "will always constant-fold"

So `empty` really looks like it was designed to constant fold from the onset



> we would do compile-time elaboration (compile-to-venom and then interpret)

I don't think this is required in this case, `empty(T)` should really be specced at the language level, and not at the bytecode level.
So it makes sense to give it an implementation that speaks the language of the front-end.

charles-cooper:
[...]

PR #5154, opened 2026-06-24
fix[stdlib]: fix panic in `abi_encode`

### LINKED ISSUE #5144: abi_encode ensure_tuple non-literal triggers compiler panic
### Version Information

* OS: Linux Ubuntu Jammy
* Python Version: Python 3.11

### What's your issue about?

Compiling a contract that passes a non-literal expression (`empty(bool)`) to the `ensure_tuple` kwarg of the `abi_encode` built-in crashes the compiler with a hard internal `CompilerPanic` that leaks a raw CPython AST `AttributeError`, instead of a graceful user-facing diagnostic. (EVM version: cancun.)

The code that causes the failure:

```vyper
# pragma version ^0.4.0
@external
def f(x: uint256) -> Bytes[64]:
    return abi_encode(x, ensure_tuple=empty(bool))
```

Command run:

```
$ vyper Vd77bf0_abi_encode_ensure_tuple_empty_ice.input.vy
```

Resulting error:

```
vyper.exceptions.CompilerPanic: unhandled exception 'Call' object has no attribute 'value'
```

The compiler exits non-zero and stderr contains `unhandled exception 'Call' object has no attribute 'value'`.

Notable asymmetry with the inverse ABI operation: the same misuse on the decode side, `abi_decode(b, uint256, unwrap_tuple=empty(bool))`, raises the milder `vyper.exceptions.UnfoldableNode: not foldable` rather than a `CompilerPanic`. Using a literal value, `abi_encode(x, ensure_tuple=False)`, compiles successfully.

A full-matrix probe of all 10 `require_literal=True` kwarg slots in `builtins/functions.py` fed `empty(T)` found `abi_encode`'s `ensure_tuple` to be the sole slot producing a `CompilerPanic`; the other 9 slots uniformly produce the milder `UnfoldableNode: not foldable`.

### How can it be fixed?

The `require_literal` kwarg path in `vyper/builtins/_signatures.py:37` reads the kwarg via `get_folded_value()`:

```python
# vyper/builtins/_signatures.py:35-39
def process_kwarg(kwarg_node, kwarg_settings, expected_kwarg_type, context):
    if kwarg_settings.require_literal:
        return kwarg_node.get_folded_value().value
    return process_arg(kwarg_node, expected_kwarg_type, context)
```

For `unwrap_tuple` (abi_decode), `empty(bool)` reaches `get_folded_value()` and surfaces as `UnfoldableNode: not foldable`. For `ensure_tuple` (abi_encode, configured at `vyper/builtins/functions.py:2223`), `abi_encode`'s own build/return path instead reads `.value` directly off the raw `Call` AST node of `empty(...)` (which has no `.value` attribute) before/without going through `get_folded_value()`, producing the Python `AttributeError` that becomes the `CompilerPanic`. Routing the `ensure_tuple` kwarg read through `get_folded_value()` (locate the `.value`-on-`Call` access in the `abi_encode` `build_IR`/`fetch_call_return` path) would converge its failure mode to the milder `UnfoldableNode` seen on the other `require_literal` slots.

Sporarum:
### What I did

Fix #5144
But still emits user-visible `NotFoldable` due to https://github.com/vyperlang/vyper/issues/5153

### Commit message

```
fixes the assumption that the argument of 'ensure_tuple' is always a
literal. this is done by querying the folded value instead of the node
itself. the xfail tests are added to document an issue surrounding
constant folding 'empty', which predates this commit. cleanup: some
comments were converted to docstrings for compiler UX.
```

### Cute Animal Picture

### DIFF
--- a/tests/functional/syntax/test_abi_decode.py
+++ b/tests/functional/syntax/test_abi_decode.py
@@ -41,8 +41,7 @@ def foo(x: Bytes[32]) -> uint256:
 @external
 def foo(x: Bytes[32]) -> uint256:
     return _abi_decode(x, uint256, unwrap_tuple=(0 < 1))
-    """
-    
+    """,
 ]
 
 
--- a/tests/functional/syntax/test_abi_decode.py
+++ b/tests/functional/syntax/test_abi_decode.py
@@ -1,7 +1,7 @@
 import pytest
 
 from vyper import compiler
-from vyper.exceptions import TypeMismatch, UnfoldableNode
+from vyper.exceptions import CompilerPanic, TypeMismatch, UnfoldableNode
 
 fail_list = [
     (
@@ -50,7 +50,8 @@ def test_abi_decode_success(good_code):
     assert compiler.compile_code(good_code) is not None
 
 
-@pytest.mark.xfail(raises=UnfoldableNode)
+# UnfoldableNode on legacy, CompilerPanic in venom
+@pytest.mark.xfail(raises=(UnfoldableNode, CompilerPanic))
 def test_abi_decode_unwrap_tuple_foldable_expr():
     code = """
 @external
--- a/tests/functional/syntax/test_abi_decode.py
+++ b/tests/functional/syntax/test_abi_decode.py
@@ -1,7 +1,7 @@
 import pytest
 
 from vyper import compiler
-from vyper.exceptions import TypeMismatch
+from vyper.exceptions import TypeMismatch, UnfoldableNode
 
 fail_list = [
     (
@@ -31,13 +31,31 @@ def test_abi_decode_fail(bad_code, exc):
         compiler.compile_code(bad_code)
 
 
-valid_list = ["""
+valid_list = [
+    """
 @external
 def foo(x: Bytes[32]) -> uint256:
     return _abi_decode(x, uint256)
-    """]
+    """,
+    """
+@external
+def foo(x: Bytes[32]) -> uint256:
+    return _abi_decode(x, uint256, unwrap_tuple=(0 < 1))
+    """
+    
+]
 
 
 @pytest.mark.parametrize("good_code", valid_list)
 def test_abi_decode_success(good_code):
     assert compiler.compile_code(good_code) is not None
+
+
+@pytest.mark.xfail(raises=UnfoldableNode)
+def test_abi_decode_unwrap_tuple_foldable_expr():
+    code = """
+@external
+def f(x: Bytes[32]) -> uint256:
+    return abi_decode(x, uint256, unwrap_tuple=empty(bool))
+    """
+    assert compiler.compile_code(code) is not None
--- a/tests/functional/syntax/test_abi_encode.py
+++ b/tests/functional/syntax/test_abi_encode.py
@@ -1,7 +1,7 @@
 import pytest
 
 from vyper import compiler
-from vyper.exceptions import InvalidLiteral, TypeMismatch
+from vyper.exceptions import InvalidLiteral, TypeMismatch, UnfoldableNode
 
 fail_list = [
     (
@@ -119,9 +119,24 @@ BAR: constant(DynArray[uint256, 5]) = [1, 2, 3, 4, 5]
 def foo() -> Bytes[224]:
     return _abi_encode(BAR)
     """,
+    """
+@external
+def foo(x: Bytes[1]) -> Bytes[96]:
+    return _abi_encode(x, ensure_tuple=(0 < 1))
+    """,
 ]
 
 
 @pytest.mark.parametrize("good_code", valid_list)
 def test_abi_encode_success(good_code):
     assert compiler.compile_code(good_code) is not None
+
+
+@pytest.mark.xfail(raises=UnfoldableNode)
+def test_abi_encode_ensure_tuple_foldable_expr():
+    code = """
+@external
+def f(x: uint256) -> Bytes[64]:
+    return abi_encode(x, ensure_tuple=empty(bool))
+    """
+    assert compiler.compile_code(code) is not None
--- a/vyper/builtins/functions.py
+++ b/vyper/builtins/functions.py
@@ -108,15 +108,19 @@ SHA256_PER_WORD_GAS = 12
 
 
 class FoldedFunctionT(BuiltinFunctionT):
-    # Base class for nodes which should always be folded
+    """
+    Base class for nodes which should always be folded
+    """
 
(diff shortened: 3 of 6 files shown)


tests/functional/syntax/test_abi_decode.py
```diff
@@ -31,13 +31,31 @@ def test_abi_decode_fail(bad_code, exc):
...
+    """,
+]
 
 
 @pytest.mark.parametrize("good_code", valid_list)
 def test_abi_decode_success(good_code):
     assert compiler.compile_code(good_code) is not None
+
+
+# UnfoldableNode on legacy, CompilerPanic in venom
```
charles-cooper, line 53: where is the compiler panic from?

vyper/builtins/functions.py
```diff
@@ -2244,7 +2249,12 @@ def infer_kwarg_types(self, node):
     def fetch_call_return(self, node):
         self._validate_arg_types(node)
         ensure_tuple = next(
-            (arg.value.value for arg in node.keywords if arg.arg == "ensure_tuple"), True
+            (
```
charles-cooper: expressions that span more than one line indicate too much complexity for me, at this point we should change this to `for/else`

vyper/codegen_venom/builtins/abi.py
```diff
@@ -14,8 +14,9 @@
 from vyper.codegen.core import calculate_type_for_external_return
 from vyper.codegen_venom.abi import abi_decode_to_buf, abi_encode_to_buf
 from vyper.codegen_venom.buffer import Buffer, Ptr
+from vyper.codegen_venom.builtins.misc import _get_bool_kwarg
```
charles-cooper: should be extracted to common location

vyper/codegen_venom/builtins/misc.py
```diff
@@ -392,8 +392,10 @@ def _get_bool_kwarg(node: vy_ast.Call, kwarg_name: str, default: bool) -> bool:
     if kw_node is None:
         return default
     kw_node = kw_node.reduced()
+    # The value should be a NameConstant (True/False)
     if isinstance(kw_node, vy_ast.NameConstant):
         return kw_node.value
+    # Could also be an Int with constant value
```
charles-cooper: hm actually i don't think it can lol, this seems like a mistake

Sporarum: I copied the comments from the deleted version to put them here
I haven't checked why they were there in the first place

vyper/codegen_venom/builtins/abi.py
Sporarum: Isn't misc the common location ?

vyper/builtins/functions.py
Sporarum: I did something cleaner, see 7eb702e8765e9f7f7ccec155699a484bbc3d99c2

tests/functional/syntax/test_abi_decode.py
Sporarum: `CompilerPanic: unfoldable boolean kwarg: unwrap_tuple`
From `_get_bool_kwarg` in misc.py

I tried to make it raise an UnfoldableNode as well, but actually the proper fix is to fix https://github.com/vyperlang/vyper/issues/5153
The front-end checks that this is a Constant, so we should never be in a situation were we need to raise in the first place

vyper/codegen_venom/builtins/abi.py
charles-cooper: no? misc is the grab bag of builtins, please check

charles-cooper: https://github.com/vyperlang/vyper/blob/cf389d96dc8dd8448f1de622d72adb1a636d131a/vyper/codegen_venom/builtins/misc.py

vyper/codegen_venom/builtins/misc.py
charles-cooper: yes, i think we should also fix the mistake

vyper/codegen_venom/builtins/abi.py
Sporarum: Reverted an opened https://github.com/vyperlang/vyper/issues/5173

vyper/codegen_venom/builtins/misc.py
Sporarum: Reverted and opened https://github.com/vyperlang/vyper/issues/5173

tests/functional/syntax/test_abi_decode.py
charles-cooper: can we fix the venom pipeline too so that it doesn't panic? what is the point of "fixing" the panic in one pipeline but not the other?

Sporarum: This PR fixes the panic on both backends for cases like:
```python
_abi_decode(x, uint256, unwrap_tuple=(0 < 1))
```

This xfail test is instead due to #5153
I did not want to burden this PR with some codegen changes just to change the type of `VyperInternalException` which is reported

tests/functional/syntax/test_abi_encode.py
```diff
@@ -119,9 +119,24 @@ def foo(x: Bytes[1]) -> Bytes[68]:
...
+    """,
 ]
 
 
 @pytest.mark.parametrize("good_code", valid_list)
 def test_abi_encode_success(good_code):
     assert compiler.compile_code(good_code) is not None
+
+
+@pytest.mark.xfail(raises=UnfoldableNode)
```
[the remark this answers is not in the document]
Sporarum: This is empirically wrong, the venom tests pass

tests/functional/syntax/test_abi_decode.py
Sporarum: Claude-authored explanation of why this happens here (and not on abi_encode):

Semantic analysis passes cleanly, and the kwarg is only consumed at codegen:

- Legacy codegen: @process_inputs → process_kwarg (vyper/builtins/_signatures.py:37) calls kwarg_node.get_folded_value().value → UnfoldableNode (same underlying failure, just later).
- Venom codegen: vyper/codegen_venom/builtins/abi.py:36-48's _get_bool_kwarg uses reduced() (which returns self when there's no folded value, no exception), then does an isinstance(kw_node, vy_ast.NameConstant | vy_ast.Int) check. empty(bool) is a Call, falls through, and hits the explicit raise CompilerPanic("unfoldable boolean kwarg: …") at line 48.

charles-cooper (approved):

PR #5156, opened 2026-06-24
fix[lang]: method_id panic on non-constant

### LINKED ISSUE #5152: method_id non-literal arg triggers CodegenPanic
### Version Information

* OS: Linux Ubuntu Jammy
* Python Version: Python 3.11

### What's your issue about?

`method_id(s)` where `s` is a runtime `String[N]` variable crashes the Vyper compiler with a `CodegenPanic` instead of producing a clean error message.

Expected behavior: a clean `InvalidType: method id must be given as a literal string` error (which the source code at `vyper/builtins/functions.py:720` is prepared to raise in `_try_fold`).

Actual behavior: the `InvalidType` raise apparently doesn't fire on this path (or is swallowed), and codegen attempts to call `MethodID.build_IR(...)`, which doesn't exist. The Python class repr leaks to the user.

This is a compiler bug — `CodegenPanic` is internal-only and shouldn't be reachable from user input on otherwise valid Vyper syntax.

Reproducer:

```vyper
@external
@view
def f(s: String[20]) -> Bytes[4]:
    return method_id(s)
```

Command:

```
$ vyper vyper-tests/expect/Vbdfc11_method_id_runtime.input.vy
```

Error output:

```
CodegenPanic: unhandled exception 'MethodID' object has no attribute 'build_IR', parse_Call
```

Source-level evidence (`vyper/builtins/functions.py:711-730`):

```python
class MethodID(BuiltinFunctionT):
    _id = "method_id"
    _inputs = [("value", StringT.any())]
    _kwargs = {"output_type": KwargSettings(TYPE_T.any(), BytesT(4))}

    def _try_fold(self, node):
        validate_call_args(node, 1, ["output_type"])

        value = node.args[0].get_folded_value()
        if not isinstance(value, vy_ast.Str):
            raise InvalidType("method id must be given as a literal string", node.args[0])
        ...
```

The class defines only `_try_fold`, `fetch_call_return`, `infer_arg_types`, and `infer_kwarg_types` — there is NO `build_IR`. The intended path is that `_try_fold` constant-folds the call into a Hex/Bytes literal at compile time so codegen never sees `method_id(...)`. For a runtime argument, this fold path doesn't terminate cleanly, and codegen tries `MethodID(...).build_IR(...)`, which panics.

The documentation at https://docs.vyperlang.org/en/stable/built-in-functions.html#method_id describes `method_id(method, output_type=Bytes[4])` as taking a "method declaration formatted as a string" but does not specify it must be a literal string, so a user passing a runtime `String[20]` argument might reasonably expect a clean rejection rather than a compiler panic.

### How can it be fixed?

Ensure the non-literal-string case is rejected with a clean `InvalidType` error before codegen is reached, rather than relying solely on `_try_fold`. Either:

- Make the semantic-analysis / folding pass propagate the `InvalidType` raised in `_try_fold` (instead of swallowing it and falling through to codegen), or
- Add a `build_IR` (or equivalent fallback) on `MethodID` that raises `InvalidType("method id must be given as a literal string")` so the panic path is unreachable.

Defense-in-depth: audit other pure-fold builtins (e.g. `min_value`, `max_value`, `epsilon`, `as_wei_value`) for the same pattern — every builtin should either have `build_IR` or raise a clean error before codegen.

Sporarum:
### What I did

Fix #5152. `method_id(s)` where `s` is a runtime `String[N]` variable crashed the compiler with a `CodegenPanic: 'MethodID' object has no attribute 'build_IR'` instead of raising the intended `InvalidType: method id must be given as a literal string`.

### How I did it

`MethodID._try_fold` called `node.args[0].get_folded_value()`, which raises when the argument can't be constant-folded. The folding pass swallows that exception and skips `_try_fold` entirely, so codegen receives the un-folded `method_id(...)` call and falls through to `MethodID.build_IR`, which does not exist.

Switch to `node.args[0].reduced()`, which returns the original node unchanged when folding fails. The existing `isinstance(value, vy_ast.Str)` guard then raises the clean `InvalidType` as intended, before codegen is reached.

### How to verify it

- New syntax test in `tests/functional/syntax/test_method_id.py` asserts that `method_id(s)` with a runtime `String[20]` argument raises `InvalidType`.
- `uv run pytest tests/functional/syntax/test_method_id.py` passes.

### Commit message

```
method_id's arg is assumed to be a constant during codegen: `_try_fold` is
meant to constant-fold the call into a Bytes literal so codegen never
sees it. but `_try_fold` used `get_folded_value()`, which raises on non-
constant args; the folding pass swallows that exception, leaving codegen
to dispatch `MethodID.build_IR`, which doesn't exist — hence the
`CodegenPanic`.

use `reduced()` instead, which returns the original node when folding
fails. the existing `isinstance(value, vy_ast.Str)` guard then raises
`InvalidType` as intended, before codegen is reached.
```

`method_id` with a non-literal (runtime) string argument now raises a clean `InvalidType` error instead of crashing the compiler with `CodegenPanic`.

### Cute Animal Picture

### DIFF
--- a/tests/functional/syntax/test_method_id.py
+++ b/tests/functional/syntax/test_method_id.py
@@ -24,6 +24,15 @@ FOO: constant(Bytes[4]) = method_id("bar ()")
     """,
         InvalidLiteral,
     ),
+    (
+        """
+@external
+@view
+def f(s: String[20]) -> Bytes[4]:
+    return method_id(s)
+    """,
+        InvalidType,
+    ),
 ]
 
 
--- a/vyper/builtins/functions.py
+++ b/vyper/builtins/functions.py
@@ -718,7 +718,7 @@ class MethodID(FoldedFunctionT):
     def _try_fold(self, node):
         validate_call_args(node, 1, ["output_type"])
 
-        value = node.args[0].get_folded_value()
+        value = node.args[0].reduced()
         if not isinstance(value, vy_ast.Str):
             raise InvalidType("method id must be given as a literal string", node.args[0])
         if " " in value.value:


charles-cooper (approved):
lgtm

PR #5159, opened 2026-06-25
fix[lang]!: reserve `unreachable` keyword

### LINKED ISSUE #5140: Missing comma defeats two reserved keyword entries
### Version Information

* OS: Linux Ubuntu Jammy
* Python Version: Python 3.11

### What's your issue about?

There is a missing comma in the `RESERVED_KEYWORDS` set literal at `vyper/ast/identifiers.py:74-77`. The entry `"flag"` (line 74) is followed by a comment and then `"unreachable"` (line 76) **with no comma between them**. Because Python implicitly concatenates adjacent string literals (whitespace/comments between them is ignored), `"flag" "unreachable"` becomes the single string `"flagunreachable"`.

The current source:

```python
RESERVED_KEYWORDS = _PYTHON_RESERVED_KEYWORDS | {
    ...
    # "class" keywords
    "interface",
    "struct",
    "event",
    "enum",
    "flag"            # <-- LINE 74: no comma!
    # EVM operations
    "unreachable",    # <-- LINE 76: gets concatenated with "flag"
    ...
}
```

Net effect: `RESERVED_KEYWORDS` contains `"flagunreachable"` instead of the two intended separate entries `"flag"` and `"unreachable"`. As a result, **neither** `flag` **nor** `unreachable` is actually reserved — both can be used as variable/function/parameter names with no compile error — while the impossible-to-naturally-write `flagunreachable` IS reserved.

Both keywords were intended to be reserved: `flag` is a type-class keyword used to declare flag types (`flag MyFlag: RED; GREEN; BLUE`), and `unreachable` corresponds to the IR-level `assert_unreachable` (which emits an INVALID opcode). Neither should be usable as an identifier.

The smoking-gun: compiling the impossible identifier `flagunreachable` is rejected, proving the concatenation occurred.

```vyper
@external
def f() -> uint256:
    flagunreachable: uint256 = 5
    return flagunreachable
```

```
$ vyper t.vy
vyper.exceptions.StructureException: 'flagunreachable' is a reserved keyword
```

Meanwhile, both `flag` and `unreachable` compile fine even though they are supposed to be reserved:

```vyper
@external
def f() -> uint256:
    flag: uint256 = 5
    return flag
```

```
$ vyper t.vy
0x6027...   # compiles!
```

The same holds for `unreachable: uint256 = 5` — it compiles. This lets a user declare a function parameter, local variable, or state variable named `flag` or `unreachable`, silently shadowing the language keyword:

```vyper
flag MyFlag:    # 'flag' here is a type-class keyword
    RED

@external
def get_flag() -> uint256:
    flag: uint256 = 5   # 'flag' here is a local var (shadowing) — silently succeeds
    return flag
```

### How can it be fixed?

Add a comma after `"flag"` on `vyper/ast/identifiers.py:74` — a one-character change that restores `"flag"` and `"unreachable"` as two separate reserved-keyword entries. A regression test in `tests/unit/ast/test_identifiers.py` asserting `'flag' in RESERVED_KEYWORDS` and `'unreachable' in RESERVED_KEYWORDS` would prevent this from recurring.

Sporarum:
### What I did

Fix https://github.com/vyperlang/vyper/issues/5140

### How I did it

The missing comma between `"flag"` and `"unreachable"` in the `RESERVED_KEYWORDS` set literal in `vyper/ast/identifiers.py` caused Python's implicit string-literal concatenation to merge them into the single string `"flagunreachable"`, so neither keyword was actually reserved. Per @charles-cooper's feedback in the PR thread, `flag` is useful as an identifier and should remain allowed; only `unreachable` needs to be reserved. So instead of adding the missing comma, I commented out the `flag` entry — this both resolves the concatenation (so `unreachable` is reserved as intended) and keeps `flag` usable as an identifier.

### How to verify it

```vyper
# `unreachable` is now rejected as a reserved keyword
@external
def f() -> uint256:
    unreachable: uint256 = 5
    return unreachable
```

```vyper
# `flag` remains usable as an identifier
@external
def f() -> uint256:
    flag: uint256 = 5
    return flag
```

Run `uv run vyper` on each snippet; the first should fail with `StructureException: 'unreachable' is a reserved keyword`, the second should compile.

### Commit message

```
a missing comma in the RESERVED_KEYWORDS set literal between "flag" and
"unreachable" let Python's implicit string-literal concatenation merge
them into the single string "flagunreachable". as a result, neither
`flag` nor `unreachable` was actually reserved — both could be used as
identifiers — while the impossible-to-write "flagunreachable" was
reserved instead.

per user feedback, `flag` is useful as an identifier and should remain
allowed; only `unreachable` needs to be reserved (it maps to the
IR-level `assert_unreachable`). comment out the `flag` entry rather than
adding the missing comma — this both fixes the concatenation (so
`unreachable` is reserved as intended) and keeps `flag` usable as an
identifier.

this is a breaking change for `unreachable`: programs that previously
used `unreachable` as an identifier will no longer compile. that is the
intended behavior.
```

breaking change: `unreachable` is now a reserved keyword and can no longer be used as an identifier. (`flag` remains usable as an identifier.)

### Cute Animal Picture

### DIFF
--- a/tests/functional/codegen/abstract/test_abstract.py
+++ b/tests/functional/codegen/abstract/test_abstract.py
@@ -84,7 +84,7 @@ SUCCESSFUL_OVERRIDES = [
     # Basic successful override
     ("x: uint256", "uint256", "x", "x: uint256", "uint256", "42", 42),
     # Boolean parameter and return type matching
-    ("_flag: bool", "bool", "_flag", "_flag: bool", "bool", "True", True),
+    ("flag_: bool", "bool", "flag_", "flag_: bool", "bool", "True", True),
     # Bytes32 parameter matching
     (
         "data: bytes32",
@@ -224,10 +224,10 @@ SUCCESSFUL_OVERRIDES = [
     # === ADDITIONAL DEFAULT EXPRESSION TYPES ===
     # Boolean literal default
     (
-        "x: uint256, _flag: bool = True",
+        "x: uint256, flag_: bool = True",
         "bool",
-        "_flag",
-        "x: uint256, _flag: bool = True",
+        "flag_",
+        "x: uint256, flag_: bool = True",
         "bool",
         "1",
         True,
@@ -863,7 +863,7 @@ FLAG_B: constant(bool) = False
 import constants_lib
 
 @abstract
-def check(_flag: bool = constants_lib.FLAG_A and not constants_lib.FLAG_B) -> bool: ...
+def check(flag_: bool = constants_lib.FLAG_A and not constants_lib.FLAG_B) -> bool: ...
     """
 
     override_contract = """
@@ -877,8 +877,8 @@ def call_check() -> bool:
     return self.check()
 
 @override(abstract_module)
-def check(_flag: bool = constants_lib.FLAG_A and not constants_lib.FLAG_B) -> bool:
-    return _flag
+def check(flag_: bool = constants_lib.FLAG_A and not constants_lib.FLAG_B) -> bool:
+    return flag_
     """
 
     input_bundle = make_input_bundle(
--- a/tests/functional/codegen/abstract/test_abstract.py
+++ b/tests/functional/codegen/abstract/test_abstract.py
@@ -84,7 +84,7 @@ SUCCESSFUL_OVERRIDES = [
     # Basic successful override
     ("x: uint256", "uint256", "x", "x: uint256", "uint256", "42", 42),
     # Boolean parameter and return type matching
-    ("flag: bool", "bool", "flag", "flag: bool", "bool", "True", True),
+    ("_flag: bool", "bool", "_flag", "_flag: bool", "bool", "True", True),
     # Bytes32 parameter matching
     (
         "data: bytes32",
@@ -224,10 +224,10 @@ SUCCESSFUL_OVERRIDES = [
     # === ADDITIONAL DEFAULT EXPRESSION TYPES ===
     # Boolean literal default
     (
-        "x: uint256, flag: bool = True",
+        "x: uint256, _flag: bool = True",
         "bool",
-        "flag",
-        "x: uint256, flag: bool = True",
+        "_flag",
+        "x: uint256, _flag: bool = True",
         "bool",
         "1",
         True,
@@ -863,7 +863,7 @@ FLAG_B: constant(bool) = False
 import constants_lib
 
 @abstract
-def check(flag: bool = constants_lib.FLAG_A and not constants_lib.FLAG_B) -> bool: ...
+def check(_flag: bool = constants_lib.FLAG_A and not constants_lib.FLAG_B) -> bool: ...
     """
 
     override_contract = """
@@ -877,8 +877,8 @@ def call_check() -> bool:
     return self.check()
 
 @override(abstract_module)
-def check(flag: bool = constants_lib.FLAG_A and not constants_lib.FLAG_B) -> bool:
-    return flag
+def check(_flag: bool = constants_lib.FLAG_A and not constants_lib.FLAG_B) -> bool:
+    return _flag
     """
 
     input_bundle = make_input_bundle(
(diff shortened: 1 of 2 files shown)


Sporarum:
Gas benchmark fails because a parameter is called `flag` in `src/snekmate/utils/mocks/merkle_proof_verification_mock.vy`

tests/functional/codegen/abstract/test_abstract.py
```diff
@@ -863,7 +863,7 @@ def test_default_param_bool_operation_override(get_contract, make_input_bundle):
 import constants_lib
 
 @abstract
-def check(flag: bool = constants_lib.FLAG_A and not constants_lib.FLAG_B) -> bool: ...
+def check(_flag: bool = constants_lib.FLAG_A and not constants_lib.FLAG_B) -> bool: ...
```
charles-cooper: style wise i think leading underscore is used for "private" variables (either variables that aren't supposed to be accessed globally or are not supposed to escape scope), let's change this to trailing underscore

charles-cooper: ffa9c58ee7458e2ffd8f4ac7e8679fa3f6a2664c

charles-cooper (approved):

charles-cooper:
@codex review

charles-cooper:
i wonder why the linter didn't catch this btw...

pcaversaccio (commented):
So snekmate uses `flag` in the following [loop](https://github.com/pcaversaccio/snekmate/blob/283693fb0cd03c1a2ad5f69f0a5e4048dd296c12/src/snekmate/utils/merkle_proof_verification.vy#L193-L215):

```vy
# At each step, the next hash is calculated from two values:
# - a value from the "main queue". If not all leaves have been used,
#   the next leaf is picked up, otherwise the next hash.
# - depending on the flag, either another value from the "main queue"
#   (merging branches) or an element from the `proof` array.
for flag: bool in proof_flags:
    if leaf_pos < leaves_length:
        a = leaves[leaf_pos]
        leaf_pos = unsafe_add(leaf_pos, 1)
    else:
        a = hashes[hash_pos]
        hash_pos = unsafe_add(hash_pos, 1)
    if flag:
        if leaf_pos < leaves_length:
            b = leaves[leaf_pos]
            leaf_pos = unsafe_add(leaf_pos, 1)
        else:
            b = hashes[hash_pos]
            hash_pos = unsafe_add(hash_pos, 1)
    else:
        b = proof[proof_pos]
        proof_pos = unsafe_add(proof_pos, 1)
    hashes.append(self._hash_pair(a, b))
```

I think `flag` is a perfectly natural loop variable name because it represents a Boolean flag. Reserving `flag` would unnecessarily break code like this if you ask me, even though there is no parsing ambiguity. TL;DR: `flag` is short, descriptive, and naturally used (as seen in snekmate) as a Boolean variable name.

charles-cooper:
yes ok -- let's remove `flag` from reserved keywords and keep unreachable

Sporarum:
@charles-cooper done

vyper/ast/identifiers.py
```diff
@@ -71,7 +71,7 @@ def validate_identifier(attr, ast_node=None):
     "struct",
     "event",
     "enum",
-    "flag"
+    # "flag", # Made into a soft keyword to preserve backwards compat
```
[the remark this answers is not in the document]
charles-cooper: i agree, it's kind of confusing what the comment is supposed to mean. it's also not about backwards compatibility, it's more that we just think that `flag` is a valid user identifier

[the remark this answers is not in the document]
Sporarum: Done, let me know what you think

vyper/ast/identifiers.py
```diff
@@ -71,7 +71,7 @@ def validate_identifier(attr, ast_node=None):
     "struct",
     "event",
     "enum",
-    "flag"
+    # "flag", # Made into a soft keyword, since it's useful as an identifier
```
charles-cooper: i don't know what a soft keyword is (as in, it's not defined by vyper anywhere in the codebase or docs)


```suggestion
    # "flag", # allowed since it's useful as an identifier per user feedback
```

PR #5168, opened 2026-06-26
fix: improve NamespaceCollision error message for module import collisions

### LINKED ISSUE #4579: Improve error message for `math` namespace collision

kushdab:
Fixes #4579.

## Problem

When two modules are imported under the same alias (e.g. `from snekmate.utils import math` followed by `import math`), Vyper raised a `NamespaceCollision` with an opaque, implementation-leaking message:

```
vyper.exceptions.NamespaceCollision: 'math' has already been declared as a
ModuleInfo(module_t=venv/Lib/site-packages/snekmate/utils/math.vy, alias='math',
ownership=<ModuleOwnership.NO_OWNERSHIP: 'no_ownership'>, ownership_decl=None)
```

This exposes internal class repr instead of helping the user understand what went wrong or how to fix it.

## Fix

In `vyper/semantics/namespace.py:validate_assignment`, detect when the already-declared name is a `ModuleInfo` and emit a user-friendly message:

```
'math' is already imported as a module (from 'snekmate/utils/math.vy').
Use a different alias, e.g. `import snekmate/utils/math.vy as math_m`
```

## Changes

- `vyper/semantics/namespace.py` — detect `ModuleInfo` in `validate_assignment` and format a clear error with the module path and an alias suggestion
- `tests/unit/semantics/test_namespace.py` — new test `test_namespace_collision_module_info_message` verifies the improved message

### DIFF
--- a/tests/unit/semantics/test_namespace.py
+++ b/tests/unit/semantics/test_namespace.py
@@ -94,3 +94,28 @@ def test_undeclared_definition_across_scopes(namespace):
             namespace["foo"] = 42
     with pytest.raises(UndeclaredDefinition):
         namespace["foo"]
+
+
+def test_namespace_collision_module_info_message(namespace):
+    """Module import collision should give a clear, actionable error message."""
+    from unittest.mock import MagicMock
+
+    # Simulate a ModuleInfo-like object: has module_t but no decl_node
+    mock_module_t = MagicMock()
+    mock_module_t._id = "snekmate/utils/math.vy"
+    mock_module_info = MagicMock()
+    mock_module_info.module_t = mock_module_t
+    mock_module_info.decl_node = None
+    del mock_module_info.decl_node  # ensure getattr returns None
+
+    with namespace.enter_scope():
+        # Directly set without triggering validate_assignment
+        super(type(namespace), namespace).__setitem__("math", mock_module_info)
+
+        with pytest.raises(NamespaceCollision) as excinfo:
+            namespace["math"] = 42
+
+    msg = str(excinfo.value)
+    assert "already imported as a module" in msg
+    assert "snekmate/utils/math.vy" in msg
+    assert "alias" in msg.lower()
--- a/tests/unit/semantics/test_namespace.py
+++ b/tests/unit/semantics/test_namespace.py
@@ -117,5 +117,7 @@ def test_namespace_collision_module_info_message(namespace):
 
     msg = str(excinfo.value)
     assert "already imported as a module" in msg
-    assert "snekmate/utils/math.vy" in msg
-    assert "alias" in msg.lower()
+    # source path now shown via context arrows, not in main message
+    assert "snekmate/utils/math.vy" not in msg
+    # hint uses the later import (attr) with valid alias syntax
+    assert "import math as math_m" in msg
--- a/vyper/semantics/namespace.py
+++ b/vyper/semantics/namespace.py
@@ -91,6 +91,17 @@ class Namespace(dict):
         if attr in self:
             prev = super().__getitem__(attr)
             prev_decl = getattr(prev, "decl_node", None)
+            module_t = getattr(prev, "module_t", None)
+            if module_t is not None:
+                # prev is a ModuleInfo: emit a friendlier message that names
+                # the already-imported module and suggests using an alias.
+                module_path = module_t._id
+                msg = (
+                    f"'{attr}' is already imported as a module "
+                    f"(from '{module_path}'). "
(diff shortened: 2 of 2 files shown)


Sporarum:
> 'math' is already imported as a module (from 'snekmate/utils/math.vy').
Use a different alias, e.g. `import snekmate/utils/math.vy as math_m`

This seems wrong, I would expect something more like:
> 'math' is already imported as a module.
> hint: use a different alias, e.g. `import math as math_m`

With both locations included in the error, this will print a couple of lines of context with nice arrows to the precise location

Note the changes:
1. we don't need `(from 'snekmate/utils/math.vy')` since it will already be shown
2. we use a hint section to describe the action to take (use `hint=`)
3. in the hint we put the later import, as it is more likely to be more recent
4. we make sure hint contains valid code `import math as math_m` as opposed to `import snekmate/utils/math.vy as math_m`

tests/unit/semantics/test_namespace.py
```diff
@@ -94,3 +94,28 @@ def test_undeclared_definition_across_scopes(namespace):
             namespace["foo"] = 42
     with pytest.raises(UndeclaredDefinition):
         namespace["foo"]
+
+
+def test_namespace_collision_module_info_message(namespace):
+    """Module import collision should give a clear, actionable error message."""
+    from unittest.mock import MagicMock
```
Sporarum, line 101: We don't use mocking, please instead provide a concrete vyper example
(see the tests in other files)

tests/unit/semantics/test_namespace.py
```diff
@@ -94,3 +94,28 @@ def test_undeclared_definition_across_scopes(namespace):
             namespace["foo"] = 42
     with pytest.raises(UndeclaredDefinition):
         namespace["foo"]
+
+
+def test_namespace_collision_module_info_message(namespace):
```
Sporarum, line 99: This feels like the wrong file for this test, I would instead put it somewhere where we test imports
(we usually group by user-facing constructs, and not by internals)

Sporarum (changes requested):

kushdab:
Thanks @Sporarum, all four points addressed:

1. Removed `(from 'snekmate/utils/math.vy')` from the main message the context arrows already show that
2. Moved the suggestion into a `hint=` section
3. The hint now uses the later import (`attr`) rather than the module path, since that is the one more likely to need renaming
4. Hint now contains valid code: `import math as math_m` instead of `import snekmate/utils/math.vy as math_m`

Updated the test assertions to match. Happy to adjust anything further.

vyper/semantics/namespace.py
```diff
@@ -93,15 +93,14 @@ def validate_assignment(self, attr):
...
-                    f"Use a different alias, e.g. `import {module_path} as {attr}_m`"
-                )
-                raise NamespaceCollision(msg, prev_decl=prev_decl)
+                # prev is a ModuleInfo: emit a friendlier message with a hint
+                # showing valid alias syntax. The (from ...) source path will
+                # already appear in the context arrows, so omit it from the
+                # main message. Use the later import (attr) in the hint as it
+                # is more likely to be the one the user needs to rename.
+                msg = f"'{attr}' is already imported as a module."
+                hint = f"use a different alias, e.g. `import {attr} as {attr}_m`"
```
Sporarum, line 102: This probably does not handle things like:
```python
from a import foo
from b import foo
```
Here we want the hint to contain `from b import foo`

vyper/semantics/namespace.py
```diff
@@ -93,15 +93,14 @@ def validate_assignment(self, attr):
...
+                raise NamespaceCollision(msg, prev_decl=prev_decl, hint=hint)
```
Sporarum, line 103: Is there something like a `current_decl` ?
Ideally both should be provided in pretty-printed error

Sporarum:
@kushdab you seem to have missed the review conversations I opened, note that I opened more

vyper/semantics/namespace.py
charles-cooper: i don't think it's available at this location in the call stack, we could probably do this and also protect the call site (of validate_assignment)?

PR #5171, opened 2026-06-28
feat[venom]: add redundant memory copy forwarding pass

harkal:
## What I did

Added a Venom optimization pass, `RedundantMemoryCopyForwardingPass`, that eliminates redundant memory staging copies. When a value is `mcopy`'d into a fresh temporary buffer and the copied bytes are only observed through safe memory reads, the pass redirects those reads to the stable source and deletes the copy.

The final pass rewrites only memory-read address operands. It does not mutate the destination's alias-producing instructions, which keeps the transformation and its correctness argument local to each read site.

## How I did it

The pass runs after `MemoryCopyElisionPass`, which forms the canonical `mcopy`, and before `ConcretizeMemLocPass`, while static allocas are still abstract.

For a bounded copy into a fixed segment of a static local alloca, forwarding is accepted only when:

- The source is a fixed, disjoint region of a tracked static alloca, or an exclusively resolved readonly internal-function memory parameter.
- Every alias that can observe the copied segment is either a recognized pointer derivation or the address operand of a memory read contained in that segment.
- Neither the source nor destination is clobbered between the staging copy and any forwarded read, as proved by MemSSA.
- Source and destination pointers do not escape into observable scalar positions such as return sizes, log topics, stored values, or opaque call arguments.
- Every runtime-sized read has an explicit upper bound from typed-codegen `memory_read_max_size` metadata or `VariableRangeAnalysis`, and the complete bounded read fits in the copied segment. Unbounded reads are rejected.

The rewrite plan contains only `(read instruction, address operand, source offset)` entries. A nonzero source offset is materialized immediately before the read, so there is no phi-specific alias rewrite or separate source-dominance machinery.

Supporting analysis was made shared and fail-closed: `MemoryParamRootResolver` now provides both may-root and exclusive-root parameter provenance; `BasePtrAnalysis` exposes allocation aliases and treats unsupported pre-SSA reassignments as untracked roots; MemSSA provides clobber queries; and instruction reachability lives in `ReachableAnalysis.is_reachable_from`, shared with `DeadStoreElimination`. Instructions preserve the frontend read bound through copying and inlining, and affected analyses are invalidated after IR mutation.

## How to verify it

- `uv run pytest -q tests/unit/compiler/venom/test_redundant_memory_copy_forwarding.py tests/unit/compiler/venom/test_memory_location.py tests/unit/compiler/venom/test_mem_ssa.py tests/unit/compiler/venom/test_base_ptr_analysis.py tests/unit/compiler/venom/test_inliner.py tests/unit/compiler/venom/test_removeunused.py`
- `test_redundant_memory_copy_forwarding.py` is marked `hevm`, so the dedicated hevm job checks pre/post equivalence.
- Full experimental O2/Prague/revm run: 12,328 passed, 6 skipped, 50 xfailed.
- The complete GitHub Actions matrix is green, including all experimental optimization levels, CodeQL, gas-bench, and bytecode-size.
- Snekmate gas comparison retains the optimization wins on bounded, non-INF contracts (for example, ERC1155 removes 62 memory copies and TimelockController removes 13). The architectural simplification introduced no gas or status changes across the measured 815-case result.

## Commit message

```
Venom frequently stages a value into a fresh temporary buffer with
`mcopy` and then only reads from that temporary, for example while
materializing a bounded Bytes/String/DynArray before ABI encoding,
hashing, or passing it to a call. When the source is already stable,
that staging copy is pure overhead: the reads can address the source
directly.

This pass recognizes that shape, redirects the reads to the source,
and deletes the staging copy. It runs after MemoryCopyElisionPass and
before ConcretizeMemLocPass, while static allocas are still abstract.
The rewrite is deliberately local: only memory-read address operands
are changed. Alias-producing assign, arithmetic, and phi instructions
are left intact.

Forwarding is gated conservatively. The destination must be a fixed
segment of a static local allocation. The source must be a fixed,
disjoint region of a tracked static allocation or an exclusively
resolved readonly internal-function memory parameter. Every use that
can observe the copied segment must be a recognized pointer derivation
or a contained memory read, and source and destination pointers may
not escape as scalar values. MemSSA must prove that neither region is
modified between the copy and any forwarded read.

Runtime-sized reads require a real bound. Typed codegen attaches a
path-sensitive memory_read_max_size where the type and runtime guards
prove one; otherwise VariableRangeAnalysis may supply the bound. The
entire bounded read must fit within the copied segment. An unknown or
oversized read keeps the original mcopy. This replaces the broader
whole-allocation assumption and does not remove any runtime check.

Readonly parameter provenance is shared through
MemoryParamRootResolver, which distinguishes may-roots from values
derived exclusively from parameters and combines every definition for
pre-SSA input. Base pointer and clobber queries fail closed on unknown
roots, pointer escapes, missing MemorySSA uses, and unresolved writes.
Affected analyses are invalidated after each applied plan.

The pass is enabled at O2, O3, and Os. It removes ordinary
materialize-then-read overhead while retaining the existing safety
checks and conservative fallback behavior.
```


vyper/venom/passes/redundant_memory_copy_forwarding.py
```diff
@@ -0,0 +1,525 @@
...
+        assert dst_loc.offset is not None and dst_loc.size is not None
+        dst_start = dst_loc.offset
+        dst_end = dst_start + dst_loc.size
+        alias_rewrites = self._segment_aliases(dst_alloca, aliases, dst_start, dst_end)
+        direct_read_rewrites = []
+        read_sites: set[IRInstruction] = set()
+        alias_read_uses: dict[IRVariable, list[tuple[IRInstruction, int]]] = {}
+
+        for alias in alias_rewrites:
+            for use in self.dfg.get_uses(alias):
```
charles-cooper: **correctness:** reads of the staged buffer are enumerated only through the DFG pointer graph (uses of the alloca root and its tracked aliases), so a read through a **literal/concrete address** is invisible — it is neither redirected nor treated as disqualifying — yet the copy is still deleted. The MemSSA guards don't catch it either: `_has_location_clobber_between` walks clobbering `MemoryDef`s, and a concrete `mload` is a `MemoryUse`.

Reproducer (well-formed Venom accepted by the `venom` CLI):

```
%tmp = alloca 96
%src = alloca 96
%s64 = add 64, %src
mstore %s64, 42
mstore 64, 7
mcopy %tmp, %src, 96
%val = mload %tmp
%hidden = mload 64
sink %val, %hidden
```

The pass nops the mcopy while leaving `%hidden = mload 64` untouched; after `ConcretizeMemLocPass` (`MemoryAllocator.FN_START == 0`, and the allocator never accounts for literal accesses) `%tmp` can land over address 64, so the baseline sinks the staged byte while the transformed program reads stale memory.

This is asymmetric with the rest of the pass: concrete *writes* are defended (`memory_location.py` treats concrete-vs-abstract as may-overlap, and there's `test_keeps_copy_from_concrete_scratch_memory` for the source side), but concrete *reads* of the destination are not, and there's no test for them.

Severity caveat: I instrumented the pass over `examples/*.vy` and frontend-generated IR never reads memory through literals at this pipeline position, so this is only reachable from hand-written/external Venom — but that's an accepted input surface. Suggested fix: treat any untracked memory read whose location may-overlap `dst_loc` as disqualifying (mirroring the write-side check).

---
_Generated by [Claude Code](https://claude.ai/code)_

vyper/venom/analysis/readonly_memory_args.py
```diff
@@ -14,7 +15,152 @@
...
+        """
+        Return param roots only if `op` is exactly param-backed.
+
+        This intentionally follows only param, assign, and phi edges. Unlike
+        `root_param_indices`, it does not treat pointer arithmetic as proof
+        that the resulting value still points inside a param region, and it
+        returns None when some path has an unknown/non-param root.
+        """
+        if not isinstance(op, IRVariable):
+            return frozenset()
```
charles-cooper: **correctness:** this returns `frozenset()` (success, empty roots) for a non-`IRVariable` operand, but the docstring above says the query "returns None when some path has an unknown/non-param root" — and a literal is exactly an unknown/non-param root. The must-analysis contract is violated on the `assign <literal>` arm of a phi walk:

`%z = assign 64` → `_exclusive_param_alias_from_inst` delegates here with `IRLiteral(64)` → `frozenset()`, so for `%s = phi(@a: %arg, @b: %z)` (with `%arg` a readonly memory param) `_combine_exclusive_param_aliases` unions to `{arg_idx}` instead of bailing with None.

The consumer leans on this contract for soundness: in `redundant_memory_copy_forwarding.py` the readonly-param branch deliberately skips `_source_may_include_untracked_root` (the design comment there says untracked sources are rejected *by this query returning None*), and then applies the weaker `_has_unresolved_param_clobber_between`, which ignores writes to tracked local allocas via the frame-separation argument. That argument doesn't hold for the literal path: address 64 can overlap a local alloca after `ConcretizeMemLocPass`, so a tracked-alloca write between the copy and a forwarded read would be skipped by the clobber filter.

I traced the whole `_candidate`/`_build_forward_plan` gauntlet and found no other gate that rejects a phi-of-(param, literal) source on this branch (a *pure* literal source bails at `_source_is_readonly_param` because the root set is empty — it's specifically the phi merge that gets through). One-line fix: return `None` here for non-variable operands (the may-analysis `root_param_indices` can keep `frozenset()`).

---
_Generated by [Claude Code](https://claude.ai/code)_

vyper/venom/passes/redundant_memory_copy_forwarding.py
```diff
@@ -0,0 +1,525 @@
...
+
+        if changed:
+            self._invalidate()
+
+    def _prepare(self) -> None:
+        self.dfg = self.analyses_cache.request_analysis(DFGAnalysis)
+        self.domtree = self.analyses_cache.request_analysis(DominatorTreeAnalysis)
+        self.base_ptr = self.analyses_cache.request_analysis(BasePtrAnalysis)
+        self.mem_alias = self.analyses_cache.request_analysis(MemoryAliasAnalysis)
+        self.mem_ssa = self.analyses_cache.request_analysis(MemSSA)
```
charles-cooper, line 106: **invalidation hygiene (new exposure introduced by this PR):** this makes MemSSA a *cached mid-pipeline artifact* — on the common no-change path `run_pass` leaves it valid in the cache. But `LowerDloadPass`, which sits between this pass and `DeadStoreElimination` in all three optimization levels, mutates memory instructions (dload→mload in place, inserted alloca+codecopy) and invalidates only Liveness/DFG/BasePtr — not MemSSA.

The masking invalidators are both conditional: `LoadElimination` (which this PR taught to invalidate MemSSA unconditionally) can be removed from the pipeline via `disable_load_elimination`, and `RemoveUnusedVariables` now only invalidates when it removed something. Under `disable_load_elimination` + a no-op RUV, DSE receives a MemSSA computed before the dload lowering. Pre-PR this couldn't happen — DSE was the only MemSSA requester, so its first run always computed fresh.

I could not build a miscompile from it today: DSE's `_has_uses` check runs against a *fresh* DFG, so a live converted mload is never nopped, and the other staleness directions are conservative (stale defs/uses only keep more stores alive). But it's the same opt-in whack-a-mole this PR already had to patch twice (LoadElimination, RUV), and it's one forgotten pass away from a real bug. Minimal fix: have `LowerDloadPass` invalidate the memory-SSA family (or have this pass not leave MemSSA cached on the no-change path). The deeper fix — a preserved-set / invalidate-by-default scheme in `IRAnalysesCache` so correctness doesn't depend on each pass author remembering the full list — is probably out of scope here but worth a follow-up issue.

---
_Generated by [Claude Code](https://claude.ai/code)_

vyper/venom/passes/redundant_memory_copy_forwarding.py
```diff
@@ -0,0 +1,525 @@
...
+        dst_ptr = self.base_ptr.ptr_from_op(copy_inst.operands[2])
+        if dst_ptr is None or dst_ptr.offset is None:
+            return None
+
+        dst_root_inst = dst_ptr.base_alloca.inst
+        if dst_root_inst is None or dst_root_inst.opcode != "alloca":
+            return None
+
+        dst_alloca = Allocation(dst_root_inst)
+        if dst_alloca.is_dynamic or copy_size > dst_alloca.alloca_size:
```
charles-cooper: **hardening:** this guard reads like a destination-bounds check but only enforces `copy_size <= alloca_size`, ignoring `dst_ptr.offset`. An mcopy at `%tmp+96` with size 64 into a 128-byte alloca writes `[96, 160)` — past the allocation — yet passes (`64 <= 128`). `dst_loc` then becomes `(alloca, 96, 64)` and `completely_contains` happily admits reads at offsets the alias machinery never models (it assumes accesses stay within their allocation).

Unreachable from well-formed IR — the pass's own comment near the full-coverage gate documents the access-within-allocation invariant, and OOB IR is broken for the whole pre-concretization framework independent of this pass — so no demonstrated miscompile. But the tightening is one token: `dst_ptr.offset + copy_size > dst_alloca.alloca_size`.

---
_Generated by [Claude Code](https://claude.ai/code)_

vyper/venom/passes/redundant_memory_copy_forwarding.py
```diff
@@ -0,0 +1,525 @@
...
+            if plan is None:
+                break
+
+            self._apply_forward_plan(plan)
+            self.updater.nop(plan.copy_inst, annotation="[redundant memory copy forwarding]")
+            changed = True
+            self._invalidate()
+
+        if changed:
+            self._invalidate()
```
charles-cooper, line 99: **perf / simplification:** two things about this loop:

1. The trailing `if changed: self._invalidate()` is redundant and actively harmful: the loop can only exit through an iteration that just ran `_prepare()` (rebuilding everything after the last in-loop `_invalidate()`) and mutated nothing, so at exit all five analyses are fresh and valid — this just discards them and forces downstream passes to recompute. It can be deleted.

2. The restart is per-forwarded-copy: `_find_forwardable_plan` returns the *first* plan, so K forwardable copies cost K+1 full DFG/DomTree/BasePtr/MemAlias/MemSSA rebuilds plus K+1 whole-function mcopy scans (`aliases_of_allocation` is itself a linear scan over `var_to_mem` per candidate). Contracts that stage many aggregates will feel this at O2. Batching non-interacting plans per rebuild isn't trivial — a plan's read site can itself be another candidate mcopy, and alias rewrites change BasePtr facts — so it'd need disjointness checks between plans; fine as a follow-up, but worth a `# TODO` given the pass runs at default-adjacent optimization levels.

---
_Generated by [Claude Code](https://claude.ai/code)_

vyper/venom/passes/redundant_memory_copy_forwarding.py
```diff
@@ -0,0 +1,525 @@
...
+        if not isinstance(src, IRVariable):
+            return False
+        return len(self.base_ptr.get_possible_ptrs(src)) > 0
+
+    def _source_may_include_untracked_root(self, src: IROperand) -> bool:
+        if not isinstance(src, IRVariable):
+            return False
+        return self.base_ptr.pointer_may_include_untracked_root(src)
+
+    def _is_after(self, copy_inst: IRInstruction, use_inst: IRInstruction) -> bool:
```
charles-cooper: **style/reuse:** `_is_after` is character-for-character identical to `InvokeCopyForwardingBase._is_after` (`invoke_copy_forwarding_common.py`), and `_prepare` above duplicates that base's dfg/domtree/base_ptr/mem_alias/`InstUpdater`/`CopyForwardingPolicy` wiring, adding only MemSSA and `MemoryParamRootResolver`.

Subclassing `InvokeCopyForwardingBase` directly is awkward (its `_prepare` force-runs `ReadonlyMemoryArgsGlobalAnalysis` per call where this pass deliberately forces it once per run; its `_has_src_clobber_between` has an incompatible signature; and it carries invoke-specific helpers), so the cleaner move is extracting the shared analysis setup + `_is_after` into a neutral base/helper — or hoisting `_is_after` onto `DominatorTreeAnalysis` next to the new `ReachableAnalysis.is_reachable_from`, which this PR already did for the analogous DSE helper. As-is, a fix to one copy (e.g. caching instruction indexes to avoid the O(n) `.index` scans) won't propagate to the other.

---
_Generated by [Claude Code](https://claude.ai/code)_

vyper/venom/memory_location.py
```diff
@@ -183,6 +183,7 @@ class InstAccessOps:
     ofst: Optional[IROperand]
     size: Optional[IROperand]
     max_size: Optional[IROperand] = None
+    ofst_index: Optional[int] = None
```
charles-cooper, line 185: **style/reuse:** `ofst_index` adds a third parallel opcode→operand-position table next to `update_read_location` and `update_write_location` in this same file. I checked every opcode pair and they agree today, but three copies of "which operand is the memory offset" invite drift (the `create` fix in this PR shows it has already happened once).

Two observations that make the cleanup easy: `update_read_location`/`update_write_location` have **zero call sites** in the repo, so they can either be deleted outright (which also removes the pre-existing shadowed duplicate `elif opcode == "call"` branch in `update_read_location`), or reimplemented as `inst.operands[memory_read_ops(inst).ofst_index] = new_op` so `ofst_index` becomes the single source of truth.

---
_Generated by [Claude Code](https://claude.ai/code)_

vyper/venom/analysis/base_ptr_analysis.py
```diff
@@ -326,3 +337,163 @@ def _get_copyable_read_location(self, inst, addr_space: AddrSpace) -> MemoryLoca
...
+            return False
+        if len(self.get_possible_ptrs(inst.output)) == 0:
+            return False
+        if inst.get_read_effects() & effects.MEMORY != effects.EMPTY:
+            return False
+        if inst.get_write_effects() & effects.MEMORY != effects.EMPTY:
+            return False
+        return True
+
+    def pointer_uses_may_touch(self, var: IRVariable, loc: MemoryLocation, mem_alias, dfg) -> bool:
```
charles-cooper: **style:** two doc/typing nits on the new public API here, per `docs/style-guide.rst`:

- `mem_alias` and `dfg` are unannotated even though the docstring names their types ("publicly exposed … **should** include PEP 484 annotations for all arguments"). A `TYPE_CHECKING`-guarded import (or string annotations) satisfies mypy without creating the import cycle the docstring is worried about.
- `pointer_may_include_untracked_root` (line 365) has no docstring ("all publicly exposed classes and methods **must** include detailed docstrings") — its explanation currently lives inside the private `_r` helper.

Relatedly, the fact that this method needs `MemoryAliasAnalysis` facts threaded in by the caller suggests it wants to live one layer up (on `MemoryAliasAnalysis`, which is built on `BasePtrAnalysis` and already holds both dependencies) — that would make the untyped-parameter workaround unnecessary entirely. Not blocking, but worth considering while it has a single caller.

---
_Generated by [Claude Code](https://claude.ai/code)_

charles-cooper (commented):
Reviewed for correctness first, then style. Overall the pass held up well under adversarial checking — pipeline position is enforced via `required_predecessors`/`required_successors`, the gating is genuinely conservative, most soundness edges (phi merges, derived pointers, back-edge clobbers, dominance for in-place rewrites) are correctly closed with matching tests, and lint/mypy are clean.

**Correctness (inline, most severe first):**
1. Reads of the staged destination through literal/concrete addresses are invisible to the DFG-based read enumeration, yet the copy is deleted — reproduced a stale read with well-formed Venom (only reachable via hand-written Venom, not frontend IR, but that's an accepted input surface).
2. `exclusive_param_alias_indices` returns `frozenset()` instead of `None` for non-variable operands, violating its documented fail-closed contract; a `phi(readonly_param, literal)` source then gets the weaker frame-separation clobber filter it isn't entitled to.
3. MemSSA becomes a cached mid-pipeline artifact, but `LowerDloadPass` mutates memory without invalidating it — currently conservative (no miscompile constructible today), but reachable staleness under `disable_load_elimination`, and the same forgotten-invalidation class this PR already patched twice.
4. The destination bounds guard ignores `dst_ptr.offset` (hardening only; unreachable from well-formed IR).

**Style/cleanup (inline):** redundant trailing `_invalidate()` and per-forwarded-copy full re-analysis in `run_pass`; `_is_after`/`_prepare` duplicated verbatim from `InvokeCopyForwardingBase`; `ofst_index` as a third parallel opcode table next to the (dead) `update_read/write_location`; missing annotations/docstring on the new `BasePtrAnalysis` public methods.

Checked and found fine along the way (so nobody re-chases them): the missing `LivenessAnalysis` invalidation in `_invalidate()` is harmless — invalidating DFG cascades to Liveness, and `_prepare()` guarantees DFG is cached when it fires; likewise `RemoveUnusedVariables`' new early-return only skips a `clear_nops()`-level mutation that can't stale liveness.

---
_Generated by [Claude Code](https://claude.ai/code)_

vyper/venom/passes/redundant_memory_copy_forwarding.py
harkal: `_is_after` moved. Left the _prepare wiring duplicated on purpose.

vyper/venom/passes/lower_dload.py
```diff
@@ -20,6 +20,7 @@ def run_pass(self):
         self.analyses_cache.invalidate_analysis(LivenessAnalysis)
         self.analyses_cache.invalidate_analysis(DFGAnalysis)
         self.analyses_cache.invalidate_analysis(BasePtrAnalysis)
+        self.analyses_cache.invalidate_analysis(MemSSA)
```
charles-cooper, line 34: **nit (latent, non-blocking):** while auditing the invalidation graph end to end, this is the one asymmetry left: `MemoryAliasAnalysis` isn't invalidated here. It's unreachable today only because `LoadElimination` immediately precedes this pass in all three pipelines and unconditionally cascades the alias analyses away — but if a MemAlias were ever cached *without* MemSSA entering this pass, the next `MemSSA.analyze` (in DSE) would embed the stale instance (`mem_ssa.py` requests the cached-or-fresh alias analysis), and its `mem_loc_insts` would omit the new mload/codecopy accesses, letting `get_all_insts` under-report readers. Since this pass inserts new memory accesses, invalidating `MemoryAliasAnalysis` alongside `MemSSA` makes it robust to future pipeline reshuffles for the cost of one line.

---
_Generated by [Claude Code](https://claude.ai/code)_

vyper/venom/analysis/base_ptr_analysis.py
```diff
@@ -326,3 +337,120 @@ def _get_copyable_read_location(self, inst, addr_space: AddrSpace) -> MemoryLoca
...
+                continue
+            if any(ptr.base_alloca != alloca for ptr in ptrs):
+                return None
+            if self.pointer_may_include_untracked_root(var):
+                return None
+            aliases.add(var)
+
+        return aliases
+
+    def pointer_may_include_untracked_root(self, var: IRVariable) -> bool:
```
charles-cooper, line 375: **nit (API hardening, non-blocking):** the fail-closed rewrite below checks out — I traced `param` and every other zero-input/non-deriving def, and none can reach the walk through the current callers because both recursion sites and both top-level callers guard on pointer facts first. But that makes the *bare* public API partial: `pointer_may_include_untracked_root(%param)` falls to the new tail and returns `any([]) == False`, contradicting this docstring's own example ("e.g. a param or calldata-derived pointer"). A `len(self.get_possible_ptrs(var)) == 0: return True` guard at the top would make the method total for future callers, matching the guards the recursive sites already have.

---
_Generated by [Claude Code](https://claude.ai/code)_

charles-cooper (commented):
Round 2 — re-reviewed the delta since e9b7b87 (commits `710d42d..6136830`). All fixes from the previous round verified, adversarially where it mattered:

- **`readonly_memory_args.py` literal roots** — `None` for non-variable operands is the right fail-closed fix, and the new `test_keeps_copy_when_readonly_param_source_merges_literal_address` exercises exactly the phi-of-(param, literal) shape that previously slipped through.
- **`LowerDload` → MemSSA invalidation** — closes the stale chain under `disable_load_elimination`. I re-audited the full O2/O3/Os invalidation graph (cascade hooks + every pass in the window between analysis cachers and consumers): **it is now closed** — every consumer of instruction-dependent MemSSA/MemAlias/BasePtr state (this pass, DSE, LoadAnalysis) sees instances computed after the last relevant mutation. The `MemoryCopyElision` → MemAlias invalidation added here was also load-bearing for that conclusion.
- **Bounds guards** — the offset-aware dst check plus the new symmetric src check leave no hole I could find; the non-fixed src path correctly needs none (no alloca to bound against), and the full-coverage gate is consistent at the boundary.
- **`is_after` hoist** — argument order verified correct at every call site (including the shim in `invoke_copy_forwarding_common.py`), semantics bit-for-bit equivalent to both old copies.
- **`pointer_uses_may_touch` move onto `MemoryAliasAnalysis`** — behaviorally identical, instance wiring is coherent through the pass's invalidate/prepare loop, fully annotated, and it resolved the CodeQL cyclic-import alerts.
- **Fail-closed tail in the untracked-root walk** — sound; the catch-all only flips results in the conservative direction, cycle handling untouched. One API nit inline.

Tests green (`test_redundant_memory_copy_forwarding.py` + the four related files: 79 passed), lint/mypy clean.

Two non-blocking nits inline (LowerDload/MemAlias symmetry, a defensive guard on `pointer_may_include_untracked_root`). One item from round 1 is still open: the **literal-address read of the staged destination** thread (reads are enumerated only via the DFG pointer graph, so a concrete `mload 64` of memory the destination alloca can later be concretized over is neither redirected nor disqualifying) — no commit addresses it yet and the thread is unresolved.

Two pre-existing fragilities surfaced by the audit that are *not* this PR's fault and don't manifest with the current pass order — noting them for follow-up issues rather than blocking: (1) the invoke copy-forwarding passes invalidate only Liveness after rewriting pointer derivations, leaving misrooted BasePtr/MemAlias facts that MemMerge #1 consumes (potentially non-conservative since `may_overlap` returns False for distinct allocas); (2) DSE's final no-change iteration leaves MemSSA/MemAlias cached, which is value-safe today only because downstream consumers use pure location math.

---
_Generated by [Claude Code](https://claude.ai/code)_

vyper/venom/passes/redundant_memory_copy_forwarding.py
```diff
@@ -0,0 +1,517 @@
...
+
+        for alias in alias_rewrites:
+            for use in self.dfg.get_uses(alias):
+                for pos, op in enumerate(use.operands):
+                    if op != alias:
+                        continue
+                    if use is copy_inst and pos == 2:
+                        continue
+                    if self._is_allowed_pointer_use(use, alias, aliases):
+                        continue
```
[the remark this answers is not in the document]
harkal: false

tests/unit/compiler/venom/test_removeunused.py
```diff
@@ -83,3 +85,23 @@ def test_removeunused_loop():
         jmp @after
     """
     _check_pre_post(pre, post)
+
+
+def test_removeunused_invalidates_memory_ssa():
```
HodanPlodky, line 90: this test does not check anything

vyper/venom/analysis/base_ptr_analysis.py
```diff
@@ -57,11 +57,22 @@ class BasePtrAnalysis(IRAnalysis):
...
+    _untracked_root_active: set[IRVariable]
 
     def analyze(self):
         self.var_to_mem = dict()
+        self.var_to_def = dict()
+        self._untracked_root_memo = dict()
+        self._untracked_root_active = set()
         self.cfg = self.analyses_cache.request_analysis(CFGAnalysis)
 
+        for bb in self.function.get_basic_blocks():
```
HodanPlodky: this should be already computed in the dfg

vyper/venom/analysis/readonly_memory_args.py
```diff
@@ -14,7 +15,152 @@
...
+
+        return frozenset()
+
+    def _root_from_add(self, inst: IRInstruction) -> frozenset[int]:
+        roots: set[int] = set()
+        for op in inst.operands:
+            roots.update(self.root_param_indices(op))
+        return frozenset(roots)
+
+    def _root_from_sub(self, inst: IRInstruction) -> frozenset[int]:
```
HodanPlodky: why is this different from add and additionally when would be mem allocation go through `sub` inst

tests/unit/compiler/venom/test_removeunused.py
harkal: true :/ added assertions: the `mload` is gone, the cached `MemSSA ` instance was invalidated, and DSE removes the dead store afterwards

vyper/venom/analysis/readonly_memory_args.py
harkal: unified into `_root_from_arith`. sub is covered for the same reason `BasePtrAnalysis` models it

harkal:
Pass simplification brief: https://gist.github.com/harkal/bf5d91428845daa4dd170ac5a0dd73d9

vyper/venom/passes/redundant_memory_copy_forwarding.py
```diff
@@ -0,0 +1,495 @@
...
+        dst_ptr = self.base_ptr.ptr_from_op(copy_inst.operands[2])
+        if dst_ptr is None or dst_ptr.offset is None:
+            return None
+
+        dst_root_inst = dst_ptr.base_alloca.inst
+        if dst_root_inst is None or dst_root_inst.opcode != "alloca":
+            return None
+
+        dst_alloca = Allocation(dst_root_inst)
+        if dst_alloca.is_dynamic:
```
HodanPlodky, line 145: this should be already checked by the previous condition (so it could be assert)

vyper/venom/passes/redundant_memory_copy_forwarding.py
```diff
@@ -0,0 +1,495 @@
...
+            return None
+        if dst_ptr.offset < 0 or dst_ptr.offset + copy_size > dst_alloca.alloca_size:
```
HodanPlodky, line 147: if the offset + copy_size is bigger then the alloca_size that sounds like the oob so it would be problem somewhere else. Maybe just assert it

vyper/venom/passes/redundant_memory_copy_forwarding.py
```diff
@@ -0,0 +1,495 @@
...
+            return None
+
+        src_loc = self.base_ptr.get_read_location(copy_inst, addr_space.MEMORY)
+        dst_loc = self.base_ptr.get_write_location(copy_inst, addr_space.MEMORY)
+        if not dst_loc.is_fixed:
+            return None
+        src_is_readonly_param = False
+        if src_loc.is_fixed:
```
HodanPlodky, line 154: same comments (that some condition could be assert) could be applied in some of the condition in these branch

vyper/venom/passes/redundant_memory_copy_forwarding.py
harkal: Good catch. I removed the redundant reconstruction/opcode check and now use `dst_ptr.base_alloca` directly. I kept the single `is_dynamic` bailout because dynamic allocas are valid at this pipeline point, but unsuported by this optimization.

redundant-memory-copy-forwarding-simplification.md
```diff
@@ -0,0 +1,215 @@
...
+  dynamic read.
+- Separate readonly-root resolution algorithms.
+- Optimistic handling of unsupported pre-SSA reassignments.
+
+The resulting pass is conservative by design. Unsupported cases retain the
+original `mcopy` rather than expanding the transformation's proof obligations.
+
+## Intentional Limits
+
+- Copy size is capped at 4096 bytes.
```
charles-cooper: why?

redundant-memory-copy-forwarding-simplification.md
```diff
@@ -0,0 +1,215 @@
...
+- A tracked source allocation is capped at 4096 bytes.
```
charles-cooper: ditto -- why?

redundant-memory-copy-forwarding-simplification.md
harkal: Bigger cause regressions (want to elaborate on the reasons on the .md?)

redundant-memory-copy-forwarding-simplification.md
```diff
@@ -0,0 +1,215 @@
...
+A dynamic-size read is forwarded only when its maximum reachable size is
+proved. The proof comes from either:
+
+- `memory_read_max_size` metadata supplied by typed code generation.
+- `VariableRangeAnalysis` when no frontend bound is available.
+
+The proven maximum is used to check that the entire possible read is contained
+within the staged segment. An unbounded dynamic read is rejected.
+
+`memory_read_max_size` is path-sensitive. It describes the maximum size on an
```
HodanPlodky: I kinda looked at the codegen and there is the `copy_memory_dynamic` which requires the `memory_read_max_size` but this example does not have it. It raises `CompilerPanic`.
```
@external
def foo(x: Bytes[INF]) -> Bytes[INF]:
    tmp: Bytes[1] = b"\x01"
    return concat(x, tmp) # concat uses copy_memory_dynamic
```
Is this supposed to work or should it be handled earlier in frontend.

harkal: Bytes[INF] is not implemented here yet, this should panic as on master

charles-cooper (approved):
lgtm pending review from @HodanPlodky

vyper/codegen_venom/builtins/misc.py
```diff
@@ -513,7 +514,7 @@ def lower_print(node: vy_ast.Call, ctx: "VenomCodegenContext") -> IROperand:
         b.mstore(payload_buf._ptr, payload_len)
 
         # Allocate schema buffer: [32 bytes length] | [data]
-        schema_buf = ctx.allocate_buffer(32 + schema_len)
+        schema_buf = ctx.allocate_buffer(32 + padded_schema_len)
```
HodanPlodky, line 517: this is existing bug which is not related to this PR specifically right?

harkal: Yes, this is a drive by. This miscompiles on master but this PR's bounds assertions made it visible

PR #5172, opened 2026-06-29
fix[lang]: forbid creation of unusable interfaces

### LINKED ISSUE #5143: Interface accepts self param but implementation rejects
### Version Information

* OS: Linux Ubuntu Jammy
* Python Version: Python 3.11

### What's your issue about?

A Vyper `interface` block accepts `self` as a parameter name, but the implementation of the same signature is rejected with `NamespaceCollision`. The interface-side parameter parsing does not perform the reserved-name / namespace-collision check that implementation-side parameter parsing performs, so an interface can declare a signature that no contract can implement as written.

Concretely:

- `interface I: def f(self: address) -> uint256: view` → **compiles**.
- `staticcall I(t).f(addr)` → **compiles** (ABI selectors use parameter types, not names).
- Implementing `def f(self: address) -> uint256: return 1` under `implements: I` → **rejected**.

Reproducer:

```vyper
interface I:
    def f(self: address) -> uint256: view

implements: I

@external
@view
def f(self: address) -> uint256:
    return 1
```

Error output:

```
NamespaceCollision: 'self' has already been declared as a
VarInfo(typ=self, location=<DataLocation.UNSET: 'unset'>,
modifiability=<Modifiability.MODIFIABLE: 'modifiable'>,
is_public=False, decl_node=None)
```

The interface declaration alone compiles cleanly; only the implementation fails. Net effect: the interface admits a declaration that no contract can satisfy with matching parameter names. A user importing an external ABI (e.g. a Solidity contract, where `self` is a normal identifier) that uses `self` as a parameter ends up with an unimplementable Vyper interface, and the docs (`interfaces.rst`) do not enumerate reserved parameter names.

Source-level: interface-side parameter parsing in `vyper/semantics/types/function.py` calls `_parse_args(funcdef, is_interface=True)`, which skips the namespace-collision check (or runs it against a namespace where `self` is not declared). Implementation-side parsing uses `is_interface=False`, where the current contract's namespace already contains a `VarInfo` for `self`, so the collision triggers at `vyper/semantics/namespace.py:100`.

### How can it be fixed?

Make the reserved-name / namespace-collision check symmetric between interface declarations and implementations. Either option restores parity:

1. Enforce the same reserved-name check at the interface declaration site, so `interface I: def f(self: address) -> uint256: view` is rejected up front with a clear "`self` is a reserved name" error.
2. Allow parameter-name shadowing in implementations (renaming `self` implicitly inside the function body, or otherwise letting the parameter shadow the contract-self reference for that function scope).

A user-facing workaround today is to (a) implement with a different parameter name (which may break `implements: I` if name matching is required), (b) dispatch via `raw_call`, or (c) rename the parameter at the interface level if the user controls the interface source.

Sporarum:
### What I did

Fix #5143

### How I did it

checking for 'self' parameters in '_parse_args' which is called when constructing 'ContractFunctionT's.
This ensures all functions are validated, even ones which come from interfaces.

### How to verify it

`pytest`

### Commit message

```
interfaces were allowed to define methods whose parameters contained
`self`, which lead to making them impossible to implement. the reason is
that interface methods are not analyzed, and so the namespace collision
check never fired. this commit fixes it by checking for `self`
parameters in `_parse_args` which is called when constructing
`ContractFunctionT`s. This ensures all functions are validated, even
ones which come from interfaces.
```

Add check that interface functions do not have `self` as parameter, as that would make them un-implementable

### Cute Animal Picture

### DIFF
--- a/tests/functional/syntax/test_interfaces.py
+++ b/tests/functional/syntax/test_interfaces.py
@@ -266,6 +266,13 @@ def f():
         """,
         InterfaceViolation,
     ),
+    (
+        """
+interface Foo:
+    def bar(self: uint256): nonpayable
+        """,
+        ArgumentException,
+    ),
 ]
 
 
@@ -525,6 +532,25 @@ from ethereum.ercs import {erc}
     assert e.value._message == f"ethereum.ercs.{erc}"
     assert e.value._hint == f"try renaming `{erc}` to `I{erc}`"
     assert "code.vy:" in str(e.value)
+
+
+def test_interface_self_param_vyi(make_input_bundle):
+    interface_code = """
+def bar(self: uint256): ...
+"""
+    input_bundle = make_input_bundle({"foo.vyi": interface_code})
+
+    code = """
+import foo as Foo
+
+@external
+def baz():
+    pass
+"""
+    with pytest.raises(ArgumentException) as e:
+        compiler.compile_code(code, input_bundle=input_bundle)
+
+    assert e.value.message == "Cannot use 'self' as a variable name in a function input"
 
 
 def test_interface_body_check(make_input_bundle):
--- a/vyper/semantics/types/function.py
+++ b/vyper/semantics/types/function.py
@@ -1083,7 +1083,8 @@ def _parse_args(
 
     for i, arg in enumerate(funcdef.args.args):
         argname = arg.arg
-        if argname in ("gas", "value", "skip_contract_check", "default_return_value"):
+        # TODO: Forbid `self`/`Self`/... everywhere, would be a breaking change
+        if argname in ("gas", "value", "skip_contract_check", "default_return_value", "self"):
             raise ArgumentException(
                 f"Cannot use '{argname}' as a variable name in a function input", arg
             )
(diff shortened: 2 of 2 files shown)


charles-cooper:
"the reason is that interface methods are not analyzed" so why don't we analyze them?

Sporarum:
> "the reason is that interface methods are not analyzed" so why don't we analyze them?

We should !

But that would be quite more involved, especially given the json interfaces
So this PR instead uses a shortcut

charles-cooper:
> > "the reason is that interface methods are not analyzed" so why don't we analyze them?
> 
> We should !
> 
> But that would be quite more involved, especially given the json interfaces So this PR instead uses a shortcut

in that case i think this PR has questionable gain because it introduces code which may need to be removed later (and whose purpose is a bit hard to understand)

Sporarum:
> in that case i think this PR has questionable gain because it introduces code which may need to be removed later (and whose purpose is a bit hard to understand)

Note that we should probably still check that parameters are not `self` (even if just as a sanity check), so it wouldn't realy be "code which needs to be removed later"

But as you prefer

charles-cooper:
yea idk. the other rejected ones are kwargs with special semantics. so it's weird to share the same codepath. other options include just rejecting self everywhere

Sporarum:
Yes as well, so what should we do ?

charles-cooper:
at least we should split out the condition to its own branch as well as add another constant `RESERVED_PARAM_NAMES`

PR #5175, opened 2026-06-30
fix[lang]: fix empty lists having incorrect type

### LINKED ISSUE #5161: Empty lists are not considered "lists of anything"
### Version Information

* vyper Version (output of `vyper --version`): 0.5.0a4+commit.751931ff
* OS: NixOS
* Python Version (output of `python --version`): Python 3.11.15

### What's your issue about?

```python
flag Foo:
  Member1

def foo():
    tmp: DynArray[DynArray[Foo, 5], 5] = [[], [Foo.Member1]]
```
raises:
```
vyper.exceptions.InvalidLiteral: Array contains multiple, incompatible types

  contract "playground/test.vy:6", function "foo", line 6:41 
       5 def foo():
  ---> 6     tmp: DynArray[DynArray[Foo, 5], 5] = [[], [Foo.Member1]]
  ------------------------------------------------^
       7
```
Even though both elements of the outer array are valid `DynArray[Foo, 5]`:
```
flag Foo:
  Member1

def foo():
    tmp1: DynArray[Foo, 5] = []
    tmp2: DynArray[Foo, 5] = [Foo.Member1]
    tmp: DynArray[DynArray[Foo, 5], 5] = [tmp1, tmp2]
```

An empty list is inferred to be a list of primitive types, when an empty list is a list of anything !

This is most likely due to `types_from_List` in `vyper/semantics/analysis/utils.py`:
```python
                # empty list literal `[]`
                # subtype can be anything
                subtypes = types.PRIMITIVE_TYPES.values()
```

The reason single-deep literals like `tmp1` do not cause this issue, is that `validate_expected_type` does not compare list literals against the expected type.
Instead, if a list type is expected, it loops over array elements and validating each one.
Since an empty array has no elements, it is considered valid for any element type.

### How can it be fixed?

The standard solution is to infer empty collections to <collection>[<bottom type>].
However this only works for covariant, and thus immutable, collections.
Since both array types are mutable, this doesn't work.

Instead there are two solutions:
1. add a new covariant type for array literals and a bottom type
2. instead of inferring the type of array literals, and then type-checking them, traverse them and deconstruct the expected type in tandem

Solution 2. seems to be what is attempted by the mutual recursion of `validate_expected_type`/`_validate_literal_array`, but there is an issue somewhere.

Sporarum:
### What I did

Fix #5161

### How I did it

Added a bottom type, this greatly simplifies the logic, now we can just let widening do its magic.

### How to verify it

`pytest`

### Commit message

```
```

### Cute Animal Picture

### DIFF
--- a/vyper/semantics/analysis/utils.py
+++ b/vyper/semantics/analysis/utils.py
@@ -24,7 +24,7 @@ from vyper.semantics import types
 from vyper.semantics.analysis.base import ExprInfo, Modifiability, ModuleInfo, VarAccess, VarInfo
 from vyper.semantics.analysis.levenshtein_utils import get_levenshtein_error_suggestions
 from vyper.semantics.namespace import get_namespace
-from vyper.semantics.types.base import TYPE_T, VyperType
+from vyper.semantics.types.base import TYPE_T, BottomT, VyperType
 from vyper.semantics.types.bytestrings import BytesT, StringT
 from vyper.semantics.types.infinity import is_bounded_length
 
@@ -349,28 +349,10 @@ class _ExprAnalyser:
 
     def types_from_List(self, node):
         # literal array
-        if _is_empty_list(node):
-            ret = []
 
-            if len(node.elements) > 0:
-                # empty nested list literals `[[], []]`
-                subtypes = self.get_possible_types_from_node(node.elements[0])
-            else:
-                # empty list literal `[]`
-                # subtype can be anything
-                subtypes = types.PRIMITIVE_TYPES.values()
-
-            for t in subtypes:
-                # 1 is minimum possible length for dynarray,
-                # can be assigned to anything
-                if isinstance(t, VyperType):
-                    ret.append(DArrayT(t, 1))
-                elif isinstance(t, type) and issubclass(t, VyperType):
-                    # for typeclasses like bytestrings, use a generic type acceptor
-                    ret.append(DArrayT(t.any(), 1))
-                else:
-                    raise CompilerPanic(f"busted type {t}", node)
-            return ret
+        if len(node.elements) == 0:
+            # can't have an empty SArrayT
+            return [DArrayT(BottomT(), 1)]
 
         types_list = get_common_types(*node.elements)
 
@@ -430,18 +412,6 @@ class _ExprAnalyser:
         # unary operation: `-foo`
         types_list = self.get_possible_types_from_node(node.operand)
         return _validate_op(node, types_list, "validate_numeric_op")
-
-
-def _is_empty_list(node):
-    # Checks if a node is a `List` node with an empty list for `elements`,
-    # including any nested `List` nodes. ex. `[]` or `[[]]` will return True,
-    # [1] will return False.
-    if not isinstance(node, vy_ast.List):
-        return False
-
-    if not node.elements:
-        return True
-    return all(_is_empty_list(t) for t in node.elements)
 
 
 def _is_type_in_list(obj, types_list):
@@ -598,17 +568,9 @@ def validate_expected_type(node, expected_type):
 
     given_types = _ExprAnalyser().get_possible_types_from_node(node)
 
-    if isinstance(node, vy_ast.List):
-        # special case - for literal arrays we individually validate each item
-        for expected in expected_type:
-            if not isinstance(expected, (DArrayT, SArrayT)):
-                continue
-            if _validate_literal_array(node, expected):
-                return
-    else:
-        for given, expected in itertools.product(given_types, expected_type):
-            if given.is_subtype_of(expected):
-                return
+    for given, expected in itertools.product(given_types, expected_type):
+        if given.is_subtype_of(expected):
+            return
 
     # validation failed, prepare a meaningful error message
     if len(expected_type) > 1:
--- a/vyper/semantics/types/base.py
+++ b/vyper/semantics/types/base.py
@@ -37,6 +37,9 @@ class _GenericTypeAcceptor:
         return self.compare_type(other) and other.compare_type(self)
 
     def compare_type(self, other):
+        if isinstance(other, BottomT):
+            return True
+
         if isinstance(other, self.type_):
             return True
         # compare two GenericTypeAcceptors -- they are the same if the base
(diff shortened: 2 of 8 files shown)


vyper/semantics/types/base.py
```diff
@@ -480,6 +486,17 @@ def __init__(self, typ, default, require_literal=False):
...
 
+class BottomT(VyperType):
+    """
+    Bottom type, the ultimate subtype: is a subtype of every other type.
+    It is unhabited: no value has this type.
+
+    It is for example the element type for empty lists: `[]: DynArray[Never, 1]`
+    """
+
+    _id = "Never"  # see python's typing.Never
```
charles-cooper, line 497: cute but does having this id have any semantic meaning? otherwise i would just call it `(bottom)`

vyper/semantics/analysis/utils.py
```diff
@@ -349,28 +349,10 @@ def types_from_IfExp(self, node):
...
-                    ret.append(DArrayT(t, 1))
-                elif isinstance(t, type) and issubclass(t, VyperType):
-                    # for typeclasses like bytestrings, use a generic type acceptor
-                    ret.append(DArrayT(t.any(), 1))
-                else:
-                    raise CompilerPanic(f"busted type {t}", node)
-            return ret
+        if len(node.elements) == 0:
+            # can't have an empty SArrayT
+            return [DArrayT(BottomT(), 1)]
```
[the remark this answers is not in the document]
charles-cooper: iiuc https://github.com/vyperlang/vyper/pull/5174 resolves this @Sporarum ?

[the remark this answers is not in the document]
Sporarum: Yes exactly, I split my changes in two to make it more readable

vyper/semantics/types/base.py
Sporarum: What do you mean by semantic meaning ?

It will probably show up in error messages, so I'd rather give it a proper name

Sporarum:
Superseded by #5177

vyper/semantics/types/base.py
charles-cooper: it shouldn't really show up in error messages but if we think that might happen then we should give it a more recognizable name like `wildcard` or `any`

Sporarum: But it's neither a wildcard nor anything !
`Never` as a name is very apt I think, it's a type that never has a value inside
It's also pretty common across programming languages: Python, TypeScript, PHP and Dart use it
And the similar `Nothing` is used by Kotlin and Scala
(source: https://en.wikipedia.org/wiki/Bottom_type#In_programming_languages)

charles-cooper: why do you think it's different from a wildcard?

Sporarum: It's at most a one-sided wildcard, like `Bytes[0]` is for `Bytes`:

You can override a method with `param: Bytes[0]` with a method which takes `param: Bytes[n]` for any `n` (including `0` and `INF`)
But you can only override a method with `-> Bytes[0]` with another method with `-> Bytes[0]`

In other words:
If I tell you, "you can give me 0 bytes", I don't have to crash if you give me more
But if I tell you "I will give you 0 bytes", giving you more would break that promise

In the same way `Bytes[0]` is the bottom type of the `Bytes` hierarchy, `Never` is the bottom of the whole type hierarchy.
And so `param: Never` can be overridden by `param: T` for any T
But `-> Never` can only be overridden by another `-> Never`

And for one-sided wildcards of the "other side", we have `Bytes[INF]` and `object`/`VyperType.any()` the top types of the `Bytes` hierarchy, and the full type hierarchy, respectively

Contrast this with `Bytes[...]` where both `param: Bytes[...]` and `-> Bytes[...]` can be overridden by any `Bytes[n]`, this acts like [`typing.Any`](https://docs.python.org/3/library/typing.html#the-any-type) (this page also explains nicely the differences between it an the top type `object`)

charles-cooper: i prefer 'bottom' to 'never' then -- and also it should never really surface in error messages tbh unless we have a bug

Sporarum: I don't see why it shouldn't surface to users, it's a real part of the type system !

For now there are not a lot of ways to make it appear, but with things like polymorphism, it will be more commonplace

charles-cooper: then let's call it `any`?

Sporarum: Definitely not !
1. `Any`/`any` is the name of the **wildcard** type in python and typescript
2. `Any`/`any` is the name of the **top** type in C++, Rust, Kotlin, Scala, Swift (source: [wikipedia](https://en.wikipedia.org/wiki/Any_type))

So we should not call our **bottom** type `any` !


P.S: By "wildcard type" I mean "type which is both a universal supertype and a universal subtype", apparently the name used in literature for this is "dynamic type"

We basically have two potential names:
1. `Never` to be consistent with python
2. `(bottom)`/`Bottom` to be faithful to it's type-theoretic name

But no language calls their bottom type "bottom", and I think that makes sense: You don't want your users to need to know about type theory

(And we should not add a wildcard type to the type system unless we **absolutely** have to, and in this case, a bottom type is a much better fit)

charles-cooper: hmm i believe bottom type is well known in haskell https://wiki.haskell.org/Bottom

we can also call it `undefined` or `void`

Sporarum: > bottom type is well known in haskell

Ah right, doesn't suprise me that haskell assumes you have type-theory knowledge tho ^^

> we can also call it `undefined` or `void`

We could, but why would we ?
Isn't the goal to stick as close to Python as possible ?
(Especially for something as simple as "how do we name this ?")

I'll leave it as-is (`Never`) in https://github.com/vyperlang/vyper/pull/5177 and we can always open a separate issue/PR for it

charles-cooper: sure we can leave it as is -- just as an anecdotal example: personally i've used haskell and python both professionally, but python much longer than haskell; i know what bottom is, i've never heard of Never.

PR #5181, opened 2026-07-07
fix[lang]: forbid assignment to values

### LINKED ISSUE #3933: Assigning a Builtin to Itself Leads to Panic
### Version Information

### Issue description

Assigning some builtin to itself is not prevented and leads to a
compiler panic.
 
#### POC

For example, compiling the following code leads the compiler to raise
the following exception: `CodegenPanic: unhandled exception ,
parse_Name`.

``` python
@external
def foo():
    convert = convert
```
credits: @trocher

### LINKED ISSUE #5138: Assign to call-return index crashes compiler
### Version Information

* OS: Linux Ubuntu Jammy
* Python Version: Python 3.11

### What's your issue about?

Assigning (`=` or augmented `+=`) to an **index or struct field of a function-call return value** leaks a bare Python `AssertionError` (no message) that is chained into `vyper.exceptions.CompilerPanic: unhandled exception`. The user sees the `AssertionError`, the Python `During handling of the above exception, another exception occurred:` boilerplate, and the panic banner directing them to file a GitHub issue.

The code that triggers it:

```vyper
@internal
def g() -> uint256[3]:
    return [1, 2, 3]

@external
def f():
    self.g()[0] = 5   # AssertionError ⇒ CompilerPanic: unhandled exception
```

Output:

```
AssertionError

vyper.exceptions.CompilerPanic: unhandled exception
```

The same failure occurs for the struct-field form and for augmented assignment:

```vyper
self.g().x = 5      # g() -> struct S
self.g()[0] += 5
self.g().x += 1
```

More generally, the trigger is **any write target whose base is a `Call` node**. The same bare `AssertionError` ⇒ `CompilerPanic` chain reproduces for:

```vyper
S(x=1).x = 5                 # struct-literal field
S(arr=[1,2,3]).arr[0] = 5    # struct-literal array field
empty(uint256[3])[0] = 5     # builtin-return index
```

**Asymmetry:** the bare form `self.g() = 5` is caught cleanly and early at parse with `SyntaxException: cannot assign to function call here. Maybe you meant '==' instead of '='?`. Only indexing/field-access into the call result slips past the parse guard and reaches the semantic-stage bare assert.

**Controls that compile cleanly** (index/field assign on a proper lvalue):

```vyper
self.a[0] = 5                          # storage
a: uint256[3] = [1, 2, 3]; a[0] = 5    # local
```

The literal-array target `[1,2,3][0] = 5` is instead rejected earlier with `StructureException: Ambiguous type`. Note also that the read-side of the same rvalue misuse already has a clean guard — `for i in self.g()` yields `IteratorException: Cannot iterate over the result of a function call` — so the call-return rvalue is recognized and cleanly rejected for iteration, but the mutation paths panic.

### How can it be fixed?

The root is in `vyper/semantics/analysis/local.py` `_handle_modification` (reached via `_assign_helper` → `visit_Assign`/`visit_AugAssign`), lines 578-579:

```python
var_access = _get_variable_access(target)
assert var_access is not None          # line 579
```

`_get_variable_access(target)` walks the target expression to find the underlying VarInfo/variable whose write should be recorded. For a `Subscript`/`Attribute` whose base is a `Call` (e.g. `self.g()`), there is no underlying variable — a call's return value is an ephemeral rvalue — so it returns `None`. The bare `assert var_access is not None` then raises an unmessaged `AssertionError`, which the top-level compile wrapper re-raises as `CompilerPanic: unhandled exception` without `from None`, leaking the Python exception chain.

Expected behavior: a clean `StructureException` (e.g. "cannot assign to an index/member of a function call", or an lvalue-required diagnostic), mirroring the clean parse-stage `SyntaxException` the bare form already gets. Concretely: replace the bare `assert ... ` with `assert ..., msg` raising a proper `StructureException`, or guard the `None` return from `_get_variable_access` and raise a proper lvalue error `from None`.

### LINKED ISSUE #5151: Writing self.balance panics under Venom codegen
### Version Information

* OS: Linux Ubuntu Jammy
* Python Version: Python 3.11

### What's your issue about?

Writing to a read-only attribute of `self` (e.g. `self.balance`, `self.codesize`) raises a `CompilerPanic` under the `--experimental-codegen` (Venom) pipeline, while the legacy pipeline raises a structured `StructureException`. Since `CompilerPanic` is conventionally interpreted as "internal compiler bug, please file an issue", the venom pipeline misleads users whose code is simply wrong.

Reproducer:

```vyper
@external
def f():
    self.balance = 100
```

Behavior:

- `vyper t.vy` (legacy) → `StructureException: Looking for a variable location, instead got a value`
- `vyper --experimental-codegen t.vy` (venom) → `CompilerPanic: Unsupported attribute target: balance`

The sibling case `self.codesize = 0` triggers the same panic with `Unsupported attribute target: codesize`.

Other env-var writes (`msg.sender = ...`, `block.timestamp = ...`, `tx.origin = ...`) raise a clean `ImmutableViolation` in BOTH pipelines. Only `self.<read-only proxy attr>` writes (balance, codesize) hit the venom panic.

The panic originates at `vyper/codegen_venom/stmt.py:471`:

```python
raise CompilerPanic(f"Unsupported attribute target: {target.attr}")  # pragma: nocover
```

This is the fall-through path of an attribute-target dispatcher in `_lower_assignment_target`. For `self.balance`:

- The varinfo lookup returns `None` (balance isn't a varinfo entry).
- The struct-field branch doesn't apply (`self` is a `SelfT`, not a `StructT`).
- Fall-through to `raise CompilerPanic(...)`.

The `# pragma: nocover` annotation indicates the developer expected this branch to be unreachable, but it IS reachable from user code.

Confirmed divergent cases:

| User code             | Legacy                                    | Venom                                                |
|-----------------------|-------------------------------------------|------------------------------------------------------|
| `self.balance = X`    | StructureException                        | CompilerPanic: Unsupported attribute target: balance |
| `self.codesize = X`   | StructureException                        | CompilerPanic: Unsupported attribute target: codesize|
| `self.code = X`       | StructureException (gated by .code rule)  | StructureException (same — .code path catches first) |
| `msg.sender = X`      | ImmutableViolation                        | ImmutableViolation                                   |
| `block.timestamp = X` | ImmutableViolation                        | ImmutableViolation                                   |
| `tx.origin = X`       | ImmutableViolation                        | ImmutableViolation                                   |

Root cause: venom's semantic-stage lvalue validator has narrower coverage than legacy's — legacy correctly recognizes `self.balance` as a non-assignable "value expression" at the semantic analysis stage, whereas venom defers to codegen, where the defensive panic catches it.

### How can it be fixed?

Two options:

1. Add a `self.<read-only attr>` check to the semantic-stage local-analysis validator (e.g. in `vyper/semantics/analysis/local.py`) that mirrors the legacy lvalue check, so the error fires as `StructureException` (or `ImmutableViolation`) before reaching codegen.
2. Upgrade the `CompilerPanic` at `vyper/codegen_venom/stmt.py:471` to a `StructureException` for user-facing clarity. The `# pragma: nocover` annotation should be removed since the branch is reachable from user code.

Sporarum:
### What I did

Fix #5138
Fix #5151
Fix #3933

In short: we didn't check what was on the lhs of assignments enough, so things like `self.g()[0] = 5` were allowed => failed asserts

### How I did it

Check that the lhs of assignments has a location (memory/transient/storage), and if not, raise an exception.
This catches even tricky cases, like the long-standing `self.nostedArray.pop().append()` which otherwise have correct type and mutability

### How to verify it

`pytest`

### Commit message

```
the left hand-side of assignments used to only be checked to have
correct mutability and location. however some expressions fulfilled
these criterion without representing a concrete offset in
memory/transient/storage, for example 'self.g()[0] = 5'. this commit
addresses that by directly checking that the lhs has a location.
```

Compiler no longer panics when trying to assign to expressions

### Cute Animal Picture

### DIFF
--- a/vyper/codegen/expr.py
+++ b/vyper/codegen/expr.py
@@ -35,7 +35,6 @@ from vyper.exceptions import (
     CodegenPanic,
     CompilerPanic,
     EvmVersionException,
-    StructureException,
     TypeCheckFailure,
     TypeMismatch,
     UnimplementedException,
--- a/vyper/codegen/expr.py
+++ b/vyper/codegen/expr.py
@@ -825,6 +825,6 @@ class Expr:
     @classmethod
     def parse_pointer_expr(cls, expr, context):
         o = cls(expr, context).ir_node
-        if not o.location:
-            raise StructureException("Looking for a variable location, instead got a value", expr)
+        # Looking for a variable location, instead got a value
+        assert o.location
         return o
--- a/vyper/semantics/analysis/local.py
+++ b/vyper/semantics/analysis/local.py
@@ -567,6 +567,11 @@ class FunctionAnalyzer(VyperNodeVisitorBase):
         if info.modifiability == Modifiability.CONSTANT:
             raise ImmutableViolation("Constant value cannot be written to.")
 
+        if info.location == DataLocation.UNSET:
+            raise StructureException(
+                f"Cannot modify `{target.node_source_code}` since it is not a reference", target
+            )
+
         var_access = _get_variable_access(target)
         assert var_access is not None
 
--- a/vyper/semantics/analysis/local.py
+++ b/vyper/semantics/analysis/local.py
@@ -187,9 +187,9 @@ def _validate_pure_access(node: vy_ast.Attribute | vy_ast.Name, typ: VyperType) 
 
 
 # analyse the variable access for the attribute chain for a node
-# e.x. `x` will return varinfo for `x`
-# `module.foo` will return VarAccess for `module.foo`
-# `self.my_struct.x.y` will return VarAccess for `self.my_struct.x.y`
+# e.x. `x` will return `VarAccess(<x>, ())`
+# `module.foo` will return `VarAccess(<module.foo>, ())`
+# `self.my_struct.x.y` will return `VarAccess(<self.my_struct>, ('x', 'y'))`
 def _get_variable_access(node: vy_ast.ExprNode) -> Optional[VarAccess]:
     path: list[str | object] = []
     info = get_expr_info(node)
@@ -210,9 +210,6 @@ def _get_variable_access(node: vy_ast.ExprNode) -> Optional[VarAccess]:
         node = node.value
         info = get_expr_info(node)
 
-    # ignore `self.` as it interferes with VarAccess comparison across modules
-    if len(path) > 0 and path[-1] == "self":
-        path.pop()
     path.reverse()
 
     return VarAccess(info.var_info, tuple(path))
--- a/vyper/semantics/analysis/local.py
+++ b/vyper/semantics/analysis/local.py
@@ -207,7 +207,6 @@ def _get_variable_access(node: vy_ast.ExprNode) -> Optional[VarAccess]:
         if (attr := info.attr) is not None:
             path.append(attr)
 
-        assert isinstance(node, (vy_ast.Subscript, vy_ast.Attribute))  # help mypy
         node = node.value
         info = get_expr_info(node)
 
--- a/vyper/semantics/analysis/local.py
+++ b/vyper/semantics/analysis/local.py
@@ -569,7 +569,7 @@ class FunctionAnalyzer(VyperNodeVisitorBase):
 
         if info.location == DataLocation.UNSET:
             raise StructureException(
-                f"Cannot modify `{target.node_source_code}` since it is not a reference", target
+                f"`{target.node_source_code}` is not a valid assignment target", target
             )
 
         var_access = _get_variable_access(target)
--- a/vyper/semantics/analysis/local.py
+++ b/vyper/semantics/analysis/local.py
@@ -169,7 +169,7 @@ def _validate_pure_access(node: vy_ast.Attribute | vy_ast.Name, typ: VyperType) 
             )
         # allow type exprs in the value node, e.g. MyFlag.A
         parent_info = get_expr_info(node.value, is_callable=True)
-        if isinstance(parent_info.typ, AddressT) and node.attr in AddressT._type_members:
+        if isinstance(parent_info.typ, AddressT) and node.attr in AddressT._builtin_members:
             raise StateAccessViolation("not allowed to query address members in pure functions")
 
     if (varinfo := info.var_info) is None:
--- a/vyper/semantics/analysis/utils.py
+++ b/vyper/semantics/analysis/utils.py
@@ -23,6 +23,7 @@ from vyper.exceptions import (
 from vyper.semantics import types
 from vyper.semantics.analysis.base import ExprInfo, Modifiability, ModuleInfo, VarAccess, VarInfo
 from vyper.semantics.analysis.levenshtein_utils import get_levenshtein_error_suggestions
+from vyper.semantics.data_locations import DataLocation
 from vyper.semantics.namespace import get_namespace
 from vyper.semantics.types.base import TYPE_T, VyperType
 from vyper.semantics.types.bytestrings import BytesT, StringT
@@ -109,6 +110,13 @@ class _ExprAnalyser:
 
             if isinstance(t, ModuleInfo):
                 return ExprInfo.from_moduleinfo(t, attr=attr)
+
+            if info.typ._type_members and attr in info.typ._type_members:
+                # things like `addr.balance` should not inherit the location of `addr`
+                # since `addr` can be assignable, while `addr.balance` never is
+                return ExprInfo(
+                    t, attr=attr, location=DataLocation.UNSET, modifiability=info.modifiability
+                )
 
             return info.copy_with_type(t, attr=attr)
 
--- a/vyper/semantics/analysis/utils.py
+++ b/vyper/semantics/analysis/utils.py
@@ -111,7 +111,7 @@ class _ExprAnalyser:
             if isinstance(t, ModuleInfo):
                 return ExprInfo.from_moduleinfo(t, attr=attr)
 
-            if info.typ._type_members and attr in info.typ._type_members:
+            if info.typ._builtin_members and attr in info.typ._builtin_members:
                 # things like `addr.balance` should not inherit the location of `addr`
                 # since `addr` can be assignable, while `addr.balance` never is
                 return ExprInfo(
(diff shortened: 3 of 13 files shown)


vyper/codegen/expr.py
```diff
@@ -825,6 +825,6 @@ def parse_value_expr(cls, expr, context):
     @classmethod
     def parse_pointer_expr(cls, expr, context):
         o = cls(expr, context).ir_node
-        if not o.location:
-            raise StructureException("Looking for a variable location, instead got a value", expr)
+        # Looking for a variable location, instead got a value
+        assert o.location
```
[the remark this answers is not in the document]
Sporarum: The correct fix for this would be to make `<address>.balance` (and other methods on `address`) have `RUNTIME_CONSTANT` mutability

But this is somewhat involved, as currently members cannot have a mutability different than what they are attached to

Sporarum:
I'm not sure what the wording should be
In other words: How do we call what's allowed on the lhs of an assignment ?
I don't want to use 'l-value' as only C programmers will be familiar with it
But then I don't have good words to describe it "value", "expression", "reference" ?
I considered "variable", but I wouldn't call `a.b[10]` a variable, since we're indexing into it.

Ideas welcome !

charles-cooper:
lvalue, assignment target or even 'left hand side of an assignment' are all ok

vyper/semantics/analysis/local.py
```diff
@@ -571,6 +567,11 @@ def _handle_modification(self, target: vy_ast.ExprNode):
         if info.modifiability == Modifiability.CONSTANT:
             raise ImmutableViolation("Constant value cannot be written to.")
 
+        if info.location == DataLocation.UNSET:
+            raise StructureException(
+                f"Cannot modify `{target.node_source_code}` since it is not a reference", target
```
charles-cooper:
```suggestion
                f"`{target.node_source_code}` is invalid as an assignment target", target
```

vyper/codegen/expr.py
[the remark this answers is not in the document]
Sporarum: Fixed by un-setting the location on "built-in members" of types

A more powerful and correct fix would be to have more granular control over the mutability of attribute accesses, this would allow us to potentially simplify the "not `uses`/`initializes`-ed" check (`other_module`.foo could be set as immutable), and to add private fields to modules (`self.counter` would be mutable in `lib`, but `lib.counter` would be immutable)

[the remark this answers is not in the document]
Sporarum: (would also need a new Mutabillity, since `addr.balance` is neither `CONSTANT` nor `RUNTIME_CONSTANT`)

vyper/semantics/analysis/utils.py
```diff
@@ -111,6 +111,13 @@ def get_expr_info(self, node: vy_ast.VyperNode, is_callable: bool = False) -> Ex
             if isinstance(t, ModuleInfo):
                 return ExprInfo.from_moduleinfo(t, attr=attr)
 
+            if info.typ._type_members and attr in info.typ._type_members:
+                # things like `addr.balance` should not inherit the location of `addr`
+                # since `addr` can be assignable, while `addr.balance` never is
+                return ExprInfo(
+                    t, attr=attr, location=DataLocation.UNSET, modifiability=info.modifiability
```
charles-cooper, line 118: why does location get overridden?

vyper/semantics/analysis/utils.py
```diff
@@ -111,6 +111,13 @@ def get_expr_info(self, node: vy_ast.VyperNode, is_callable: bool = False) -> Ex
             if isinstance(t, ModuleInfo):
                 return ExprInfo.from_moduleinfo(t, attr=attr)
 
+            if info.typ._type_members and attr in info.typ._type_members:
+                # things like `addr.balance` should not inherit the location of `addr`
```
charles-cooper, line 115: the condition does not match the comment

vyper/semantics/analysis/utils.py
Sporarum: There are two reasons:
1. Conceptually, `addr.balance` is not stored anywhere, it's a query we execute
2. Practically, having no location is what I use to emit the error, so without it `addr.balance = something` would pass the front-end

vyper/semantics/analysis/utils.py
Sporarum: Not sure what you mean
`balance` on `address` is a built-in member (aka in `_type_members`)
So this branch really does match "things like `addr.balance`"

Oh and what we do in the branch is to reset the location, matching the "should not inherit the location" part

vyper/semantics/analysis/utils.py
charles-cooper: > There are two reasons:
> 
> Conceptually, addr.balance is not stored anywhere, it's a query we execute
> Practically, having no location is what I use to emit the error, so without it addr.balance = something would pass the front-end

i understand that's the case for `balance` and how you have emitted the error. does it apply unconditionally for anything in `_type_members`?

Sporarum: Yes, `_type_member` are the built-in members, and all built-in members are in some way magical and can't be assigned to

vyper/semantics/analysis/utils.py
charles-cooper: no? like it's perfectly conceivable that some builtin member should inherit the location

Sporarum: I don't know, currently there are none where that makes sense, they are all magical in some way

The only thing I could imagine would be things of this sort:
```
struct Example:
  x: uint256
  y: uint256

def foo():
  ex: Example = Example(0, 0)
  ex.__bytes__ = b"\x01\x02"
  assert ex.x = 1
  assert ex.y = 2
```

But is anything like this planned ?
(Note that mutation methods on built-in types are regular `members` and not `_type_members`)

charles-cooper: not planned but i could envision it. i think the code as is just ties two concepts together that should not be

Sporarum: Okay, I think I know how to do it better, but I need #5204 merged first
(It makes types register as immutable, which I need for this)

charles-cooper (approved):

Issue #4957, opened 2026-05-08
ICE in venom stack-order analysis: `AssertionError` in `_reorder`


jubnzv:
The compiler panics with `AssertionError` in [`vyper/venom/analysis/stack_order.py:131`](https://github.com/vyperlang/vyper/blob/7d73c46878e850987891ac21430fc4d29af33d5a/vyper/venom/analysis/stack_order.py#L131) when compiling a `for` loop with two `if`/`break` exits, only one of which writes a local, under `--experimental-codegen`.

MRE:
```vyper
@external
def f(_t: uint256) -> uint256:
    s: uint256 = 0
    for i: uint256 in range(32):
        if i == _t:
            break
        if i == _t:
            s = 1
            break
    return s
```

Reproduce steps:
```
vyper --experimental-codegen mre.vy
```

Output:
```
Error compiling: mre.vy
AssertionError: ([6_for_cond, %alloca_15_1, 7_for_body, %alloca_15_1, 13_then, %alloca_15_1:2], [6_for_cond, 7_for_body, %alloca_15_1, 13_then, %alloca_15_1:2])
```

The default codegen (`vyper mre.vy`) compiles successfully.

Expected behavior: successful compilation under `--experimental-codegen`, matching the default codegen.

Git commit: 7d73c46878e850987891ac21430fc4d29af33d5a
vyper 0.4.4+commit.5ea0f386
Python 3.13.5

PR #5183, opened 2026-07-09
fix[venom]: fix stack ordering pass for phi instructions

HodanPlodky:
### What I did
Fix the issue with the `phi` handling that was shown here https://github.com/vyperlang/vyper/issues/4957, this issue was also addressed by  https://github.com/vyperlang/vyper/pull/5101. However that PR basically ignored the the issue instead of the doing proper semantics of `phi`

### Commit message

```
`StackOrderAnalysis` simulates a stack per basic block to compute the
preferred variable ordering at block boundaries. `phi` instructions were
handled by the generic `_handle_inst` path, which pushes every operand
onto the simulated stack and then asserts (in `_reorder`) that the stack
top matches the full operand list. This does not match phi semantics: at
runtime only one incoming value is on the stack (whichever predecessor
branch was taken), and `venom_to_assembly` lowers a phi by renaming that
single stack slot in place rather than consuming all operands. When the
same variable flowed in from multiple predecessors, the simulated stack
held fewer entries than the operand list, tripping the `_reorder`
assertion (GH 4957).

Fix by giving phi its own handler that mirrors the `venom_to_assembly`
lowering: record all incoming variables as needed, but simulate only a
single chosen operand on the stack, then replace it with the phi's
output. Since any incoming value must produce the same stack shape,
simulating one representative operand is sufficient.
```

### Cute Animal Picture

### DIFF
--- a/vyper/venom/analysis/stack_order.py
+++ b/vyper/venom/analysis/stack_order.py
@@ -43,6 +43,7 @@ class StackOrderAnalysis(IRAnalysis):
     _from_to: dict[tuple[IRBasicBlock, IRBasicBlock], Needed]
 
     def analyze(self):
+        print(self.function)
         self._from_to = dict()
         self.liveness = self.analyses_cache.request_analysis(LivenessAnalysis)
         self.cfg = self.analyses_cache.request_analysis(CFGAnalysis)
@@ -55,7 +56,10 @@ class StackOrderAnalysis(IRAnalysis):
             if inst.opcode == "assign":
                 self._handle_assign(inst)
             elif inst.opcode == "phi":
-                self._handle_inst(inst)
+                self._handle_phi(inst)
+                self.stack.pop()
+                self.stack.append(inst.output)
+                continue
             elif inst.is_bb_terminator:
                 self._handle_terminator(inst)
             else:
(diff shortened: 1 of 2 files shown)


charles-cooper:
paging @harkal for review

vyper/venom/analysis/stack_order.py
```diff
@@ -55,7 +55,10 @@ def analyze_bb(self, bb: IRBasicBlock) -> Needed:
             if inst.opcode == "assign":
                 self._handle_assign(inst)
             elif inst.opcode == "phi":
-                self._handle_inst(inst)
+                self._handle_phi(inst)
+                self.stack.pop()
+                self.stack.append(inst.output)
```
harkal: Move these in _handle_phi() to make it more contained?

harkal:
other than the small comment above lgtm

harkal (approved):

charles-cooper:
LGTM. A few non-blocking nits as potential follow-ups:

1. **Simulation diverges slightly from the backend**: `_reorder([chosen])` swaps the chosen operand to the top before renaming, while `venom_to_assembly` pokes the phi output in place at depth (`stack.poke(depth, ret)`). After the phi, the simulated stack's internal order can differ from what codegen produces. Heuristic-only impact since this analysis only feeds DFT scheduling.

2. **`needed` pollution across edges**: `_handle_phi` adds *all* incoming variables to `needed`, but `analyze_bb` copies the same `needed` list to every `(pred, bb)` edge — so a predecessor is told it "needs" a variable that only flows in along a different edge. Downstream this gets filtered naturally (DFT ignores vars not produced in the block; `_merge` just shortens the common prefix), so again heuristic degradation at worst. Per-edge `needed` would be more precise.

3. Cosmetic: comment typos ("varible", and the first comment reads garbled — "add vars to needed if it is necessary as in other instructions"); also `next(inst.phi_operands)` creates a second generator when the preceding loop already iterates the operands — could grab the first pair inside that loop instead.

PR #5185, opened 2026-07-11
fix[parser]: reject bare await keyword

### LINKED ISSUE #5142: User await leaks bare Python KeyError
### Version Information

* OS: Linux Ubuntu Jammy
* Python Version: Python 3.11

### What's your issue about?

A user-typed `await` keyword causes the compiler to raise a bare Python `KeyError: (line, col)` instead of a proper `vyper.exceptions.SyntaxException`. There is no source-location label, no caret arrow, and no diagnostic message — just the bare Python exception.

The following minimal contract triggers the bug:

```vyper
@external
def f():
    await something
```

Error output:

```
KeyError: (3, 4)
```

Other unsupported Python keywords are handled cleanly with a `SyntaxException`, for example:

```
try:          → SyntaxException: unsupported Try AST node
lambda x: x   → SyntaxException: unsupported Lambda AST node
f"..."        → SyntaxException: unsupported JoinedStr AST node
with X:       → SyntaxException: unsupported With AST node
match x:      → SyntaxException: unsupported Match AST node
yield x       → SyntaxException: yield keyword is not allowed
del x         → SyntaxException: Deleting is not supported
x[a:b]        → SyntaxException: Vyper does not support slicing
```

Root cause is in `vyper/ast/parse.py:402-404`:

```python
def visit_Await(self, node):
    start_pos = (node.lineno, node.col_offset)
    node.ast_type = self._pre_parser.keyword_translations[start_pos]
```

This code assumes that every `Await` AST node produced by Python's `ast.parse` was synthesized by Vyper's pre-parser as a syntactic vehicle for Vyper-specific keywords (`extcall foo(...)`, `staticcall foo(...)`, `log Event(...)`, etc., which look like `<keyword> <expression>` and are parsed by Python as `Await` expressions). The pre-parser populates `keyword_translations[(line, col)]` with the original Vyper keyword name for each rewritten site.

When a user types `await` directly (which is NOT a Vyper-known keyword), no entry was inserted at `(line, col)`, and the dict lookup raises Python's bare `KeyError` with no formatting, no Vyper exception class, and no message.

### How can it be fixed?

Guard the dict lookup with a membership check and raise a proper `SyntaxException` when the position is missing. For example, in `vyper/ast/parse.py:402-404`:

```python
def visit_Await(self, node):
    start_pos = (node.lineno, node.col_offset)
    if start_pos not in self._pre_parser.keyword_translations:
        _raise_syntax_exc("`await` is not a Vyper keyword", node, ...)
    node.ast_type = self._pre_parser.keyword_translations[start_pos]
```

It would also be worth auditing all consumers of `keyword_translations` for the same pattern, since the pre-parser uses Python's `Await` AST node (and potentially others) as a syntactic vehicle for Vyper-specific keywords and the same class of bug could occur elsewhere.

Functionhx:
### What I did

Fixes #5142.

User-written `await` now reports a normal `SyntaxException` at its source position instead of leaking a `KeyError`.

### How I did it

The AST visitor checks whether an `Await` node has a pre-parser keyword translation. Nodes produced for `extcall`, `staticcall`, and `log` keep their existing behavior; a native `await` without a translation raises the parser diagnostic.

### How to verify it

`pytest tests/unit/ast/test_parser.py -q`

The regression test checks the message, line, column, and source annotation. Related AST, external-call, logging, and event tests also pass.

### Commit message

```
`await` is a Python keyword that parses into an `Await` AST node, but
vyper only produces `Await` nodes internally via the pre-parser as a
carrier for `extcall`, `staticcall`, and `log`. `visit_Await`
unconditionally indexed `keyword_translations` by source position, so a
user-written `await` — for which no translation was registered —
surfaced as a bare `KeyError: (line, col)` with no diagnostic or source
location.

Guard the lookup: if the position is missing from the translation map,
the node was written by the user, so raise a `SyntaxException` pinned to
the `await` token. Pre-parser-synthesized `Await` nodes still hit the
translation path unchanged.

Fixes GH 5142.
```

Report a compiler syntax error for unsupported `await` expressions instead of a raw Python `KeyError`.

### Cute Animal Picture

Cute baby python reviewing code

### DIFF
--- a/tests/unit/ast/test_parser.py
+++ b/tests/unit/ast/test_parser.py
@@ -1,5 +1,8 @@
+import pytest
+
 from tests.ast_utils import deepequals
 from vyper.ast.parse import parse_to_ast
+from vyper.exceptions import SyntaxException
 
 
 def test_ast_equal():
@@ -34,3 +37,19 @@ def test() -> int128:
     ast2 = parse_to_ast(code2)
 
     assert not deepequals(ast1, ast2)
+
+
+def test_await_raises_syntax_exception():
+    code = """@external
+def f():
+    await something
+"""
+
+    with pytest.raises(SyntaxException) as exc_info:
+        parse_to_ast(code)
+
+    exc = exc_info.value
+    assert exc.message == "The `await` keyword is not allowed."
+    annotation = exc.annotations[0]
+    assert (annotation.lineno, annotation.col_offset) == (3, 4)
+    assert annotation.full_source_code == code
--- a/vyper/ast/parse.py
+++ b/vyper/ast/parse.py
@@ -409,6 +409,13 @@ class AnnotatingVisitor(python_ast.NodeTransformer):
     def visit_Await(self, node):
         start_pos = node.lineno, node.col_offset
         self.generic_visit(node)
+        if start_pos not in self._pre_parser.keyword_translations:
+            raise SyntaxException(
+                "The `await` keyword is not allowed.",
+                self._source_code,
+                node.lineno,
+                node.col_offset,
+            )
         node.ast_type = self._pre_parser.keyword_translations[start_pos]
         return node
 


Sporarum (approved):
LGTM !

charles-cooper (approved):

PR #5186, opened 2026-07-12
feat[venom]: add `noinline` function annotation

harkal:
### What I did

Add a `noinline` function-header annotation to Venom that opts a function out of `FunctionInlinerPass`. Previously the only control over inlining was the global `disable_inlining` flag; there was no way to keep a specific function out-of-line — in particular, a function with a single call site is always inlined unconditionally.

### How I did it

- Reuse the existing bracketed function-header annotation list (previously FMP-only): `function f [noinline] { ... }`.
- `parser.py`: add `noinline` to `_KNOWN_ANNOTATIONS`; `_apply_annotations` sets a bool on `IRFunction`, kept separate from the FMP signature (the flag carries no calling-convention facts). Annotations without `fmp_lowered` no longer error unconditionally, since `noinline` can now appear alone.
- `function.py`: new `noinline` attribute, carried through `copy()`; `FmpSignature.annotation` refactored into an `attrs` list so `__repr__` can merge FMP attributes with `noinline` into a single bracket, making the flag round-trip through the text format.
- `function_inliner.py`: `_select_inline_candidate` skips flagged functions, covering both the single-call-site and cost-threshold branches.
- `tests/venom_utils.py`: `assert_fn_eq` now compares `noinline` so round-trip tests aren't vacuous.

### How to verify it

```bash
pytest tests/functional/venom/parser/test_parsing.py::test_noinline_annotation
pytest tests/unit/compiler/venom/test_inliner.py::test_noinline_annotation
```

The parser test checks parsing and printer/parser round-trip of the flag (alone and combined with FMP annotations); the inliner test checks that a single-call-site function — which would otherwise always be inlined — survives when flagged, and is inlined without the flag.

### Commit message

```
the inliner offered only a global opt-out (`disable_inlining`); there
was no per-function way to keep a call out-of-line while leaving the
rest of the inlining heuristics intact. this matters in particular for
hand-written venom (tests, debugging passes that run downstream of the
inliner) where a call site must survive: a function with a single call
site is unconditionally inlined, so it cannot be kept alive by
restructuring alone.

add a `noinline` flag, carried by the existing bracketed function-header
annotation list (`function f [noinline] { ... }`). the parser sets a
bool on `IRFunction`, kept separate from the FMP signature since the
flag carries no calling-convention facts; the printer emits it so the
flag round-trips through the text format. `FunctionInlinerPass` skips
flagged functions during candidate selection, which covers both the
single-call-site and cost-threshold branches.
```

Add `noinline` function-header annotation to Venom, opting a function out of the function inliner.

### Cute Animal Picture

### DIFF
--- a/tests/functional/venom/parser/test_parsing.py
+++ b/tests/functional/venom/parser/test_parsing.py
@@ -400,14 +400,17 @@ def test_noinline_annotation():
     source = """
     function main {
         main:
-            invoke @f
             stop
     }
 
     function f [noinline] {
         f:
-            %retpc = param
-            ret %retpc
+            stop
+    }
+
+    function g [fmp_lowered, noinline] {
+        g:
+            stop
     }
     """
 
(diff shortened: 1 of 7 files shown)


charles-cooper:
why don't we have a `--inline-factor` tuning flag?

harkal:
> why don't we have a `--inline-factor` tuning flag?

Different purposes. We can have a `--inline-factor` on top of this —if exposing that factor is the right idea.. not really sure--

charles-cooper:
Isn't inline factor of 0 or inf equivalent to noinline?

harkal:
> Isn't inline factor of 0 or inf equivalent to noinline?

This is a per function annotation

charles-cooper (approved):

charles-cooper:
oh gotcha, lgtm then

PR #5188, opened 2026-07-14
feat[venom]: reduce spilling for deep dups

harkal:
### What I did

Reduced the cost of `DUP`ing operands deeper than 16 in the Venom stack spiller. Previously the spiller moved the entire stack segment above and including the target to memory and reloaded everything; now it spills only the prefix that blocks `DUP16` reachability.

### How I did it

- spill the top `dup_idx - 16` items instead of all `dup_idx` items, then duplicate the now-reachable operand with `DUP16`
- when the spilled prefix fits in swap range (`spill_count <= 16`), reload it in a permuted order (deepest item last) and finish with a single `SWAPn`, avoiding a `SWAP1` per reloaded item; deeper prefixes keep the per-item `SWAP1`
- emit the `DUP` opcode from the computed depth rather than a hardcoded literal

Cost per deep dup drops from `4 * dup_idx + 2` to `4 * (dup_idx - 16) + 2` ops for dup depths up to 32 (e.g. `DUP17`: 70 → 6 ops), and peak spill memory shrinks by 16 slots.

### How to verify it

`tests/unit/compiler/venom/test_stack_spill.py`:

- `test_dup_stack_model` is parametrized over dup depths 16–64, covering the no-spill case, both restore paths, and the `spill_count == 16/17` boundary; asserts stack model, opcode shape, and cost
- `test_partial_deep_dup_integration` checks the exact emitted opcode sequence around a spill end-to-end
- slot reuse, dry-run state preservation, and peak-memory tracking each have dedicated tests

### Commit message

```
duplicating an operand deeper than 16 previously spilled the entire
stack segment above and including the target to memory (dup_idx items)
and reloaded all of them plus the duplicate, costing 4 * dup_idx + 2 ops
per dup.

only the items that block DUP16 reachability actually need to move:
spill the top dup_idx - 16 items, duplicate the now-reachable operand
with DUP16, then reload the spilled prefix. when the prefix fits in swap
range (spill_count <= 16, i.e. dup depth <= 32), reload it in a permuted
order (deepest item last) so a single SWAPn drops it into the
duplicate's old position while lifting the duplicate to the top; deeper
prefixes fall back to one SWAP1 per reloaded item.

cost drops from 4 * dup_idx + 2 to 4 * (dup_idx - 16) + 2 ops for dup
depths up to 32 (e.g. DUP17: 70 -> 6 ops), and peak spill memory shrinks
by 16 slots per deep dup.
```

perf: spill fewer stack items when duplicating deep stack operands in Venom

### DIFF
--- a/tests/unit/compiler/venom/test_stack_spill.py
+++ b/tests/unit/compiler/venom/test_stack_spill.py
@@ -95,16 +95,146 @@ def test_dup_spills_deep_stack() -> None:
     assert isinstance(depth, int) and depth < -16
     dup_idx = 1 - depth
 
-    compiler.spiller.dup(assembly, stack, depth)
+    cost = compiler.spiller.dup(assembly, stack, depth)
 
     expected = before.copy()
     expected.append(target)
     assert stack._stack == expected
 
     ops_str = _ops_only_strings(assembly)
-    assert ops_str.count("MSTORE") == dup_idx
-    assert ops_str.count("MLOAD") == dup_idx + 1
-    assert all(int(op[3:]) <= 16 for op in ops_str if op.startswith("DUP"))
+    spill_count = dup_idx - 16
+    assert ops_str.count("MSTORE") == spill_count
+    assert ops_str.count("MLOAD") == spill_count
+    assert ops_str.count("SWAP1") == spill_count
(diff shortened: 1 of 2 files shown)


charles-cooper (approved):
lgtm

PR #5189, opened 2026-07-15
feat[venom]: halting paths stack optimization

harkal:
### What I did

Skip dead-value stack cleanup (POPs) at basic-block entry when the block is in
a must-halt region — every path from it ends the message call — and it is
provably safe to leave the dead slots on the stack.

### How I did it

Two new analyses plus a codegen hook:

- `MustHaltAnalysis` (per-function): the set of blocks from which every CFG
  path halts the call. Computed as a least fixed point over DFS postorder that
  deliberately excludes cycles — a loop with a halting exit is not guaranteed
  to take it.
- `StackCleanupSafety` (global): proves retained dead slots cannot overflow
  the 1024-slot EVM stack. It composes a bound from the maximum caller stack
  height (via the FCG), per-function frame growth (one persistent slot per
  distinct SSA value plus the largest per-instruction transient), and regional
  growth from the block onward, including transitive callees. Any CFG or
  call-graph cycle makes the bound unknown and disables the elision.
- `venom_to_assembly`: in `clean_stack_from_cfg_in`, retain only dead slots
  that sit *below* every live slot, so operand depths are unchanged and every
  omitted POP is an unconditional gas saving. Retained slots are replaced with
  anonymous `_DeadStackItem`s in the stack model so a later halting phi join
  cannot mistake junk for an incoming value. The first elision on a path
  establishes a whole-region height promise that later elisions reuse.

The bounds are verified defensively: codegen measures the actual peak stack
height of every assembly fragment it emits and raises `CompilerPanic` if a
promise was underestimated — a bad bound is a compile-time failure, never a
runtime stack overflow.

### How to verify it

`tests/unit/compiler/venom/test_stack_cleanup.py` covers: direct and
transitive halts, loops with halting exits (not must-halt), internal `ret`
not counting as a message halt, recursion and unknown-entry fallbacks, the
caller-frame term in callee bounds, identity-loss before halting phi joins,
fallback near the stack limit, and that elision induces no spilling.

### Commit message

```
at CFG splitter edges, codegen pops stack items that are dead in the
successor block. when every path from the block ends the message call
(return/revert/stop/...), those POPs are wasted gas — the EVM discards
the stack at halt anyway. but skipping cleanup unconditionally is not
safe: retained slots deepen the physical stack, which could overflow the
1024-slot limit, and dead slots sitting above live values would deepen
operand access and add SWAP/DUP work.

introduce `MustHaltAnalysis`, which computes the blocks from which every
path halts the call (a least fixed point that deliberately excludes
cycles — a loop with a halting exit is not guaranteed to take it), and
`StackCleanupSafety`, a global analysis that proves retained slots
cannot overflow the stack. it bounds the worst-case physical height from
three composable terms: the maximum caller stack height (via the
function call graph), per-function frame growth (one persistent slot per
distinct SSA value plus the largest per-instruction transient), and
regional growth from the block onward, including transitive callees. any
CFG or call-graph cycle makes the bound unknown and disables the
elision.

codegen only retains dead slots that form a prefix below every live
slot, so operand depths are unchanged and every omitted POP is an
unconditional gas saving. retained slots are replaced with anonymous
dead items in the stack model so a later halting phi join cannot mistake
junk for an incoming value. the analysis bounds are verified
defensively: codegen measures the actual peak stack height of the
assembly it emits and panics if a promise was underestimated, turning a
bad bound into a compile-time failure instead of a possible runtime
stack overflow.
```

### Cute Animal Picture

### DIFF
--- a/tests/unit/compiler/venom/test_stack_cleanup.py
+++ b/tests/unit/compiler/venom/test_stack_cleanup.py
@@ -195,33 +195,30 @@ def test_stack_cleanup_elision_requires_a_known_entry_function():
     assert not safety.can_skip_cleanup(halt, 0)
 
 
-def test_stack_bound_rejects_recursive_internal_calls():
-    def candidate_is_safe(callee_body):
-        ctx = parse_venom(f"""
-            function test {{
-            entry:
-                jnz 1, @candidate, @other
-            candidate:
-                invoke @callee
-                stop
-            other:
-                stop
-            }}
+def test_stack_bound_allows_internal_calls():
+    ctx = parse_venom("""
+        function test {
+        entry:
+            jnz 1, @candidate, @other
+        candidate:
+            invoke @callee
+            stop
+        other:
+            stop
+        }
 
-            function callee {{
-            callee:
-                %retpc = param
-                {callee_body}
-                ret %retpc
-            }}
-        """)
-        fn, _, must_halt, safety = _get_stack_cleanup_analyses(ctx)
-        candidate = fn.get_basic_block("candidate")
-        assert candidate in must_halt.must_halt
-        return safety.can_skip_cleanup(candidate, 0)
+        function callee {
+        callee:
+            %retpc = param
+            nop
+            ret %retpc
+        }
+    """)
+    fn, _, must_halt, safety = _get_stack_cleanup_analyses(ctx)
+    candidate = fn.get_basic_block("candidate")
 
-    assert candidate_is_safe("nop")
-    assert not candidate_is_safe("invoke @callee")
+    assert candidate in must_halt.must_halt
+    assert safety.can_skip_cleanup(candidate, 0)
 
 
 def test_callee_stack_bound_includes_the_caller_frame():
--- a/tests/unit/compiler/venom/test_stack_cleanup.py
+++ b/tests/unit/compiler/venom/test_stack_cleanup.py
@@ -1,24 +1,22 @@
 import pytest
 
 from vyper.compiler.settings import OptimizationLevel
-from vyper.ir.compile_ir import Label
+from vyper.evm.assembler.instructions import Label
 from vyper.venom import generate_assembly_experimental
 from vyper.venom.analysis import CFGAnalysis, IRAnalysesCache, LivenessAnalysis, MustHaltAnalysis
-from vyper.venom.basicblock import IRLabel, IRVariable
+from vyper.venom.basicblock import IRLabel
 from vyper.venom.context import IRContext
 from vyper.venom.parser import parse_venom
-from vyper.venom.stack_model import StackModel
-from vyper.venom.stack_safety import StackCleanupSafety, _EVM_STACK_LIMIT
+from vyper.venom.stack_safety import StackCleanupSafety
 from vyper.venom.venom_to_assembly import VenomCompiler
 
 
 def _get_stack_cleanup_analyses(ctx, fn_name="test"):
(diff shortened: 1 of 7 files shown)


vyper/venom/stack_safety.py
```diff
@@ -0,0 +1,221 @@
...
+    def _stack_growth_summary(
+        self, bb: IRBasicBlock, active_blocks: set[IRBasicBlock], active_functions: set[IRFunction]
+    ) -> tuple[frozenset[IRVariable], int] | None:
+        if bb in self._block_summaries:
+            return self._block_summaries[bb]
+        if bb in active_blocks:
+            return None
+
+        active_blocks.add(bb)
+        try:
```
HodanPlodky, line 168: unused try?

vyper/venom/stack_safety.py
```diff
@@ -0,0 +1,221 @@
...
+            variables: set[IRVariable] = set()
+            max_transient = 0
+            for inst in bb.instructions:
+                variables.update(inst.get_input_variables())
+                variables.update(inst.get_outputs())
+                inst_transient = len(inst.operands) + 2
```
HodanPlodky, line 174: Why  `+2`?

tests/unit/compiler/venom/test_stack_cleanup.py
```diff
@@ -15,3 +40,312 @@ def test_cleanup_stack():
...
+                ret %retpc
+            }}
+        """)
+        fn, _, must_halt, safety = _get_stack_cleanup_analyses(ctx)
+        candidate = fn.get_basic_block("candidate")
+        assert candidate in must_halt.must_halt
+        return safety.can_skip_cleanup(candidate, 0)
+
+    assert candidate_is_safe("nop")
+    assert not candidate_is_safe("invoke @callee")
```
HodanPlodky: I dont think that would be valid venom since recursion should not be allowed

vyper/venom/stack_safety.py
```diff
@@ -0,0 +1,221 @@
...
+        growth = len(variables) + max_transient
+        self._function_frame_growth[fn] = growth
+        return growth
+
+    def _max_growth_from_function(
+        self, fn: IRFunction, active_functions: set[IRFunction]
+    ) -> int | None:
+        if fn in self._function_growth:
+            return self._function_growth[fn]
+        if fn in active_functions:
```
HodanPlodky: this is guarding against the recursion (right?) but some other parts of the venom expect it never appears and also the vyper should not generate it

vyper/venom/stack_safety.py
harkal: The try is used by the finally below to remove bb from active_blocks, including on early returns

vyper/venom/stack_safety.py
harkal: It covers temporary stack values added during lowering. `invoke` is the worst case because it pushes the return label and target label before `JUMP`

vyper/venom/stack_safety.py
HodanPlodky: yes but what can raise in in this `try` block? Why cannot be the `try/final` replace by just doing what is in the `try` block and then what is in the `final` block?

harkal: it's essentially removing the bb befere every return in the try block (finally runs before the return in python)

HodanPlodky (approved):

PR #5193, opened 2026-07-17
feat[lang]: forbid positional args for custom errors

### LINKED ISSUE #5133: Emit warning if custom `error`s don't use keyword arguments
```vyper
error MyError:
    arg: uint256

@external
@pure
def foo():
    raise MyError(1)
```

doesn't currently emit a compiler warning to use keyword arguments (i.e. `MyError(arg=1)`. We should follow the same pattern as for events, where we emit a warning (see https://github.com/vyperlang/vyper/pull/4257).

Sporarum:
### What I did

Fix https://github.com/vyperlang/vyper/issues/5133

Emit an `InstantiationException` when a custom `error` is instantiated with
positional arguments (e.g. `MyError(1, 2)`), pointing users to the
kwarg form (`MyError(a=1, b=2)`).

### How I did it

Check if a constructor call to an error has positional arguments, and raise a error.

### How to verify it

`pytest`

### Commit message

```
instantiating a custom error with positional arguments now emits an
`InstantiationException`, suggesting the equivalent kwargs form if
possible.

the suggestion is only emitted in the case every argument is positional,
since in that case we can reasonably assume the user meant the
parameters in declaration order. Exampe: for `MyError(1, 2)` we will
suggest `MyError(a=1,b=2)`.
```

(shouldn't appear in changelog, since it modifies a feature not present in the previous release)

### Cute Animal Picture


vyper/semantics/types/user.py
```diff
@@ -421,19 +421,17 @@ def from_ErrorDef(cls, base_node: vy_ast.ErrorDef) -> "ErrorT":
...
     def _ctor_call_return(self, node: vy_ast.Call) -> "ErrorT":
-        if len(node.keywords) > 0:
-            if len(node.args) > 0:
-                raise InstantiationException(
-                    "Error instantiation requires either all keyword arguments "
-                    "or all positional arguments",
-                    node,
-                )
+        if len(node.args) > 0:
+            correct_kwargs = list(zip(self.arguments.keys(), node.args)) + node.keywords
```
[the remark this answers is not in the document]
Sporarum: Done: https://github.com/vyperlang/vyper/pull/5193/commits/59c3629842a03b42f95aacbacfa6ef3957664c05

vyper/semantics/types/user.py
```diff
@@ -421,19 +421,19 @@ def from_ErrorDef(cls, base_node: vy_ast.ErrorDef) -> "ErrorT":
...
     def _ctor_call_return(self, node: vy_ast.Call) -> "ErrorT":
-        if len(node.keywords) > 0:
-            if len(node.args) > 0:
-                raise InstantiationException(
-                    "Error instantiation requires either all keyword arguments "
-                    "or all positional arguments",
-                    node,
-                )
+        if len(node.args) > 0:
+            positional = list(zip(self.arguments.keys(), node.args))
```
charles-cooper: zip truncates to shorter of two lists

charles-cooper (changes requested):

vyper/semantics/types/user.py
Sporarum: Hmm I don't really know how to handle the following cases:
```python
error E:
  a: uint256
  b: uint256

def foo():
  raise E(1, 2, 3) # suggests E(a=1, b=2), correct but loses the '3'
  raise E(1) # suggets E(a=1), incorrect, 'b' is not assigned
```

Sporarum: Maybe it's good enough like this ?
(Note that only the suggestion is affected, we raise no matter what `positional` is)

charles-cooper: the suggestion is very likely for the programmer (human or agent) to just copy and paste directly into the code. i think we should maybe do this check after we have done the other length matching checks so that by the time we get to this branch we know that len(args) == len(expected args)

Sporarum: The issue is, even with a perfect length match, we still have:
```python
error E:
  a: uint256
  b: uint256
  c: uint256

def foo():
  raise E(1, a=2, b=3)
```
Should we suggest `E(a=2, b=3, c=1)` or `E(a=1, a=2, b=3)` ?

The easiest seem to be:
1. if the user 1) uses only positional and 2) uses the correct number of positional parameters, provide a suggestion
2. in all other invalid cases, do not try to provide a suggestion, as there are multiple coherent ways to guess the user intention
(but the error will have the definition included, so they can see the names they need to use)

charles-cooper:
```quote
The easiest seem to be:

1. if the user 1) uses only positional and 2) uses the correct number of positional parameters, provide a suggestion
2. in all other invalid cases, do not try to provide a suggestion, as there are multiple coherent ways to guess the user intention
(but the error will have the definition included, so they can see the names they need to use)
```
yes i agree with this

charles-cooper (changes requested):
please implement per https://github.com/vyperlang/vyper/pull/5193/changes#r3902669093

vyper/semantics/types/user.py
Sporarum: Done: 61063a3f04a2d8bd098ff0472c4b609cecbe1a70

vyper/semantics/types/user.py
```diff
@@ -421,19 +421,23 @@ def from_ErrorDef(cls, base_node: vy_ast.ErrorDef) -> "ErrorT":
...
+                # Don't try to guess what the user meant
+                hint = ""
 
-            validate_kwargs(node, self.arguments, self.typeclass)
-        else:
-            validate_call_args(node, len(self.arguments))
-            for arg, expected in zip(node.args, self.arguments.values()):
-                validate_expected_type(arg, expected)
+            msg = "Instantiating errors with positional arguments is not allowed"
+            raise InstantiationException(msg, node, hint=hint)
```
charles-cooper: i would include the error definition in the exception

```suggestion
            raise InstantiationException(msg, node, self.ast_def, hint=hint)
```

charles-cooper: sorry maybe `self.decl_node`

vyper/semantics/types/user.py
```diff
@@ -421,19 +421,23 @@ def from_ErrorDef(cls, base_node: vy_ast.ErrorDef) -> "ErrorT":
...
 
     def _ctor_call_return(self, node: vy_ast.Call) -> "ErrorT":
-        if len(node.keywords) > 0:
-            if len(node.args) > 0:
-                raise InstantiationException(
-                    "Error instantiation requires either all keyword arguments "
-                    "or all positional arguments",
-                    node,
+        if len(node.args) > 0:
+            if len(self.arguments) == 0 and len(node.args) == len(node.keywords):
```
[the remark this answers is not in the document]
Sporarum: I swapped `self.arguments` and `node.keywords` by mistake ...
Fixed: 6261ede0a2a2fc61bdb50c3f5d7c47554ea397f9

vyper/semantics/types/user.py
Sporarum: Done: 1cbed8e182b69e716af9eb00e41ebf8c7a7d97b9

vyper/semantics/types/user.py
```diff
@@ -421,19 +421,23 @@ def from_ErrorDef(cls, base_node: vy_ast.ErrorDef) -> "ErrorT":
...
+                # There are only positional arguments, and there is the correct number
+                # Assume the user put them in the right order
+                correct_kwargs = ", ".join(
+                    f"{argname}={val.node_source_code}"
+                    for argname, val in zip(self.arguments.keys(), node.args)
                 )
+                hint = f"did you mean `{node.func.node_source_code}({correct_kwargs})` ?"
+            else:
+                # Don't try to guess what the user meant
+                hint = ""
```
charles-cooper, line 435: `hint = None` i believe

Sporarum: Done: f9fd33f52e3b51bb9d17c6ebf680e06e49a9e019

charles-cooper (approved):
lgtm, thanks!

PR #5194, opened 2026-07-21
fix[lang]: reject assigning to a builtin instead of panicking

### LINKED ISSUE #3933: Assigning a Builtin to Itself Leads to Panic

beardthelion:
### What I did

Assigning a builtin function to itself panics the compiler. For example:

```vyper
@external
def foo():
    convert = convert
```

crashes with `CodegenPanic: unhandled exception, parse_Name` instead of a clean error (#3933).

### How I did it

The left-hand `convert` resolves to the builtin's type (a `BuiltinFunctionT`). Because the right-hand side is the same builtin type, `validate_expected_type` sees the value matching the target and lets the assignment through. It then reaches codegen, which has no representation for a bare builtin name and panics in `parse_Name`.

A builtin is callable, not an assignable location, so `_handle_modification` now rejects an assignment whose target is a `BuiltinFunctionT`, raising a `StructureException` during semantic analysis. The error becomes a clear compile-time message rather than an internal panic. Legitimate builtin calls such as `convert(x, uint256)` are unaffected.

### How to verify it

Added the self-assignment cases to `tests/functional/syntax/names/test_variable_names.py`; they now expect `StructureException` where they previously raised `CodegenPanic`. The full `tests/functional/syntax` suite passes (1301 passed, 12 xfailed).

Closes #3933

### Commit message

```
Assigning a builtin to itself panics the compiler. The left-hand
`convert` resolves to the builtin's type, and because the right-hand
side is that same type, the assignment satisfies the typechecker. It
then reaches codegen, which has no representation for a bare builtin
name, and panics in parse_Name.

The mutation checker does not stop it either. That checker reasons about
variables, and a builtin has no VarInfo, so get_expr_info returns a bare
ExprInfo carrying the dataclass defaults: location unset, modifiability
modifiable. Every guard in _handle_modification tests for something a
builtin does not have, so all of them fall through.

Reject an assignment whose target is a BuiltinFunctionT during semantic
analysis. Calling a builtin is unaffected, since only the assignment
target is checked. One visible consequence is that assigning a literal
to a builtin, or augmenting one, now reports the new error rather than
TypeMismatch, because the guard runs ahead of the type check.

Fixes GH 3933
```

Reject assigning to a builtin (e.g. `convert = convert`) with a clear error instead of an internal compiler panic.

### Cute Animal Picture

### DIFF
--- a/tests/functional/syntax/names/test_variable_names.py
+++ b/tests/functional/syntax/names/test_variable_names.py
@@ -23,6 +23,18 @@ def foo(i: int128) -> int128:
     false : int128 = i
     return false
     """,
+    # a builtin cannot be an assignment target; assigning one to itself used to
+    # slip past the typechecker and panic in codegen (GH issue #3933).
+    """
+@external
+def foo():
+    convert = convert
+    """,
+    """
+@external
+def foo():
+    as_wei_value = as_wei_value
+    """,
 ]
 
 
--- a/vyper/semantics/analysis/local.py
+++ b/vyper/semantics/analysis/local.py
@@ -535,6 +535,15 @@ class FunctionAnalyzer(VyperNodeVisitorBase):
         func_t = self.func
         info = get_expr_info(target)
 
+        # a builtin (e.g. `convert`) is callable, not an assignable location.
+        # without this, `convert = convert` slipped past the typechecker (the
+        # right-hand side matched the target's builtin type) and panicked in
+        # codegen (GH issue #3933). lazy import avoids a builtins<->semantics cycle.
+        from vyper.builtins._signatures import BuiltinFunctionT
+
+        if isinstance(info.typ, BuiltinFunctionT):
+            raise StructureException(f"Cannot assign to '{info.typ}'", target)
+
         if isinstance(info.typ, HashMapT):
             raise StructureException(
                 "Left-hand side of assignment cannot be a HashMap without a key"
--- a/vyper/semantics/analysis/local.py
+++ b/vyper/semantics/analysis/local.py
@@ -556,6 +556,15 @@ class FunctionAnalyzer(VyperNodeVisitorBase):
         func_t = self.func
         info = get_expr_info(target)
 
+        # a builtin (e.g. `convert`) is callable, not an assignable location.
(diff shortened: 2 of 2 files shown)


Sporarum:
Why isn't this already caught by the mutation checker ?

beardthelion:
> Why isn't this already caught by the mutation checker ?

Because the mutation checker only reasons about variables, and a builtin name never becomes one. There is no `VarInfo` for `convert`, so `get_expr_info` returns a bare `ExprInfo(t)` with the dataclass defaults, and every guard in `_handle_modification` is testing something a builtin doesn't have:

```
typ           = Convert
location      = unset
modifiability = modifiable

isinstance(info.typ, HashMapT)      -> False
location in (STORAGE, TRANSIENT)    -> False
location == CALLDATA                -> False
modifiability == RUNTIME_CONSTANT   -> False
modifiability == CONSTANT           -> False
```

The typechecker doesn't stop it either, but only in the self-assign form. `convert = 1` is a clean `TypeMismatch` on master, while `convert = convert` has the same type on both sides, so it passes and reaches `parse_Name` in codegen, which has no handler for a builtin.

On master six forms panic: `convert = convert` plain, inside an `if`, inside a `for`, inside an internal function, and the same for `as_wei_value` and `empty`. With the patch all six raise `StructureException`.

One visible change beyond the panics: `convert = 1` and `convert += 1` move from `TypeMismatch` to the new message, since the guard runs before the type check. I think it reads better, but say the word if you'd rather those keep the old error. Everything else I checked is identical before and after, including the HashMap, constant, immutable, calldata and view guards, and `uint256 = 1`.

Sporarum:
Sorry I just noticed this is already fixed by #5181, I added your tests to that PR (877ecfcb961694a43dc03818fbb4f2354e731fa9), thank you !

Issue #4609, opened 2025-04-25
Builtins Variable Arguments Are Not Semantically Checked


cyberthirst:
### Version Information

* vyper Version (output of `vyper --version`): b635696f9291ea16d4e98efbf4938d65e0ee4f24

### What's your issue about?

Per @trocher:

Builtins that accept variable arguments (varargs) do not semantically check them. In case incorrect objects are passed (types, environment reserved keywords, etc.), the compiler will fail for diverse reasons only at code generation.

The following code shows different examples of this issue (`lib1` is an empty module here):

```python
import lib1

initializes: lib1

a: HashMap[uint256, address]

@external
def foo():
    # CodegenPanic: unhandled exception 'IntegerT' object has
    #              no attribute 'location', parse_Call
    a: address = raw_create(b'hello world', uint256)

    # TypeCheckFailure: assigning HashMap[uint256, address] to
    #                  HashMap[uint256, address] 160 <#internal> 0 <self.a>
    b: address = raw_create(b'hello world', self.a)

    # CompilerPanic: Method must be implemented by the inherited class
    c: Bytes[1000] = abi_encode(msg)

    # CompilerPanic: Method must be implemented by the inherited class
    d: Bytes[1000] = abi_encode(lib1)
```

PR #5195, opened 2026-07-21
fix[lang]: reject type names passed as builtin varargs

beardthelion:
### What I did

Builtins that take variable arguments never semantically checked them, so a bare type name passed as a vararg reached codegen and crashed the compiler with an internal panic. This rejects that case during semantic analysis. Part of https://github.com/vyperlang/vyper/issues/4609

### How I did it

`BuiltinFunctionT._validate_arg_types` inferred each vararg's type only for its side effects and discarded the result. A bare type name like `uint256` infers to `TYPE_T`, which is exactly inferable, so it passed the check and slipped through to codegen, where `TYPE_T` has no `abi_type`/`typ`/`location` and the compiler panicked. The vararg loop now inspects the inferred type and rejects `TYPE_T` with `InvalidReference`, the same "not a variable or literal" error the language already raises for a type name in value position (`x: uint256 = uint256`).

### How to verify it

On master both of these panic:

```vyper
@external
def foo():
    x: Bytes[32] = abi_encode(uint256)   # CompilerPanic: TYPE_T does not implement abi_type

@external
def bar():
    print(uint256)                       # CodegenPanic: 'IntegerT' object has no attribute 'typ'
```

After this change both raise `InvalidReference: not a variable or literal: 'uint256'` at compile time. Regression tests are in `tests/functional/syntax/test_abi_encode.py` and `tests/functional/syntax/test_print.py`.

### Commit message

```
Builtins with variable arguments (abi_encode, print, raw_create,
create_from_blueprint) inferred each vararg's type for its side effects
but never checked the argument was a runtime value. A bare type name
infers to TYPE_T, which passed the vararg check and slipped through to
codegen, where it panicked (`TYPE_T does not implement abi_type`,
`'IntegerT' object has no attribute 'typ'`).

Reject a TYPE_T vararg during semantic analysis with the same "not a
variable or literal" error the language already raises for a type name
in value position (e.g. `x: uint256 = uint256`).

Part of GH 4609
```

Reject a bare type name passed as a builtin vararg (e.g. `abi_encode(uint256)`) with a clear error instead of an internal compiler panic.

### Cute Animal Picture

### DIFF
--- a/tests/functional/syntax/test_abi_encode.py
+++ b/tests/functional/syntax/test_abi_encode.py
@@ -1,7 +1,7 @@
 import pytest
 
 from vyper import compiler
-from vyper.exceptions import InvalidLiteral, TypeMismatch, UnfoldableNode
+from vyper.exceptions import InvalidLiteral, InvalidReference, TypeMismatch, UnfoldableNode
 
 fail_list = [
     (
@@ -71,6 +71,14 @@ def foo() -> Bytes[132]:
     return abi_encode(x, y, method_id=b"")
         """,
         InvalidLiteral,  # len(method_id) must be 4
+    ),
+    (
+        """
+@external
+def foo() -> Bytes[32]:
+    return abi_encode(uint256)
+    """,
+        InvalidReference,  # a type name is not a value (vararg), see #4609
     ),
 ]
 
--- a/tests/functional/syntax/test_abi_encode.py
+++ b/tests/functional/syntax/test_abi_encode.py
@@ -1,7 +1,13 @@
 import pytest
 
 from vyper import compiler
-from vyper.exceptions import InvalidLiteral, InvalidOperation, TypeMismatch, UnfoldableNode
+from vyper.exceptions import (
+    InvalidLiteral,
+    InvalidOperation,
+    InvalidReference,
+    TypeMismatch,
+    UnfoldableNode,
+)
 
 fail_list = [
     (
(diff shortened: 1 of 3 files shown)


tests/functional/syntax/test_abi_encode.py
```diff
@@ -72,6 +72,14 @@ def foo() -> Bytes[132]:
         """,
         InvalidLiteral,  # len(method_id) must be 4
     ),
+    (
+        """
+@external
+def foo() -> Bytes[32]:
+    return abi_encode(uint256)
+    """,
+        InvalidReference,  # a type name is not a value (vararg), see #4609
```
Sporarum: @charles-cooper what's our policy on references to PRs/issues ?

Sporarum (approved):
Looks very good !

One small potential nitpick below

tests/functional/syntax/test_abi_encode.py
charles-cooper: url is best

Sporarum: @beardthelion please convert these references to urls

beardthelion: > please convert these references to urls

Done, both now point at https://github.com/vyperlang/vyper/issues/4609. The one here moved above the entry since the URL doesn't fit in a trailing comment under the 100 column limit. I also converted the reference in the commit message and the PR description.

Branch is rebased on master while I was in here, it was conflicting on the `InvalidOperation` import in both test files.

charles-cooper: for commit msg + PR body please follow skills/contributing.md in root of the repo

beardthelion: > for commit msg + PR body please follow skills/contributing.md in root of the repo

Done. The commit message section is the body only now, since the title becomes the subject on squash, wrapped at 72 through `fmt_commit_msg.py`, and the issue reference is `GH 4609` rather than a URL.

Made the same pass over #5194, which was missing the commit message, changelog and picture sections entirely, and #5257.

PR #5198, opened 2026-07-22
fix[lang]: fix errors when shadowing a built-in

Sporarum:
### What I did

Improve error messages when trying to shadow a built-in:

Before:
```python
int128: Bytes[3]
# NamespaceCollision: 'int128' has already been declared as a int128

def foo():
    msg: bool = True
    # NamespaceCollision: 'msg' has already been declared as a VarInfo(typ=msg, location=<DataLocation.UNSET: 'unset'>, modifiability=<Modifiability.RUNTIME_CONSTANT: 'runtime_constant'>, is_public=False, decl_node=None)
```

After:
```python
int128: Bytes[3]
# NamespaceCollision: 'int128' is already the name of a built-in

def foo():
    msg: bool = True
    # NamespaceCollision: 'msg' is already the name of a built-in
```

### How I did it

When there is a name collision, check if the previous definition has a source location, if it doesn't: it must be a built-in.

### How to verify it

`pytest`
New tests added

### Commit message

```
before this commit, the error message users got when trying to override
a built-in contained the __repr__ of the node, which was very noisy.
instead this commit uses a bespoke error message for these cases.
```

Improve error message when trying to shadow a built-in

### Cute Animal Picture


vyper/semantics/namespace.py
```diff
@@ -85,15 +85,17 @@ def clear(self):
...
 
+    # TODO: This should also include the location of the conflicting definition
     def validate_assignment(self, attr):
         validate_identifier(attr)
 
         if attr in self:
             prev = super().__getitem__(attr)
             prev_decl = getattr(prev, "decl_node", None)
-            msg = f"'{attr}' has already been declared"
             if prev_decl is None:
```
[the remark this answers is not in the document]
Sporarum: Fixed: https://github.com/vyperlang/vyper/pull/5198/commits/3cf1695923781a13ab70a502bb0739750e719d1e

vyper/semantics/analysis/base.py
```diff
@@ -95,7 +95,7 @@ class AnalysisResult:
 class ModuleInfo(AnalysisResult):
     module_t: "ModuleT"
     alias: str
-    # import_node: vy_ast._ImportStmt # maybe could be useful
+    decl_node: vy_ast.alias
```
charles-cooper, line 97: do we still need `alias: str` on the previous line then?

Sporarum: Technically no, we can instead do `module_info.decl_node.asname`, but `module_info.alias` is I think much cleaner
(especially given the above can easily be confused with `module_info.decl_node.name` which is the un-aliased name)

So I'll keep the member, but make it a property instead

Sporarum: Actually is equivalent to `module_info.decl_node.asname or module_info.decl_node.name`

Sporarum: See 1ca34807636171baab6c1695f6e4150a698cc2b0

vyper/semantics/analysis/module.py
```diff
@@ -903,21 +903,23 @@ def visit_ImportFrom(self, node):
         self._add_import(node)
 
     def _add_import(self, node: vy_ast.Import | vy_ast.ImportFrom) -> None:
-        for import_info in node._metadata["import_infos"]:
+        for alias_node, import_info in zip(node.names, node._metadata["import_infos"]):
```
charles-cooper: never use zip like this because it silently truncates to shorter of two lists

Sporarum: Done: https://github.com/vyperlang/vyper/pull/5198/commits/b9b286a3f6b5c0b3366e410c87a2416876ebe658

charles-cooper (approved):

vyper/semantics/namespace.py
```diff
@@ -85,15 +85,17 @@ def clear(self):
...
-                msg += f" as a {prev}"
+                msg = f"'{attr}' is already the name of a built-in"
```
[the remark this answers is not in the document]
Sporarum: I think the fix for this would be to populate `decl_node` for things comming from JSON ABIs, as this would also allow us to provide location for error messages involving members from json abis

PR #5202, opened 2026-07-26
fix[ux]: show readable type names in type-mismatch errors

### LINKED ISSUE #4955: The `len(<module>)` TypeMismatch leaks GenericTypeAcceptor Python class paths in error message
### Version Information

* Vyper Version: 0.4.4+commit.bdf00d31
* OS: linux
* Python Version: Python 3.11.0rc1

### What's your issue about?

Calling `len()` on an imported module correctly raises a `TypeMismatch`, but the error message exposes Vyper-internal Python class paths through `GenericTypeAcceptor(...)` reprs instead of showing human-readable Vyper type names.

The issue is with the error formatting, not with rejecting the code. `len()` should only accept `String`, `Bytes`, or `DynArray`, but the message currently reports those expected types as internal Python classes.

Command run:

```bash
vyper V7c15c5_len_module.input.vy
```

Reproducer:

```vyper
import V7c15c5_dir.lib as lib

uses: lib

@external
@view
def f() -> uint256:
    return len(lib)
```

Where `V7c15c5_dir/lib.vy` can be a minimal module, for example:

```vyper
counter: uint256
```

Observed error excerpt:

```text
vyper.exceptions.TypeMismatch: Given reference has type ...lib.vy, expected one of GenericTypeAcceptor(<class 'vyper.semantics.types.bytestrings.StringT'>), GenericTypeAcceptor(<class 'vyper.semantics.types.bytestrings.BytesT'>), GenericTypeAcceptor(<class 'vyper.semantics.types.subscriptable.DArrayT'>)
```

Expected behavior:

```text
TypeMismatch: Given reference has module type, expected one of String, Bytes, DynArray
```

or similar human-readable Vyper type names.

The current message leaks compiler implementation details such as:

```text
vyper.semantics.types.bytestrings.StringT
vyper.semantics.types.bytestrings.BytesT
vyper.semantics.types.subscriptable.DArrayT
```

This is confusing for users because the message requires knowledge of Vyper's internal Python type hierarchy to understand what the compiler expected.

Running with `--verbose` produces the same underlying issue: the expected types are displayed through `GenericTypeAcceptor(<class ...>)` instead of user-facing type names.

### How can it be fixed?

The likely fix is to make `GenericTypeAcceptor` render as a user-facing Vyper type name when included in error messages.

For example, define or adjust `__str__` / `__repr__` on `GenericTypeAcceptor` so that these:

```text
GenericTypeAcceptor(<class 'vyper.semantics.types.bytestrings.StringT'>)
GenericTypeAcceptor(<class 'vyper.semantics.types.bytestrings.BytesT'>)
GenericTypeAcceptor(<class 'vyper.semantics.types.subscriptable.DArrayT'>)
```

are rendered as something like:

```text
String
Bytes
DynArray
```

The `len()` builtin already knows the accepted type classes, so the error path should format those acceptors using their Vyper-facing type identifiers instead of Python `repr()` output.

cristianizzo:
Fixes #4955.

### What I did

`_GenericTypeAcceptor` (used to represent the set of types a builtin accepts) rendered as its internal Python class repr in error messages. For example, `len()` on a wrong type or a module produced:

```
expected one of GenericTypeAcceptor(<class 'vyper.semantics.types.bytestrings.StringT'>), GenericTypeAcceptor(<class 'vyper.semantics.types.bytestrings.BytesT'>), GenericTypeAcceptor(<class 'vyper.semantics.types.subscriptable.DArrayT'>)
```

Now it produces:

```
expected one of String, Bytes, DynArray
```

### How

The type-mismatch message renders acceptors via `str(...)` (`analysis/utils.py`), which fell back to `__repr__`. I added a `__str__` that resolves the user-facing type name — `_id` when it's a plain class string (`String`/`Bytes`/`DynArray`), falling back to `typeclass` for parametric types whose `_id` is a property (e.g. `IntegerT` → `integer`), then the class name — so it is always a string. `__repr__` is left as the unambiguous debug form (which also keeps it total, so e.g. `TYPE_T.any()` doesn't raise).

### Tests

Added two regression tests to `tests/functional/syntax/test_len.py`: one asserting a `len()` type-mismatch message shows `String`/`Bytes`/`DynArray` (and no `GenericTypeAcceptor`), and one for the `typeclass` fallback branch (indexing with the wrong type → `integer`). Both fail on the current code and pass with this change; `black`/`isort`/`flake8` are clean.

### DIFF
--- a/tests/functional/syntax/test_concat.py
+++ b/tests/functional/syntax/test_concat.py
@@ -117,6 +117,24 @@ def test_block_fail(assert_compile_failed, get_contract, bad_code, exc):
     assert_compile_failed(lambda: get_contract(bad_code), exc)
 
 
+def test_concat_type_mismatch_message_uses_readable_type_names():
+    # `concat`'s accepted `bytesM` type must be shown by its user-facing name,
+    # not the internal `bytes_m` typeclass (nor a `GenericTypeAcceptor` repr).
+    # See issue #4955.
+    code = """
+@external
+def foo() -> Bytes[64]:
+    return concat(123, b"x")
+    """
+    with pytest.raises(TypeMismatch) as exc_info:
+        compiler.compile_code(code)
+
+    message = str(exc_info.value)
+    assert "GenericTypeAcceptor" not in message
+    assert "bytes_m" not in message
+    assert "bytesM" in message
+
+
 valid_list = [
     """
 @external
--- a/tests/functional/syntax/test_concat.py
+++ b/tests/functional/syntax/test_concat.py
@@ -118,9 +118,6 @@ def test_block_fail(assert_compile_failed, get_contract, bad_code, exc):
 
 
 def test_concat_type_mismatch_message_uses_readable_type_names():
-    # `concat`'s accepted `bytesM` type must be shown by its user-facing name,
-    # not the internal `bytes_m` typeclass (nor a `GenericTypeAcceptor` repr).
-    # See issue #4955.
     code = """
 @external
 def foo() -> Bytes[64]:
--- a/tests/functional/syntax/test_concat.py
+++ b/tests/functional/syntax/test_concat.py
@@ -126,13 +126,12 @@ def test_concat_type_mismatch_message_uses_readable_type_names():
 def foo() -> Bytes[64]:
     return concat(123, b"x")
     """
-    with pytest.raises(TypeMismatch) as exc_info:
+    with pytest.raises(TypeMismatch) as e:
         compiler.compile_code(code)
 
-    message = str(exc_info.value)
-    assert "GenericTypeAcceptor" not in message
-    assert "bytes_m" not in message
-    assert "bytesM" in message
+    assert e.value.message == (
+        "Expected one of Bytes, String, bytesM but literal can only be cast as int104 or uint96."
+    )
 
 
 valid_list = [
--- a/vyper/semantics/types/base.py
+++ b/vyper/semantics/types/base.py
@@ -17,10 +17,6 @@ from vyper.exceptions import (
 from vyper.semantics.analysis.levenshtein_utils import get_levenshtein_error_suggestions
 from vyper.semantics.data_locations import DataLocation
 
-# user-facing spelling for internal `typeclass` identifiers that differ from
-# the syntax users write (e.g. the `bytesM` family: `bytes1`..`bytes32`)
-_TYPECLASS_DISPLAY_NAMES = {"bytes_m": "bytesM"}
-
 
 # Some fake type with an overridden `compare_type` which accepts any RHS
 # type of type `type_`
@@ -30,17 +26,15 @@ class _GenericTypeAcceptor:
 
     def __str__(self):
         # User-facing type name (e.g. `String`, `Bytes`, `DynArray`) used in
-        # error messages, instead of the internal Python class repr. `_id` is a
-        # plain class attribute on most types, but a property on parametric
-        # types (e.g. `IntegerT`, `BytesM_T`), so fall back to `typeclass`
-        # (mapped to its user-facing spelling), then the class name, keeping
-        # this total (never returns a non-string).
+        # error messages, instead of the internal Python class repr. Prefer the
+        # class-level `_id`; parametric types (e.g. `IntegerT`, `BytesM_T`)
+        # define `_id` as a property, so it has no value at the class level that
+        # `.any()` hands us -- fall back to their `_generic_id`.
         name = getattr(self.type_, "_id", None)
         if not isinstance(name, str):
-            typeclass = getattr(self.type_, "typeclass", None)
-            name = _TYPECLASS_DISPLAY_NAMES.get(typeclass, typeclass)
+            name = self.type_._generic_id
         if not isinstance(name, str):
-            name = self.type_.__name__
+            raise CompilerPanic(f"{self.type_.__name__} has no user-facing name")
         return name
 
     def __init__(self, type_):
@@ -102,6 +96,10 @@ class VyperType:
     typeclass: str = None  # type: ignore
 
     _id: str  # rename to `_name`
+    # user-facing name for parametric types whose `_id` is an instance
+    # property (e.g. `bytesM`, `integer`), so it has no value at the class
+    # level. see `_GenericTypeAcceptor.__str__`.
+    _generic_id: str = None  # type: ignore
     _type_members: Optional[Dict] = None
     _valid_literal: Tuple = ()
     _invalid_locations: Tuple = ()
@@ -517,6 +515,8 @@ def map_void(typ: Optional[VyperType]) -> VyperType:
 # position, ex. constructors (events, interfaces and structs), and also
 # certain builtins which take types as parameters
 class TYPE_T(VyperType):
+    _generic_id = "type"
+
     def __init__(self, typedef):
         super().__init__()
 
--- a/vyper/semantics/types/base.py
+++ b/vyper/semantics/types/base.py
@@ -17,12 +17,31 @@ from vyper.exceptions import (
 from vyper.semantics.analysis.levenshtein_utils import get_levenshtein_error_suggestions
 from vyper.semantics.data_locations import DataLocation
 
+# user-facing spelling for internal `typeclass` identifiers that differ from
+# the syntax users write (e.g. the `bytesM` family: `bytes1`..`bytes32`)
+_TYPECLASS_DISPLAY_NAMES = {"bytes_m": "bytesM"}
+
 
 # Some fake type with an overridden `compare_type` which accepts any RHS
 # type of type `type_`
 class _GenericTypeAcceptor:
     def __repr__(self):
         return f"GenericTypeAcceptor({self.type_})"
+
+    def __str__(self):
+        # User-facing type name (e.g. `String`, `Bytes`, `DynArray`) used in
+        # error messages, instead of the internal Python class repr. `_id` is a
+        # plain class attribute on most types, but a property on parametric
+        # types (e.g. `IntegerT`, `BytesM_T`), so fall back to `typeclass`
+        # (mapped to its user-facing spelling), then the class name, keeping
+        # this total (never returns a non-string).
+        name = getattr(self.type_, "_id", None)
+        if not isinstance(name, str):
+            typeclass = getattr(self.type_, "typeclass", None)
+            name = _TYPECLASS_DISPLAY_NAMES.get(typeclass, typeclass)
+        if not isinstance(name, str):
+            name = self.type_.__name__
+        return name
 
     def __init__(self, type_):
         self.type_ = type_
--- a/vyper/semantics/types/base.py
+++ b/vyper/semantics/types/base.py
@@ -25,11 +25,6 @@ class _GenericTypeAcceptor:
         return f"GenericTypeAcceptor({self.type_})"
 
     def __str__(self):
-        # User-facing type name (e.g. `String`, `Bytes`, `DynArray`) used in
-        # error messages, instead of the internal Python class repr. Prefer the
-        # class-level `_id`; parametric types (e.g. `IntegerT`, `BytesM_T`)
-        # define `_id` as a property, so it has no value at the class level that
-        # `.any()` hands us -- fall back to their `_generic_id`.
         name = getattr(self.type_, "_id", None)
         if not isinstance(name, str):
             name = self.type_._generic_id
@@ -96,10 +91,10 @@ class VyperType:
     typeclass: str = None  # type: ignore
 
     _id: str  # rename to `_name`
-    # user-facing name for parametric types whose `_id` is an instance
-    # property (e.g. `bytesM`, `integer`), so it has no value at the class
-    # level. see `_GenericTypeAcceptor.__str__`.
-    _generic_id: str = None  # type: ignore
+    # user-facing name for types whose `_id` is the fully applied type
+    # (e.g. `uint256`, `bytes4`) rather than the type constructor, so there
+    # is no generic name to read off the class. see `_GenericTypeAcceptor`.
+    _generic_id: Optional[str] = None
     _type_members: Optional[Dict] = None
     _valid_literal: Tuple = ()
     _invalid_locations: Tuple = ()
@@ -515,8 +510,6 @@ def map_void(typ: Optional[VyperType]) -> VyperType:
 # position, ex. constructors (events, interfaces and structs), and also
 # certain builtins which take types as parameters
 class TYPE_T(VyperType):
-    _generic_id = "type"
-
     def __init__(self, typedef):
         super().__init__()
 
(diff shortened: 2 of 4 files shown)


vyper/semantics/types/base.py
```diff
@@ -24,6 +24,19 @@ class _GenericTypeAcceptor:
...
+        # error messages, instead of the internal Python class repr. `_id` is a
+        # plain class attribute on most types, but a property on parametric
+        # types (e.g. `IntegerT`), so fall back to `typeclass`, then the class
+        # name, keeping this total (never returns a non-string).
+        name = getattr(self.type_, "_id", None)
+        if not isinstance(name, str):
+            name = getattr(self.type_, "typeclass", None)
+        if not isinstance(name, str):
+            name = self.type_.__name__
+        return name
```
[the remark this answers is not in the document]
cristianizzo: Good catch, the raw `typeclass` fallback leaks internal names. I added a small `_TYPECLASS_DISPLAY_NAMES` map (`bytes_m` → `bytesM`) so `concat(...)` and friends now show `bytesM` instead of `bytes_m`, and added a regression test in `test_concat.py`. (`SArrayT.any()` is never used in an error path and `IntegerT`/`DecimalT` already render as `integer`/`decimal`, so `bytesM` was the only leaking case.)

vyper/semantics/types/base.py
```diff
@@ -17,13 +17,32 @@
...
+        # types (e.g. `IntegerT`, `BytesM_T`), so fall back to `typeclass`
+        # (mapped to its user-facing spelling), then the class name, keeping
+        # this total (never returns a non-string).
+        name = getattr(self.type_, "_id", None)
+        if not isinstance(name, str):
+            typeclass = getattr(self.type_, "typeclass", None)
+            name = _TYPECLASS_DISPLAY_NAMES.get(typeclass, typeclass)
+        if not isinstance(name, str):
+            name = self.type_.__name__
+        return name
```
Sporarum, line 33: I'm not sure the `typeclass` is a good default name (somewhat highlighted by the need for `_TYPECLASS_DISPLAY_NAMES`), as far as I understand it's the name used when linearizing to json, and thus needs to be unique for us, but not a good user-facing name.

And the user-facing name might evolve differently than the `typeclass`, for backward compat reasons

I'm not sure what the "correct" solution is then, maybe have `_id` also work on `String` etc ?

Sporarum (commented):
Thank you for your work!

A couple points:

Given the tests are explicitly about error messages, I would prefer the check was of the form `assert message ==` to have the full expected content on display, you can look how we did it in some other tests
(notably I don't think we do `str(exc_info.value)`)

cristianizzo:
Yeah, fair point. `typeclass` is really the serialization id (it's what we dump in `to_dict`), so using it for display and then patching it with `_TYPECLASS_DISPLAY_NAMES` is fixing the wrong layer, and you're right that it could drift from the user-facing name for backwards-compat reasons.

The reason it breaks is just that `_id` is a normal class attr on most types (`String`, `Bytes`, `DynArray`...) but a property on `IntegerT` and `BytesM_T`, so there's no value at the class level, which is what `.any()` hands us.

So instead of touching `typeclass`, I'd give those two an explicit class-level name:

```python
class BytesM_T(_PrimT):
    _generic_id = "bytesM"

class IntegerT(...):
    _generic_id = "integer"
```

and make `__str__` go `_generic_id` -> class-level `_id` -> `__name__`. That drops both the `typeclass` fallback and the map.

Or if you'd rather make `_id` itself resolve at the class level for every type, I can do it that way instead. Lmk which you prefer and I'll push it.

Sporarum:
> make `__str__` go `_generic_id` -> class-level `_id` -> `__name__`. That drops both the `typeclass` fallback and the map.

Not sure I understand, but I think we should use `_id` if available and `_generic_id` otherwise (and never need `__name__`!)

> Or if you'd rather make `_id` itself resolve at the class level for every type

I'm not sure what that would entail, if it can be done cleanly, it conceptually looks like the cleaner option

cristianizzo:
Yeah, `_id` first then `_generic_id` makes more sense, I'll flip it.

On dropping `__name__`: mostly agree, one gotcha though — `TYPE_T` has no `_id` at all (only reachable via `TYPE_T.any()` in a couple of builtins), so it'd still hit `__name__`. If I give it a `_generic_id` too (like `"type"`), then we can drop `__name__` entirely like you want — everything reachable via `.any()` then has either a real `_id` (String/Bytes/DynArray) or a `_generic_id` (bytesM/integer/type).

On folding it into `_id`: I looked at it and it's not really cleaner in practice. The parametric `_id`s are properties that need instance state (`self.m`, `self.bits`), so to make `Cls._id` return a generic name while `instance._id` stays specific you'd need a custom class/instance descriptor on each of them — more magic than a one-line `_generic_id` attr imo. So I'd lean `_generic_id`.

I'll push that (flip the order, add `_generic_id` to `BytesM_T`/`IntegerT`/`TYPE_T`, drop the map + `__name__`) unless you object.

vyper/semantics/types/base.py
cristianizzo: Agreed — I dropped `typeclass` from the display path entirely (it is the json-linearization id and shouldn't double as a user-facing name), along with the `_TYPECLASS_DISPLAY_NAMES` map and the `__name__` fallback.

Instead, `__str__` now resolves the class-level `_id` first and falls back to a new explicit `_generic_id`:

```python
name = getattr(self.type_, "_id", None)
if not isinstance(name, str):
    name = self.type_._generic_id
if not isinstance(name, str):
    raise CompilerPanic(f"{self.type_.__name__} has no user-facing name")
```

`_generic_id: str = None  # type: ignore` is declared on `VyperType` (mirroring the adjacent `typeclass` declaration) and set on exactly the three types that `.any()` can hand us without a class-level `_id`: `BytesM_T` (`_id` is a `property`) → `"bytesM"`, `IntegerT` (`_id` is a `cached_property`) → `"integer"`, and `TYPE_T` (no `_id` at all) → `"type"`. I checked every type reachable via `.any()` and they all now resolve to a readable name, so the `CompilerPanic` is unreachable but keeps a future parametric type from silently leaking a class repr instead of quietly degrading.

I also switched the tests to full-message equality as you asked (`assert e.value.message == ...`, no `str(exc_info.value)` — that appends the source location), e.g.:

```python
assert e.value.message == "Given reference has type int128, expected one of String, Bytes, DynArray"
```

tests/functional/syntax/test_concat.py
```diff
@@ -117,6 +117,23 @@ def test_block_fail(assert_compile_failed, get_contract, bad_code, exc):
     assert_compile_failed(lambda: get_contract(bad_code), exc)
 
 
+def test_concat_type_mismatch_message_uses_readable_type_names():
+    # `concat`'s accepted `bytesM` type must be shown by its user-facing name,
+    # not the internal `bytes_m` typeclass (nor a `GenericTypeAcceptor` repr).
+    # See issue #4955.
```
Sporarum: Nit:
I think the added tests are small enough they don't need an explanation

vyper/semantics/types/base.py
```diff
@@ -498,6 +515,8 @@ def map_void(typ: Optional[VyperType]) -> VyperType:
 # position, ex. constructors (events, interfaces and structs), and also
 # certain builtins which take types as parameters
 class TYPE_T(VyperType):
+    _generic_id = "type"
```
Sporarum: Hmm this doesn't seem great, can you add a test so we can see what the error looks like ?

I think TYPE_T should instead use the name of the underlying object (event, interface, etc)

vyper/semantics/types/primitives.py
```diff
@@ -266,6 +267,7 @@ class IntegerT(NumericT):
     """
 
     typeclass = "integer"
+    _generic_id = "integer"
```
Sporarum, line 270: I'm not sure this is reachable, since the .any() does not go through a GenericTypeAcceptator
Please look into it

vyper/semantics/types/base.py
```diff
@@ -83,6 +96,10 @@ class VyperType:
     typeclass: str = None  # type: ignore
 
     _id: str  # rename to `_name`
+    # user-facing name for parametric types whose `_id` is an instance
+    # property (e.g. `bytesM`, `integer`), so it has no value at the class
+    # level. see `_GenericTypeAcceptor.__str__`.
+    _generic_id: str = None  # type: ignore
```
Sporarum: Nit:
If it doesn't have the right type, I'm sure I'll get confused at some point, let's instead:
```suggestion
    _generic_id: Optional[str] = None
```

vyper/semantics/types/base.py
```diff
@@ -24,6 +24,19 @@ class _GenericTypeAcceptor:
     def __repr__(self):
         return f"GenericTypeAcceptor({self.type_})"
 
+    def __str__(self):
+        # User-facing type name (e.g. `String`, `Bytes`, `DynArray`) used in
+        # error messages, instead of the internal Python class repr. Prefer the
+        # class-level `_id`; parametric types (e.g. `IntegerT`, `BytesM_T`)
+        # define `_id` as a property, so it has no value at the class level that
+        # `.any()` hands us -- fall back to their `_generic_id`.
```
Sporarum: I don't think this needs a comment:
`__str__` is for user-facing stuff
The rest just paraphrases the body

Also it seems incorrect:
IntegerT and BytesM_T are not parametric (`uint256`, `bytes4`)
(well as python values they are, but then so are `String` et al)

vyper/semantics/types/base.py
Sporarum: Why does `_id` on `String` (et al) work ?
Since if anything `String` is more parametrized than `IntegerT`

cristianizzo: Good question, and the answer showed my comment was wrong (fixed below).

It is not about parametricity. It is about *what `_id` names*: the type constructor, or the fully applied type.

```
StringT   _id = "String"     (class attr)      __repr__ -> "String[5]"
BytesT    _id = "Bytes"      (class attr)      __repr__ -> "Bytes[5]"
DArrayT   _id = "DynArray"   (class attr)      __repr__ -> "DynArray[uint256, 3]"
IntegerT  _id = cached_property -> "uint256"
BytesM_T  _id = property        -> "bytes4"
```

So `String` is parameterized, as you say, but its `_id` holds the *constructor* name and the parameters are appended in `__repr__`. `IntegerT`/`BytesM_T` instead bake the parameters into `_id` itself, so there is no constructor-level name to read off the class — which is exactly what `.any()` needs. That is the gap `_generic_id` fills, and I have reworded the declaration comment accordingly.

vyper/semantics/types/base.py
cristianizzo: Right on both counts — comment removed, `__str__` is now just the three lines of body.

And thanks for catching the inaccuracy: "parametric" was the wrong axis (`String` is parameterized too). The actual distinction is constructor name vs fully applied name in `_id`; I have written that up in your other thread and corrected the remaining comment on the `_generic_id` declaration.

vyper/semantics/types/base.py
cristianizzo: Done, applied as suggested: `_generic_id: Optional[str] = None`, and the `# type: ignore` is gone with it. `mypy -p vyper` is clean.

tests/functional/syntax/test_concat.py
cristianizzo: Removed from both tests. The one comment I kept is a single line on the index test, noting it is the `_generic_id` path rather than the `_id` path, since that is not obvious from reading the contract snippet.

vyper/semantics/types/primitives.py
cristianizzo: I looked into it — `IntegerT` *is* reachable, so I kept it. `.any()` is `_GenericTypeAcceptor(cls)` (base.py), and `IntegerT.any()` has five call sites.

To be sure the string actually comes from this path rather than somewhere else, I traced `_GenericTypeAcceptor.__str__` and recorded every render:

```
x[b"ab"]           TypeMismatch: Expected integer but literal can only be cast as Bytes[2].
                   rendered via __str__: [(IntegerT, "integer")]
shift(1, b"ab")    TypeMismatch: Expected integer but literal can only be cast as Bytes[2].
                   rendered via __str__: [(IntegerT, "integer")]
```

Two independent paths (`validate_expected_type(node, IntegerT.any())` in `subscriptable.py` for index validation, and the `shift` builtin). Without `_generic_id` these hit the `CompilerPanic`. The index case is covered by `test_index_type_mismatch_message_uses_readable_type_names`.

Note that `IntegerT.any()` inside a *tuple* of expected types behaves differently — `uint2str(b"ab")` expands to `Expected one of uint8, uint16, ...`, which never reaches `__str__`. So only the standalone uses matter here.

vyper/semantics/types/base.py
cristianizzo: You were right that this was not great — I removed `TYPE_T._generic_id` entirely.

I could not produce an error that displays it. I tried every `TYPE_T.any()` call site with a variety of bad arguments:

```
empty(1)                            InvalidType: '1' is not a type!
empty(self.bar)                     InvalidType: 'self.bar' is not a type!
empty("uint256")                    InvalidType: '"uint256"' is not a type!
epsilon(b"ab")                      InvalidType: 'b"ab"' is not a type!
abi_decode(b, b"ab")                InvalidType: 'b"ab"' is not a type!
method_id("f()", output_type=b"ab") InvalidType: 'b"ab"' is not a type!
extract32(b, 0, output_type=b"ab")  InvalidType: 'b"ab"' is not a type!
```

In all of them the argument is resolved as a type *before* the expected-type name is ever formatted, so `__str__` is never called (I confirmed this with the same tracing as in the `IntegerT` thread — zero renders). The only other use, `TYPE_T.any().compare_type(...)` in `_signatures.py`, does not stringify either. So there is nothing to write a test against; the name was dead code, and the `CompilerPanic` now documents that invariant.

On using the underlying object's name (event, interface, ...): that is not available here. `.any()` is a classmethod and hands `_GenericTypeAcceptor` the *class*, not a `TYPE_T` instance, so there is no `typedef` to read a name from at that point. If a code path ever does need to display it, the natural fix would be an instance-level name on `TYPE_T` rather than a class-level `_generic_id` — but today nothing needs it.

vyper/semantics/types/base.py
Sporarum: Okay thanks, then I think we should instead do the same, and have `_id` not have parameters baked-in, and the repr outputing the full name

cristianizzo:
@charles-cooper — on your suggestion to align `IntegerT`/`BytesM_T` with the bytestring types (`_id` as the family name, `__repr__` emitting the applied name), I've implemented it as a follow-up: **#5238**.

One thing I found worth flagging: the applied name isn't only used for display — `uint256`/`bytes4` are also the keys in the primitive-type namespace, the `name` in `to_dict()` AST output, and the tokens used to synthesize public-variable getter args. So the realign keeps those working through a small `VyperType.serialization_name` property (defaults to `_id`; the two parametric primitives override it), which leaves AST output and the namespace byte-identical.

So my plan: **#5238** carries the realign (and removes `_generic_id` entirely), while **#5202** stays the minimal user-facing fix for #4955. If that split works for you I'd suggest merging #5202 first — I'll rebase #5238 on top so its diff is just the realign.

Please let me know if you'd rather have the realign folded into #5202 instead.

Sporarum:
@cristianizzo What are you talking about, @charles-cooper never commented on this PR ?

Also please don't open a different PR to fix an issue with this one !

PR #5204, opened 2026-07-29
refactor[lang]: streamline type inference

Sporarum:
### What I did

`_ExprAnalyser` used to do three different things:
1. compute the possible types for an expression (its `get_possible_types_from_node`)
2. check if there is only one possible type  (its `get_exact_type_from_node`)
3. compute the expr info for an expression (its `get_expr_info`)

This made it hard to reason about what were its duties and guarantees.

`get_common_types` used to get the types of the nodes, combine them, and filter the result.

This commit splits these responsibilities:
1.  `_TypeSynthesizer` (a `VyperNodeVisitorBase`) computes the possible types
2. `get_exact_type_from_node` checks there is only one possible type
3. `_get_expr_info_helper` computes the expr info, and `get_expr_info` memoizes it
4. `supremum` combines types together
4. `get_common_types` filters the result of getting the supremum of the possible types for each node

Along the way, the different caches have been simplified.
For example there used to be two distinct `get_possible_types_from_node` caches depending on `include_type_exprs`

The `include_type_exprs` was set inconsistently: defaulted to False in `_ExprAnalyzer.get_exact_type_from_node`, but force-set to True in `get_exact_type_from_node`. And some methods used different names for it.
It is now named `allow_type_exprs` across the board.

### How I did it

In addition to the thing above, I updated `VyperNodeVisitorBase` to be able to return a value (since `_TypeSynthesizer` returns one)
There might be other classes with mro-based dispatch logic which could be refactored, this is out of scope of this PR

### How to verify it

Still passes all the tests, a couple new ones were added

### Commit message

```
`_ExprAnalyzer` used to do three different things:

  1. compute the possible types for an expression
     (`get_possible_types_from_node`)
  2. check that there is only one possible type
     (`get_exact_type_from_node`)
  3. compute the expr info for an expression (`get_expr_info`)

this made its duties and guarantees hard to reason about.
`get_common_types` was similarly overloaded: it fetched the types of the
nodes, combined them, and filtered the result.

this commit splits these responsibilities into focused pieces:

  - `_TypeSynthesizer` (a `VyperNodeVisitorBase`) computes the possible
    types
  - `get_exact_type_from_node` checks there is only one possible type
  - `_get_expr_info_helper` computes the expr info, and `get_expr_info`
    memoizes it
  - `supremum` combines types together
  - `get_common_types` filters the result of taking the supremum of the
    possible types of each node

this commit additionally simplifies the corresponding caches. it also
uniformizes the different "should we allow type exprs" parameters under
a single name: `allow_type_exprs`. finally, it adds the ability for
`VyperNodeVisitorBase`'s `visit` to return a value.
```

(refactor, no user-facing changelog)

### Cute Animal Picture

### DIFF
--- a/vyper/semantics/analysis/common.py
+++ b/vyper/semantics/analysis/common.py
@@ -1,16 +1,14 @@
-from typing import Tuple
+from typing import Generic, TypeVar
 
 from vyper.exceptions import StructureException, tag_exceptions
 
+Result = TypeVar("Result")
 
-class VyperNodeVisitorBase:
-    ignored_types: Tuple = ()
+
+class VyperNodeVisitorBase(Generic[Result]):
     scope_name = ""
 
-    def visit(self, node, *args):
-        if isinstance(node, self.ignored_types):
-            return
-
+    def visit(self, node, *args) -> Result:
         # iterate over the MRO until we find a matching visitor function
         # this lets us use a single function to broadly target several
         # node types with a shared parent
--- a/vyper/semantics/analysis/common.py
+++ b/vyper/semantics/analysis/common.py
@@ -8,7 +8,7 @@ Result = TypeVar("Result")
 class VyperNodeVisitorBase(Generic[Result]):
     scope_name = ""
 
-    def visit(self, node, *args) -> Result:
+    def visit(self, node, *args, **kwargs) -> Result:
         # iterate over the MRO until we find a matching visitor function
         # this lets us use a single function to broadly target several
         # node types with a shared parent
@@ -18,7 +18,7 @@ class VyperNodeVisitorBase(Generic[Result]):
             with tag_exceptions(node):
                 visitor_fn = getattr(self, f"visit_{ast_type}", None)
                 if visitor_fn:
-                    return visitor_fn(node, *args)
+                    return visitor_fn(node, *args, **kwargs)
 
         node_type = type(node).__name__
         raise StructureException(
--- a/vyper/semantics/analysis/utils.py
+++ b/vyper/semantics/analysis/utils.py
@@ -22,6 +22,7 @@ from vyper.exceptions import (
 )
 from vyper.semantics import types
 from vyper.semantics.analysis.base import ExprInfo, Modifiability, ModuleInfo, VarAccess, VarInfo
+from vyper.semantics.analysis.common import VyperNodeVisitorBase
 from vyper.semantics.analysis.levenshtein_utils import get_levenshtein_error_suggestions
 from vyper.semantics.namespace import get_namespace
 from vyper.semantics.types.base import TYPE_T, VyperType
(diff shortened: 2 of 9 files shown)


vyper/semantics/analysis/utils.py
```diff
@@ -62,142 +63,57 @@ def uses_state(var_accesses: Iterable[VarAccess]) -> bool:
...
+            # we don't pass `allow_type_exprs` to the `visit_...` nodes, as they shouldn't handle it
+            possible_types = super().visit(node)
 
-            if all(isinstance(i, IntegerT) for i in ret):
-                # for numeric types, sort according by number of bits descending
+            if all(isinstance(t, IntegerT) for t in possible_types):
+                # for numeric types, sort according to number of bits descending
                 # this ensures literals are cast with the largest possible type
-                ret.sort(key=lambda k: (k.bits, not k.is_signed), reverse=True)
+                def sorting_function(t: VyperType):
```
charles-cooper: name length should roughly be negatively correlated with scope (so for this, which is just used in one place 3 lines later, use a short name like `k` or `fn`)

vyper/semantics/analysis/utils.py
```diff
@@ -499,19 +403,103 @@ def get_exact_type_from_node(node):
...
+
+    # in case we cached a value when `allow_type_exprs` was True
+    if not allow_type_exprs:
+        _raise_if_type(node._expr_info.typ, node)
+
     return node._expr_info
 
 
-def get_common_types(*nodes: vy_ast.VyperNode, filter_fn: Callable = None) -> List:
+def supremum(*branches: List[VyperType]) -> List[VyperType]:
```
charles-cooper: `branches` is a bit confusing of a name, is it a reference to something? i would use `*possible_types` or just `*types`

charles-cooper: i would also clarify which supremum, i recommend `type_sup` for the function name

vyper/semantics/types/base.py
```diff
@@ -543,5 +543,5 @@ def get_member(self, key, node):
         raise UnknownAttribute("Value is not attributable", node)
 
 
-def is_type_t(x: VyperType, t: type) -> bool:
+def is_type_t(x: VyperType, t: type) -> TypeGuard["TYPE_T"]:
```
charles-cooper: i don't think the quotes are necessary bc of the `from __future__ import annotations`

vyper/semantics/analysis/common.py
```diff
@@ -1,16 +1,14 @@
-from typing import Tuple
+from typing import Generic, TypeVar
 
 from vyper.exceptions import StructureException, tag_exceptions
 
+Result = TypeVar("Result")
 
-class VyperNodeVisitorBase:
-    ignored_types: Tuple = ()
```
charles-cooper, line 7: what happened to `ignored_types`?

charles-cooper: oh i see, it's causing most of the control flow to be missed / redundant with `if visitor_fn` clause

charles-cooper (approved):
lgtm once naming changes are addressed

vyper/semantics/analysis/common.py
Sporarum: 1. It was only used a single time
2. It was incompatible with the new "returns a T" model, since we don't know what to put in that case
3. It can easily be replaced by a `visit_<node type that was ignored>`

So I removed it

vyper/semantics/analysis/utils.py
Sporarum: The issue is that it is a list of list of (possible) types, so I would rather not call it just `types`
Maybe `typess` could work but I find it ugly
I went with `multitypes` the idea that a multitype is the list of types of a node
(anyways this should not stay long, as I intend to make it so any node can have a single type)

charles-cooper: but the annotation says it is a list of types, not a list of lists

Sporarum: Yes but it's a vararg !

Sporarum:
Merging with master leads to a failing test, will be fixed/fixable with #5234
Therefore, waiting for that to be merged

PR #5208, opened 2026-08-05
fix[venom]: ctor return empty deploy fixes #5048

### LINKED ISSUE #5048: Venom: constructor `return` can deploy empty runtime bytecode

harkal:
### What I did

Fixes two bugs, one exposed by the other.

1. A contract compiled through the venom pipeline whose `__init__` contains a reachable `return` deploys with **empty runtime code**. The `return` lowered to a `stop`, halting the deploy code before the epilogue that returns the runtime bytecode.

2. `SimplifyCFGPass` could leave behind a `jnz` whose two arms point at the same block, which then crashes `BranchOptimizationPass`.

Both are venom-pipeline only; legacy codegen is unaffected.

```vyper
x: public(uint256)

@deploy
def __init__():
    self.x = 1
    return   # <- account deployed with no code
```

### How I did it

The constructor now gets a dedicated exit block which holds the nonreentrant unlock and the deploy epilogue. `return` inside `__init__` is dispatched to a new `_lower_ctor_return` (a third case alongside internal/external return, keyed on `ctx.ctor_exit_label`) which jumps to that block instead of halting. This makes the deploy epilogue the constructor's only terminator by construction, rather than only when the body happens to fall through — mirroring legacy codegen, where the `deploy` fragment positionally follows the ctor body.

That change made a previously unreachable CFG shape reachable: `if cond: return` in a constructor produces two arms that both end up jumping to the ctor exit. When `SimplifyCFGPass` bypasses the now-empty arm blocks, it rewrites the `jnz` operands to the same label, and `BranchOptimizationPass` — which unpacks `cfg_out(bb)` into exactly two successors — raises on it. The `jnz` is now canonicalized to a `jmp` in the same rewrite that creates the degeneracy.

### Commit message

```
in the venom pipeline, `return` inside `@deploy def __init__()` was
lowered through the same path as an external function return, which for
a void return emits `stop`. halting the deploy code returns zero bytes
of runtime code, so any constructor with a reachable `return` deployed
an account with empty code. legacy codegen cannot hit this: there the
`deploy` fragment positionally follows the constructor body, so exiting
the body falls through into it.

give the constructor a single exit block holding the nonreentrant unlock
and the deploy epilogue, and lower `return` in `__init__` as a jump to
it. the epilogue is then the constructor's only terminator by
construction, instead of only when the body happens to end in
fallthrough.

this exposed a latent bug in `SimplifyCFGPass`. when it bypasses an
empty block by rewriting a predecessor's jump target, both arms of a
`jnz` can end up pointing at the same block - an `if` whose body is a
bare `return` now leaves both arms converging on the ctor exit.
`BranchOptimizationPass` unpacks `cfg_out(bb)` into exactly two
successors and blew up on such a branch. canonicalize the degenerate
`jnz` to a `jmp` as part of the same rewrite.
```

Fix a venom codegen bug where a `return` in the constructor deployed a contract with empty runtime code.

### Cute Animal Picture

### DIFF
--- a/tests/unit/compiler/venom/test_simplify_cfg.py
+++ b/tests/unit/compiler/venom/test_simplify_cfg.py
@@ -276,7 +276,7 @@ def test_merge_jump_dedup_phi_when_direct_edge():
 
     entry:
         %x = source
-        jnz %cond, @join, @join
+        jmp @join
 
     other:
         %o = source
@@ -317,6 +317,36 @@ def test_merge_jump_conflicting_phi_operands():
     """
 
     _check_no_change(pre, hevm=False)
+
+
+def test_merge_jump_converging_arms():
+    """
+    Regression test: bypassing both arms of a `jnz` can make them converge on
+    the same block; the branch must become a `jmp` so that it does not end up
+    with a single cfg successor.
+    """
+    pre = """
+    _global:
+        %cond = source
+        jnz %cond, @then, @else
+
+    then:
+        jmp @join
+
+    else:
+        jmp @join
+
+    join:
+        sink %cond
+    """
+
+    post = """
+    _global:
+        %cond = source
+        sink %cond
+    """
+
+    _check_pre_post(pre, post)
 
 
 def test_data_section_label_chain():
--- a/tests/unit/compiler/venom/test_simplify_cfg.py
+++ b/tests/unit/compiler/venom/test_simplify_cfg.py
@@ -247,6 +247,9 @@ def test_merge_jump_target_has_no_phi():
 def test_merge_jump_dedup_phi_when_direct_edge():
     """
     If the bypassed block's target is already a successor, avoid duplicating phi labels.
+    Both arms of `entry`'s `jnz` then converge on `@join`, so it must be
+    canonicalized to a `jmp` - a `jnz` with a single cfg successor breaks
+    consumers like `BranchOptimizationPass`.
     """
     pre = """
     _global:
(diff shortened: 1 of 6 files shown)


tests/unit/compiler/venom/test_simplify_cfg.py
```diff
@@ -319,6 +319,36 @@ def test_merge_jump_conflicting_phi_operands():
     _check_no_change(pre, hevm=False)
 
 
+def test_merge_jump_converging_arms():
```
HodanPlodky: I dont think this test does what doc comment says and it passes in the master as well. The case that is described is exercised in `test_merge_jump_dedup_phi_when_direct_edge` in this file

HodanPlodky (commented):
Just one thing otherwise LGTM

tests/unit/compiler/venom/test_simplify_cfg.py
charles-cooper: @harkal

charles-cooper:
related: https://github.com/vyperlang/vyper/pull/5066

tests/unit/compiler/venom/test_simplify_cfg.py
harkal: It was already addressed

HodanPlodky (approved):

PR #5209, opened 2026-08-05
fix[parser]: fix source-map end positions

### LINKED ISSUE #5027: Invalid source map: negative length emitted for inline `a ** b` (runtime exponent) used as a call argument
### Version Information

* vyper Version (output of `vyper --version`): `0.4.3+commit.bff19ea2`
* OS: linux
* Python Version (output of `python --version`): `Python 3.10.6`

### What's your issue about?

Vyper emits an invalid source map with a negative length when an external-call argument is an inline runtime exponentiation `a ** b` (where the exponent is a non-constant, e.g. an `immutable`/storage var). Source-map lengths must be non-negative — the only legal "negative" token is the `-1:-1:-1` sentinel — so this produces a malformed artifact.

The compilation itself succeeds (no compiler error, `--verbose` shows nothing relevant); the bug is in the emitted source map, visible via `-f source_map`.

**Command:**
`vyper -f source_map repro.vy`

**Minimal repro (`repro.vy`):**
```
# @version 0.4.3

interface Vault:
    def convertToAssets(shares: uint256) -> uint256: view

d: immutable(uint256)

@deploy
def __init__():
    d = 18

@external
@view
def get(v: address) -> uint256:
    return staticcall Vault(v).convertToAssets(10 ** d)
```

**Output (trimmed):** `pc_pos_map_compressed` contains the token `248:-3:0` (start `248`, length `-3`, source `0`). The uncompressed `pc_pos_map` shows the cause - the span for the `10 ** d` argument is `[15, 47, 15, 44]` (`[lineno, col, end_lineno, end_col]`), i.e. `end_col` **44 is before** `col` **47**, so the compressed `start:length` becomes negative:
```
"pc_pos_map_compressed": "...;-1:-1:-1;248:-3:0;-1:-1:-1;248:7:0;...",
"pc_pos_map": { "104": [15, 47, 15, 44], ... }
```

(On line 15, `10 ** d` starts at column 47; the node's recorded end column of 44 is bogus.)

**Downstream impact:** consumers that parse the source map reject it. Foundry errors with `failed to parse source map: [N] unexpected negative number` and aborts `forge script` for the entire project (it loads all artifacts to build a trace decoder), breaking even scripts that never reference this contract.

**Workaround:** hoist the exponentiation out of the call site into a precomputed value (e.g. an `immutable` assigned in `__init__`) and pass that variable to the call — the resulting source map is well-formed.

### How can it be fixed?

The source-position annotation for the inline `a ** b` (Pow/BinOp) call-argument node has its end offset before its start offset, yielding a negative length when compressed to `start:length:source`. Fix the span attached to that argument node so it covers the actual `a ** b` source range (or at minimum clamp so `end >= start`/`length >= 0`). Source-map lengths should never be negative except for the `-1` sentinel.

Sporarum:
### What I did

1. Fix https://github.com/vyperlang/vyper/issues/5027 (token end positions are not offset properly)
2. Make every `get_contract`-based test sanity-check the produced source maps

### How I did it

1. Change how we keep track of pre-parser adjustments from a dict of exact position -> offset, to a set of positions where the offset changes. 
  This fixes the issue since now end positions are updated correctly (previously not recorded in the dict)
2. Add a validation method which is called every time `get_contract` is called

### How to verify it

`pytest` (new tests in file `tests/unit/compiler/test_source_map.py`)

Also can check that `test_source_map_no_negative_length_after_keyword_rewrite` fails on master

### Commit message

```
the pre-parser rewrites keywords like extcall and staticcall into
shorter Python stand-ins, and shifts subsequent columns to compensate.
end columns went unadjusted because shifts were only recorded at the
exact column of each rewrite start, producing negative-length spans in
the compressed source map.

this commit replaces the exact-position dict with a per-line sorted list
of (threshold, shift) points looked up via binary search, so shifts
apply to every column past a rewritten keyword.

this commit also makes the get_contract-based tests sanity-check
produced source maps.
```

Fix exported source map sometimes being invalid

### Cute Animal Picture


vyper/ast/parse.py
```diff
@@ -238,13 +238,8 @@ def generic_visit(self, node):
...
 
-        adjustments = self._pre_parser.adjustments
-
-        adj = adjustments.get((node.lineno, node.col_offset), 0)
-        node.col_offset += adj
-
-        adj = adjustments.get((node.end_lineno, node.end_col_offset), 0)
-        node.end_col_offset += adj
+        node.col_offset += self._pre_parser.shift_for(node.lineno, node.col_offset)
+        node.end_col_offset += self._pre_parser.shift_for(node.end_lineno, node.end_col_offset)
```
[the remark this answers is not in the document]
Sporarum: Fixed: c1c24b23dd302d6183ff31707ab0268e5ce3ef47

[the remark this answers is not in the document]
Sporarum: Fixed: 6fcaaf80f02abdda57b06027156229b21cdfd2ce

vyper/ast/pre_parser.py
```diff
@@ -351,9 +353,21 @@ def _parse(self, code: str):
...
+    def shift_for(self, lineno: int, col: int) -> int:
+        """
+        Given a python ast lineno and col, return the difference with the original vyper col
+        """
+        pts = self._shift_points.get(lineno)
+        if not pts:
+            return 0
+        # bisect_right on col: the last shift point with col_key <= col is at
+        # index i - 1 (0 means no shift point applies yet).
+        i = bisect.bisect_right(pts, (col, float("inf")))
```
charles-cooper, line 372: why are we bisecting here?

Sporarum: We're finding the element of `pts` which has the biggest `col_key` that is smaller than `col`
Equivalent to something like:
```
res: int
for (col_key, shift) in pts:
  if (col_key <= col):
    res = shift
  else:
    break

return res
```

Sporarum: In other words, we're looking to find which shift applies at this location

charles-cooper: Maybe I didn't read carefully enough but why wouldn't the line just be in the dict?

Sporarum: The most complete example is this:
```
x = extcall foo(extcall bar())
```
Which will be pre-parsed to this:
```
x = await foo(await bar())
```
And so:
* `x =` should not be shifted
* the first `await` should have its start not shifted, but its end shifted by 2
* `foo(` should be shifted by 2
* the second `await` should have it's start shifted by 2, and it's end shifted by 4
* `bar())` should be shifted by 4

So a single line has actually 3 different shifts

Sporarum: (oh and in this case we could replace `extcall` by `await  ` and avoid having to shift entirely, but there are cases where the shift goes the other way around, so that wouldn't work)

vyper/ast/parse.py
```diff
@@ -216,20 +216,46 @@ def generic_visit(self, node):
...
+            # node has the position fields, adjust them
+
+            # Modules have their positions manually set
+            if not isinstance(node, python_ast.Module):
+                node.col_offset += self._pre_parser.shift_for(node.lineno, node.col_offset)
+                node.end_col_offset += self._pre_parser.shift_for(
+                    node.end_lineno, node.end_col_offset
+                )
+        else:
+            assert not hasattr(node, "col_offset")
```
charles-cooper: i would move these assertions into a single stmt in the loop below. `assert not hasattr(node, field_name)`

vyper/ast/parse.py
```diff
@@ -216,20 +216,46 @@ def generic_visit(self, node):
...
+            assert not hasattr(node, "end_col_offset")
+            # node doesn't have the position fields, copy from parent
+            # (they will already have been adjusted)
+
+            assert len(self._parents) > 0
+            parent = self._parents[-1]
+
+            for field_name in LINE_INFO_FIELDS:
+                parent_field = getattr(parent, field_name)
+                assert parent_field is not None
```
charles-cooper, line 252: i think this assertion is unreachable (getattr will raise unless a fallback will provided) if my memory serves me correctly

vyper/ast/parse.py
```diff
@@ -216,20 +216,46 @@ def generic_visit(self, node):
...
+            assert not hasattr(node, "end_lineno")
+            assert not hasattr(node, "end_col_offset")
+            # node doesn't have the position fields, copy from parent
+            # (they will already have been adjusted)
+
+            assert len(self._parents) > 0
+            parent = self._parents[-1]
+
+            for field_name in LINE_INFO_FIELDS:
+                parent_field = getattr(parent, field_name)
```
charles-cooper, line 251:
```suggestion
                assert not hasattr(node, field_name)
                parent_field = getattr(parent, field_name)
```

vyper/ast/parse.py
```diff
@@ -216,20 +216,46 @@ def generic_visit(self, node):
...
+        elif hasattr(node, "lineno"):
+            assert hasattr(node, "col_offset")
+            assert hasattr(node, "end_lineno")
+            assert hasattr(node, "end_col_offset")
+            # node has the position fields, adjust them
+
+            # Modules have their positions manually set
+            if not isinstance(node, python_ast.Module):
+                node.col_offset += self._pre_parser.shift_for(node.lineno, node.col_offset)
+                node.end_col_offset += self._pre_parser.shift_for(
```
charles-cooper: bad formatting, please extract the rhs expression into a short lived variable

vyper/ast/pre_parser.py
charles-cooper: i don't think we need bisect for algorithmic efficiency. the argument for bisect vs loop is for clarity (which i don't have a strong opinion on)

charles-cooper (approved):
approved pending resolution of some nits

vyper/ast/pre_parser.py
charles-cooper: also comparing ints to inf looks kind of bad

Sporarum: Okay, I refactored it: 17f48a0adddd90c025957f82a3bfc411abfb37f8

vyper/ast/parse.py
Sporarum: But it can still return `None` if the field is present but set to `None`, no ?

vyper/ast/parse.py
Sporarum: Done: b353fcd2dc655a5401160524b860a77defe694ef

vyper/ast/parse.py
Sporarum: Done: 006a37d6cbd7362a13c800b3401a07395029255a

charles-cooper (approved):

PR #5210, opened 2026-08-06
fix[venom]: support `O1` minimal passes

harkal:
### What I did

Reinstated O1 as a distinct lowering-only IR optimization level instead of mapping it to O2.

This PR also:

  - Keeps `NONE` and O1 as distinct optimization levels.
  - Gives `NONE` an independent copy of the O1 IR pass pipeline.
  - Runs final assembly optimization at O1, while `NONE` remains fully unoptimized.
  - Preserves per-function static memory boundaries across Venom text round trips.
  - Places stack-spill regions above every static memory frame and keeps function spill regions disjoint.
  - Places the initial free-memory pointer above both static frames and spill regions.
  - Adds explicit pass-order constraints and optimization-specific test skips.

### How I did it

Added an O1 pipeline containing only the passes required to lower frontend Venom into legal assembly.

Introduced `uses_lowering_only_ir()` for decisions shared by `NONE` and O1, while keeping assembly optimization policy separate.

Added `[eom=N]` serialization and parsing for each function's static memory boundary. `ConcretizeMemLocPass` now preserves finalized `[fmp_lowered, eom=N]` functions and rejects contradictory functions that still contain `alloca`.

Updated stack spilling to allocate disjoint regions above the largest static frame and account for those regions when selecting the initial free-memory pointer.

### Commit message

```
reinstate o1 with the minimum venom passes required to produce legal
assembly while keeping none as a separate level that also disables final
assembly optimization. preserve each function's static memory boundary
when venom is serialized and parsed again so rerunning the pipeline cannot
place stack spills over existing memory. allocate spill regions above all
static frames, keep them disjoint between functions, and move the initial
free-memory pointer beyond those regions. add regression coverage for pass
ordering, memory-layout round trips, spilling, and optimization policies
```

### Cute Animal Picture

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -71,6 +71,7 @@ xfail_strict = true
 markers = [
     "fuzzing: Run Hypothesis fuzz test suite (deselect with '-m \"not fuzzing\"')",
     "requires_evm_version(version): Mark tests that require at least a specific EVM version and would throw EvmVersionException otherwise",
+    "requires_optimization(what): Mark tests whose expectation is produced by an optimization, so they cannot hold at the minimal levels (--optimize O1/none)",
     "venom_xfail: mark a test case as a regression (expected to fail) under the venom pipeline",
     "hevm: run tests marked for symbolic execution",
 ]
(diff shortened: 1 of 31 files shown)


charles-cooper (approved):
sgtm mergeable pending @HodanPlodky review

PR #5221, opened 2026-08-24
refactor[types]: remove unused `_NodeMetadataJournal`

Sporarum:
### What I did

_NodeMetadataJournal was unused since ddfce5273 (from that commit message):
> we also no longer need to use the typechecker speculation
machinery for inferring the type of the loop variable. however, the
NodeMetadata code is not removed because it might come in handy at a
later date.

It is now more than 2 years later, and it hasn't been used since.
(and this commit can always be reverted if that need comes)

### How I did it

Remove `_NodeMetadataJournal`, `NodeMetadata` (became a no-op dict wrapper), and associated tests.

### How to verify it

`pytest`

### Commit message

```
_NodeMetadataJournal was unused since ddfce5273 (from that commit
message): "we also no longer need to use the typechecker speculation
machinery for inferring the type of the loop variable. however, the
NodeMetadata code is not removed because it might come in handy at a
later date."

this commit removes it, `NodeMetadata` (became a no-op dict wrapper),
and the associated test file.
```

(refactor, no user-visible change)

### Cute Animal Picture


charles-cooper (approved):

charles-cooper (changes requested):
hang on, it is used though? like we are using NodeMetadata. the journal functionality is unused, but that doesn't really bother me..

Sporarum:
@charles-cooper
As you prefer, I don't see a value in the journal functionality.
I think it's usually better to avoid mutations, rather than recording them to undo them

There is also be a performance impact to having that journal around, but I have not tested to see if it is significant
(and I have no intuition for whether it would be)

Feel free to close as you see fit

charles-cooper:
> @charles-cooper As you prefer, I don't see a value in the journal functionality. I think it's usually better to avoid mutations, rather than recording them to undo them
> 
> There is also be a performance impact to having that journal around, but I have not tested to see if it is significant (and I have no intuition for whether it would be)
> 
> Feel free to close as you see fit

oh yea, i'm not concerned about the performance impact. and yes we do not need the journal anymore.. we could keep NodeMetadata as a class but delete the journal part

Sporarum:
@charles-cooper 
But then what's the point of NodeMetadata ?

charles-cooper:
> @charles-cooper But then what's the point of NodeMetadata ?

type safety / hygiene / better signatures

charles-cooper:
relatedly makes it easier to refactor if we ever want to add or change functionality to the node metadata

Sporarum:
Done: f43f0839f7f229b45e6ea155b43a6f0e87152c6f

But as you can see, I'm not sure this adds much
Since `NodeMetadata` is essentially untyped

charles-cooper (approved):

charles-cooper:
@Sporarum approved, please fix lint and i think we should also add a `typechecker` scope in pull-request.yaml workflow

Sporarum:
> @Sporarum approved, please fix lint

I had forgotten to push it ...

> i think we should also add a `typechecker` scope in pull-request.yaml workflow

I agree !

However, I don't think it applies for this PR, the issue being neither does any others :/

Issue #5226, opened 2026-08-28
archive output is silently broken when relative imports cross above cwd


charles-cooper:
_co-authored by claude fable 5_

### Version Information

* vyper Version (output of `vyper --version`): master (c1b6130 and later)
* OS: all
* Python Version: all

### What's your issue about?

`-f archive` can silently produce an archive that fails to recompile, when a source file is reached via a **relative import crossing above the cwd**, and that file is covered by a search path above the cwd (e.g. `-p ..`).

Repro:

```
/tmp/esc/escaped.vy      # x: constant(uint256) = 1
/tmp/esc/pkg/main.vy     # from .. import escaped
                         # y: constant(uint256) = escaped.x

$ cd /tmp/esc/pkg
$ vyper -p .. -f archive -o out.zip main.vy   # succeeds
$ vyper -f bytecode out.zip
vyper.exceptions.ModuleNotFound: escaped

  contract "main.vy:1", line 1:0
  ---> 1 from .. import escaped
  -------^
```

Mechanism:

- Archive member names are `_anonymize(safe_relpath(resolved_path))` (`vyper/compiler/output_bundle.py`), which rewrites `..` segments to positional digits: `../escaped.vy` is stored as `0/escaped.vy`. Zip members can't carry `..`, so some rewrite is unavoidable.
- Anonymized *search paths* get the same rewrite (`..` → `0`), so **absolute** imports resolved through an above-cwd search path round-trip fine (`0/lib/foo.vy` is found via search path `0/lib`).
- **Relative** imports don't use search paths at all — they resolve from the importing module's parent (`vyper/semantics/analysis/imports.py`, `_load_file` with `level != 0`). On recompile, `main.vy`'s parent is the archive root, so the import requests `../escaped.vy`, which doesn't exist in the zip.

Note the rewrite is not broken for *all* above-cwd sources: a relative import whose importer and importee share the same leading `..` prefix still round-trips (e.g. `../a.vy` and `../b.vy` become `0/a.vy` and `0/b.vy`, and `from . import b` resolves from parent `0`). The failure is an import edge that *crosses* the anonymized prefix — importer and importee at different `..` depths, most commonly inside-cwd importing outside-cwd as above. Any fix should reject/repair only crossing edges, not all above-cwd bundles.

Related: this is the same failure mode that #4706's `.`-fallback would have produced. When the file is *not* covered by any search path, `used_search_paths` errors out at bundle time; the case here bypasses that check because the file *is* inside a search path.

[...]

PR #5227, opened 2026-08-28
fix[tool]: report output bundle path errors

charles-cooper:
_co-authored by claude fable 5_

Supersedes #4706.

### What I did

Raise a user-facing error instead of `CompilerPanic` when a source file escapes all search paths during output bundle construction (e.g. a relative import reaching above the search path root). The panic implied a compiler bug and told users to file an issue, but this is a user-configuration problem.

Why not #4706's approach (append `.` to the bundle search paths and emit the bundle anyway): it produces a silently-broken archive. Relative imports resolve from the importing module's parent, not from search paths, and `_anonymize` rewrites `..` segments (`../escaped.vy` → `0/escaped.vy`), so on recompilation the import requests `../escaped.vy`, which doesn't exist in the zip. A reproducible-build artifact that fails verification downstream is worse UX than an upfront error.

The error hint (run the compiler from a common parent directory of all sources) was verified to round-trip: with `..`-free relpaths, anonymization is the identity and the archive recompiles.

Note there is a residual pre-existing hole this PR does not close: adding e.g. `-p ..` puts the escaping file inside a search path, bypassing this check and silently emitting a broken archive. Tracked in #5226.

### How I did it

- new `BundleError(VyperException)` in `vyper/exceptions.py`
- `OutputBundle.used_search_paths` raises it (with hint) instead of `CompilerPanic`

### How to verify it

`test_output_bundle_escaping_file` covers it: compilation of the escaping layout succeeds, `-f archive` raises `BundleError`. Manually:

```
$ cd pkg && vyper -f archive main.vy   # main.vy has `from .. import escaped`
vyper.exceptions.BundleError: cannot bundle `/tmp/esc/escaped.vy`: it is not contained in any search path
  (hint: run the compiler from a common parent directory of all source files)
$ cd .. && vyper -f archive -o out.zip pkg/main.vy && vyper -f bytecode out.zip   # round-trips
```

### Commit message

```
fix[tool]: report output bundle path errors

raise a user-facing `BundleError` instead of `CompilerPanic` when a
source file escapes all search paths during output bundle construction
(e.g. a relative import reaching above the search path root). the
error hints to run the compiler from a common parent directory of all
sources, which makes relpaths `..`-free so the archive recompiles.
```

Bundling a file outside all search paths now raises a `BundleError` with a recovery hint instead of `CompilerPanic`.

### DIFF
--- a/tests/unit/cli/vyper_compile/test_compile_files.py
+++ b/tests/unit/cli/vyper_compile/test_compile_files.py
@@ -15,7 +15,7 @@ from vyper.compiler.input_bundle import FilesystemInputBundle, JSONInputBundle
 from vyper.compiler.output_bundle import OutputBundle
 from vyper.compiler.phases import CompilerData
 from vyper.compiler.settings import Settings
-from vyper.exceptions import TypeMismatch
+from vyper.exceptions import BundleError, TypeMismatch
 from vyper.utils import sha256sum
 
 TAMPERED_INTEGRITY_SUM = sha256sum("tampered integrity sum")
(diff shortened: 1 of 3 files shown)


Sporarum (approved):
LGTM

pcaversaccio (approved):

PR #5229, opened 2026-08-29
fix[codegen]: preserve consts when pruning dead code

harkal:
### What I did

Fix an codegen bug causing (`KeyError: CONSTREF mem_deploy_end`) when compiling a contract which has an immutable and a constructor that always reverts.

Repro, on master with default settings:

```vyper
V: public(immutable(uint256))

@deploy
def __init__():
    self.V = 1
    raise "nope"
```

```
KeyError: CONSTREF mem_deploy_end
```

### Commit message

```
`_prune_unreachable_code` deletes every assembly item between an
unconditional terminator and the next JUMPDEST. when the constructor
always reverts, the deploy sequence which follows the constructor body
is unreachable, so it gets pruned. but that sequence also *declares*
`mem_deploy_end`, and the `istore`/`iload` sequences for immutables --
which are in reachable code -- refer to it via CONSTREF. pruning the
declaration leaves the reference dangling and symbol resolution fails.

CONST items are declarations, not code: they emit no bytes and their
position in the assembly list carries no meaning. so rather than
special-casing `mem_deploy_end` or hoisting its declaration out of the
deploy sequence, keep CONST items in place and prune only the items
which actually emit bytecode. this covers any zero-byte declaration that
lands in an unreachable range, not just this one.

only the legacy pipeline is affected -- venom compiles the same contract
fine. `-O none` is also unaffected, since the assembly optimizer does
not run at that level.
```

### Cute Animal Picture

### DIFF
--- a/tests/functional/codegen/features/test_immutable.py
+++ b/tests/functional/codegen/features/test_immutable.py
@@ -1,6 +1,7 @@
 import pytest
 
 from vyper.compiler.settings import OptimizationLevel
+from vyper.utils import method_id
 
 
 @pytest.mark.parametrize(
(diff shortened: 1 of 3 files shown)


charles-cooper (approved):
Probably should prune if no dangling refs

PR #5230, opened 2026-09-01
refactor[parser]: remove `ParserException`

Sporarum:
### What I did

Remove `ParserException`
The one error that used it is now a `SyntaxException`

### How I did it

It was only used for a single error: when a source file contains null bytes
Since the python parser already checks for it, we can instead rely on that

### How to verify it

See new tests in tests/unit/ast/test_parser.py

### Commit message

```
`ParserException` was only used when a source file contains null bytes.
this was checked before running the python parser, which emits a similar
error message. this commit removes `ParserException` by leveraging that
fact, as a result the error goes from `ParserException("No null bytes
(\\x00) allowed in the source code.")` to `SyntaxException("source code
string cannot contain null bytes")`
```

(refactor, no visible change)

### Cute Animal Picture

### DIFF
--- a/tests/unit/ast/test_parser.py
+++ b/tests/unit/ast/test_parser.py
@@ -1,3 +1,5 @@
+import sys
+
 import pytest
 
 from tests.ast_utils import deepequals
@@ -5,7 +7,9 @@ from vyper.ast.parse import parse_to_ast
 from vyper.compiler import compile_code
 from vyper.exceptions import SyntaxException
 
-_NULL_BYTE_MSG = "source code string cannot contain null bytes"
+_NULL_BYTE_MSG = (
+    f"source code {'string ' if sys.version_info < (3, 12) else ''}cannot contain null bytes"
+)
 
 
 def test_ast_equal():
--- a/tests/unit/ast/test_parser.py
+++ b/tests/unit/ast/test_parser.py
@@ -2,7 +2,10 @@ import pytest
 
 from tests.ast_utils import deepequals
 from vyper.ast.parse import parse_to_ast
+from vyper.compiler import compile_code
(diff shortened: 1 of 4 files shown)


charles-cooper (approved):

tests/unit/ast/test_parser.py
```diff
@@ -1,9 +1,16 @@
...
+
 import pytest
 
 from tests.ast_utils import deepequals
 from vyper.ast.parse import parse_to_ast
+from vyper.compiler import compile_code
 from vyper.exceptions import SyntaxException
 
+_NULL_BYTE_MSG = (
+    f"source code {'string ' if sys.version_info < (3, 12) else ''}cannot contain null bytes"
```
charles-cooper, line 11: oh, no ternary hidden inside of an f-string, please! just use a branch at top level (kind of standard style for version branches, they should be as toplevel and visible as possible)

Sporarum: Done: 322abadc417b5b4ca4be4e3d7e1c1c59c23fdd29

tests/unit/ast/test_parser.py
```diff
@@ -53,3 +56,50 @@ def f():
     annotation = exc.annotations[0]
     assert (annotation.lineno, annotation.col_offset) == (3, 4)
     assert annotation.full_source_code == code
+
+
+def test_null_byte_in_main_file():
+    code = "a: uint256 = 1\x00\n"
+    with pytest.raises(SyntaxException) as exc_info:
```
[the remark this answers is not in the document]
Sporarum: Done: 8a8627deaff8b727c89bd0614728e59fa6169a7e

charles-cooper (approved):
approved, ty

PR #5219, opened 2026-08-14
fix[venom]: fuzz venom pass ordering

HodanPlodky:
### What I did
Created the fuzz test that showed few error in the implementation. There error are fixed and discuses in the https://github.com/vyperlang/vyper/pull/5231 

### How to verify it
`pytest -m "fuzzing" --experimental-codegen tests/functional/examples/thirdparty/test_thirdparty_fuzz.py --optimize O3 -k test_compile_pass_fuzz`

### Commit message

```
This commit adds the fuzz test to check the interaction between the passes by disabling the subset of the passes and compiling the thirdparty examples.
```

### Cute Animal Picture

Put a link to a cute animal picture inside the parenthesis-->

PR #5231, opened 2026-09-01
fix[venom]: fix various well formedness issues

HodanPlodky:
### What I did
Implementation of the fixes that were shown by the https://github.com/vyperlang/vyper/pull/5219. 

`venom_to_assembly.py`: There was dependency on liveness in the `clean_stack_from_cfg_in`. For the `offset` and `phi` instructions the output could have been in the stack even if was never used.

`BranchOptimizationPass`: In the case that the branch has a literal as a condition (this happens when SCCP is not run) this pass failed on assertion. If thats the case short circuit.
 
`RemoveUnusedVariablesPass`: The `load` instructions could be removed by this pass but did no invalidate their respective `memalias` analyses

`FunctionInlinerPass`: `retpc_param` should not be needed after if function inline. There was an error where because if was still there the label that does no longer exist was still referenced in the code.

### Commit message

```
Implementation of the fixes that were shown by the https://github.com/vyperlang/vyper/pull/5219. 
```

### Cute Animal Picture


vyper/venom/passes/function_inliner.py
```diff
@@ -132,12 +132,18 @@ def _inline_call_site(self, func: IRFunction, call_site: IRInstruction) -> None:
...
         binding_ops = InvokeLayout(self.ctx, call_site).bound_params
 
+        retpc_op = None
         for bb in func_copy.get_basic_blocks():
             bb.parent = call_site_func
             call_site_func.append_basic_block(bb)
             param_idx = 0
             for inst in bb.instructions:
                 if inst.is_param:
+                    if inst.opcode == "retpc_param":
```
harkal: `FunctionCallLayout.return_pc_param()` which already resolves retpc properly, could we use that here instead of matching the opcode? Raw written venom with plain 'param" inlines correctly on master.

harkal (changes requested):

vyper/venom/passes/function_inliner.py
HodanPlodky: Used the `FunctionCallLayout.retpc_param_opcode_inst` since the retpc_param should exist in the inliner. And added the test to show the problematic case outside of the fuzz test in the other PR

PR #5232, opened 2026-09-02
feat[venom]: amortize unbounded `DynArray` append

harkal:
### What I did

Make `append()` on `DynArray[T, INF]` locals amortized linear. Previously every append reallocated an exact-size payload and copied the whole array, so a loop of n appends was O(n^2).

### How I did it

Widen the pointer cell of INF locals from one word to two: payload pointer and capacity of the owned payload. Append writes in place while `length < capacity`, otherwise allocates `max(2 * capacity, length + 1)` elements and copies. Only append sets a nonzero capacity; every other producer stores 0, which forces the first append to copy (the payload may not be owned, e.g. an internal callee's parameter cell aliasing the caller's buffer) and keeps the first append after a large ingest exact-size rather than doubling.

Three commits: centralize cell stores into `store_pointer_cell`, widen the cell and update all writers, add the capacity-aware append lowering and tests.

### How to verify it

```
uv run pytest tests/functional/codegen/features/test_inf_dynarray_runtime.py --experimental-codegen
```

Append-loop gas, `uint256` elements:

| n | master | this PR |
|---|---|---|
| 100 | 128,706 | 41,573 |
| 1000 | 494,512,315 | 211,452 |

Single append after a 500-element calldata copy: +144 gas (the extra cell word).

### Commit message

```
appending to a `DynArray[T, INF]` local reallocated an exact-size
payload and copied the whole array on every call, so building an array
with n appends did O(n^2) copying. beyond a few hundred elements this
was unusable: 1000 appends of `uint256` cost ~494M gas.

widen the pointer cell of INF locals from one word to two: the payload
pointer and the capacity of the owned payload. append writes in place
while `length < capacity`, otherwise it allocates `max(2 * capacity,
length + 1)` elements and copies, so repeated appends are amortized
linear (1000 appends: ~211k gas).

capacity is only ever set nonzero by append itself; every other producer
(assignment, decode from calldata, kwarg default, internal call
parameter) stores 0. this doubles as an ownership sentinel: a 0-capacity
payload may alias memory the cell does not own (an internal callee's
parameter cell points into the caller's buffer), so the first append
always copies before writing. it also means the first append after
ingesting a large array allocates the exact new size rather than
doubling, avoiding a 2x memory jump on the ingest path; doubling starts
from there.

the cell layout is defined in one place (`store_pointer_cell`) and all
cell writers go through it. `Bytes[INF]`/`String[INF]` cells share the
layout with capacity permanently 0 since bytestrings have no append. all
size arithmetic in the grow path is checked. cost is one extra word per
INF pointer cell and one extra `mstore` per cell store.
```

`append()` on `DynArray[..., INF]` locals is now amortized linear instead of copying the whole array on every call.

### Cute Animal Picture

### DIFF
--- a/vyper/codegen_venom/expr.py
+++ b/vyper/codegen_venom/expr.py
@@ -1869,7 +1869,8 @@ class Expr:
 
         new_length = self.builder.add(length, IRLiteral(1))
         self.builder.mstore(new_ptr, new_length)
-        self.ctx.store_pointer_cell(var.value.operand, new_ptr)
+        # the exact-sized payload has room for exactly new_length elements
+        self.ctx.store_pointer_cell(var.value.operand, new_ptr, new_length)
         return VyperValue.from_stack_op(IRLiteral(0), VOID_TYPE)
 
     def _lower_dynarray_pop(self) -> VyperValue:
--- a/vyper/codegen_venom/expr.py
+++ b/vyper/codegen_venom/expr.py
@@ -1809,12 +1809,17 @@ class Expr:
         return VyperValue.from_stack_op(IRLiteral(0), VOID_TYPE)
 
     def _lower_unbounded_dynarray_append(self) -> VyperValue:
-        """Lower append on exact-sized DynArray[..., INF] pointer-cell locals.
+        """Lower append on DynArray[..., INF] pointer-cell locals.
 
-        The current representation keeps no spare capacity: each append
-        allocates the exact new size and copies the old payload before writing
-        the new element. Repeated appends are therefore linear per append and
-        quadratic in aggregate.
+        The pointer cell carries the owned payload capacity next to the
+        payload pointer (see VenomCodegenContext.store_pointer_cell). When
+        the payload has spare room the element is written in place; otherwise
+        a payload of `max(2 * capacity, length + 1)` elements is allocated
+        and the old contents copied over, making repeated appends amortized
+        linear in aggregate. Non-append stores leave capacity at 0, so the
+        first append after an assignment/decode allocates the exact new size
+        (no 2x memory jump on a large ingested array) and doubling starts
+        from there.
         """
         node = self.node
         assert isinstance(node, vy_ast.Call)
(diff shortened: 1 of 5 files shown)


charles-cooper (approved):
approved in principle pending clanker and @HodanPlodky correctness review

vyper/codegen_venom/expr.py
```diff
@@ -1848,28 +1853,53 @@ def _lower_unbounded_dynarray_append(self) -> VyperValue:
...
+        b.append_block(fast_bb)
+        b.set_block(fast_bb)
+        b.mstore(target_cell._ptr, old_ptr)
+        b.jmp(join_bb.label)
+
+        b.append_block(grow_bb)
+        b.set_block(grow_bb)
+        doubled = self.ctx.checked_add(capacity, capacity)
+        min_cap = self.ctx.checked_add(length, IRLiteral(1))
+        new_cap = b.select(b.lt(doubled, min_cap), min_cap, doubled)
```
HodanPlodky, line 1884: just check this should be only case when the `capacity` is zero right?

vyper/codegen_venom/module.py
```diff
@@ -1107,7 +1107,7 @@ def _store_abi_arg_to_existing_ptr(
         assert elem_src.location is not None, "src must have a location for ABI decoding"
         hi = _abi_arg_hi(ctx, elem_src.location)
         val = decode_unbounded_sequence_to_scratch(ctx, elem_src, arg.typ, hi, arg.name)
-        ctx.builder.mstore(dst, val.operand)
+        ctx.store_pointer_cell(dst, val.operand, IRLiteral(0))
```
HodanPlodky, line 1110: should we default to 0?

HodanPlodky (approved):
Just leaving comments about starting size otherwise it looks good

vyper/codegen_venom/expr.py
harkal: A nonzero capacity is only ever written by this grow path, always with `length <= capacity`, and between stores length only changes via pop or the fast path (guarded by `length < capacity`). So reaching grow with `capacity > 0` implies `length == capacity`, and `2 * capacity >= length + 1` always wins. The `length + 1` arm fires exactly when `capacity == 0` (the first append after an assign/decode/param entry)

vyper/codegen_venom/module.py
harkal: Can you elaborate?

harkal: settled offline

Issue #5199, opened 2026-07-22
Panic on empty list literal expression


Sporarum:
### Version Information

* vyper Version (output of `vyper --version`): current master (2a6f8a84cd7fd76675e814bc769a46b51a707276) and #5177
* OS: NixOS
* Python Version (output of `python --version`): Python 3.12.13

### What's your issue about?

When empty list literals are used as part of an expression, the compiler panics:

```python
def foo(c: bool):
    tmp: DynArray[uint256, 1] = [] if c else []
```

```
vyper.exceptions.CompilerPanic: cannot dereference non-pointer type

  contract "playground/test.vy:3", function "foo", line 3:4 
       2 def foo(c: bool):
  ---> 3     tmp: DynArray[uint256, 1] = [] if c else []
  -----------^


This is an unhandled internal compiler error. Please create an issue on Github to notify the developers!
```

Note that this only happens if the list:
1. is in an expression, `tmp0: ... = []` works
2. is a literal, `tmp0 if c else []` and `[] if c else tmp0` work
3. is empty, `[1] if c else []` works, but `[] if c else [1]` doesn't

Seems like it doesn't happen in venom

### How can it be fixed?

No idea ^^'

Issue #5017, opened 2026-05-27
Simplify Typer Internals


Sporarum:
Currently the typer is not located in one single spot, but instead spread over multiple files and classes.

The goal is to restructure the typing infrastructure to make it easier to understand, more standard and therefore easier to extend with features like generics.

### Bidirectional Type Checking

Consists of two components:
* the *type checker*: Given an expression `e` and an expected type `T`, it checks whether `e: T`. This is top-down, the expected type trickling down to the atoms.
* the *type synthesizer*: Given an expression `e`, it tries to *infer* a type `T` such that `e: T`. This is generally bottom-up, types are inferred for the atoms, and then propagated up.

One of its strengths is that it offers type inference by default, while being compatible with subtyping (`<:`).

The type checker can call the type synthesizer, for example when type checking `1 + 1` against type `uint256`: The type synthesizer is called on `1 + 1` and the inferred type is then compared to the expected type.

The type synthesizer can call the type checker, for example when synthesizing a type for `foo(bar)`: We first check that `bar` conforms to the type of the first parameter of `foo`, and then set the inferred type for the whole expression to the return type of `foo`.

This mutual recursion allows for each component to have a single mission, while being able to handle complex expressions, for example ones with complicated generics.

## The current state

In Vyper, type inference is only done inside expressions, things like variables do not get their types inferred: `bar = 4` is not a valid way to declare variable `bar`, unlike `bar: int8 = 4`.

Vyper already has a kind of bi-directional type checker:
* `get_{exact_type,possible_types}_from_node` powered by `_ExprAnalyser` in `semantics/analysis/util.py`, acts as the *type synthesizer*: It takes an expression as input, and outputs its type(s), or raises if unable.
* `validate_expected_type` in `semantics/analysis/utils.py` and `ExprVisitor` in `semantics/analysis/local.py` acts as the *type checker*. (But `ExprVisitor` also does other things)
* Subtyping is handled by:
	* expression types: by the types themselves, through methods `is_subtype_of`,  `is_equivalent_to`, `is_supertype_of`, and the deprecated `compare_type` .
	* `ContractFunctionT`: 
		* when implementing interfaces: member `implements`.
		* when overriding an abstract method: function `_compute_override_discrepancies` in `semantics/analysis/module.py` .
* Type joins(?) are handled by `get_common_types` in `semantics/analysis/utils.py`.

The current system however does a lot of redundant work, since every expression is type-checked, and type-checking calls type synthesis:

[...]

Issue #3480, opened 2023-06-05
Ternary operator unable to return `empty` type


pcaversaccio:
### Version Information

* vyper Version (output of `vyper --version`): `0.3.9`
* OS: `linux`
* Python Version (output of `python --version`): `3.11.3`

### What's your issue about?

The following snippet will throw the error `Exception: Weird code element: ~empty <empty(uint256)>`:

```vyper
@external
@pure
def ceil_div(x: uint256, y: uint256) -> uint256:
    assert y != empty(uint256), "Math: ceil_div division by zero"
    return empty(uint256) if (x == empty(uint256)) else unsafe_add(unsafe_div(x - 1, y), 1)
```

To fix it, simply replace `empty(uint256)` with `0`.

```vyper
@external
@pure
def ceil_div(x: uint256, y: uint256) -> uint256:
    assert y != empty(uint256), "Math: ceil_div division by zero"
    return 0 if (x == empty(uint256)) else unsafe_add(unsafe_div(x - 1, y), 1)
```

### How can it be fixed?

Allow returning `empty` type with the ternary operator.

trocher:
Some minimal repro for the tests when the issue will be fixed:

``` Vyper
@external
def foo():
    success: address = empty(address) if True else empty(address)
```

cyberthirst:
@trocher further extends the issue:

`IfExp` AST nodes containing on both side values that have no location compile to an incorrect IR node.

This is due to the following branch in `parse_IfExp` not being entered as both sides of the `IfExp` have no location.

```python
if body.location != orelse.location or body.value == "multi":
    body = ensure_in_memory(body, self.context)
    orelse = ensure_in_memory(orelse, self.context)
```

The branch can be bypassed if either:

* The two sides are "empty" values, i.e. `empty(T)` (see static array example below).
* The left side is an "empty" value, and the right side is a "literal" (multi) value. (see dynamic array example below).

---

### Static arrays:

```python
@external
def test():
    a: uint256[256] = empty(uint256[256]) if True else empty(uint256[256])
```

```
vyper.exceptions.CompilerPanic: unreachable!
```

---

### Dynamic arrays:

```python
@external
def test():
    a: DynArray[uint256, 256] = [] if True else []
    b: DynArray[uint256, 256] = empty(DynArray[uint256, 256]) if True else [1]
```

```
vyper.exceptions.CompilerPanic: cannot dereference non-pointer type
```

---

### Structs:

```python
struct S:
    a: bool

[...]

PR #5234, opened 2026-09-03
feat[lang]: allow `DynArray`s of length 0

Sporarum:
### What I did

Allow `DynArray`s of length 0, including in user-written annotations
Change the type of `[]` to `DynArray[Never, 0]` (was `DynArray[Never, 1]`)

Note: Does **not** allow `Bytes[0]` and `String[0]` (a different PR can do that).

### How I did it

Change bounds check on lengths for DynArrays from `>` to `>=`

### How to verify it

`pytest`, also see new tests

### Commit message

```
this commit allows DArrayT to have a length of zero, and changes the
type of `[]`  to `DArrayT(BottomT(), 0)` (was `DArrayT(BottomT(), 1)`).
```

Allow `DynArray[T, 0]` type (which is only inhabited by `[]`)

### Cute Animal Picture


charles-cooper (approved):
excellent!

harkal:
@Sporarum  tests seem to fail under venom

vyper/codegen/abi_encoder.py
```diff
@@ -101,12 +101,16 @@ def _encode_dyn_array_helper(dst, ir_node, context):
...
         static_ofst = ["mul", i, static_elem_size]
-        loop_body = _encode_child_helper(
-            dst, child_location, static_ofst, "dyn_child_ofst", context
-        )
-        loop = ["repeat", i, 0, len_, ir_node.typ.count, loop_body]
-
-        x = ["seq", loop, "dyn_child_ofst"]
+        if ir_node.typ.count == 0:
+            # we know statically the loop is empty
+            # outside the efficiency reasons, there are issues with bound-0 `repeat`
```
charles-cooper: what are the issues? i would much rather see an optimization which deletes bound 0 repeats rather than whack-a-mole modifying every usage

charles-cooper (changes requested):

vyper/ir/compile_ir.py
charles-cooper:
```suggestion
            if rounds_bound.value == 0:
                # stack: i, rounds
                # if (0 == rounds) { goto end_dest; }
                o.extend(["DUP1", "ISZERO", *JUMPI(exit_dest)])
            elif rounds != rounds_bound:
```

vyper/ir/compile_ir.py
```diff
@@ -376,6 +376,24 @@ def _height_of(varname):
...
+                    o.extend(self._compile_r(start, height))
+                    o.extend(["POP"])
+                if isinstance(rounds.value, int):
+                    assert rounds.value == 0  # rounds <= bound, guaranteed by codegen
+                else:
+                    o.extend(self._compile_r(rounds, height))
+                    # assert rounds == 0
+                    o.extend(self._assert_false())
+                return o
+
```
charles-cooper:
```suggestion
```

vyper/semantics/types/subscriptable.py
```diff
@@ -120,7 +120,7 @@ class _SequenceT(_SubscriptableT):
 
     def __init__(self, value_type: VyperType, length: LengthUpperBound):
         if is_bounded_length(length):
-            if not 0 < length < 2**256:
+            if not 0 <= length < 2**256:
                 raise InvalidType("Array length is invalid")
```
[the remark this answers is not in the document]
Sporarum: Unrelated to the current changes, see https://github.com/vyperlang/vyper/issues/5199

vyper/ir/compile_ir.py
```diff
@@ -388,7 +388,7 @@ def _height_of(varname):
             # stack: i
 
             # assert rounds <= round_bound
-            if rounds != rounds_bound:
+            if rounds != rounds_bound or rounds_bound.value == 0:
```
charles-cooper: the new clause in the condition needs a detailed explanation for why `rounds_bound.value == 0` requires the code generated below

charles-cooper (approved):
lgtm. mergeable pending 2nd pair of eyes from @HodanPlodky

Sporarum:
It turned out to be simpler to refactor the code than to explain the existing one, see: 0b527205cc7f478bb62543d7952754155440b16d

vyper/ir/compile_ir.py
Sporarum: See 0b527205cc7f478bb62543d7952754155440b16d

HodanPlodky (approved):

tests/functional/codegen/types/test_array_indexing.py
```diff
@@ -341,3 +341,19 @@ def foo(i: uint256) -> uint256:
     with tx_failed():
         c.foo(0)
     pytest.xfail("compilation succeeded with correct bytecode, but should have rejected `[][i]`")
+
+
+@pytest.mark.xfail(raises=InvalidOperation)
+def test_index_all_empty_lists_variable_index(get_contract, tx_failed):
```
HodanPlodky, line 347: are we planning to support this in future, since the return of the single of DynArray[Never, 0] will be handled correctly?
And also
```
@external
def foo(i: uint256) -> DynArray[uint256, 5]:
    return [[], [1]][i]
```
would work as I would expect so what is the blocking this one to compile?

HodanPlodky (commented):
This one test just seems bit odd to me but otherwise looks good

vyper/ir/compile_ir.py
```diff
@@ -374,6 +374,9 @@ def _height_of(varname):
             rounds_bound = code.args[3]
             body = code.args[4]
 
+            if isinstance(rounds.value, int):
+                assert isinstance(rounds_bound.value, int)
+                assert 0 <= rounds.value <= rounds_bound.value
```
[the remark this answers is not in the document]
Sporarum: @charles-cooper @HodanPlodky do we care about direct VyperIR inputs ?
(by care I mean preserve backwards compat)

tests/functional/codegen/types/test_array_indexing.py
Sporarum: > are we planning to support this in future, since the return of the single of DynArray[Never, 0] will be handled correctly?

I think we should

> ```
> @external
> def foo(i: uint256) -> DynArray[uint256, 5]:
>     return [[], [1]][i]
> ```
> 
> would work as I would expect so what is the blocking this one to compile?

It does work as you would expect (0 -> [], 1 -> [1], 2+ -> reverts)

The issue is that we use the element type's abi to compile even empty lists
With `[[], [1]]` we infer type `SArrayT(DArrayT(uint256, 1), 2)` for the whole expression, so `[]` is compiled as a `DArrayT(uint256, 2)` (or `, 0]`, doesn't matter)
But with `[[], []]` we infer the type to `SArrayT(DArrayT(BottomT, 0), 2)`, so we don't know how to compile the `[]` elements

This can be fixed in two ways:
1. For nodes of type `DArrayT[T, 0]`, don't fetch the abi-encoding of the elements, just output an empty (discussed in private with @harkal as a possibility)
2. Annotate nodes with the expected type, and not the infered type, requires/part of #5017

And I think we should do both (but outside of scope for this PR)

vyper/ir/compile_ir.py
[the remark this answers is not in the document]
HodanPlodky: I tried it and it fails even on master in `vyper/codegen/ir_node.py`. There is a check `repeat` without `0` bound so why would this be a change against the current version, what am I missing?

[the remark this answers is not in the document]
charles-cooper: no i think the bot comment is being too pedantic

HodanPlodky (approved):

pcaversaccio (commented):
hmmm, if you use `empty` it doesn't compile:

```vy
@external
@pure
def foo(x: DynArray[uint256, empty(uint256)]) -> DynArray[uint256, empty(uint256)]:
    return x
```

charles-cooper:
> hmmm, if you use `empty` it doesn't compile:
> 
> ```vyper
> @external
> @pure
> def foo(x: DynArray[uint256, empty(uint256)]) -> DynArray[uint256, empty(uint256)]:
>     return x
> ```

separate issue, having to do with compile-time elaboration of `empty()`

pcaversaccio:
> > hmmm, if you use `empty` it doesn't compile:
> > ```vyper
> > @external
> > @pure
> > def foo(x: DynArray[uint256, empty(uint256)]) -> DynArray[uint256, empty(uint256)]:
> >     return x
> > ```
> 
> separate issue, having to do with compile-time elaboration of `empty()`

do we have an open issue on this? i found my old one here: https://github.com/vyperlang/vyper/issues/3480

charles-cooper:
i think you could create a new issue -- that one is not about availability of empty in the type system

pcaversaccio:
> i think you could create a new issue -- that one is not about availability of empty in the type system

https://github.com/vyperlang/vyper/issues/5272

PR #5236, opened 2026-09-03
fix[codegen]: widen ternary arms to the result type layout

harkal:
### What I did

Fix a silent miscompile in both codegen pipelines: a ternary (`x if c else y`) over an aggregate returned data laid out per the arm's type but read with the widened result type. For example

```vyper
@external
def f(c: bool, a: DynArray[Bytes[10], 5], b: DynArray[Bytes[512], 5]) -> DynArray[Bytes[512], 5]:
    return a if c else b
```

returned `[b"a", b"", b""]` for `a = [b"a", b"0123456789", b""]`. Same for nested DynArray capacity widening, static arrays of DynArrays, String elements and tuple members (via locals or internal call results). Legacy and venom codegen are both affected.

### How I did it

Each arm is converted to the result layout inside its own branch, before the join. Legacy `parse_IfExp` copies the arm into an internal variable of the result type with `make_setter`; venom `lower_IfExp` uses its typed memory store. Both already widen element-wise.

The gate is a shared structural layout check, `same_memory_layout` in `codegen/core.py`, not type equality. Same-type arms, bytestring arms with different bounds, capacity-only widening and unbounded result types produce byte-identical IR to master on both pipelines.

### Commit message

```
both codegen pipelines lower a ternary by lowering each arm to a pointer
into the arm's own memory layout and typing the result with the expected
(widened) type that semantic analysis stamps on the node. semantic
analysis only stamps the root node; the arms keep their exact types.
when an arm is narrower than the result -- e.g. `DynArray[Bytes[10], 5]`
for a `DynArray[Bytes[512], 5]` result -- every consumer walks the
narrow data with the wide element stride and reads garbage. the same
happens for nested dynarray capacity, static arrays of dynarrays, string
elements and widened tuple members reached through locals or internal
call results. the legacy pipeline (`parse_IfExp`) and the venom pipeline
(`lower_IfExp`) have the same defect.

convert each arm to the result layout inside its own branch, before the
join, so the result pointer genuinely has the result type's layout.
legacy does this with `make_setter` into an internal variable of the
result type; venom with its typed memory store. both already widen
element-wise. the gate is a shared structural layout check
(`same_memory_layout` in codegen/core.py) comparing element and member
layout of dynarrays, static arrays and tuples, rather than type
equality: same-type arms, bytestring arms with different bounds,
capacity-only widening and unbounded result types stay copy-free and
produce identical IR, and the check works for tuples even though
`TupleT.__eq__` always returns True.
```

Fix ternary expressions over aggregates whose arms are narrower than the result type.

### DIFF
--- a/tests/functional/codegen/features/test_ternary_widening.py
+++ b/tests/functional/codegen/features/test_ternary_widening.py
@@ -139,26 +139,6 @@ def bar(c: bool, a: String[3], b: String[64]) -> String[64]:
     assert c.bar(False, "abc", "y" * 64) == "y" * 64
 
 
-# a ternary passed straight to an internal call returns the second arm even
-# when the condition is true, also for arms of identical type
-@pytest.mark.xfail(strict=True)
-def test_ternary_as_internal_call_argument(get_contract):
-    code = """
-@internal
-def _echo(xs: DynArray[Bytes[512], 5]) -> DynArray[Bytes[512], 5]:
-    return xs
-
-@external
-def foo(
-    c: bool, a: DynArray[Bytes[10], 5], b: DynArray[Bytes[512], 5]
-) -> DynArray[Bytes[512], 5]:
-    return self._echo(a if c else b)
-    """
-    c = get_contract(code)
-    assert c.foo(True, SHORT, LONG) == SHORT
-    assert c.foo(False, SHORT, LONG) == LONG
-
-
 def test_ternary_as_event_argument(get_contract, get_logs):
     code = """
 event Picked:
--- a/tests/functional/codegen/features/test_ternary_widening.py
+++ b/tests/functional/codegen/features/test_ternary_widening.py
@@ -357,15 +357,15 @@ def whole(
     return self._short(a) if c else self._long(b)
 
 @external
-def member(
-    c: bool, a: DynArray[DynArray[uint256, 2], 3], b: DynArray[DynArray[uint256, 4], 3]
+def member_flipped(
+    c: bool, a: DynArray[DynArray[uint256, 4], 3], b: DynArray[DynArray[uint256, 2], 3]
 ) -> DynArray[DynArray[uint256, 4], 3]:
-    return (self._long(b) if c else self._short(a))[1]
+    return (self._long(a) if c else self._short(b))[1]
     """
     a = [[1, 2], [3], []]
     b = [[4, 5, 6, 7], [], [8]]
     c = get_contract(code)
     assert c.whole(True, a, b) == ([1], a)
     assert c.whole(False, a, b) == ([2], b)
-    assert c.member(True, a, b) == b
-    assert c.member(False, a, b) == a
+    assert c.member_flipped(True, b, a) == b
+    assert c.member_flipped(False, b, a) == a
--- a/tests/functional/codegen/features/test_ternary_widening.py
+++ b/tests/functional/codegen/features/test_ternary_widening.py
@@ -1,3 +1,5 @@
+import pytest
+
 from vyper.compiler import compile_code
 
 SHORT = [b"a", b"0123456789", b""]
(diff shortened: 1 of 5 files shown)


vyper/codegen_venom/context.py
```diff
@@ -566,6 +566,45 @@ def materialize_value(
...
+        DynArray[Bytes[10], 5] vs DynArray[Bytes[512], 5] (element stride 64
+        vs 544). A wider bytestring bound or DynArray capacity at the top level
+        does not change the layout of the data that is present.
+        """
+        if isinstance(dst_typ, (DArrayT, SArrayT)):
+            assert isinstance(src_typ, (DArrayT, SArrayT))
+            pairs = [(src_typ.value_type, dst_typ.value_type)]
+        elif isinstance(dst_typ, TupleT):
+            assert isinstance(src_typ, TupleT)
+            pairs = list(zip(src_typ.member_types, dst_typ.member_types))
```
charles-cooper: zip silently truncates to shorter of two lists. honestly we should ban zip in this repo

vyper/codegen_venom/context.py
```diff
@@ -566,6 +566,45 @@ def materialize_value(
         self.store_vyper_value(vv, ret.operand, typ)
         return ret
 
+    @classmethod
+    def same_memory_layout(cls, src_typ: VyperType, dst_typ: VyperType) -> bool:
```
charles-cooper: maybe rename to `aliasable` or `punnable`

vyper/codegen/core.py
```diff
@@ -1235,8 +1262,9 @@ def ensure_in_memory(ir_var, context):
     return create_memory_copy(ir_var, context)
 
 
-def create_memory_copy(ir_var, context):
-    typ = ir_var.typ
+def create_memory_copy(ir_var, context, typ=None):
+    if typ is None:
+        typ = ir_var.typ
```
Sporarum, line 1267: I think it might be possible to avoid the need for this (and maybe more) by changing the `node._metadata["type"] =` line in `ExprVisitor`

Right now set it to the expected type, maybe setting it to the expression type would work better for downstream cases like this ?
(We again would need to be mindful of non-abi-encodable types)

This is kind of what #5127 and #5233 are doing

vyper/codegen/core.py
```diff
@@ -1224,6 +1224,33 @@ def _complex_make_setter(left, right, hi=None):
...
+        assert len(src_typ.member_types) == n
+        pairs = [(src_typ.member_types[i], dst_typ.member_types[i]) for i in range(n)]
+    else:
+        # primitive words, bytestrings ([length][data]) and (nominal) structs
+        return True
+
+    # nested values also need equal sizes: they determine strides and offsets
+    return all(
+        s.memory_bytes_required == d.memory_bytes_required and punnable(s, d) for s, d in pairs
+    )
```
Sporarum, line 1251: I think this can be simplified by moving the length check to the beginning:

```suggestion
def punnable(src_typ, dst_typ):
    """
    Return True if memory laid out as `src_typ` can be read (punned) as `dst_typ`.

    Compatible types can differ in element or member layout, e.g.
    DynArray[Bytes[10], 5] vs DynArray[Bytes[512], 5] (element stride 64
    vs 544). A wider bytestring bound or DynArray capacity at the top level
    does not change the layout of the data that is present.
    """
    if src_typ.memory_bytes_required == dst_typ.memory_bytes_required:
        return False
    
    if isinstance(dst_typ, (DArrayT, SArrayT)):
        assert isinstance(src_typ, (DArrayT, SArrayT))
        pairs = [(src_typ.value_type, dst_typ.value_type)]
    elif isinstance(dst_typ, TupleT):
        assert isinstance(src_typ, TupleT)
        n = len(dst_typ.member_types)
        assert len(src_typ.member_types) == n
        pairs = [(src_typ.member_types[i], dst_typ.member_types[i]) for i in range(n)]
    else:
        # primitive words, bytestrings ([length][data]) and (nominal) structs
        return True

    # nested values also need equal sizes: they determine strides and offsets
    return all(punnable(s, d) for s, d in pairs)
```
Which then allows us to recurse immediately:
```suggestion
def punnable(src_typ, dst_typ):
    """
    Return True if memory laid out as `src_typ` can be read (punned) as `dst_typ`.

    Compatible types can differ in element or member layout, e.g.
    DynArray[Bytes[10], 5] vs DynArray[Bytes[512], 5] (element stride 64
    vs 544). A wider bytestring bound or DynArray capacity at the top level
    does not change the layout of the data that is present.
    """
    if src_typ.memory_bytes_required == dst_typ.memory_bytes_required:
        return False
    
    if isinstance(dst_typ, (DArrayT, SArrayT)):
        assert isinstance(src_typ, (DArrayT, SArrayT))
        return punnable(src_typ.value_type, dst_typ.value_type)
    elif isinstance(dst_typ, TupleT):
        assert isinstance(src_typ, TupleT)
        assert len(src_typ.member_types) == len(dst_typ.member_types)
        return all(punnable(s, d) for s,d in zip(src_typ.member_types, dst_typ.member_types))
    else:
        # primitive words, bytestrings ([length][data]) and (nominal) structs
        return True
```

tests/functional/codegen/features/test_ternary_widening.py
```diff
@@ -0,0 +1,193 @@
...
+def via_return(
+    c: bool, a: DynArray[Bytes[10], 5], b: DynArray[Bytes[512], 5]
+) -> DynArray[Bytes[512], 5]:
+    return a if c else b
+    """
+    c = get_contract(code)
+    assert c.via_local(True, SHORT, LONG) == SHORT
+    assert c.via_local(False, SHORT, LONG) == LONG
+    assert c.via_return(True, SHORT, LONG) == SHORT
+    assert c.via_return(False, SHORT, LONG) == LONG
```
Sporarum, line 41: Should we also test `...(cond, LONG, SHORT)` ?

(and same flipping question for the other test cases)

tests/functional/codegen/features/test_ternary_widening.py
```diff
@@ -0,0 +1,193 @@
...
+    ir = ir["ir_runtime"]
+    if experimental_codegen:
+        assert "typed_dyn_copy" not in str(ir)
+    else:
+        ternaries = list(_ternaries(ir, "a if c else b"))
+        assert len(ternaries) == 2
+        for ternary in ternaries:
+            _, body, orelse = ternary.args
+            # each arm is the variable itself, not a copy of it
+            assert body.is_literal and orelse.is_literal
```
Sporarum:
```suggestion
            assert body.is_literal
            assert orelse.is_literal
```

Sporarum (changes requested):
Good find !

It might be possible to remove the need for a lot of added logic by setting the type in metadata in a smarter way instead, see one of the comments below

HodanPlodky (approved):

vyper/codegen/core.py
harkal: Stamping the expression type would only remove the conversion when both arms have the same type. When the arms differ from each other (e.g. `DynArray[Bytes[10], 5]` vs `DynArray[Bytes[40], 5]`), the result still needs a single layout, so at least one arm has to be converted inside the ternary, and `punnable` plus the copy stay either way.

It would also move correctness to every consumer's own widening path (return, call args, events, abi encode). On the venom side that path is already broken for tuples: `store_memory` gates on `src_typ != typ`, and `TupleT.__eq__` always returns True, so tuple ternaries would regress. Happy to switch that gate to `punnable` as a follow-up.

Does #5233 change the stamp for `IfExp` specifically? If so I can rebase on it and drop whatever becomes redundant.

vyper/codegen/core.py
harkal: The top-level size is deliberately not compared. `Bytes[10]` vs `Bytes[40]` and `DynArray[uint256, 2]` vs `DynArray[uint256, 5]` are punnable despite different sizes (the same-type test asserts no copy for those), and `size_in_bytes` panics for a `DynArray[..., INF]` result. Sizes only matter one level down, where they fix the stride.

Also the suggested early return is inverted (returns False on equal sizes), and the second version brings `zip` back, which @charles-cooper asked to remove above.

vyper/codegen/core.py
Sporarum: #5233 does not apply to IfExp, I was more saying that the same approach might work here, but as you have explained, it wouldn't !

vyper/codegen/core.py
Sporarum: I see, thanks

tests/functional/codegen/features/test_ternary_widening.py
```diff
@@ -0,0 +1,371 @@
...
+@external
+def via_local_flipped(
+    c: bool, a: DynArray[Bytes[10], 5], b: DynArray[Bytes[512], 5]
+) -> DynArray[Bytes[512], 5]:
+    ys: DynArray[Bytes[512], 5] = b if c else a
+    return ys
+
+@external
+def via_return_flipped(
+    c: bool, a: DynArray[Bytes[10], 5], b: DynArray[Bytes[512], 5]
```
Sporarum: Very nitpicky:
It would be clearer (ro me) what's going on with:

```suggestion
    c: bool, a: DynArray[Bytes[512], 5], b: DynArray[Bytes[10], 5]
```
(and the body swapped as well)

That way the applications visually look to have their parameters swapped (`..(..., Long, Short)`)

Sporarum (approved):

PR #5237, opened 2026-09-03
fix[venom]: keep invoke arg copies from multiply assigned vars

harkal:
### What I did

Fix a miscompile in the invoke copy forwarding passes: an aggregate ternary passed directly as an internal call argument always received the else arm.

```vyper
@internal
def _echo(xs: DynArray[uint256, 5]) -> DynArray[uint256, 5]:
    return xs

@external
def f(c: bool, a: DynArray[uint256, 5], b: DynArray[uint256, 5]) -> DynArray[uint256, 5]:
    return self._echo(a if c else b)
```

`f(True, [1, 2], [3, 4])` returned `[3, 4]`. Correct at `-Onone`; independent of inlining. Present since the passes started running pre-SSA (#4811).
                                                                                                                                    ### How I did it
                                                                                                                                    `ReadonlyInvokeArgCopyForwardingPass` and `InternalRen once before `MakeSSA`. They walk assign chainsthrough `DFGAnalysis`, which keeps one producing instruction per variable, so a variable assigned in both arms of a branch resolved to only its last definition. The argument was forwardand the copy dropped; the merged variable was thendead at the join, no phi was emitted, and the branch on the condition was removed.                                               
The shared base now collects multiply defined variables up front and fails closed when an assign chain or alias set reaches one, keeping the copy. The post-SSA run is unaffected (the internal return pass had the same defect through amultiply defined alias of the destination and is covered by the same check.

### How to verify it

```
pytest tests/unit/compiler/venom/test_invoke_arg_copy
pytest tests/functional/codegen/features/test_ternary.py --experimental-codegen
```

The two new unit tests and `test_ternary_as_internal_on master. `examples/*.vy` compile to byte-identicalasm vs master.

### Commit message

```
the invoke copy forwarding passes (`ReadonlyInvokeArg
and `InternalReturnCopyForwardingPass`) run once before `MakeSSA`, where
a variable can be assigned on several paths. they wal
through `DFGAnalysis`, which records a single producing instruction per
variable, so for a variable assigned in both arms of
follows only the last definition. for an aggregate ternary passed
directly as an internal call argument this forwarded
else-arm buffer and dropped the copy; the merged variable was then dead
at the join, `MakeSSA` emitted no phi, and later pass
branch on the condition entirely, so the call always received the else
arm. present since the passes started running pre-SSA

collect the multiply defined variables when the pass
closed whenever an assign chain or alias set reaches one: the root is
unknown, so the copy is kept. the post-SSA run is una
set is empty in SSA form. the internal return pass had the same defect
through a multiply defined alias of the destination a
the same check.
```

Fix internal call arguments that are ternary expressiing the wrong arm (venom pipeline).

### DIFF
--- a/tests/functional/codegen/features/test_ternary.py
+++ b/tests/functional/codegen/features/test_ternary.py
@@ -307,3 +307,67 @@ def foo() -> Bytes[10]:
 
     c = get_contract(source)
     assert c.foo() == b"\x01\x02"
+
+
+@pytest.mark.parametrize("test", [True, False])
+def test_ternary_as_internal_call_argument(get_contract, test):
+    code = """
+@internal
+def _echo(xs: DynArray[uint256, 5]) -> DynArray[uint256, 5]:
+    return xs
+
+@external
+def direct(c: bool, a: DynArray[uint256, 5], b: DynArray[uint256, 5]) -> DynArray[uint256, 5]:
+    return self._echo(a if c else b)
+
+@external
+def via_local(
+    c: bool, a: DynArray[uint256, 5], b: DynArray[uint256, 5]
+) -> DynArray[uint256, 5]:
+    xs: DynArray[uint256, 5] = a if c else b
+    return self._echo(xs)
+    """
+    c = get_contract(code)
+    a = [1, 2]
+    b = [3, 4, 5]
+    expected = a if test else b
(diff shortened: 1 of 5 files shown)


charles-cooper (approved):

charles-cooper:
approved pending clanker correctness review

PR #5239, opened 2026-09-04
fix[venom]: type state variable reads with the declared type

harkal:
### What I did

Fix a venom-only miscompile when a storage, transient or immutable variable is
read into a wider type, e.g. `ys: DynArray[Bytes[512], 5] = self.stored` with
`stored: DynArray[Bytes[10], 5]`. Same bug family as #5236.

The semantic pass stamps the `self.x` / `lib.X` node with the *expected*
(widened) type. `codegen_venom` used that type to describe the borrowed pointer
to the state variable, but the data has the declared type's layout. Two things
went wrong:

- slot over-read: `load_storage_to_memory` sized the copy loop from the wide
  type (86 slots for `DynArray[Bytes[512], 5]` vs the declared 11), reading
  slots that belong to neighbouring variables;
- layout skew: the copied words were then walked with the wide element stride
  (544 bytes vs 64), so every element after the first was garbage.

For `[b'a', b'0123456789', b'']` this returned `[b'a', b'', b'']`; other shapes
halted out of gas. Immutables had the same defect via
`load_immutable_to_memory`. Legacy codegen is unaffec
these IRnodes with `varinfo.typ`.

### How I did it

Type the pointer returned by `lower_Attribute` (state variable reads) and
`lower_Name` (immutable reads) with the variable's de
`node._metadata["type"]`. The existing typed copies on the consumer side
(`store_memory(..., src_typ=...)`, `_store_complex_ty
return encoding, call args) then widen element-wise. No new copy machinery.

Not covered here: `for x: <wide> in self.stored` copies each element linearly
in `_lower_iter_loop` and has the same class of bug f
widenings. That is a separate code path and will be a separate PR.

### Commit message

```
state variable reads in codegen_venom wrapped the storage/transient/immutable
pointer with the node's expected type, which can be w
declared type. the storage-to-memory copy then read storage_size_in_words
of the wide type (slots past the variable) and the re
wide element stride, corrupting every element after the first.

type the pointer with varinfo.typ; the existing typed stores widen the
value at the consumer.
```

### DIFF
--- a/tests/functional/codegen/features/test_state_variable_widening.py
+++ b/tests/functional/codegen/features/test_state_variable_widening.py
@@ -1,0 +1,262 @@
+import pytest
+
+SHORT = [b"a", b"0123456789", b""]
+NESTED = [[1, 2], [], [3]]
+
+
+def test_storage_dynarray_bytes_widening(get_contract):
+    """
+    A `DynArray[Bytes[10], 5]` storage variable read into a
+    `DynArray[Bytes[512], 5]` local, return value or argument used to be
+    copied with the wide element layout, walking past the variable's slots
+    and returning garbage.
+    """
+    code = """
+stored: DynArray[Bytes[10], 5]
+
+@internal
+def _identity(xs: DynArray[Bytes[512], 5]) -> DynArray[Bytes[512], 5]:
+    return xs
+
+@external
+def store(xs: DynArray[Bytes[10], 5]):
+    self.stored = xs
+
+@external
+def via_local() -> DynArray[Bytes[512], 5]:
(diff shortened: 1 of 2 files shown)


charles-cooper (commented):
hmm it's a bit surprising that this changes behavior. @Sporarum can you take a look?

HodanPlodky (approved):

Sporarum (commented):
LGTM

The one question I have, is why only those specific cases ?
I'm also having issues with ternary ifs, and I think the root cause is the same

Sporarum (approved):

PR #5240, opened 2026-09-04
feat[venom]: allow abi-dynamic elements in inf dynarrays

harkal:
### What I did

- Allow ABI-dynamic element types (`Bytes[N]`, `String[N]`, `DynArray[T, N]`,
  structs with bytestring members) in `DynArray[T, INF]`. This makes a
  Solidity-compatible `multicall(DynArray[Bytes[N], INF]) -> DynArray[Bytes[N], INF]`
  expressible (0.5.0b1 feedback from the aave port). Follows GH 5232 (amortized
  INF append). Element types must still be bounded: `DynArray[Bytes[INF], INF]`
  is still rejected.
- Fix an element-widening miscompile on the venom INF copy paths that the
  relaxation exposes (`DynArray[Bytes[10], 5]` copied into
  `DynArray[Bytes[512], INF]` at the wrong stride). ABI-static elements never
  hit it because identical element types have identical strides.
- Document the decode cost model (memory amplification from aliased head
  offsets, who pays per ingress path) and the one case where INF decoding is
  stricter than the bounded calldata decoder (truncated final tail is rejected
  rather than zero-filled).

### How I did it

- semantics: drop the ABI-static element check in `DArrayT`. Unbounded types
  nested in elements (`DynArray[Bytes[INF], INF]`) are still rejected by the
  existing nested-unbounded check. Wildcard `DynArray[Bytes[10], ...]` now
  resolves to `DynArray[Bytes[10], INF]` instead of erroring.
- encoder: `dynarray_runtime_abi_size` returns a per-element bound
  (`count * (head + tail bound of T)`) for INF arrays with ABI-dynamic elements
  instead of an exact size. `runtime_abi_size_for_arg` / `_for_encode` document
  that the result is allocation-only and that the encoded length comes from the
  encoder's return value. Every consumer was audited; the local names at those
  sites (`abi_encode`, `print`, external returns, custom errors) are renamed
  `alloc_size` / `*_bound` so the contract is visible. The `abi_encode` padding
  zero write still lands on the allocation's last word because the bound is a
  sum of word multiples.
- decoder: remove the panic. The count bound
  `count <= payload / embedded_static_size(T)` already generalizes because the
  embedded static size of an ABI-dynamic element is its 32-byte head word. A
  comment block records the soundness argument and the amplification exposure:
  scratch is reserved as `count * memsize(T)` before per-element validation and
  head offsets may alias, so a payload can cost up to `memsize(T) / 32` times
  its size in memory expansion (17x for `Bytes[512]`, 129x for `Bytes[4096]`).
  This is the same exposure as a bounded `DynArray[T, N]` return with `N` = the
  claimed count and as Solidity's `bytes[]` memory decoding, so no tail
  pre-scan is emitted; the cost is documented instead.
- codegen: when source and destination element types differ on an INF copy,
  copy element-wise through the typed store helpers
  (`copy_sequence_to_scratch`, internal `dret` returns, dynamic tuple frame
  members). The static-array slow path now shares the same element loop
  (`_copy_elements_typed`). External returns of an INF type size the buffer by
  the declared type's bound but encode from the source layout instead of
  converting first (there is no bounded temporary for an INF type); the
  widened-frame external return reads the frame with the source tuple type.
  The external-return normalization block now gates on INF explicitly instead
  of relying on `TupleT.__eq__` always returning True.
  Ternary arms whose result is an INF array are converted through the same INF
  copy path, since there is no bounded temporary for an INF type.
- docs: `docs/types.rst` describes the new element rule, the per-element memory
  and ABI-buffer reservation, the amplification cost model and the
  truncated-tail divergence from bounded calldata decoding.

Known cost (from the end-to-end multicall tests, py-evm): frame memory is not
reclaimed between iterations of a `results.append(raw_call(...))` loop (the
argument stride plus the doubled append payloads stay pinned), so per-call gas
grows with batch size: exec-only gas for n=5/20/40 calls is ~14.4k/53.9k/136.9k.
Append copying is amortized; memory is not. Fine for realistic batches; a
follow-up can teach the FMP lowering that the previous payload is dead after
`store_pointer_cell`.

### Commit message

```
Allow ABI-dynamic element types (`Bytes[N]`, `String[N]`, `DynArray[T,
N]`, structs with bytestring members) in `DynArray[T, INF]`, so that a
Solidity-compatible `multicall(bytes[]) -> bytes[]` can be written. Drop
the ABI-static element check in `DArrayT`; unbounded element types are
still rejected. Make `dynarray_runtime_abi_size` return an allocation
bound for ABI-dynamic elements, with every consumer taking the encoded
length from the encoder's return value, and remove the decoder panic,
since its count bound already covers ABI-dynamic elements.

Fix element widening on the INF copy paths, which copied at the
destination stride and misplaced the elements: copies into an INF
destination with a different element type, internal and external
returns, and ternary arms now go element by element through the typed
store helpers. Document the element rule and the decode cost model in
`docs/types.rst`.
```

### DIFF
--- a/docs/types.rst
+++ b/docs/types.rst
@@ -584,9 +584,12 @@ may use ``INF`` as the length bound:
         return ys
 
 ``Bytes[INF]`` and ``String[INF]`` can hold any runtime length. ``DynArray[T, INF]``
-can hold any runtime item count, but ``T`` must have an ABI-static layout, such
-as ``uint256``, ``bytes32``, static arrays, or structs made only from ABI-static
-members.
+can hold any runtime item count; ``T`` must be a bounded type, but it may be
+ABI-dynamic, such as ``Bytes[512]``, ``DynArray[uint256, 3]`` or a struct with
+bytestring members. Decoding a ``DynArray[T, INF]`` with an ABI-dynamic ``T``
+reserves memory for every claimed element before the elements themselves are
+validated, so malformed input costs the caller memory expansion gas in
+proportion to the claimed element count.
 
 Unbounded sequence values are supported for memory locals, function arguments,
 function returns, event and custom error members, ABI encoding and decoding, and
@@ -597,7 +600,7 @@ for example ``(uint256, Bytes[INF])``.
 Unbounded sequences are not supported in storage, transient storage, immutable
 module variables, struct members, static arrays, mappings, or nested inside
 another dynamic layout. For example,
-``DynArray[Bytes[INF], INF]`` and ``DynArray[Bytes[10], INF]`` are rejected.
+``DynArray[Bytes[INF], INF]`` and ``DynArray[DynArray[uint256, INF], 3]`` are rejected.
 Tuple arguments and local tuple variables containing unbounded sequence members
 are also rejected.
 
--- a/docs/types.rst
+++ b/docs/types.rst
@@ -586,10 +586,13 @@ may use ``INF`` as the length bound:
 ``Bytes[INF]`` and ``String[INF]`` can hold any runtime length. ``DynArray[T, INF]``
 can hold any runtime item count; ``T`` must be a bounded type, but it may be
 ABI-dynamic, such as ``Bytes[512]``, ``DynArray[uint256, 3]`` or a struct with
-bytestring members. Decoding a ``DynArray[T, INF]`` with an ABI-dynamic ``T``
-reserves memory for every claimed element before the elements themselves are
-validated, so malformed input costs the caller memory expansion gas in
-proportion to the claimed element count.
+bytestring members. Each element then occupies the full memory size of ``T``
+regardless of its actual length, and ABI buffers reserve the full encoded bound
+of ``T`` per element, so valid but non-canonical input (element offsets that
+alias one another) can cost memory proportional to the payload size times the
+memory size of ``T`` in words. ``INF`` decoders bound every element against the
+readable region (``calldatasize`` or the returndata size), so truncated trailing
+data is rejected where a bounded array in calldata would be zero-filled.
 
 Unbounded sequence values are supported for memory locals, function arguments,
 function returns, event and custom error members, ABI encoding and decoding, and
--- a/docs/types.rst
+++ b/docs/types.rst
@@ -584,15 +584,29 @@ may use ``INF`` as the length bound:
         return ys
 
 ``Bytes[INF]`` and ``String[INF]`` can hold any runtime length. ``DynArray[T, INF]``
-can hold any runtime item count; ``T`` must be a bounded type, but it may be
+can hold any runtime item count. ``T`` itself must be bounded, but it may be
 ABI-dynamic, such as ``Bytes[512]``, ``DynArray[uint256, 3]`` or a struct with
-bytestring members. Each element then occupies the full memory size of ``T``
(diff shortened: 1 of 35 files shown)


vyper/codegen_venom/builtins/misc.py
HodanPlodky, line 1: All of the changes in the file are just the var name changes right. We could leave it as is

docs/types.rst
```diff
@@ -584,9 +584,15 @@ may use ``INF`` as the length bound:
         return ys
 
 ``Bytes[INF]`` and ``String[INF]`` can hold any runtime length. ``DynArray[T, INF]``
-can hold any runtime item count, but ``T`` must have an ABI-static layout, such
-as ``uint256``, ``bytes32``, static arrays, or structs made only from ABI-static
-members.
+can hold any runtime item count; ``T`` must be a bounded type, but it may be
+ABI-dynamic, such as ``Bytes[512]``, ``DynArray[uint256, 3]`` or a struct with
+bytestring members. Each element then occupies the full memory size of ``T``
```
Sporarum: I had a bit of trouble following this, I think an example would be useful

docs/types.rst
```diff
@@ -597,7 +603,7 @@ for example ``(uint256, Bytes[INF])``.
 Unbounded sequences are not supported in storage, transient storage, immutable
 module variables, struct members, static arrays, mappings, or nested inside
 another dynamic layout. For example,
-``DynArray[Bytes[INF], INF]`` and ``DynArray[Bytes[10], INF]`` are rejected.
+``DynArray[Bytes[INF], INF]`` and ``DynArray[DynArray[uint256, INF], 3]`` are rejected.
```
Sporarum, line 620: Is `DynArray[DynArray[uint256, INF], 3]` an example of 
> Unbounded sequences are not supported in storage, transient storage, immutable
module variables, struct members, static arrays, mappings, or nested inside
another dynamic layout.

?

vyper/codegen_venom/abi/abi_decoder.py
```diff
@@ -374,13 +374,32 @@ def decode_unbounded_sequence_to_scratch(
...
+        # payload can claim close to payload/32 elements that all validate, and
+        # decoding can cost memory expansion of up to `elem_mem_size / 32`
+        # times the payload size (17x for Bytes[512], 129x for Bytes[4096]).
+        # Who pays depends on the ingress path: for calldata the sender does;
+        # for extcall returndata and abi_decode of foreign bytes the decoding
+        # contract pays for data it did not author. That is the same exposure
+        # as a bounded DynArray[T, N] return type with N = the claimed count
+        # (it reserves N * memsize(T) per call) and as Solidity's memory
+        # decoding of bytes[] (aliased heads copy per element there too), so
+        # no pre-scan of the tails is emitted here.
```
Sporarum: This is very hard to parse, would it be possible to focus it a little more ?

(It feels like it over-explains obvious parts, and under-explain complex parts)

Sporarum (commented):
LGTM !

(but that is bounded by my knowledge of these areas)

charles-cooper (approved):

docs/types.rst
harkal: yes, and I updated the comment

PR #5242, opened 2026-09-04
fix[lang]: reject wildcard tuple returns passed as arguments

### LINKED ISSUE #5241: Panic on wildcard tuple
#5127 (d3c150e4a318b70f2290234619c1d3dc2527ad8a) introduced a panic on the following example:
```python
interface Foo:
    def source() -> (Bytes[...], DynArray[uint256, ...]): nonpayable

    def sink(input: (Bytes[10], DynArray[uint256, ...])): nonpayable

@external
def bar(foo: Foo):
    extcall foo.sink(extcall foo.source())
```
```
vyper.exceptions.CodegenPanic: unhandled exception unsupported operand type(s) for %: 'Inf' and 'int', lower_ExtCall
```
This is documented in an xfail test in #5233

harkal:
### What I did

Fix https://github.com/vyperlang/vyper/issues/5241. Passing the result of an
external call with a wildcard tuple return type (e.g. `(Bytes[...],
DynArray[uint256, ...])`) directly to a wildcard tuple parameter, to `print`,
`raw_create` or `create_from_blueprint` crashed the venom pipeline with
`CodegenPanic: unsupported operand type(s) for %: 'Inf' and 'int'`. It is now
rejected at analysis with the existing "... arguments cannot contain unbounded
sequence types inside aggregate types" errors.

### How I did it

A wildcard call return is resolved against the expected type when the caller
provides a bounded one, and to `INF` otherwise. The `INF` case is a valid
return shape (a discarded call or a subscript keeps working), but a tuple with
`INF` members is not a valid argument type — the same restriction that
declaration-time checks, memory variables and `abi_encode` already enforce.

The consumer-side argument checks read the argument type with
`get_exact_type_from_node`, which returns the unresolved wildcard type; the
nested-unbounded predicate matches only `INF`, so they never fired and the
resolved tuple reached codegen. `abi_encode` already resolved wildcards before
its check; this PR does the same in the call-argument loop (only when the
parameter itself is a wildcard, so a bounded parameter still resolves the
return to the parameter type) and in the base varargs inference (which
covers `print` and `create_from_blueprint` and subsumes `abi_encode`'s
override) plus `raw_create`'s own constructor-argument list.

This is a semantics fix rather than codegen support: the restriction on
`INF` members inside tuples is documented and applies at every other
argument site. Discarding the call (`extcall foo.source()` as a statement)
and subscripting its result intentionally stay legal and are covered by
tests.

### Commit message

```
the wildcard-return branch of `visit_Call` resolves a wildcard external
call return against the expected type when the caller provides a bounded
one, and to `INF` otherwise. the `INF` case is fine for a discarded call
or a subscript, where a tuple of `INF` members is a valid return shape,
but not when the value is passed on as an argument: tuples with `INF`
members are not valid argument types (GH 5241).

the consumer-side checks that enforce this (the call argument loop in
`visit_Call`, and the varargs builtins `print`, `raw_create` and
`create_from_blueprint`) all read the argument type with
`get_exact_type_from_node`, which returns the declared, unresolved
wildcard type. the nested-unbounded predicate only matches `INF`, so the
checks never fired and the resolved `INF` tuple reached codegen, which
panics computing the size bound of a `Bytes[INF]` member.

resolve the wildcards before those checks, as `abi_encode` already did,
so the existing checks reject the value with their user-facing message.
the resolution is applied only where the expected type cannot bound the
return: a wildcard parameter, or a vararg. a bounded parameter still
resolves the return to the parameter type, which is validated at its
declaration. `abi_encode`'s own override is subsumed by resolving in the
base varargs inference. discarding the call and subscripting its result
stay legal. codegen support for `INF` tuple arguments was not pursued:
the restriction is documented and applies to every other argument site.
```

Reject wildcard tuple call returns passed as arguments instead of crashing the
compiler.

### DIFF
--- a/vyper/semantics/analysis/local.py
+++ b/vyper/semantics/analysis/local.py
@@ -79,14 +79,29 @@ from vyper.semantics.types.infinity import (
 from vyper.semantics.types.utils import type_from_annotation
 
 
-def _expr_contains_unbounded_sequence(node: vy_ast.VyperNode) -> bool:
-    if isinstance(node, (vy_ast.Tuple, vy_ast.List)):
-        return any(_expr_contains_unbounded_sequence(item) for item in node.elements)
+def _expr_contains_unbounded_sequence(node: vy_ast.VyperNode, typ: VyperType) -> bool:
+    # walk literals alongside the expected type. a literal whose shape does
+    # not match `typ` is rejected later, when it is visited
+    if isinstance(node, vy_ast.Tuple) and isinstance(typ, TupleT):
+        return any(
+            _expr_contains_unbounded_sequence(item, item_typ)
+            for item, item_typ in zip(node.elements, typ.member_types)
+        )
+    if isinstance(node, vy_ast.List) and isinstance(typ, (SArrayT, DArrayT)):
+        return any(
+            _expr_contains_unbounded_sequence(item, typ.value_type) for item in node.elements
+        )
 
     try:
-        return type_contains_unbounded_sequence(get_exact_type_from_node(node))
+        actual_typ = get_exact_type_from_node(node)
     except VyperException:
         return False
+
+    if typ.has_wildcard:
+        # resolve a wildcard call return against the element's expected
+        # type, the same way `ExprVisitor.visit_Call` does for arguments
+        actual_typ = actual_typ.resolve_wildcard()
+    return type_contains_unbounded_sequence(actual_typ)
 
 
 def analyze_functions(vy_module: vy_ast.Module) -> None:
@@ -1052,7 +1067,7 @@ class ExprVisitor(VyperNodeVisitorBase):
 
             for arg, arg_typ in zip(node.args, func_type.argument_types):
                 if isinstance(arg, (vy_ast.Tuple, vy_ast.List)):
-                    has_nested_unbounded = _expr_contains_unbounded_sequence(arg)
+                    has_nested_unbounded = _expr_contains_unbounded_sequence(arg, arg_typ)
                 else:
                     try:
                         actual_arg_typ = get_exact_type_from_node(arg)
--- a/vyper/semantics/analysis/local.py
+++ b/vyper/semantics/analysis/local.py
@@ -1059,6 +1059,12 @@ class ExprVisitor(VyperNodeVisitorBase):
                     except VyperException:
                         has_nested_unbounded = False
                     else:
+                        if arg_typ.has_wildcard:
+                            # a wildcard call return resolves to the parameter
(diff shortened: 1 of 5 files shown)


charles-cooper (approved):
lgtm pending review from @Sporarum (and usual clanker review)

vyper/semantics/analysis/local.py
```diff
@@ -79,15 +79,30 @@
...
     try:
-        return type_contains_unbounded_sequence(get_exact_type_from_node(node))
+        actual_typ = get_exact_type_from_node(node)
     except VyperException:
         return False
 
+    if typ.has_wildcard:
+        # resolve a wildcard call return against the element's expected
+        # type, the same way `ExprVisitor.visit_Call` does for arguments
+        actual_typ = actual_typ.resolve_wildcard()
```
Sporarum, line 103:
```suggestion
    actual_typ = actual_typ.resolve_wildcard()
```
`resolve_wildcard` is a no-op on a type which does not contain wildcards

vyper/semantics/analysis/local.py
```diff
@@ -1052,13 +1067,19 @@ def visit_Call(self, node: vy_ast.Call, typ: VyperType) -> None:
...
                         actual_arg_typ = get_exact_type_from_node(arg)
                     except VyperException:
                         has_nested_unbounded = False
                     else:
+                        if arg_typ.has_wildcard:
+                            # a wildcard call return resolves to the parameter
+                            # type, unless the parameter is itself a wildcard,
+                            # in which case it resolves to INF (see the
+                            # external call handling below)
+                            actual_arg_typ = actual_arg_typ.resolve_wildcard()
```
Sporarum, line 1082: Same as above, the branch is unnecessary

harkal: The check is on the parameter type (`typ` and `arg_typ`), not on the type being resolved. If the parameter is bounded, a wildcard return takes the parameter's type, so resolving it to INF here would be wrong. Without the check, two cases in `test_wildcard_tuple_arg_roundtrip` fail. Same for above.

Sporarum: Oh my bad ...

Makes sense then !

vyper/semantics/analysis/local.py
Sporarum: See https://github.com/vyperlang/vyper/pull/5242#discussion_r3969099001

Sporarum (approved):
LGTM

HodanPlodky (approved):

PR #5248, opened 2026-09-05
fix[venom]: keep venom memory allocation order deterministic

### LINKED ISSUE #5245: Venom memory-allocation ties produce nondeterministic runtime bytecode with fixed PYTHONHASHSEED
### Version information

- Vyper: `0.5.0b2+commit.5764199a`, master at `5764199a9ab64e13bc64d8f41371378db4fa6ab0` (checked September 5, 2026).
- Also observed at the port's pinned compiler, `8af5e83c10af4f065cc660fe20293dac11fdff83`. The intervening commit changes only `.github/workflows/codeql.yml`.
- Python 3.13.14; macOS 26.6.2, arm64.
- Venom (`--experimental-codegen`), Cancun; bytecode metadata disabled.

### Observed behavior

Two fresh compiler processes with identical source, import path, optimization settings, and `PYTHONHASHSEED=0` produce different runtime bytecode. Metadata is disabled. This is a compilation-only reproducer; no contract execution is needed.

Source: [Aave V4 Vyper Spoke and its checked-in imports](https://github.com/aave/aave-v4/blob/1f7129042fe57e24bfc4c934be5088514cb93503/vyper/src/spoke/SpokeInstance.vy), fixed at `1f7129042fe57e24bfc4c934be5088514cb93503`.

Fresh O3 results against the master commit above (SHA-256 of decoded runtime bytecode):

| Process | Runtime bytes | SHA-256 |
|---|---:|---|
| 1 | 23,299 | `3a74132d1c7621d95384bf3a9bff895191397c315b91b0e64f78eb2500abe767` |
| 2 | 23,299 | `d94cb129eb5e259b4169f35bd6ac7934c324a01be4066f8d71dac845bd3415ae` |

Exact hashes may vary between reproductions. The failure is that they differ between otherwise identical fresh processes.

### Reproduction

```sh
git clone https://github.com/aave/aave-v4.git aave-v4-vyper-feedback
cd aave-v4-vyper-feedback
git checkout 1f7129042fe57e24bfc4c934be5088514cb93503
python3.13 -m venv .venv
.venv/bin/python -m pip install 'vyper @ git+https://github.com/vyperlang/vyper.git@5764199a9ab64e13bc64d8f41371378db4fa6ab0'
```

Run the following command twice, using a fresh process each time, saving the outputs as `run-1.txt` and `run-2.txt`:

```sh
PYTHONHASHSEED=0 PYTHONDONTWRITEBYTECODE=1 .venv/bin/vyper \
  -p vyper/src --evm-version cancun --experimental-codegen \
  --disable-bytecode-metadata -O 3 -f bytecode_runtime,ir_runtime \
  vyper/src/spoke/SpokeInstance.vy > run-1.txt
```

The first output line is runtime hex; the remaining output is IR. Compare the first lines and the IR. In the current rerun, for example, the same IR temporary `%192` is assigned memory offset `256` in one build and `416` in the other. Other concrete memory offsets change too.

The reproducer currently uses the full source module; it has not been reduced to a standalone small contract. It requires no Solidity compilation or submodule initialization.

### Narrowing the cause

The same behavior was investigated at the pinned compiler, whose compiler source is identical to current master:

1. [`Allocation`](https://github.com/vyperlang/vyper/blob/5764199a9ab64e13bc64d8f41371378db4fa6ab0/vyper/venom/memory_location.py#L10) is a frozen dataclass containing an `IRInstruction`. The instruction uses object-identity hashing, which a fixed `PYTHONHASHSEED` does not stabilize across processes.
2. `MemLivenessAnalysis` iterates ordinary sets, including `escaped` and the result of `_find_base_ptrs`, and transfers their order into ordered collections / the liveness mapping.
3. `ConcretizeMemLocPass` sorts allocations by `(not escaped, liveness-set length)`. Ties preserve the incoming order and can therefore receive different concrete memory locations.

A temporary, process-only diagnostic added the IR output-variable name as a third sort key. Four fresh pinned-compiler processes then produced identical 23,304-byte runtime output, SHA-256 `58bcca118ebc7141b4a388f189b3b4c5a5861d602419b9fe1279e1c17de161c5`.

[Diagnostic explanation and commands](https://github.com/aave/aave-v4/blob/1f7129042fe57e24bfc4c934be5088514cb93503/vyper/compiler-feedback/determinism/README.md), [process-only shim](https://github.com/aave/aave-v4/blob/1f7129042fe57e24bfc4c934be5088514cb93503/vyper/compiler-feedback/determinism/compile.py), [recorded hashes](https://github.com/aave/aave-v4/blob/1f7129042fe57e24bfc4c934be5088514cb93503/vyper/compiler-feedback/determinism/results.json), [pinned-compiler IR diff](https://github.com/aave/aave-v4/blob/1f7129042fe57e24bfc4c934be5088514cb93503/vyper/compiler-feedback/determinism/ir-diff.txt).

This isolates allocation tie ordering for this case. The diagnostic is not a proposed, validated production fix, and differing bytecode alone does not establish differing contract semantics. A compiler fix should define stable allocation ordering and include fresh-process reproducibility checks plus the normal semantic and allocation regressions.

### Expected behavior

The same source and compiler settings should produce identical runtime bytecode across fresh processes. Reproducibility matters for artifact verification and reliable gas/size comparisons.

harkal:
### What I did

Fixes https://github.com/vyperlang/vyper/issues/5245

### Commit message

```
use `OrderedSet` so `Ptr` and `Allocation` that hash by `IRInstruction` identity
don't result in non-deterministic builds. 
```

### Cute Animal Picture

### DIFF
--- a/tests/functional/venom/test_memory_allocation_stability.py
+++ b/tests/functional/venom/test_memory_allocation_stability.py
@@ -1,0 +1,62 @@
+import os
+import subprocess
+import sys
+
+# the pointers of the `Bytes` locals escape into `self._f`, and each
+# if/else arm gives its local the same live range. the memory allocator
+# must resolve such ties by program order, not by object-identity hash
+# order, which differs between processes.
+CONTRACT = """
+@internal
+def _f(a: Bytes[64]) -> uint256:
+    s: uint256 = 0
+    for i: uint256 in range(len(a), bound=64):
+        s += convert(slice(a, i, 1), uint256)
+    return s
+
+@external
+def foo(x: Bytes[64], y: Bytes[64], flags: bool[4]) -> uint256:
+    n: uint256 = 0
+    if flags[0]:
+        a0: Bytes[64] = x
+        n += self._f(a0)
+    else:
+        b0: Bytes[64] = y
+        n += self._f(b0)
+    if flags[1]:
+        a1: Bytes[64] = x
+        n += self._f(a1)
+    else:
+        b1: Bytes[64] = y
+        n += self._f(b1)
+    if flags[2]:
+        a2: Bytes[64] = x
+        n += self._f(a2)
+    else:
+        b2: Bytes[64] = y
+        n += self._f(b2)
+    if flags[3]:
+        a3: Bytes[64] = x
+        n += self._f(a3)
+    else:
+        b3: Bytes[64] = y
+        n += self._f(b3)
+    return n
+"""
+
+
+def test_memory_allocation_stability(tmp_path):
+    # object-identity hashes are not fixed by PYTHONHASHSEED, so the
+    # bytecode must be compared across fresh compiler processes. varying
+    # the seed changes allocation patterns and thereby object addresses.
+    path = tmp_path / "foo.vy"
+    path.write_text(CONTRACT)
+
+    outputs = set()
+    for seed in range(8):
+        env = {**os.environ, "PYTHONHASHSEED": str(seed)}
+        cmd = [sys.executable, "-m", "vyper", "--experimental-codegen", "-f", "bytecode_runtime"]
+        res = subprocess.run(cmd + [str(path)], env=env, capture_output=True, text=True, check=True)
+        outputs.add(res.stdout.strip())
+
+    assert len(outputs) == 1, f"{len(outputs)} distinct outputs across processes"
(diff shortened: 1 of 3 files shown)


charles-cooper (approved):
lgtm, thanks!

PR #5249, opened 2026-09-07
feat[venom]: invalidate only aliased loads at fixed-size memory copies

harkal:
### What I did

`LoadAnalysis` used to forget every forwardable store as soon as it saw any memory write
that was not an `mstore`, including `mcopy`, `codecopy`, `calldatacopy`, `dloadbytes`,
`returndatacopy` and `extcodecopy`. For a copy with a known destination and a literal
size, it now removes only the entries that may alias the written interval, the same way
the `mstore` branch already does. Copies with an unknown destination or size still clear
everything.

### How I did it

One new branch in `LoadAnalysis` next to the existing `mstore` case: look up the
instruction's write location through `BasePtrAnalysis`; if it is fixed, call
`remove_aliases` on it, otherwise fall back to `clear()`. The set of opcodes is the
memory-copy instructions whose only write effect is `MEMORY`.

### Commit message

```
invalidate only aliased loads at fixed-size memory copies
```

### DIFF
--- a/vyper/venom/analysis/load_analysis.py
+++ b/vyper/venom/analysis/load_analysis.py
@@ -21,7 +21,7 @@ from .mem_alias import (
 
 _LatticeKey = IROperand | MemoryLocation
 
-# bulk memory writes whose destination is known when their size is a literal
+# bulk memory writes whose destination operand names the written location
 _MEMORY_COPY_OPS = frozenset(
     ["mcopy", "calldatacopy", "codecopy", "dloadbytes", "returndatacopy", "extcodecopy"]
 )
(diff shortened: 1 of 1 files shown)


vyper/venom/analysis/load_analysis.py
```diff
@@ -226,6 +231,14 @@ def get_write(self, inst: IRInstruction) -> IROperand | MemoryLocation:
                 return memloc
         return self._normalize_operand(inst.operands[1])
 
+    def _fixed_copy_location(self, inst: IRInstruction, eff: Effects) -> Optional[MemoryLocation]:
+        if eff != Effects.MEMORY or inst.opcode not in _MEMORY_COPY_OPS:
+            return None
+        memloc = self.base_ptrs.get_write_location(inst, self.space)
+        if not memloc.is_fixed:
+            return None
+        return memloc
```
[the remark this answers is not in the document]
harkal: Fixing the pre-existing at the proper level is here: https://github.com/vyperlang/vyper/pull/5250

vyper/venom/analysis/load_analysis.py
```diff
@@ -247,7 +260,17 @@ def _handle_bb(
                 lattice.remove_aliases(memloc, alias_set)
                 lattice[ptr] = OrderedSet([val])
             elif isinstance(eff, Effects) and eff in inst.get_write_effects():
-                lattice.clear()
+                # a fixed-size memory copy is a store to a known location:
+                # drop what it may alias, exactly like an mstore. anything
+                # else (calls, unknown sizes) clobbers everything
+                copy_loc = self._fixed_copy_location(inst, eff)
```
HodanPlodky: the mem_alias set for the for non fixed mem location should already contain all of the mem location that should be removed so it should be possible to only check for the viable opcodes and do not do the fixed check.
It could improve the cases where it is not fixed but it is in the mem from alloca which could just invalidate the mem locs from the same allocation and not the full memory

HodanPlodky (changes requested):

vyper/venom/analysis/load_analysis.py
harkal: Right, it's going to need the mixed root PR too. https://github.com/vyperlang/vyper/pull/5250

HodanPlodky (approved):

PR #5252, opened 2026-09-09
fix[venom]: make `BranchOptimizationPass` not dependant on `SCCP`

HodanPlodky:
### What I did
Implementation of the fix that was shown by the https://github.com/vyperlang/vyper/pull/5219.

In the case that the branch has a literal as a condition (this happens when SCCP is not run) this pass failed on assertion. If thats the case short circuit.

### Commit message

```
Removed the assumption from branch prediction that the 
condition is variable, which is only true if the SCCP run 
before hand. This commits fixes that by skipping the 
optimization if the condition is not variable.
```

### Cute Animal Picture


harkal (approved):

harkal:
@HodanPlodky needs a squash commit message

PR #5254, opened 2026-09-09
fix[venom]: retpc_param handle in inliner fix

HodanPlodky:
### What I did
Implementation of the fix that was shown by the https://github.com/vyperlang/vyper/pull/5219.

`retpc_param` should not be needed after if function inline. There was an error where because if was still there the label that does no longer exist was still referenced in the code.

### Commit message

```
This commit removes the retpc_param from the body of the function if it is inlined. This could create an error if the label would survived up until the bytecode emit it created the CompilerPanic
```

### Cute Animal Picture


harkal (changes requested):
codex comments seem legit, all of those are valid raw IR that master inlines fine and now trip the asserts (retpc through a phi and no retpc/no ret too). fixture changes look like they're just there to dodge the assert

wouldn't it be simpler to just keep master's code and not bind retpc to the label?

`binding_ops = (*binding_ops[:-1], IRLiteral(0))`

your new test passes with that, no fixture changes needed, bytecode identical on everything I tried. the noinline ret fix is good tho, I'd keep that

PR #5256, opened 2026-09-09
fix[venom]: widen tuples by memory layout, not type equality

harkal:
### What I did

Venom-only miscompile: assigning a tuple to a tuple with wider members (e.g. `x: (Bytes[10], uint256)`, `y: (Bytes[40], uint256) = x`) copied `x` as a flat block, so the second member of `y` sits at the wrong offset and is read back wrong. Same for storage/transient/immutable tuple writes and for internal returns of tuples that have unbounded members. Legacy codegen (`make_setter`) is correct.

Fixes the venom side of the tuple widening path that #5236 relies on.

### How I did it

- `VenomCodegenContext.store_memory`, `Stmt._store_complex_type` and `Stmt._emit_dynamic_tuple_internal_return` chose between a flat copy and the member-wise copy (`_store_memory_typed`) with `src_typ != typ`. `TupleT.__eq__` compares `members`, which is never populated (`_equality_attrs = ("members",)`), so any two tuple types compare equal and the member-wise path never fired for tuples.
- New `same_memory_layout(src_typ, dst_typ)` in `codegen_venom/context.py` = `punnable(src_typ, dst_typ) and src_typ.memory_bytes_required == dst_typ.memory_bytes_required`; the three sites gate on it. `punnable` alone (`vyper/codegen/core.py`, from #5236) is not enough: it accepts a wider top-level DynArray capacity or bytestring bound in the destination (the data present has the same layout), but a flat copy of the destination size would read past the source. The size conjunct keeps every such case on the copy path it takes today (`_copy_dynarray_memory_typed` / the bytestring branch) and leaves all non-tuple codegen unchanged.
- `ensure_memory_layout` now falls back to `materialize_value` (`store_vyper_value` -> `store_memory`) for bounded types; the comment that justified bypassing it was exactly this broken gate. The unbounded-sequence route added in #5240 is unchanged.
- In `_emit_dynamic_tuple_internal_return` the `type_contains_unbounded_sequence` checks are evaluated before `same_memory_layout`, since unbounded members have no `memory_bytes_required`.
- Left as is (identity checks, or unreachable for tuples): `_copy_dynarray_memory_typed`'s fast path (elements are never tuples), `_lower_dynarray_append`, `_lower_external_return`'s `ret_src_typ != ret_typ` (`can_encode_from_src` is True first for tuples), `builtins/abi.py`'s `wrapped_typ != output_typ`.
- Not changed: `TupleT.__eq__` itself. Fixing it in semantics touches type-checker equality; separate PR if wanted.

### Commit message

```
gate the flat-copy fast path in `store_memory`, `_store_complex_type`
and `_emit_dynamic_tuple_internal_return` on a new `same_memory_layout`
predicate (`punnable` plus equal `memory_bytes_required`) instead of
`src_typ != typ`, which is always False for tuples because
`TupleT.__eq__` compares the never-populated `members` dict. widened
tuple copies (e.g. `(Bytes[10], uint256)` into `(Bytes[40], uint256)`)
now take the member-wise path.

`ensure_memory_layout` falls back to `materialize_value` for bounded
types now that `store_memory` gates correctly. add runtime tests for
widened tuples through locals, returns, internal-call arguments, struct
members, nested tuples, DynArray members and state variables.
```

### Cute Animal Picture


vyper/codegen_venom/context.py
```diff
@@ -46,6 +46,22 @@ class Constancy(Enum):
...
+    laid out as `src_typ` form a valid `dst_typ` value.
+
+    `punnable` accepts a wider top-level DynArray capacity or bytestring
+    bound in `dst_typ` (the data present has the same layout), but a flat
+    copy of the destination size would then read past the source.
+
+    Not `src_typ != dst_typ`: `TupleT` compares by its never-populated
+    `members` dict, so any two tuple types are equal.
+    """
+    return punnable(src_typ, dst_typ) and (
```
charles-cooper: bit ugly, please either split into multiple fast-pathing statements if applicable or use temporary variables

vyper/codegen_venom/context.py
```diff
@@ -48,6 +48,22 @@ class Constancy(Enum):
     Constant = 1
 
 
+def same_memory_layout(src_typ: VyperType, dst_typ: VyperType) -> bool:
```
HodanPlodky, line 51: Negation of this should be a super set of `src_typ != dst_typ` right? meaning if the is true that should imply that the types at the point of the call should be same (since it already passed through type analysis) or am I missing something? 

Why I though that should be the case is because there should not be the case where the this is `True` and the types are no the same it would be the case that the we could do the widening with old condition but not with the new.

Thats not the case though this does not have that property
```
@external
def foo() -> String[32]:
    x: String[32] = "tres bien!"
    return x
```

Is that some issue or is that not concern?

HodanPlodky (commented):
I just have one question about corner case (if it is ever relevant)

vyper/codegen_venom/context.py
harkal: Not a concern. It's `True` for differing types only when the layouts are byte-identical (`String[10]` and `String[32]`, both 64 bytes), so a flat copy is already a valid `dst_typ` value and there's nothing to widen. The old != just took the typed copy unnecessarily. (Top-level bytestrings like your example don't reach this check. it matters for nested ones, like `DynArray[String[10], 2]` -> `DynArray[String[32], 2]`)

HodanPlodky (approved):

Issue #4733, opened 2025-08-15
unhandled internal error


logorithm360:
### Version Information

* **Vyper Version** (output of `vyper --version`): x.x.x
* **OS**: Linux
* **Python Version** (output of `python --version`): x.x.x

#### Error Output
```bash
## Error compiling: morning_contract.vy  
---AttributeError: type object 'StringT' has no attribute 'typ'

---During handling of the above exception, another exception occurred:

---vyper.exceptions.CompilerPanic: unhandled exception type object 'StringT' has no attribute 'typ'

 - contract "morning_contract.vy:11", line 11:0 
       10
  ---> 11 message_store: String[String[1000]]
  --------^
       12
##This is an unhandled internal compiler error. Please create an issue on Github to notify the developers!  
##https://github.com/vyperlang/vyper/issues/new?template=bug.md
```

#### Command Run
```bash
vyper morning_contract.vy
```

#### Problematic Code
```vyper
# pragma version ^0.4.0
# @license MIT

## struct for user info
user_info: HashMap[String[50], uint256]

struct User:
    name: String[50]
    message: String[100]

message_store: String[100][5] # store messages in a list of strings

## initializing the contract
@deploy
def __init__(_name: String[60], message: String[100], message_store: String[10][5], user_info: HashMap(String[50], uint256)):
    """
    Constructor to initialize the contract with a user's name and message.
    """
    self.person: User = User(
        name= _name,
        message = message
    )
    self.message_store = message_store
    self.user_info = user_info
    self.user_info[_name] = 0  # Initialize user info with zero messages
```

#### Verbose Output
```bash
vyper.exceptions.VyperException: Compilation failed with the following errors:

StructureException: arrays of String[10] are not allowed!

  contract "morning_contract.vy:15", line 15:0 
       14 @deploy
  ---> 15 def __init__(_name: String[60], message: String[100], message_store: String[10][5], user_info: HashMap(String[50], uint256)):
  --------^
       16     """


StructureException: arrays of String[100] are not allowed!

  contract "morning_contract.vy:11", line 11:0 
       10
  ---> 11 message_store: String[100][5] # store messages in a list of strings
  --------^
       12
```

### How can it be fixed?
I don't know yet - can you help fixing it?

PR #5257, opened 2026-09-09
fix[lang]: reject parameterized types used as values

### LINKED ISSUE #3246: `AttributeError` when using types as argument
### Version Information

* vyper Version (output of `vyper --version`): 0.3.8+commit.4ae2527f
* OS: OSX

### What's your issue about?

The compiler outputs an `AttributeError` when trying to use a type as an argument to a function for the following types while it would return a `InvalidReference` for other types:

- Bytes Arrays
- Strings
- Dynamic Arrays
- HashMaps

For example:
```Vyper
@internal
def bar(a:uint256):
    pass

@external
def foo():
    self.bar(String[5])
```

```Bash
  File "/Library/Frameworks/Python.framework/Versions/3.8/bin/vyper", line 8, in <module>
    sys.exit(_parse_cli_args())
    compiled = compile_files(
    return fn(*args, **kwargs)
    raise exception
  File "/Users/trocher/Documents/thesis/vyper/vyper/compiler/output.py", line 248, in build_bytecode_output
    val = self.func(instance)
  File "/Users/trocher/Documents/thesis/vyper/vyper/compiler/phases.py", line 150, in bytecode
    self.assembly, is_runtime=False, no_bytecode_metadata=self.no_bytecode_metadata
    val = self.func(instance)
  File "/Users/trocher/Documents/thesis/vyper/vyper/compiler/phases.py", line 141, in assembly
    val = self.func(instance)
  File "/Users/trocher/Documents/thesis/vyper/vyper/compiler/phases.py", line 117, in global_ctx
    return GlobalContext(self.vyper_module_folded)
  File "/Users/trocher/Documents/thesis/vyper/vyper/compiler/phases.py", line 107, in vyper_module_folded
    module, storage_layout = self._folded_module
    val = self.func(instance)
  File "/Users/trocher/Documents/thesis/vyper/vyper/compiler/phases.py", line 101, in _folded_module
    return generate_folded_ast(
  File "/Users/trocher/Documents/thesis/vyper/vyper/compiler/phases.py", line 229, in generate_folded_ast
    validate_semantics(vyper_module_folded, interface_codes)
  File "/Users/trocher/Documents/thesis/vyper/vyper/semantics/analysis/__init__.py", line 13, in validate_semantics
    validate_functions(vyper_ast)
  File "/Users/trocher/Documents/thesis/vyper/vyper/semantics/analysis/local.py", line 59, in validate_functions
    FunctionNodeVisitor(vy_module, node, namespace)
  File "/Users/trocher/Documents/thesis/vyper/vyper/semantics/analysis/local.py", line 179, in __init__
    self.visit(node)
  File "/Users/trocher/Documents/thesis/vyper/vyper/semantics/analysis/local.py", line 222, in visit
    super().visit(node)
  File "/Users/trocher/Documents/thesis/vyper/vyper/semantics/analysis/common.py", line 20, in visit
    visitor_fn(node, *args)
  File "/Users/trocher/Documents/thesis/vyper/vyper/semantics/analysis/local.py", line 508, in visit_Expr
    return_value = fn_type.fetch_call_return(node.value)
  File "/Users/trocher/Documents/thesis/vyper/vyper/semantics/types/function.py", line 475, in fetch_call_return
    validate_expected_type(arg, expected)
  File "/Users/trocher/Documents/thesis/vyper/vyper/semantics/analysis/utils.py", line 492, in validate_expected_type
    given_types = _ExprAnalyser().get_possible_types_from_node(node)
  File "/Users/trocher/Documents/thesis/vyper/vyper/semantics/analysis/utils.py", line 143, in get_possible_types_from_node
    ret = fn(node)
  File "/Users/trocher/Documents/thesis/vyper/vyper/semantics/analysis/utils.py", line 338, in types_from_Subscript
    t = self.get_exact_type_from_node(node.value)
  File "/Users/trocher/Documents/thesis/vyper/vyper/semantics/analysis/utils.py", line 116, in get_exact_type_from_node
    types_list = self.get_possible_types_from_node(node, include_type_exprs=include_type_exprs)
  File "/Users/trocher/Documents/thesis/vyper/vyper/semantics/analysis/utils.py", line 143, in get_possible_types_from_node
    ret = fn(node)
  File "/Users/trocher/Documents/thesis/vyper/vyper/semantics/analysis/utils.py", line 324, in types_from_Name
    return [t.typ]
AttributeError: type object 'StringT' has no attribute 'typ'
```

### How can it be fixed?

This seems to happen when trying to get the type of the `value` of the Subscript representing the type. The `isinstance` test shown below returns `False` as `t` is not an instance of `VyperType` but a class inheriting it (In the case above: `<class 'vyper.semantics.types.bytestrings.StringT'>`.

https://github.com/vyperlang/vyper/blob/02339dfda0f3caabad142060d511d10bfe93c520/vyper/semantics/analysis/utils.py#L314-L322

### LINKED ISSUE #4434: AttributeError: type object 'DArrayT' has no attribute 'typ'
### Version Information

* vyper Version (output of `vyper --version`): 0.4.1b2+commit.5f04c0e22

### What's your issue about?
was testing some ugly code and this crashes the compiler
```vyper
MAX: constant(uint256) = DynArray[uint256, 10]
```
```
AttributeError: type object 'DArrayT' has no attribute 'typ'

vyper.exceptions.CompilerPanic: unhandled exception type object 'DArrayT' has no attribute 'typ'

  contract "tests/custom/test.vy:1", line 1:0 
  ---> 1 MAX: constant(uint256) = DynArray[uint256, 10]
  -------^


```

beardthelion:
### What I did

Fixes #3246, fixes #4434. Also covers #4733.

Using a parameterized type (`Bytes`, `String`, `DynArray`, `HashMap`) anywhere a value is expected crashes the compiler instead of reporting a bad reference:

```vyper
@external
def foo():
    x: uint256 = Bytes[32]   # CompilerPanic: type object 'BytesT' has no attribute 'typ'
```

### How I did it

Concrete types are in the namespace as instances (`uint256` is an `IntegerT`), but parameterized types are there as the type *class* (`Bytes` is `BytesT` itself). `_ExprAnalyser.types_from_Name` returns `[t]` for a `TYPE_T` and `[TYPE_T(t)]` for a `VyperType` instance; a class matches neither, so it fell through to the `return [t.typ]` fallback meant for `VarInfo` and read a class attribute that does not exist. The resulting `AttributeError` is not a `VyperException`, so it escaped the check that turns a type expression in value position into `InvalidReference`, and surfaced as a panic.

The fix rejects a type class there with the same `InvalidReference` a concrete type name already raises.

One follow-on: `get_index_value` inspects the subscript first to produce a better message, but caught only `StructureException`, so the new `InvalidReference` escaped it and `Bytes[String]` reported differently from `Bytes[uint256]`. It now catches both, so the annotation-position message is consistent. It is deliberately not widened to `VyperException`, which would swallow `UndeclaredDefinition` and lose the better message for `uint256[NOPE]`.

### How to verify it

Every position below panics on master and raises a clean error with this patch:

```vyper
x: uint256 = Bytes[32]                  # assignment
return Bytes[32]                        # return
abi_encode(Bytes[32])                   # builtin vararg
concat(b"a", Bytes[32])                 # a different vararg builtin
raw_call(t, b"", max_outsize=Bytes[32]) # builtin kwarg
a[Bytes[32]]                            # subscript index
DynArray.foo                            # attribute base
def foo(a: uint256 = Bytes[32]):        # default argument
S(a=Bytes[32])                          # struct kwarg
log E(a=Bytes[32])                      # event kwarg
MAX: constant(uint256) = DynArray[u, 1] # module scope
initializes: HashMap                    # module level statement
x: Bytes[String] = b""                  # annotation subscript
```

I ran a 57-case matrix over the base and the patched tree, each case in a fresh interpreter since the namespace is global: 34 panics before, 0 after, and 22 cases byte-identical either way. Those 22 are 12 legitimate uses that must keep compiling (`empty(Bytes[32])`, `abi_decode(b, Bytes[32])`, `convert(x, String[10])`, annotations, storage `HashMap`, `public(DynArray)`, struct members, nested `DynArray`) and 10 diagnostics that must not change (bare `Bytes`/`DynArray`/`HashMap` annotations keep their own messages, `uint256[NOPE]` keeps `UndeclaredDefinition`, `Bytes[uint256]` keeps `InvalidType`, structs/flags/concrete types keep theirs).

Regression tests go in the existing exception test files. The negative direction is covered too: three tests that unparameterized annotations keep their specific messages, and one that an undeclared name in a subscript still reports `UndeclaredDefinition`. I checked each new test is load-bearing by reverting each change in turn: removing the analysis guard turns 19 red, removing the `get_index_value` catch turns exactly 1 red, and widening that catch to `VyperException` turns exactly 1 red.

Full suite: 13009 passed, 0 failed.

Note for whoever sequences this: #5204 rewrites this function, and I confirmed the panic still reproduces on its head (d5d5bfd6), so the two will conflict on the rename but the bug is not fixed there.

### Commit message

```
Parameterized types (`Bytes`, `String`, `DynArray`, `HashMap`) live in
the namespace as the type class, while concrete types like `uint256`
live there as instances. `types_from_Name` matched neither the TYPE_T
branch nor the VyperType branch for a class, so it fell through to the
`return [t.typ]` fallback meant for VarInfo and read a class attribute
that does not exist. The resulting AttributeError is not a
VyperException, so it escaped the check that turns a type expression in
value position into InvalidReference, and surfaced as an internal
compiler panic instead.

Reject a type class there with the same InvalidReference a concrete type
name already raises. The panic was reachable from every value position:
assignment, return, builtin argument and kwarg, module-scope constant,
struct and event kwargs, default arguments, and the index of another
subscript.

`get_index_value` inspects the subscript first to give a better
diagnostic, but caught only StructureException, so a parameterized type
name in an annotation subscript escaped it. Catch InvalidReference there
as well, so `Bytes[String]` reports the same "Subscript must be a
literal integer" that `Bytes[uint256]` already did. It is deliberately
not widened to VyperException, which would swallow UndeclaredDefinition
and lose the better message for `uint256[NOPE]`.

Fixes GH 3246 and GH 4434
```

Reject a parameterized type name used as a value (e.g. `x: uint256 = Bytes[32]`) with a clear error instead of an internal compiler panic.

### Cute Animal Picture

### DIFF
--- a/tests/functional/syntax/exceptions/test_structure_exception.py
+++ b/tests/functional/syntax/exceptions/test_structure_exception.py
@@ -114,3 +114,18 @@ interface Bar:
         arg1 = 3
     """,
 ]
+
+
+# a parameterized type name is only a bad reference in value position. in type
+# position it keeps its own message, which says how to parameterize it
+@pytest.mark.parametrize(
+    "bad_code,expected",
+    [
+        ('@external\ndef foo():\n    x: Bytes = b""\n', "without a maximum length"),
+        ("@external\ndef foo():\n    x: DynArray = []\n", "base type and max length"),
+        ("m: HashMap\n", "key type and a value type"),
+    ],
+)
+def test_unparameterized_type_annotation(bad_code, expected):
+    with pytest.raises(StructureException, match=expected):
+        compiler.compile_code(bad_code)
--- a/tests/functional/syntax/exceptions/test_invalid_reference.py
+++ b/tests/functional/syntax/exceptions/test_invalid_reference.py
@@ -44,6 +44,107 @@ a: public(constant(uint256)) = 1
 def foo():
     b: uint256 = self.a
     """,
+    # parameterized types live in the namespace as the type class rather than an
+    # instance, so they used to miss the TYPE_T check and panic on `t.typ`
+    # instead of reporting a bad reference
+    # see https://github.com/vyperlang/vyper/issues/3246
+    """
+@external
+def foo():
+    x: uint256 = Bytes[32]
+    """,
+    """
+@external
+def foo():
+    x: uint256 = String[10]
+    """,
+    """
+@external
+def foo():
+    x: uint256 = DynArray[uint256, 3]
+    """,
+    """
+@external
+def foo():
+    x: uint256 = HashMap[uint256, uint256]
+    """,
+    """
+@external
+def foo() -> uint256:
+    return Bytes[32]
+    """,
+    """
+@external
+def foo() -> Bytes[64]:
+    return abi_encode(Bytes[32])
+    """,
+    # module scope reaches the same guard through constant folding rather than
+    # function body analysis
+    # see https://github.com/vyperlang/vyper/issues/4434
+    """
+MAX: constant(uint256) = DynArray[uint256, 10]
+    """,
+    # an unsubscripted name arrives at types_from_Name directly instead of
+    # through types_from_Subscript
+    """
+@external
+def foo():
+    x: uint256 = DynArray
+    """,
+    # as the index of another subscript
+    # see https://github.com/vyperlang/vyper/issues/4733
+    """
+@external
+def foo(a: uint256[3]) -> uint256:
+    return a[Bytes[32]]
+    """,
+    # attribute access is the one route that asks for types with
+    # include_type_exprs enabled
+    """
+@external
+def foo():
+    x: uint256 = DynArray.foo
+    """,
+    # module level statements resolve names without entering a function body
+    """
+initializes: HashMap
+    """,
+    """
+uses: String
+    """,
+    """
+exports: DynArray
+    """,
+    # a default argument is analyzed with the signature, not the body
+    """
+@external
+def foo(a: uint256 = Bytes[32]):
+    pass
+    """,
+    # keyword arguments take their own validation route
+    """
+struct S:
+    a: uint256
+
+@external
+def foo():
+    s: S = S(a=Bytes[32])
+    """,
+    """
+event E:
+    a: uint256
+
+@external
+def foo():
+    log E(a=Bytes[32])
+    """,
+    # a vararg builtin other than abi_encode, to show the guard is not specific
+    # to one builtin's argument handling
+    """
+@external
+def foo() -> Bytes[64]:
+    return concat(b"a", Bytes[32])
+    """,
 ]
 
 
@@ -51,3 +152,16 @@ def foo():
 def test_invalid_reference_exception(bad_code):
     with pytest.raises(InvalidReference):
         compiler.compile_code(bad_code)
+
+
+def test_parameterized_type_message():
+    # the guard reuses the wording concrete type names already get, so a
+    # regression that routed one of these through a different path would
+    # still raise InvalidReference but with different text
+    code = """
+@external
+def foo():
+    x: uint256 = DynArray[uint256, 3]
+    """
+    with pytest.raises(InvalidReference, match="not a variable or literal: 'DynArray'"):
+        compiler.compile_code(code)
--- a/tests/functional/syntax/exceptions/test_invalid_type_exception.py
+++ b/tests/functional/syntax/exceptions/test_invalid_type_exception.py
@@ -145,3 +145,15 @@ def foo():
     """
     with pytest.raises(InvalidType, match="is not a type"):
         compile_code(code)
+
+
+def test_parameterized_type_as_subscript():
+    # a parameterized type name as a subscript is the same user mistake as any
+    # other non-integer subscript, so it gets the same message
+    code = """
+@external
+def foo():
+    x: Bytes[String] = b""
+    """
+    with pytest.raises(InvalidType, match="Subscript must be a literal integer"):
+        compile_code(code)
--- a/tests/functional/syntax/exceptions/test_undeclared_definition.py
+++ b/tests/functional/syntax/exceptions/test_undeclared_definition.py
@@ -69,3 +69,15 @@ def test_undeclared_def_exception(bad_code):
     with pytest.raises(UndeclaredDefinition) as e:
         compiler.compile_code(bad_code)
     assert "(hint: )" not in str(e.value)
+
+
+def test_undeclared_name_as_subscript():
+    # get_index_value also catches InvalidReference so a parameterized type name
+    # reports as a bad subscript. that must not hide an undeclared name
+    code = """
+@external
+def foo():
+    x: uint256[NOPE] = [0]
+    """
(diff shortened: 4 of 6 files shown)


charles-cooper (commented):
fix looks a little odd, but the mechanism is a little odd to begin with. maybe @Sporarum you have some ideas?

beardthelion:
> fix looks a little odd, but the mechanism is a little odd to begin with

Agreed on the mechanism. `_get_primitive_types` and `_get_sequence_types` deliberately put the class in the namespace for the parametrizable types ("since bytestrings are parametrizable, the *class* objects are in the namespace instead of concrete type objects"), so `types_from_Name` meets a class where every other name gives it an instance, and both isinstance branches miss.

I tried the tidier version first: return `TYPE_T(t)` and let the existing type-expr check at the top of `get_possible_types_from_node` do the rejecting, rather than raising a second time further down. Two things went wrong with it.

- `exports: DynArray` still panics, with `CompilerPanic: unreachable! <class DArrayT>` from `get_expr_info`.
- the message renders the typedef, which for a class is the Python repr: `not a variable or literal: '<class 'vyper.semantics.types.bytestrings.BytesT'>'`.

Fixing both means teaching the shared error path to print `_id` and handling the class in `get_expr_info` as well, which is more surface than the guard is worth.

The version that actually removes the asymmetry is to stop putting classes in the namespace, but that reaches into annotation handling and seemed like a lot to attach to a panic fix. Happy to go that way if you'd prefer it.

One note for sequencing: #5204 renames `types_from_Name` to `visit_Name` and carries the `return [t.typ]` fallthrough over unchanged, so the panic still reproduces on its head (d5d5bfd6). Whichever lands first, the other needs a small rebase.

tests/functional/syntax/exceptions/test_structure_exception.py
```diff
@@ -114,3 +114,18 @@ def set_lucky(arg1: int128):
...
+    "bad_code,expected",
+    [
+        ('@external\ndef foo():\n    x: Bytes = b""\n', "without a maximum length"),
+        ("@external\ndef foo():\n    x: DynArray = []\n", "base type and max length"),
+        ("m: HashMap\n", "key type and a value type"),
+    ],
+)
+def test_unparameterized_type_annotation(bad_code, expected):
+    with pytest.raises(StructureException, match=expected):
+        compiler.compile_code(bad_code)
```
Sporarum, line 131: Nice thinking!

Is this not already tested elsewhere however ?

Sporarum (approved):
LGTM

Sporarum:
> fix looks a little odd, but the mechanism is a little odd to begin with. maybe @Sporarum you have some ideas?

Yeah I think this is good as-is

We need to tidy the "namespace contains both (type) instances and (type) types" problem at some point, but this PR seems orthogonal and sensible

PR #5267, opened 2026-09-12
feat[venom]: literal bytestring codecopy

harkal:
### What I did

Emit constant `Bytes`/`String` literals in `codegen_venom` as a `codecopy` from a
data section instead of a chain of `mstore`s, when a code size estimate says the
copy is smaller. Supersedes #3501 (legacy, closed) and the venom-pass approach
discussed there.

### How I did it

`vyper/codegen_venom/bytestring_literal.py` prices the two forms:

- the mstore chain, using the cheapest literal encoding the current optimization
  level can produce — `ReduceLiteralsCodesize` runs only in the O3 and Os pass
  lists, so PUSH+NOT / PUSH+SHL are priced only there, plain PUSHn otherwise;
- the codecopy, as the data item plus PUSH1 size / PUSH1 dst / PUSH2 label /
  SWAP1 / CODECOPY.

Both estimates are biased towards the chain, so a mispredicted case can only fall
back to today's code. Literals of one word or less never convert.

`const_bytestring_value` then emits either form. The data item is word-padded at
gas levels, and exact plus one `mstore 0` over the last word at codesize levels
(32 bytes of data-section padding costs more than the 4-byte store, but the store
costs gas). Both forms leave memory bit-identical to the mstore chain, which is
required because word-rounded bulk copies (`store_storage`, `_copy_complex_type`)
propagate bytes past the string length into storage. The length store is emitted
after the copy so length `mload`s stay foldable.

`generate_deploy_venom` now appends the `runtime_begin` and `cbor_metadata`
sections after the constructor is lowered, because lowering can now append data
sections of its own and the metadata must stay at the end of the initcode.

Known limitation: if the literal is dead, DSE drops the codecopy but the data item
stays in the bytecode.

### Commit message

```
every `Bytes`/`String` literal is materialized word by word, one push +
`mstore` per 32 bytes of payload, costing roughly 36 bytes of code per
32 bytes of data. copying the bytes out of a data section instead costs
the data plus ~9 bytes of fixed overhead, so from three dense words up
it is smaller and never more gas.

the gate is a code size estimate rather than a word count, because
`ReduceLiteralsCodesize` turns sparse words into cheap PUSH+NOT/PUSH+SHL
forms that still occupy a full 32 bytes in the data section, and a fixed
threshold regressed exactly the string-heavy contracts it was meant to
help; the estimate is biased towards the chain, so a wrong guess only
leaves today's code. the data item is word-padded at gas levels and
exact plus one `mstore 0` over the last word at codesize levels, both of
which leave memory bit-identical to the chain -- word-rounded bulk
copies propagate bytes past the length into storage, where they are
observable on-chain. the length store now comes after the copy so length
loads keep folding, and the deploy context appends
`runtime_begin`/`cbor_metadata` after lowering so the metadata stays at
the end of the initcode.
```

### Cute Animal Picture

### DIFF
--- a/vyper/codegen_venom/context.py
+++ b/vyper/codegen_venom/context.py
@@ -266,15 +266,17 @@ class VenomCodegenContext:
                 offset = self.builder.add(val.operand, IRLiteral(32 + i))
                 self.builder.mstore(offset, IRLiteral(word))
 
-        # the length store goes last: LoadAnalysis forgets every forwardable
-        # store at a non-mstore memory write, so a codecopy after it would
-        # keep the length from being constant-folded by its consumers
+        # the length store goes last so that loads of the length stay
+        # forwardable no matter how precisely LoadAnalysis models the copy
         self.ptr_store(val.ptr(), IRLiteral(len(data)))
 
         return val
 
     def _codecopy_bytestring_literal(self, ptr: IRVariable, data: bytes, padded: bool) -> None:
         """Copy the data of a constant bytestring from a new data section."""
+        # known limitation: if the literal turns out to be dead, dead store
+        # elimination removes the codecopy but the data item stays in the
+        # bytecode
         item = data
         if padded:
             item = data.ljust(ceil32(len(data)), b"\x00")
--- a/vyper/codegen_venom/context.py
+++ b/vyper/codegen_venom/context.py
@@ -19,7 +19,7 @@ from vyper.codegen.core import punnable
 from vyper.codegen_venom.buffer import Buffer, Ptr
 from vyper.codegen_venom.bytestring_literal import should_codecopy
 from vyper.codegen_venom.value import VyperValue
-from vyper.compiler.settings import _opt_codesize, _opt_lowering_only_ir
+from vyper.compiler.settings import _opt_codesize, _opt_lowering_only_ir, get_global_settings
 from vyper.evm.opcodes import version_check
 from vyper.exceptions import CompilerPanic, MemoryAllocationException, StateAccessViolation
 from vyper.semantics.data_locations import DataLocation
@@ -39,8 +39,10 @@ from vyper.semantics.types.module import ModuleT
 from vyper.semantics.types.subscriptable import DArrayT, SArrayT
 from vyper.semantics.types.user import StructT
 from vyper.utils import IDENTITY_PRECOMPILE, ceil32
+from vyper.venom import OPTIMIZATION_PASSES
 from vyper.venom.basicblock import IRLabel, IRLiteral, IROperand, IRVariable
 from vyper.venom.builder import VenomBuilder
+from vyper.venom.passes import ReduceLiteralsCodesize
 
 
 class Constancy(Enum):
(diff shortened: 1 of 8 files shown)


charles-cooper:
Bit shocked there are almost no bytecode size changes

charles-cooper (approved):

harkal:
> Bit shocked there are almost no bytecode size changes

it's the padding that for gas opt is kept in the data, and for codesize the literal optimization again make is not be selected. Do we care if we pay a little gas extra gas in reverts? I don't think it's comon to be used in loops, right?

charles-cooper:
> > Bit shocked there are almost no bytecode size changes
> 
> it's the padding that for gas opt is kept in the data, and for codesize the literal optimization again make is not be selected. Do we care if we pay a little gas extra gas in reverts? I don't think it's comon to be used in loops, right?

yes i think we should definitely optimize for codesize at the cost of gas in reverts (revert can only happen once per call and it's the "non fast" path so i don't think people care very much about gas usage in that path

charles-cooper:
i would expect any string larger than 32 bytes to get optimized -- so either there are very few long strings in the sample data or there is sth wrong with the heuristic.

harkal:
> i would expect any string larger than 32 bytes to get optimized -- so either there are very few long strings in the sample data or there is sth wrong with the heuristic.

It's the sample data. Also, for 2-word strings the copy isn't smaller than the chain... :/

charles-cooper:
Hmm how about deduplication (interning)?

harkal:
Don't know if we are going to see change int he dataset we have, but I guess it makes sense. I would go with updating the heuristic based on the dub count too, if you think it's ok too.

vyper/codegen_venom/context.py
```diff
@@ -100,6 +102,13 @@ class VenomCodegenContext:
     # Range expression context - set to True when evaluating range/iterator expressions
     in_range_expr: bool = False
 
+    # set while lowering a revert reason or custom error arguments
+    on_revert_path: bool = False
+
+    # the literal data items of the code segment being lowered; None only
+    # for the dispatcher contexts, which lower no expressions
+    literal_pool: Optional[LiteralPool] = None
```
HodanPlodky, line 110: Why can't it default to the newly created `LiteralPool`? or is it just design decision?

vyper/codegen_venom/context.py
```diff
@@ -858,6 +883,15 @@ def range_scope(self):
         finally:
             self.in_range_expr = prev_value
 
+    @contextmanager
+    def revert_scope(self):
```
HodanPlodky, line 887: I looked at the `revert` emits and it looks all necessary location for this optimization are covered, that said it could be beneficial to wrap them also if that is possible for possible future optimizations that could benefit from `on_revert_path`. Might be out of scope though

HodanPlodky (approved):
Just two NITs otherwise looks good

vyper/codegen_venom/context.py
harkal: Design decision... the pool is owned and finalized by the code segment in `module.py`, so a default one would never be finalized so `None` plus the assert catches a missed wiring.

vyper/codegen_venom/context.py
harkal: Looks like it's possible to do (if I don't miss some case/place) I will note for afterwards!

PR #5271, opened 2026-09-16
feat[venom]: support `max_outsize=INF` in raw_call

### LINKED ISSUE #5246: Support runtime-sized successful returndata in raw_call (follow-up to #4856)
### Version information

- Vyper: `0.5.0b2+commit.5764199a`, master at `5764199a9ab64e13bc64d8f41371378db4fa6ab0` (checked September 5, 2026).
- Also observed at the port's pinned compiler, `8af5e83c10af4f065cc660fe20293dac11fdff83`. The intervening commit changes only `.github/workflows/codeql.yml`.
- Python 3.13.14; macOS 26.6.2, arm64.
- Venom (`--experimental-codegen`), Cancun; bytecode metadata disabled.

### Summary and relation to existing work

This is a focused implementation request for the optional `raw_call` return-value work already described in #4856, motivated by the Aave V4 port. Please link or fold it into an existing tracking issue if preferred.

A generic forwarding or multicall implementation needs to capture the complete successful return bytes, with memory allocation based on the actual return length, without choosing a finite compile-time maximum.

### Current limitation

This capability probe is rejected on the master commit above:

```vyper
# pragma version 0.5.0b2
@external
def forward(target: address, data: Bytes[INF]) -> Bytes[INF]:
    return raw_call(target, data, max_outsize=INF)
```

```sh
vyper --evm-version cancun --experimental-codegen -O 2 -f bytecode runtime_return_data.vy
```

Diagnostic:

```text
vyper.exceptions.TypeMismatch: Given reference has type INF, expected uint256
```

This is not a claim that `max_outsize=INF` is documented syntax that is broken. The request is for the underlying capability; the final syntax/API is a language-design decision.

[Probe source](https://github.com/aave/aave-v4/blob/1f7129042fe57e24bfc4c934be5088514cb93503/vyper/compiler-feedback/cases/runtime_return_data.vy), [pinned diagnostics](https://github.com/aave/aave-v4/blob/1f7129042fe57e24bfc4c934be5088514cb93503/vyper/compiler-feedback/capabilities.json), and [port context](https://github.com/aave/aave-v4/blob/1f7129042fe57e24bfc4c934be5088514cb93503/gas-snapshots/COMPILER_FEEDBACK_DIARY.md). The same rejection was rechecked on current master.

### Motivation and requested semantics

Choosing a larger finite buffer imposes a maximum response size absent from a generic Solidity `bytes` forwarding API. Rejecting oversized results makes that restriction explicit but still narrows the accepted behavior.

Desired behavior:

- Capture the complete successful return bytes as a runtime-sized value, suitable for returning, hashing, or aggregating.
- Account for memory using the actual response length, subject to normal EVM gas limits.
- Preserve a bounded option for callers that deliberately want a limit; specify whether any new limiting API truncates or rejects.
- Preserve existing failure-propagation behavior. Native revert bubbling already preserves complete revert data; successful return capture is the remaining requirement here.

This is distinct from selecting a raw output ABI/type (as discussed in #3856 and #4893). It concerns acquiring all successful returndata in the first place. It is also separable from nested unbounded collections: even a single `Bytes[INF]` forwarder needs it.

An implementation test should exercise empty and nonempty successful results, several runtime-selected lengths, and exact byte preservation without setting a larger compile-time maximum for each test.

harkal:
### What I did

Closes: https://github.com/vyperlang/vyper/issues/5246

`raw_call(..., max_outsize=INF)` returns the whole return data as `Bytes[INF]`, or `(bool, Bytes[INF])` with `revert_on_failure=False`, sized by `returndatasize` at runtime.

`max_outsize` was a compile-time cap: proxies, forwarders and multicall wrappers had to guess a bound or truncate silently. Venom already has `Bytes[INF]` locals and unbounded `extcall` returns; `raw_call` was the last return-data ingress without an unbounded form. Venom only; the legacy pipeline rejects it.

```vyper
#pragma experimental-codegen

@external
@payable
def forward(_target: address) -> Bytes[INF]:
    return raw_call(_target, msg.data, max_outsize=INF, value=msg.value)
```

### How I did it

**Semantics** (`vyper/builtins/functions.py`, `vyper/builtins/_signatures.py`). `RawCall._is_unbounded_outsize` recognises the `INF` name on `max_outsize` and nothing else, so `gas=INF` and `value=INF` still fail the `uint256` check.`BuiltinFunctionT._validate_arg_types` had its per-kwarg loop body extracted into `_validate_kwarg` so `RawCall` canoverride it for that one kwarg without duplicating the loop; no behaviour change for other builtins. `RawCall.infer_kwarg_types` annotates the kwarg with `INF`'s own type because local analysis re-visits every kwarg value against those types. `fetch_call_return` returns `BytesT(INF)` / `TupleT([BoolT(), BytesT(INF)])`.

**Which type drives codegen.** The unbounded branch in `lower_raw_call` is keyed on the call's own return type, not on the kwarg and not on the node annotation. The `Call` node's `_metadata["type"]` is the consumer-widened type: `x: Bytes[INF] = raw_call(t, d, max_outsize=32)` annotates the call as `Bytes[INF]`, and branching on it would drop the cap. Codegen therefore recomputes `node.func._metadata["type"].fetch_call_return(node)`, which is pure. There is no stored accessor for a builtin call's own return type: `call_return_type` is only written for external `ContractFunctionT` calls, and the other widenable builtins (`concat`, `slice`, `abi_encode`, ...) legitimately produce the widened buffer type, so extending the metadata would add a second writer for a single consumer. Two runtime tests pin the bounded cap through a `Bytes[INF]` consumer; they fail if the annotation is used.

**Lowering** (`vyper/codegen_venom/builtins/system.py`, `context.py`). On the unbounded path the call is emitted with `retsz=0` (the existing zero-output path). The `ok` block then calls `materialize_returndata_bytes`: `returndatasize` -> checked `32 + ceil32(len)` -> scratch allocation -> length store -> padding zero -> `returndatacopy`. With `revert_on_failure=False` the `(bool, Bytes[INF])` result is a dynamic tuple frame around that value; the bytes member isthe return data on both success and failure, as on the bounded path. The bounded branches are textually unchanged apart from an assertion that the semantic type matches the literal, so bounded output is byte-identical (verified below).

**Cost.** The return data is copied raw, no decode: the caller pays memory expansion for the actual `returndatasize`only, and the callee can only produce what the caller's gas allowed. `x: Bytes[INF] = raw_call(...)` copies the value twice (scratch, then the local), the same as unbounded `extcall` returns today.

**Legacy.** The existing unbounded-sequence gate (`_validate_legacy_function_body_no_unbounded_sequences`) already rejects every position, declaration and expression-only (`len(...)`, `keccak256(...)`, `[...][0]`), with `unbounded sequence types require --experimental-codegen`. Tests pin that message and that the same code compiles under Venom.

**Docs.** `raw_call` entry in `docs/built-in-functions.rst`, plus a `types-unbounded` label in `docs/types.rst` for the cross-reference.

### How to verify it

- `tests/functional/codegen/features/test_inf_raw_call.py` (Venom only, 17 tests): empty and 10 KB return data; assignment to a `Bytes[INF]` local; external return; forwarding through an internal function (`dret`, inlining disabled) and into an `extcall` argument; `(bool, Bytes[INF])` unpack, tuple return and direct subscript on success and on failure (revert data returned, not raised); `is_static_call` and `is_delegate_call`; `msg.data` forwarding proxy, including return data shorter than the forwarded calldata and not word-aligned; bounded `max_outsize` returned through a `Bytes[INF]` consumer keeps its cap.
- `tests/unit/semantics/types/test_inf.py`: resolved return type for `revert_on_failure` x plain/static/delegate; ABI `bytes` / `(bool, bytes)`.
- `tests/functional/syntax/test_raw_call.py`: `gas=INF` / `value=INF` rejected; narrowing into `Bytes[100]` / `(bool, Bytes[100])` rejected with the exact message; legacy rejection in four positions; bounded forms unchanged.

### Commit message

```
`max_outsize` is a compile-time cap, so proxies, forwarders and
multicall wrappers have to guess a bound on the returndata or truncate
it silently. venom already has `Bytes[INF]` locals and unbounded
`extcall` returns; `raw_call` was the last returndata ingress without an
unbounded form.

accept `max_outsize=INF`: the call then returns `Bytes[INF]`, or `(bool,
Bytes[INF])` with `revert_on_failure=False`, sized by the actual
returndatasize. only `max_outsize` accepts the name, via a per-kwarg
validation hook factored out of `BuiltinFunctionT._validate_arg_types`,
so `gas=INF` and `value=INF` still fail the uint256 check.

codegen keys the unbounded branch on the call's own return type rather
than on the node annotation. the annotation is the consumer-widened
type, so `x: Bytes[INF] = raw_call(t, d, max_outsize=32)` annotates the
call `Bytes[INF]` and branching on it would drop the cap. the unbounded
call is emitted with no output buffer, and the whole returndata is
copied into a runtime-sized scratch buffer once it returns;
`revert_on_failure=False` wraps that in a dynamic tuple frame. the
bounded path keeps its static buffer and compiles to the same bytecode
as before. the legacy pipeline rejects every use through the existing
unbounded sequence gate.
```

`raw_call` accepts `max_outsize=INF` under the experimental code generator and returns the whole return data as `Bytes[INF]` (or `(bool, Bytes[INF])` with `revert_on_failure=False`).

### Cute Animal Picture


charles-cooper (approved):
surprisingly simple! lgtm pending clanker + @Sporarum @HodanPlodky review

vyper/builtins/functions.py
```diff
@@ -1035,13 +1035,35 @@ class RawCall(BuiltinFunctionT):
...
+            return False
+        return outsize.id == "INF"
+
+    def _validate_kwarg(self, kwarg) -> None:
+        if self._is_unbounded_outsize(kwarg):
+            return
+        super()._validate_kwarg(kwarg)
+
+    # local analysis visits each kwarg value against these types; uint256 would reject INF
+    def infer_kwarg_types(self, node):
```
HodanPlodky, line 1049: isn't this possible to handle through the `_kwargs` field. And shouldn't be something done with that field either way since the `max_outsize` is still marked as `uint256` there?

harkal: Good catch the` _kwargs` takes tuple of types... patching!

docs/types.rst
```diff
@@ -563,6 +563,8 @@ In the ABI, they are represented as ``_Type[]``. For instance, ``DynArray[int128
 
 .. index:: unbounded sequences, INF
 
+.. _types-unbounded:
```
Sporarum: Why not `_unbounded_types` ?

tests/functional/syntax/test_raw_call.py
```diff
@@ -195,3 +214,65 @@ def foo():
...
+    """,
+]
+
+
+@pytest.mark.parametrize("code", unbounded_outsize_list)
+def test_raw_call_unbounded_outsize_requires_experimental_codegen(code):
+    with pytest.raises(StructureException) as e:
+        compile_code(code, settings=Settings(experimental_codegen=False))
+    assert e.value.message == "unbounded sequence types require --experimental-codegen"
+    assert compile_code(code, settings=Settings(experimental_codegen=True)) is not None
```
Sporarum: This should be

```suggestion
    if <condition to check if legacy>:
        with pytest.raises(StructureException) as e:
            compile_code(code)
        assert e.value.message == "unbounded sequence types require --experimental-codegen"
    else:
        assert compile_code(code) is not None
```

Sporarum (approved):
Approve pending nits

Sporarum:
Oh, and should close https://github.com/vyperlang/vyper/issues/5246

(add `Closes: #5246` somewhere in the PR description, I usually put it in "What I did")

PR #5276, opened 2026-09-21
chore[venom]: remove dead code, clarify invariants, cleanups

harkal:
### What I did

Remove unreachable and redundant paths from Venom codegen and fix loop-variable scope cleanup.

### How I did it

- Remove dead expression and ABI encoding branches; reject impossible byte-builtin source types explicitly.
- Unify runtime/deploy entry-point names and tighten annotations, constants, and documentation.
- Register loop variables inside their loop scope while preserving allocation placement. Add coverage for nested loops and reused names.

Clean up Venom codegen internals and fix loop-variable scope cleanup.

### Commit message

```
venom codegen contains branches ruled out by semantic analysis or the
calling convention. Together with inconsistent entry-point names and
incomplete annotations, these paths obscure the invariants that lowering
relies on.

remove unreachable paths and redundant checks, reject impossible builtin
source types explicitly, and align naming and annotations with the
existing contracts. Mark the allocation limit as a class constant so it
is no longer exposed as a dataclass constructor argument.

loop variables were registered before entering loop_scope, leaving them
in the codegen context after the loop ended. Register them within the
scope while keeping allocations in their original positions to preserve
generated code. Add regression coverage for nested range and array loops
and reused loop-variable names
```


vyper/codegen_venom/abi/abi_encoder.py
```diff
@@ -456,7 +443,6 @@ def _abi_encode_to_buf(
     # Fast path: if ABI encoding matches Vyper memory layout, just copy
     if abi_encoding_matches_vyper(src_typ):
         size = src_typ.memory_bytes_required
-        assert abi_t.embedded_static_size() == size
```
HodanPlodky, line 459: why was this removed?

vyper/codegen_venom/context.py
```diff
@@ -1092,13 +1092,11 @@ def _copy_dynarray_memory_typed(
...
         src_elem_t = src_typ.value_type
         dst_elem_size = dst_elem_t.memory_bytes_required
-        src_elem_size = src_elem_t.memory_bytes_required
-
         src_data = self._with_byte_offset(src, 32)
         dst_data = self._with_byte_offset(dst, 32)
 
         # Fast path when element layouts match: copy exactly `length` elements.
-        if src_elem_t == dst_elem_t and src_elem_size == dst_elem_size:
+        if src_elem_t == dst_elem_t:
```
HodanPlodky, line 1099:
```suggestion
        if src_elem_t == dst_elem_t:
            assert src_elem_t.size == dst_elem_size
```

HodanPlodky: Just for sanity

vyper/codegen_venom/expr.py
```diff
@@ -1580,14 +1572,8 @@ def _lower_internal_call(self) -> VyperValue:
             arg_t = func_t.arguments[i]
 
             if pass_via_stack_dict[arg_t.name]:
```
HodanPlodky, line 1574:
```suggestion
            if pass_via_stack_dict[arg_t.name]:
                assert arg_t.typ._is_prim_word
```

HodanPlodky (commented):
Got few NITs otherwise looks good

vyper/codegen_venom/context.py
harkal: agree, it'd be `assert src_elem_t.memory_bytes_required == dst_elem_size` though. I will push

vyper/codegen_venom/abi/abi_encoder.py
harkal: sideline casualty :S I will be adding it back

HodanPlodky (approved):

